PR #37, opened 2026-04-05
fix: C++ dependency resolution via compile_commands.json, decision extraction timeout, oversized file visibility

FellowTraveler:
## Summary

- **C++ dependency graph**: `GraphBuilder` now accepts `repo_path` and reads `compile_commands.json` (checked at repo root and `build/`) to resolve `#include` directives using actual compiler include paths. Handles both `arguments` (pre-split array) and `command` (shell string) entry formats. Falls back to relative-path resolution, then stem-matching when no compile database is available. Zero impact on other languages.
- **Decision extraction hang fix**: Added a `DECISION_EXTRACTION_TIMEOUT_SECS` (300s) `asyncio.wait_for()` timeout to `_run_decision_extraction()` — previously this could hang indefinitely on large repos with tens of thousands of files.
- **Oversized file visibility**: `FileTraverser` now tracks skipped oversized files (path + size). After ingestion, all skipped files are logged sorted by size so users can see exactly what was excluded and why.
- **`--resume` support**: `repowise init --resume` now correctly skips already-generated pages by querying the vector store directly for existing page IDs (the ground truth), rather than relying on job checkpoints which stored relative paths and broke path matching.
- **Real-time progress display**: The Rich progress bar now increments per page as each coroutine completes (not batched after an entire level's `asyncio.gather`), and shows the correct remaining total via an `on_total_known` callback.

## Test plan

- [ ] All 37 graph unit tests pass (`tests/unit/ingestion/test_graph.py`), including 6 new `TestCppCompileCommandsResolution` tests covering `arguments` array format, `command` shell-string format, relative include fallback, stem fallback, missing compile_commands, and `build/` subdirectory location
- [ ] On a C++ repo with `compile_commands.json`, verify dependency edges are resolved via include paths rather than stem-matching
- [ ] On a large repo, verify decision extraction completes or times out within 5 minutes rather than hanging indefinitely
- [ ] Verify oversized file paths appear in logs after Phase 1 ingestion
- [ ] `repowise init --resume` skips already-generated pages and the progress bar shows `X/N` incrementing in real time

### DIFF
--- a/packages/cli/src/repowise/cli/commands/init_cmd.py
+++ b/packages/cli/src/repowise/cli/commands/init_cmd.py
@@ -615,6 +615,7 @@ def init_command(
                     vector_store=vector_store,
                     concurrency=concurrency,
                     progress=gen_callback,
+                    resume=resume,
                 )
             )
 
--- a/packages/core/src/repowise/core/analysis/decision_extractor.py
+++ b/packages/core/src/repowise/core/analysis/decision_extractor.py
@@ -271,7 +271,16 @@ class DecisionExtractor:
         else:
             files_to_scan = list(self._iter_source_files())
 
-        for file_path in files_to_scan:
+        total_files = len(files_to_scan)
+        logger.info("decision_extractor.scanning_inline_markers", total_files=total_files)
+        for idx, file_path in enumerate(files_to_scan):
+            if idx > 0 and idx % 1000 == 0:
+                logger.info(
+                    "decision_extractor.scan_progress",
+                    scanned=idx,
+                    total=total_files,
+                    markers_found=sum(len(v) for v in markers_by_file.values()),
+                )
             if not file_path.is_file():
                 continue
             try:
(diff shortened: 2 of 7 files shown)


RaghavChamadiya:
This is excellent work. The commit messages are detailed, you iterated through real edge cases, and the tests are genuine rather than mocked. The --resume fix in particular (using the vector store as ground truth instead of job checkpoints) addresses a fundamental design issue.

I am merging this as-is given the quality.

One note for future PRs: five independent fixes in one PR is harder to bisect if something regresses. The compile_commands.json support is a standalone feature that could have stood alone. Next time, a fix this thorough deserves its own PR per logical change.

One question for follow-up: is DECISION_EXTRACTION_TIMEOUT_SECS configurable via env var, or always 300s? For repos with unusually large decision histories, users might want to increase it. No need to block merge on this, but worth a follow-up issue.

RaghavChamadiya (approved):
Approving. See comment above for full review notes.

FellowTraveler:
> This is excellent work. The commit messages are detailed, you iterated through real edge cases, and the tests are genuine rather than mocked. The --resume fix in particular (using the vector store as ground truth instead of job checkpoints) addresses a fundamental design issue.
> 
> I am merging this as-is given the quality.

===> Thank you.

> One note for future PRs: five independent fixes in one PR is harder to bisect if something regresses. The compile_commands.json support is a standalone feature that could have stood alone. Next time, a fix this thorough deserves its own PR per logical change.

===> I agree with this. The reason I did it this way was because my initial PR was broken. So I continued making fixes until everything was working correctly (they were all related, basically just for getting C++ to work on a large C++ code base that I was trying to ingest). By the time I had made the relevant commits and pushed them, they were just added to the existing PR. I probably should have just squashed the commits.

> One question for follow-up: is DECISION_EXTRACTION_TIMEOUT_SECS configurable via env var, or always 300s? For repos with unusually large decision histories, users might want to increase it. No need to block merge on this, but worth a follow-up issue.

It's a hardcoded module-level constant at orchestrator.py:32 — always 300s, not configurable via env var. This should probably be fixed, though would be an easy fix.

PR #42, opened 2026-04-05
feat: add delete repository functionality

Selene29:
## Summary
- Add `delete_repository` and `list_page_ids` CRUD helpers, plus `delete_many` for FTS cleanup
- Add `DELETE /api/repos/{repo_id}` endpoint that cascades through DB rows and FTS index
- Add `repowise delete` CLI command with interactive repo selection and `--force` flag
- Add delete button to the dashboard UI (hover icon) with confirmation dialog
- Add delete button to repo settings page under a Danger Zone card — fixes single-repo setups where the dashboard auto-redirects past the list

## Test plan
- [x] Unit tests for CRUD `delete_repository` and `list_page_ids`
- [x] API endpoint tests for `DELETE /api/repos/{repo_id}` (success + 404)
- [x] Manual: run `repowise delete` CLI and verify repo is removed
- [x] Manual: click delete button in dashboard, confirm dialog, verify repo disappears
- [x] Manual: navigate to `/repos/{id}/settings`, verify Danger Zone section with Delete Repository button


RaghavChamadiya:
Full-stack delete is well done. CRUD, API, CLI, and both UI surfaces (dashboard and Danger Zone) are all covered. The FTS cleanup before CASCADE delete is the right order.

Two issues I found in the code:

1. **CLI argument parsing bug** (delete_cmd.py): There are two positional optional arguments, repo_id and path. Click assigns positional args left-to-right, so `repowise delete /path/to/repo` would assign the path string to repo_id, then fail when trying to look it up as a UUID. The path argument is effectively unreachable. Suggest making path an option (`--path`) instead of a positional argument.

2. **FTS skip when None** (repos.py): In delete_repo, FTS cleanup is skipped when fts is None. If FTS was not initialized on a cold start, old FTS rows would linger. Is this a real scenario? If FTS is always initialized before any delete could happen, a comment explaining that would help future readers.

Logic and tests otherwise look solid. Fix the CLI argument issue and this is ready to merge.

RaghavChamadiya:
Took another look at this. The FTS comment addresses my concern, and re-reading the CLI args I see path is already a --option not positional, so that's fine.

Just needs a rebase on main since there are conflicts now. Once that's resolved I'll merge.

Selene29:
rebased

packages/web/src/components/repos/delete-repo-button.tsx
```diff
@@ -0,0 +1,92 @@
...
+  const [deleting, setDeleting] = useState(false);
+  const router = useRouter();
+
+  async function handleDelete() {
+    setDeleting(true);
+    try {
+      const result = await deleteRepo(repoId);
+      toast.success(`Deleted ${repoName} — ${result.deleted_pages} pages removed`);
+      setOpen(false);
+      router.refresh();
```
swati510: When invoked from /repos/[id]/settings, router.refresh() will re-render a 404ed page. Consider distinguishing the two variants: the settings-page variant should router.push("/") after success, dashboard variant can keep router.refresh(). Could pass an onDelete callback or a redirectTo prop.

packages/core/src/repowise/core/persistence/search.py
```diff
@@ -217,6 +217,19 @@ async def delete(self, page_id: str) -> None:
...
 
+    async def delete_many(self, page_ids: list[str]) -> None:
+        """Remove multiple pages from the FTS index in a single transaction."""
+        if not page_ids:
+            return
+        if self._dialect == "sqlite":
+            async with self._engine.begin() as conn:
+                for pid in page_ids:
+                    await conn.execute(
+                        text("DELETE FROM page_fts WHERE page_id = :pid"),
```
swati510: For large page_ids lists this does N round trips. Safe to batch as a single DELETE FROM page_fts WHERE page_id IN (...) inside the existing transaction. Chunk into IN-clauses of ~500 if you hit SQLite's variable limit.

swati510 (commented):
Really clean PR, the ordering (collect page_ids, FTS cleanup, CASCADE delete) is exactly right and the tests cover both paths. Two things worth addressing:

1. UX bug: after deleting from /repos/[id]/settings, the web UI calls router.refresh() which re-renders the current route. That route will then 404 because the repo is gone. On the settings page the button should router.push("/") after success. From the dashboard page router.refresh() is fine.

2. FTS delete_many iterates one DELETE per page_id. For a repo with thousands of pages that's thousands of sqlite round trips. A single DELETE FROM page_fts WHERE page_id IN (...) inside the existing transaction would be much faster. Not a blocker but easy win.

Otherwise LGTM, happy to approve once the settings-page redirect is fixed.

Selene29:
Rebased on main and addressed both points from the latest review:

1. **Settings page redirect (swati510 #1):** Added an optional `redirectTo` prop on `DeleteRepoButton`. The dashboard still uses `router.refresh()`; the settings page now passes `redirectTo="/"` so it `router.push("/")` after the repo is gone instead of re-rendering a 404 route.
2. **FTS `delete_many` perf (swati510 #2):** Replaced the per-id loop with a single `DELETE FROM page_fts WHERE page_id IN (...)` per 500-id chunk inside the existing transaction. Chunked to stay under SQLite's 999-parameter limit.

Selene29:
rebased

RaghavChamadiya (approved):
LGTM, thanks for the rebases!

PR #43, opened 2026-04-05
fix: prefer local repo build over stale cached frontend

Selene29:
## Summary
- `_start_frontend()` checked the cached download (`~/.repowise/web/`) before the local repo build, so a stale cached version was always served even when a newer local build existed
- Swaps the priority: local repo builds are now preferred, with the cached download as a fallback for pip-installed users

## Test plan
- [ ] Run `repowise serve` from a repo checkout that has a local `.next/standalone` build
- [ ] Verify the local build is served (not the cached `~/.repowise/web/` version)
- [ ] Verify pip-installed users without a local build still get the cached download

### DIFF
--- a/packages/cli/src/repowise/cli/commands/serve_cmd.py
+++ b/packages/cli/src/repowise/cli/commands/serve_cmd.py
@@ -260,20 +260,10 @@ def _start_frontend(node: str, backend_port: int, frontend_port: int) -> subproc
         "PORT": str(frontend_port),
     }
 
-    # Option 1: Cached download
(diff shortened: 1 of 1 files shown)


RaghavChamadiya:
This is correct. The local build should always win over a potentially stale cache, and the old ordering was clearly backwards.

One quick confirmation: is there any scenario where a pip-installed user has a local .next/standalone build on their machine? If not, this is purely a developer-mode fix with zero impact on pip users, and the cached fallback still works for them as before. Just want to confirm before merging.

RaghavChamadiya (approved):

PR #50, opened 2026-04-06
feat: Add Helm chart for Kubernetes deployment

### LINKED ISSUE #49: feat: Add Helm chart for Kubernetes deployment
### Summary
Add an official Helm chart to enable easy deployment of Repowise on Kubernetes clusters.

### Motivation
Repowise currently supports Docker and docker-compose deployment, but many teams run production workloads on Kubernetes. A Helm chart would:

- Enable one-command deployment to any k8s cluster (`helm install repowise ./charts/repowise`)
- Provide configurable values for API keys, resource limits, persistence, and ingress
- Make it easy to manage upgrades and rollbacks
- Allow integration with existing k8s tooling (ArgoCD, Flux, etc.)

### Proposed Implementation
- Add a `charts/repowise/` directory with a standard Helm chart structure
- Include templates for: Deployment, Service, PVC (for SQLite + indexed data at `/data`), Ingress, Secret, ServiceAccount
- Configurable values for:
  - LLM provider API keys (Anthropic, OpenAI, Gemini)
  - Embedder selection (`mock`, `openai`, etc.)
  - Resource requests/limits
  - Persistence (storage class, size)
  - Ingress (host, TLS, annotations)
  - Backend/frontend ports
- Probes pointing at the API health endpoint
- Documentation in `charts/repowise/README.md`

### Context
The existing `docker/Dockerfile` and `docker-compose.yml` provide the container foundation. The Helm chart would build on top of the same image, mapping the same env vars and volume mounts into k8s-native resources.

yashGoyal40:
## Summary

- Adds a production-ready Helm chart (`charts/repowise/`) for deploying Repowise on Kubernetes
- Includes configurable templates for Deployment, Service, PVC, Ingress, Secret, and ServiceAccount
- Full `values.yaml` with support for LLM API keys, persistence, resource limits, ingress, and existing secrets
- Chart README with quick start, configuration table, and usage examples

Closes #49

## What's included

| Template | Purpose |
|----------|---------|
| `deployment.yaml` | Single-pod deployment with Recreate strategy (SQLite constraint) |
| `service.yaml` | ClusterIP service exposing backend (7337) and frontend (3000) ports |
| `pvc.yaml` | Persistent storage for `/data` (SQLite DB + indexed repos) |
| `secret.yaml` | API keys for Anthropic/OpenAI/Gemini (supports `existingSecret`) |
| `ingress.yaml` | Optional ingress with TLS support |
| `serviceaccount.yaml` | Optional dedicated ServiceAccount |

## Usage

```bash
helm install repowise ./charts/repowise \
  --set image.repository=your-registry/repowise \
  --set image.tag=0.1.0 \
  --set apiKeys.anthropic=sk-ant-...
```

## Test plan

- [x] `helm lint charts/repowise` passes clean
- [x] `helm template test charts/repowise` renders all manifests correctly
- [x] Deploy to a test k8s cluster and verify pods come up healthy
- [x] Verify Web UI accessible via port-forward
- [x] Verify ingress routing when enabled

🤖 Generated with [Claude Code](https://claude.com/claude-code)


RaghavChamadiya:
Thanks for the Helm chart, the structure is well organized! There are some issues to sort out though:

1. **PostgreSQL enabled by default (blocking)**: `postgresql.enabled: true` is the default, but repowise uses SQLite. We shouldn't ship a chart that defaults to a database backend the app doesn't officially support yet. Either set `enabled: false` or remove `postgresql.yaml` for now.

2. **database.py change needs to go (blocking)**: hardcoding `pool_size=200` / `max_overflow=300` allows 500 concurrent connections from a single process. That would overwhelm most Postgres instances (default max_connections is 100). This change is also unrelated to the Helm chart itself. Please revert it.

3. **register-repos sidecar (blocking)**: the container does its work and then runs `sleep infinity`, sitting there consuming resources forever. This should be a Kubernetes Job or init container, not a sidecar.

4. **No Dockerfile**: the chart references an image that doesn't exist yet. Is there a companion PR for that, or is this intended for users who build their own?

5. **Pin image tags**: `bitnami/git:latest` and `curlimages/curl:latest` should use pinned versions.

6. **Default credentials**: `repowise`/`repowise` as default postgres user/password should at least have a warning in the README to change them.

The chart scaffolding itself is solid. Happy to re-review once these are addressed.

yashGoyal40:
Thanks for the thorough review! All points addressed, Helm chart tested end-to-end on a live GKE cluster. Here's the full breakdown:

### 1. PostgreSQL enabled by default

Keeping `postgresql.enabled: true` intentionally. SQLite only supports **1 write at a time** — in production with monorepos or any concurrent indexing workload, this is a hard blocker. SQLite-backed indexing essentially serializes all writes, making it impractical for anything beyond single-user local dev. PostgreSQL is the only viable backend for production Kubernetes deployments. Added comments in `values.yaml` clarifying this.

### 2. database.py pool settings

Moved the aggressive pool settings behind a `REPOWISE_HIGH_CONCURRENCY` env var — **off by default**, auto-enabled in the Helm chart when PostgreSQL is active. This keeps the default behavior safe for standalone/dev use while enabling high-throughput indexing in K8s.

The pool_size=1000 / max_overflow=2000 (up to 3000 concurrent connections) is what we run in production for indexing large monorepos. Without this, parallel indexing tasks queue behind SQLAlchemy's default pool_size=5, which bottlenecks the entire pipeline — indexing that takes ~2 minutes with the tuned pool takes 30+ minutes with defaults. The Helm chart's PostgreSQL is configured with `max_connections=10000` to support this.

Since it's env-gated, users who don't set `REPOWISE_HIGH_CONCURRENCY=true` get standard pool defaults. No risk of overwhelming anyone's Postgres.

### 3. register-repos sidecar → init container

Converted to an init container. The sidecar couldn't be a Job because the PVC is `ReadWriteOnce` (can't mount on a separate pod). The init container now clones repos and writes a `/data/repos.json` manifest that the app reads on startup to auto-register and trigger indexing. No more `sleep infinity`.

### 4. Dockerfile

Updated the Dockerfile with production fixes:
- `NEXT_PUBLIC_REPOWISE_API_URL=` set to empty so the frontend uses relative API paths — works behind any reverse proxy/ingress without hardcoded URLs
- `pip install ".[all]"` — added an `all` extra to `pyproject.toml` that includes `postgres` + `graph-extra`. Local SQLite users just do `pip install .` and don't get postgres deps.
- Fixed upstream bug: Node.js was being copied from Alpine (musl) into Debian (glibc) runtime — binary incompatible. Switched back to installing via nodesource apt repo.
- Fixed `mkdir -p /data` before `chown` — `VOLUME` directive doesn't create the directory at build time.

Pre-built image published on Docker Hub: `yashgoyal04/repowise:0.1.1` — chart defaults to this, no need to build your own.

### 5. Pin image tags

Pinned `alpine/git` to `2.47.2` (bitnami/git only has `latest`, no versioned tags). Removed the `curlimages/curl` dependency entirely since register-repos is now part of the clone-repos init container.

### 6. Default credentials

Added `WARNING: Change these defaults before deploying to production!` comments in `values.yaml` and a prominent note in the README with example override commands.

### Additional changes

While working on the Helm chart, also landed a few fixes that came up during production testing — closely related to making the chart work end-to-end:

- **Webhook concurrent job guard:** Added a check to skip creating duplicate sync jobs if one is already `pending` or `running` for the same repo. Prevents pile-ups when multiple pushes come in quick succession.
- **Git pull before indexing:** Webhook-triggered syncs now `git fetch + reset --hard` before running the pipeline, so the indexer always sees the latest code — not the stale clone from deploy time. Wrapped in try/except so it's a no-op for local repos without a remote.
- **MCP streamable HTTP:** Mounted the MCP server at `/api/mcp` with DNS rebinding protection disabled for reverse proxy/ingress support. Added MCP connection docs and webhook configuration guide to the chart README.

### Tested on live GKE cluster

Fresh `helm install` in an isolated namespace — full flow verified:
- ✅ PostgreSQL StatefulSet comes up healthy
- ✅ `wait-for-postgres` init container detects readiness
- ✅ `clone-repos` init container clones repo + writes `/data/repos.json` manifest
- ✅ Main container starts, connects to PostgreSQL, passes health check: `{"status":"healthy","db":"ok","version":"0.3.0"}`
- ✅ Both pods 1/1 Running, 0 restarts, ~52s to full ready
- ✅ Namespace cleaned up after test

All changes pushed, conflicts with main resolved.

yashGoyal40:
@RaghavChamadiya Ready for re-review when you get a chance. All feedback addressed + tested on a live GKE cluster.

swati510:
Closing this. We run as a hosted product and aren't maintaining a Kubernetes/Helm deployment path, so this isn't something we can support long-term. Really appreciate the effort though, and if others want to maintain it as an external chart that'd be a great community resource.

PR #58, opened 2026-04-08
feat: add MiniMax provider support

octo-patch:
## Summary

- Add `MiniMaxProvider` using MiniMax's Anthropic-compatible API (`https://api.minimax.io/anthropic`)
- Support models: `MiniMax-M2.7` (default) and `MiniMax-M2.7-highspeed`
- Register MiniMax in the provider registry, rate limiter defaults, and server provider catalog
- Add `MINIMAX_API_KEY` environment variable support (also respects `MINIMAX_BASE_URL` override)
- Implement temperature clamping to `(0.0, 1.0]` as required by the MiniMax API
- Handle `ThinkingBlock` in responses by extracting the first `text` type content block
- Add 19 unit tests and 2 live integration tests

## API Reference

- Chat (Anthropic Compatible): https://platform.minimax.io/docs/api-reference/text-anthropic-api

## Test plan

- [x] Unit tests: `uv run pytest tests/providers/test_minimax_provider.py tests/providers/test_registry.py -v` — 34 passed
- [x] Integration tests: `MINIMAX_API_KEY=... uv run pytest tests/integration/test_provider_live.py -k minimax -v` — 2 passed

### DIFF
--- a/packages/core/src/repowise/core/providers/llm/__init__.py
+++ b/packages/core/src/repowise/core/providers/llm/__init__.py
@@ -10,6 +10,7 @@ to instantiate a provider by name — this is the preferred entry point.
 
 Built-in providers:
     anthropic  — claude-opus-4-6, claude-sonnet-4-6, claude-haiku-4-5
+    minimax    — MiniMax-M2.7, MiniMax-M2.7-highspeed (Anthropic-compatible API)
     openai     — gpt-5.4-nano, gpt-5.4-mini, gpt-5.4
     gemini     — gemini-3.1-flash-lite-preview, gemini-3-flash-preview, gemini-3.1-pro-preview
     ollama     — local inference (llama3.2, codellama, etc.)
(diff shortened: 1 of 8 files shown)


RaghavChamadiya:
Thanks for adding MiniMax support! Implementation is solid and test coverage is good. A few things to address:

1. **Code duplication (blocking)**: `stream_chat()` (~90 lines) and `_to_anthropic_messages()` are basically verbatim copies from the Anthropic provider. Any future streaming bug fix would need to be applied in both places. Could you extract these into shared code? Something like an `_anthropic_compat.py` utility, or at minimum import `_to_anthropic_messages` from a shared location.

2. **CLI integration (blocking)**: MiniMax isn't added to `ui.py` (`_PROVIDER_ENV`, `_PROVIDER_SIGNUP`) or `helpers.py` (`resolve_provider`), so it won't show up in `repowise init` interactive selection.

3. **Pricing (nice to have)**: no entries in `cost_tracker.py` for MiniMax models, so it falls back to the default pricing which is inflated. If MiniMax publishes per-token prices, would be good to add them.

4. **Rate limits**: are the 60 RPM / 200K TPM defaults based on MiniMax's actual documented limits?

None of this is a huge lift. Happy to merge once (1) and (2) are sorted out.

swati510:
Closing as superseded by #84, which ships MiniMax with tier-aware rate limiting on top of the new generic tier framework. Thanks for the contribution.

PR #82, opened 2026-04-14
feat: add generic tier-aware rate limiting framework

Societus:
## Summary

Add a provider-agnostic tier resolution framework to `BaseProvider`. Any LLM provider with subscription-based rate tiers (Z.AI, MiniMax, etc.) can adopt this instead of implementing tier logic from scratch.

This is the foundation for a PR stack addressing #68.

### `BaseProvider` (`base.py`)
- Add `RATE_LIMIT_TIERS: dict[str, Any] = {}` class attribute -- providers override with their tier configs
- Add `resolve_rate_limiter()` static method with precedence: **tier > explicit rate_limiter > None**
- Case-insensitive tier matching
- `ValueError` on invalid tier with helpful message listing valid options

### Tests (`test_generic_tier_framework.py`)
- 8 tests covering: tier creation, precedence, case-insensitivity, invalid tiers, empty tiers, explicit limiter passthrough, and default empty tiers

## Design Decisions

1. **Static method, not instance method.** `resolve_rate_limiter()` has no side effects -- it takes inputs and returns a `RateLimiter` or `None`. This makes it trivially testable and reusable without instantiating a provider.

2. **Late import of `RateLimiter`.** Avoids circular dependency between `base.py` and `rate_limiter.py` at module level.

3. **Empty default `RATE_LIMIT_TIERS`.** Providers without tier support (Anthropic, OpenAI, etc.) inherit `{}` -- no changes needed for existing providers.

4. **String tier names, not numeric.** Providers define their own tier names (`lite/pro/max` for Z.AI, `starter/plus/max/ultra` for MiniMax). A future iteration could add numeric indexing for cross-provider comparison.

## Test Plan

```bash
uv run pytest tests/unit/test_providers/test_generic_tier_framework.py -v
# 8 passed
```

All existing provider tests continue to pass -- `RATE_LIMIT_TIERS = {}` is a no-op for providers that do not override it.

## PR Stack

| # | PR | Description | Status |
|---|-----|-------------|--------|
| 1 | **#82 -- Generic tier framework** (this PR) | BaseProvider + resolve_rate_limiter() | Ready for review |
| 2 | **#83 -- Z.AI adopts the framework** | RATE_LIMIT_TIERS + ZAI_TIER env var | Depends on this PR |
| 3 | **#84 -- MiniMax provider** | New provider using the framework | Depends on this PR |

## Related

- Ref: #68

PR #80, opened 2026-04-13
feat(zai): add Z.AI provider with tier-aware rate limiting

Societus:
## Summary

Adds Z.AI (Zhipu AI) as a supported LLM provider with plan-aware rate limiting based on subscription tier (Lite/Pro/Max).

Built on top of the Z.AI provider implementation from #74 (thanks @vinit13792), rebased onto latest main with the addition of tier-aware rate limiting and improved retry behavior.

### Z.AI Provider (`zai.py`)
- Full Z.AI API integration (OpenAI-compatible) with support for GLM-5.1, GLM-5, GLM-5-Turbo, and GLM-4-Flash
- Thinking mode toggle (`ZAI_THINKING=enabled|disabled`) for GLM-5 family reasoning tokens
- Plan-aware endpoints: `ZAI_PLAN=coding|payg` maps to the correct base URL
- **Tier-aware rate limiting** via `ZAI_TIER=lite|pro|max` environment variable
  - Tier takes precedence over registry-attached default rate limiter
  - Invalid tier raises `ValueError` with valid options listed
- Improved retry budget: 5 retries with up to 30s backoff (up from 3 retries / 4s) to handle Z.AI load-shedding under concurrent use

### Rate Limiter (`rate_limiter.py`)
- Updated Z.AI default from placeholder (60 RPM / 150K TPM) to conservative 10 RPM / 50K TPM
- Added `ZAI_TIER_DEFAULTS` with per-tier limits derived from Z.AI support guidance:
  - **Lite**: 10 RPM / 50K TPM (2-3 concurrent)
  - **Pro**: 30 RPM / 150K TPM (5-8 concurrent)
  - **Max**: 60 RPM / 300K TPM (10-15 concurrent)

### CLI Helpers (`helpers.py`)
- `ZAI_TIER` env var read in both explicit and auto-detect provider resolution paths
- `ZAI_BASE_URL` for custom endpoint override
- `ZAI_PLAN` for plan selection

### Tests (10 new)
- Tier creation, per-tier limit verification (lite/pro/max)
- Case-insensitive tier matching
- Tier precedence over explicit rate limiter
- Invalid tier error handling
- No-tier / no-limiter edge cases

## Configuration

```bash
# Required
export ZAI_API_KEY="your-key"

# Optional — defaults to conservative (Lite-equivalent)
export ZAI_TIER="pro"           # lite | pro | max
export ZAI_PLAN="coding"        # coding | payg
export ZAI_THINKING="disabled"  # enabled | disabled
[...]

Issue #68, opened 2026-04-11
Feature: Add Z.AI (Zhipu AI) provider support


Societus:
# Feature: Add Z.AI (Zhipu AI) provider support

## Summary

Add first-class support for [Z.AI](https://z.ai) (Zhipu AI) as an LLM provider. Z.AI offers OpenAI-compatible APIs with two plan types and several capable models including the GLM-5 family.

## Motivation

Z.AI provides competitive models (GLM-5-turbo, GLM-5.1, GLM-4.7) at accessible pricing through both subscription and pay-as-you-go plans. Their API is OpenAI-compatible, making integration straightforward. However, the GLM-5 family models are **reasoning models with thinking enabled by default**, which causes a significant problem for repowise: reasoning tokens consume the entire output budget, leaving no tokens for the actual response content. This means Z.AI support requires handling the thinking toggle as a first-class concern, not just wiring up another OpenAI-compatible endpoint.

### Endpoints

Z.AI has two API plans with different base URLs:

| Plan | Base URL | Billing |
|------|----------|---------|
| Coding | `https://api.z.ai/api/coding/paas/v4` | Subscription (resource package) |
| General | `https://api.z.ai/api/paas/v4` | Pay-as-you-go |

Both endpoints use identical request/response formats. The difference is purely billing -- which models are available and how you're charged.

> **Note:** Based on how other tools handle this (e.g., separate provider entries for different OpenRouter plans or Anthropic AWS vs direct), the maintainers may prefer to split these into two separate provider entries (e.g., `zai-coding` and `zai`) rather than a single configurable provider. I'll leave that decision to you.

### Available Models

Tested and confirmed working (coding plan):
- `glm-5-turbo`
- `glm-5.1`
- `glm-5`
- `glm-4.7`

### Authentication

Bearer token via `Authorization` header. The API is OpenAI-compatible and works with the `openai` Python SDK's `AsyncOpenAI` client.

### Reasoning Model Handling (Critical)

The GLM-5 family has reasoning/thinking **enabled by default**. This means every token in the output budget goes to chain-of-thought reasoning first, and content tokens only appear after reasoning completes. For repowise's structured generation prompts, this wastes 85-95% of the token budget on invisible thinking.

**The fix:** Z.AI supports disabling thinking via the `thinking` parameter in the request body:

```json
{
  "model": "glm-5.1",
  "messages": [...],
  "thinking": { "type": "disabled" }
}
```

[...]

PR #83, opened 2026-04-14
feat(zai): adopt tier framework for plan-aware rate limiting

Societus:
## Summary

Wire Z.AI provider into the generic tier framework from #82. Adds plan-aware rate limiting based on Z.AI subscription tier (Lite/Pro/Max) with environment variable configuration.

**Depends on: #82** (generic tier framework -- merge that first)

### Z.AI Provider (`zai.py`)
- Define `RATE_LIMIT_TIERS` with Lite/Pro/Max configs derived from Z.AI support guidance (April 2026)
- Use `resolve_rate_limiter()` from `BaseProvider` in constructor
- Add `tier` parameter to constructor and docstring
- Bump retry budget: 5 retries / 30s max wait (from 3/4s) for Z.AI load-shedding tolerance

### Rate Limiter (`rate_limiter.py`)
- Update `PROVIDER_DEFAULTS["zai"]` to conservative Lite-tier default (10 RPM / 50K TPM)

### CLI Helpers (`helpers.py`)
- Add `ZAI_TIER` env var reading in both explicit and auto-detect provider resolution paths

### LiteLLM Provider (`litellm.py`) -- behavioral change
- When `api_base` is set without an `api_key`, auto-adds `openai/` prefix to the model name for LiteLLM proxy routing
- Injects `sk-dummy` as a placeholder API key when `api_base` is set without `api_key` (LiteLLM/OpenAI SDK requires a non-empty key even for unauthenticated local proxies)
- **Downstream impact:** Users of LiteLLM-proxy with unauthenticated local endpoints will no longer need to manually add `openai/` prefix or pass a dummy key. Existing configs with explicit keys are unaffected.

### Tests (`test_zai_provider.py`, `test_litellm_provider.py`)
- 13 new Z.AI tier tests: tier creation, per-tier limits (lite/pro/max), case-insensitive matching, tier precedence over explicit limiter, invalid tier error, no-tier edge case, explicit limiter without tier, tier stored, and cross-provider empty tiers check
- 3 new LiteLLM tests: api_base adds `openai/` prefix, prefix deduplication, and dummy key injection

## Rate Limit Context

Z.AI support provided the following guidance (April 2026):

- **Lite**: 2-3 concurrent, lower tolerance -> 10 RPM / 50K TPM
- **Pro**: 5-8 concurrent, moderate tolerance -> 30 RPM / 150K TPM
- **Max**: 10-15 concurrent, highest tolerance -> 60 RPM / 300K TPM

Key facts:
- Limits are **aggregate across all models** (not per-model)
- Advanced models (GLM-5 family) consume **2-3x quota per prompt** due to reasoning tokens
- Limits are dynamic and load-dependent; these are conservative estimates
- Ref: https://docs.z.ai/devpack/usage-policy

## Configuration

```bash
# Required
export ZAI_API_KEY=***
# Optional -- defaults to conservative (Lite-equivalent)
export ZAI_TIER="pro"           # lite | pro | max
export ZAI_PLAN="coding"        # coding | general
export ZAI_THINKING="disabled"  # enabled | disabled
export ZAI_BASE_URL="..."       # override plan-based URL
```

## Test Plan

```bash
uv run pytest tests/unit/test_providers/test_zai_provider.py -v
# 34 passed (21 existing + 13 new tier tests)

uv run pytest tests/unit/test_providers/test_litellm_provider.py -v
# 10 passed (7 existing + 3 new api_base tests)
```

## PR Stack

| # | PR | Description | Status |
|---|-----|-------------|--------|
| 1 | **#82 -- Generic tier framework** | BaseProvider + resolve_rate_limiter() | Ready for review |
| 2 | **#83 -- Z.AI adopts the framework** (this PR) | RATE_LIMIT_TIERS + ZAI_TIER env var | Depends on #82 |
| 3 | **#84 -- MiniMax provider** | New provider using the framework | Depends on #82 |

## Related

- **Depends on: #82** (generic tier framework)
- Supersedes: #80 (monolithic PR, now split into layered approach)
- Ref: #68


packages/core/src/repowise/core/providers/llm/zai.py
```diff
@@ -0,0 +1,364 @@
...
+        if tier is not None and self._rate_limiter is not None:
+            log.info("zai.tier_rate_limiter", tier=tier.lower(), rpm=self._rate_limiter.config.requests_per_minute)
+
+        # Resolve base URL: explicit base_url > plan lookup
+        effective_base_url = base_url or _PLAN_BASE_URLS.get(plan, _PLAN_BASE_URLS["coding"])
+
+        # Normalize base URL for OpenAI SDK
+        effective_base_url = effective_base_url.rstrip("/")
+        if not effective_base_url.endswith("/v1"):
+            effective_base_url += "/v1"
```
swati510: Confirm this works against the live Z.AI API. Their OpenAI-compatible endpoint is /v4 as-is, the SDK tacks on /chat/completions. Adding /v1 here produces /v4/v1/chat/completions which I'd expect to 404.

swati510 (commented):
Nice work on the generic tier framework, it's the right shape. Two things worth fixing before merge:

1. ZAI_TIER_DEFAULTS in packages/core/src/repowise/core/rate_limiter.py duplicates the same values as ZAIProvider.RATE_LIMIT_TIERS. Nothing imports ZAI_TIER_DEFAULTS since resolve_rate_limiter reads from the class attribute. It's dead code waiting to drift. Drop it or have the provider read from it, one source of truth.

2. The base URL normalization in zai.py concerns me. _PLAN_BASE_URLS values end with /v4, then __init__ force-appends /v1, so the client hits https://api.z.ai/api/coding/paas/v4/v1/chat/completions. Has this been tested against the live API? Z.AI's OpenAI-compatible endpoint is /paas/v4 as-is, the SDK adds /chat/completions itself.

RaghavChamadiya:
Consolidating the open feedback so there's a single checklist to work from before this can land. The framework shape is good and worth keeping.

Blockers:

1. Base URL bug. The values in `_PLAN_BASE_URLS` already end with `/v4` (Z.AI's OpenAI-compatible endpoint), but `__init__` then unconditionally appends `/v1`, so the client ends up calling `https://api.z.ai/api/coding/paas/v4/v1/chat/completions` which is the wrong path. Either skip the `/v1` append when the configured base already ends in `/v4`, or drop the append entirely and document that callers pass full paths. Worth a smoke test against the live API to confirm the corrected path actually works for both the lite and coding plans.

2. `ZAI_TIER_DEFAULTS` in `packages/core/src/repowise/core/rate_limiter.py` duplicates the same values as `ZAIProvider.RATE_LIMIT_TIERS` and nothing imports it (`resolve_rate_limiter` reads off the class attribute). Either drop `ZAI_TIER_DEFAULTS` or have the provider read from it so there's a single source of truth.

3. Trailing-newline at EOF on `zai.py`, `test_zai_provider.py`, and `test_litellm_provider.py`. Will fail `ruff format --check`.

4. The PR also bundles a non-trivial LiteLLM behavioral change (auto `openai/` prefix and `sk-dummy` injection when `api_base` is set without a key). It's a useful fix but please call it out in the PR description so reviewers and downstream LiteLLM-proxy users know to look for it.

Once 1 and 2 are sorted I'll re-review and we can merge. #84 is queued behind this one.

Societus:
All four review items addressed:

1. **Base URL bug** -- Code was already correct (no `/v1` append), but tests were stale. Updated all 6 base URL assertions to match. Smoke-tested both `coding` and `general` plan endpoints against the live Z.AI API -- both return 401 (auth), not 404, confirming the paths are correct.

2. **`ZAI_TIER_DEFAULTS` dead code** -- Already absent from `rate_limiter.py`. Single source of truth is `ZAIProvider.RATE_LIMIT_TIERS`.

3. **Trailing newlines** -- Added missing EOF newlines on `zai.py`, `test_zai_provider.py`, and `test_litellm_provider.py`.

4. **LiteLLM behavioral change** -- Added a dedicated section to the PR description documenting the `openai/` prefix auto-addition and `sk-dummy` injection for unauthenticated local proxies, with downstream impact notes.

All 54 tests pass (34 Z.AI + 12 LiteLLM + 8 generic tier framework).

swati510:
A few things make this hard to merge as it stands.

- It conflicts with main. The branch is well behind, `PROVIDER_DEFAULTS` in `rate_limiter.py` has changed, and `openrouter`/`deepseek` were added, so it needs a rebase. The diff also bundles the #82 framework, so merging #82 first would double-apply `base.py`.
- `zai.py` records cost with a fire-and-forget `create_task(...)`. `openai.py` awaits this on purpose and documents why: a detached task can be mid-write when the loop tears down, which produces "event loop is closed" errors, and an unstored task can be garbage-collected, which silently drops the cost record. Matching the `openai.py` pattern avoids both.
- If `ZAI_API_KEY` is unset, `api_key` is `None`, and the OpenAI SDK falls back to `OPENAI_API_KEY` from the environment and sends it to `api.z.ai`. Validating a non-empty Z.AI key in the constructor and raising would prevent that.
- The retry predicate `retry_if_exception_type(ProviderError)` also catches 400/401/404, so a permanent auth failure retries 5 times over roughly 2 minutes. Narrowing it to 429 and 5xx would help.
- `stream_chat` never calls the rate limiter, so chat and agent mode run unthrottled, and that's where Z.AI's low aggregate limits matter most.
- The LiteLLM changes (`openai/` prefix, dummy key, menu entry) affect all existing proxy users, so they'd be cleaner as their own PR.

Could you split this into three PRs, each rebased on current main: the generic tier framework, the zai provider, and the litellm tweaks?

RaghavChamadiya:
@Societus this has been open since April and I owe you a straight answer rather than more silence.

The tier framework idea in here is good and the Z.AI request in #68 is still one we want. What has happened is that the branch has fallen a long way behind: it is based on a `registry.py` that predates openrouter, deepseek and kimi, so `PROVIDER_DEFAULTS` and the registry maps have all moved underneath it. GitHub reports real conflicts, and the code-health check is now failing on top of that.

There is also unanswered review from July: the fire-and-forget `create_task` for cost recording where `openai.py` awaits it, no non-empty check on `ZAI_API_KEY` so it silently falls back to `OPENAI_API_KEY`, a retry predicate broad enough to retry 400, 401 and 404, and `stream_chat` never touching the rate limiter. The suggestion at the time was to split this into three PRs, the generic tier framework, the zai provider, and the litellm changes, and I still think that is the right shape. The tier framework alone would probably go in quickly, because it is useful independent of any one provider.

I am not going to close this over you. If you want to pick it up, say so and I will keep it open and review promptly this time. If I do not hear anything in the next few weeks I will close it as stale, with #68 staying open so the Z.AI request itself does not get lost.

PR #89, opened 2026-04-20
feat(luau): add Luau/Roblox language support (issue #52)

CyanoTex:
## Summary

Draft skeleton for [issue #52](https://github.com/repowise-dev/repowise/issues/52). Adds a `luau` language tier so Rojo-synced Roblox projects produce a real dependency graph instead of the reported 21 disconnected nodes.

This PR wires the full language path end-to-end with the exception of Rojo `default.project.json`-aware resolution for `game.<Service>...` absolute instance paths, which is deliberately scoped to a follow-up.

### What's in scope
- `LanguageSpec(tag="luau", ...)` in the central registry, replacing the existing git-blame-only `lua` passthrough spec. One source of truth for both `.lua` and `.luau` (Luau's grammar is a strict superset of Lua 5.1).
- `"luau"` added to `LanguageTag` Literal in `models.py` so the extension flows through `FileTraverser`/`ASTParser` instead of being filtered out at `models.py:61`.
- `LANGUAGE_CONFIGS["luau"]` entry in `parser.py` mapping `function_declaration` → function and `type_definition` → type_alias.
- `queries/luau.scm` with captures for symbols, `require(...)` arguments, and call sites.
- `resolvers/luau.py` handling:
  - String literals: `require("rel/path")` — relative then stem fallback.
  - `script` / `script.Parent` chains — Roblox semantics where the first `Parent` is identity (the importer's container directory) and each subsequent `Parent` walks up.
  - `game.<Service>.Path...` — registers as external node; full Rojo-tree resolution is the next PR.
- `tree-sitter-luau>=1.2,<2` added to `pyproject.toml` dependencies.
- `docs/LANGUAGE_SUPPORT.md` updated to document the new tier.
- Unit tests: `TestLuauParser` in `test_parser.py`, `test_luau_resolver.py` for the resolver. Absolute instance-path resolution is pinned with an `xfail(strict=True)` test documenting the expected end state.

### What's deliberately NOT in scope
- Reading `default.project.json` to map service subtrees to filesystem directories. The shape is documented in the resolver's module docstring and the failing `xfail` test. Planned home: `core/ingestion/dynamic_hints/rojo.py`.
- Bindings / heritage extractors. Luau modules typically don't use class inheritance; revisit if needed.
- `.lua` files in pure-vanilla-Lua projects (not Rojo) will now be AST-parsed where before they were git-blame-only. This is aligned with issue #52's stated goal, but flagging explicitly since it's a behavior change for non-Roblox Lua repos.
- The pre-existing broken `repowise.core.ingestion.parsers` reference in `pyproject.toml`'s `packages` list (unrelated to this PR; surfaced in PR #75 review). Left untouched per the maintainer's preference to keep that fix in its own PR.

### Verification
```
uv run pytest tests/unit/ingestion/ -q
# 227 passed, 1 xfailed
uv run ruff check packages/core/.../luau.py .../registry.py .../parser.py .../models.py .../resolvers/__init__.py tests/.../test_luau*.py tests/.../test_parser.py
# All checks passed
```

### Open questions / risks
1. **tree-sitter-luau ABI pin.** The published wheel's `[core]` extra pins `tree-sitter~=0.22`, while repowise requires `>=0.23,<1`. Installing `tree-sitter-luau` without the `[core]` extra (which is what the standard dep resolution does) worked locally against tree-sitter 0.23.x — verified by `test_parses_require_imports` exercising the loader. Worth a CI run against each matrix slot to confirm.
2. **`.lua` ownership.** Taking `.lua` into the `luau` spec is a judgment call. Pure-Lua projects will now get AST-parsed; vanilla Lua parses cleanly as a Luau subset. Open to splitting the extensions if preferred.
3. **Resolver for `game.<Service>...`.** Currently returns an external node rather than silently matching the nearest-stem file. This matches the maintainer's "no silent fallback" norm from PRs #56 / #83, but means these edges won't land in the graph until the Rojo reader ships.

## Test plan
- [x] `uv run pytest tests/unit/ingestion/ -q` — green (230 passed, 2 xfailed)
- [x] `uv run ruff check` on changed files — clean
- [x] `uv run ruff format --check` on changed files — clean
- [x] Run against a real Rojo project locally via `repowise init` — done against a private Rojo project + upstream luau-lang/luau (see comment for numbers); Rojo-tree resolution remains pending the follow-up PR
- [ ] Confirm CI ABI compatibility across the matrix


CyanoTex:
Three follow-up commits pushed on top of `6e5c199` — all in the spirit of test-plan box 4 (running `init` against real Luau projects). Numbers and rationale below.

### Commits

- **`fix(luau): string-literal requires never entered the resolver`** — `parser.py:705` strips quotes from the captured `@import.module` text before calling the resolver, but the resolver's string-literal branch keyed off `arg.startswith('"')`. Result: every `require("./path")` fell through to the external-node fallback in production. The unit tests passed because they fed explicitly-quoted strings, which the production parser never emits. Fix detects literal paths by elimination (not `script.*`, not `game.*`, not `@alias`) and rewrites the tests to mirror what the parser actually produces.
- **`feat(luau): resolve :WaitForChild / :FindFirstChild require chains`** — On real Rojo code, `script.Parent:WaitForChild("Foo")` is vastly more common than the bare dot-chain form the original `_SCRIPT_RELATIVE` regex matched (it's the race-safe idiom against Rojo's filesystem-sync ordering). Normalizes both method-call forms to the dot shape before regex matching; the path-walking logic in `_resolve_script_relative` is reused unchanged. Variable-aliased chains (`SharedFolder:WaitForChild(...)`) still go external — that needs lightweight flow tracking and is deliberately out of scope.
- **`test(luau): pin .luaurc @alias resolution as a follow-up (xfail)`** — Second known follow-up alongside the existing Rojo xfail. `.luaurc` is Luau's official require-by-string alias mechanism (parallel to tsconfig paths). Needs a `.luaurc` reader; shape is documented in the xfail.

### Runtime-verified numbers

Ran `repowise init --index-only` against two Luau codebases before and after:

**Private Rojo project (161 `.luau` files across `src/client`, `src/server`, `src/shared`):**

| | Before | After |
|---|---|---|
| Internal (luau→luau) import edges | 13 (2.8%) | **333 (72.2%)** |
| External edges | 448 | 128 |

~26× more resolutions. The 128 remaining externals are all known-unsupported shapes: variable-aliased chains (out of scope here), and service-based lookups (`ReplicatedStorage:WaitForChild(...)`, `game:GetService(...)`) which are in the same family as the existing Rojo follow-up.

**Upstream `luau-lang/luau` (237 Luau files, mostly `tests/` and `bench/`):**

| | Before | After |
|---|---|---|
| Internal (luau→luau) import edges | 1 (0.4%) | **113 (45.0%)** |
| External edges | 250 | 138 |

~113× more resolutions. The 138 remaining break down as:
- 107× `external:script.Parent.bench_support` — correct behavior; these are the intentionally-non-matching arm of the `require(script.Parent.bench_support) or require("../bench_support")` or-chain idiom in `bench/*`. The `../bench_support` branch now resolves (was previously 0 internal, now all 107), so the pattern works end-to-end.
- 24× `@alias` references — the `.luaurc` follow-up (pinned as xfail by the third commit).
- ~7× scattered test-fixture edge cases.

### Suite

```
uv run pytest tests/unit/ingestion/ -q
# 230 passed, 2 xfailed
```

Both xfails are the strict-mode documentation-only tests pinning the two deferred follow-ups (Rojo project-tree resolution, `.luaurc` alias resolution).

### One side-finding worth flagging separately

While running `init` to generate the numbers above, I hit what looked like a serious data-loss bug in `packages/cli/src/repowise/cli/mcp_config.py` unrelated to this PR — JSON parse errors on an existing `~/.claude/settings.json` silently discard the user's config. Filed as #93 with repro and recommended fix, so it's out of this PR's scope and you can triage independently.

-Opus 4.7

RaghavChamadiya (approved):
Pattern matches the existing 14-language work cleanly. The :WaitForChild / :FindFirstChild normalization to a single regex path is nice, and pinning the two unsupported require shapes as xfail(strict=True) is the right call to keep them on the radar without blocking. Approving once CI is green.

PR #97, opened 2026-04-26
fix(build): correct setuptools packages list

RaghavChamadiya:
## Summary

The `[tool.setuptools].packages` list got out of sync with the actual source tree:

- It still listed `repowise.core.ingestion.parsers`, which no longer exists.
- It was missing `repowise.core.ingestion.extractors`, `languages`, and `resolvers`, which were added during the recent language-support refactor.

This caused `setuptools.build_meta.build_editable` to fail with `package directory does not exist`, which is what has been turning every PR's CI red for the past couple of weeks.

## What changed

Removed the stale `parsers` entry and added the three missing subpackages.

## Verification

Built locally with `uv build --wheel` and confirmed the resulting wheel ships all four ingestion subpackages:

```
repowise/core/ingestion/__init__.py
repowise/core/ingestion/dynamic_hints/__init__.py
repowise/core/ingestion/extractors/__init__.py
repowise/core/ingestion/languages/__init__.py
repowise/core/ingestion/resolvers/__init__.py
```

Once this lands, the open PRs can be rebased and CI will actually run.


swati510 (approved):
lgtm

PR #99, opened 2026-04-26
feat: output language support for wiki generation

RaghavChamadiya:
Supersedes #75 by @maxfrank76. Picking up the long-running output-language work, rebased onto current main and polished so we can land it.

## What this adds

- `language` field on `GenerationConfig` (default `en`).
- `--language` is read from `.repowise/config.yaml` and threaded through `init` and `update` to `PageGenerator`.
- `PageGenerator` prepends a small system-prompt instruction in the configured language while keeping code, paths, and symbol names untranslated.
- 15-language code-to-name table (en, ru, es, fr, de, zh, ja, ko, it, pt, nl, pl, tr, ar, hi). Unknown codes log a warning and fall back to English.
- Language code is included in the generation cache key so different output languages do not collide.
- Input is sanitized (alphanumeric + underscore) so a config-supplied language string cannot inject newlines or extra instructions into the system prompt.

## Polish on top of #75

- Restored `package-lock.json next` pin to match main (`~15.5.15`).
- Re-grouped `_LANGUAGE_NAMES` with module-level constants and restored the missing blank line before the per-type generation methods section.
- Extracted the language-instruction construction into a small `_build_system_prompt` helper so `_call_provider` stays focused on caching plus dispatch.
- Switched the unknown-language warning from an f-string to structlog keyword args, and stopped logging the fallback when the configured language is already `en`.
- Made `run_generation`'s new `generation_config` kwarg optional so `run_pipeline` and the workspace update path keep working when a config is not threaded through.
- Removed a broken `generation_config=config` call inside `run_pipeline` that referenced an undefined local.
- Hoisted `dataclasses.replace` to the module-level imports.

## Tests

Added five unit tests in `tests/unit/generation/test_page_generator.py`:

- English passthrough leaves the prompt untouched.
- Non-English prepends the named-language instruction.
- Unknown language code falls back silently to English.
- Newline-injection in the language string is stripped and the prompt is not poisoned.
- Cache keys differ across output languages.

Credit to @maxfrank76 for the design and the bulk of the implementation.


swati510 (approved):
lgtm

PR #84, opened 2026-04-14
feat(minimax): add MiniMax provider with tier-aware rate limiting

Societus:
## Summary

Add MiniMax as a built-in LLM provider using the generic tier framework from #82.

This PR is a straightforward application of the same pattern as #83. Both MiniMax and Z.AI are OpenAI-compatible APIs with subscription tiers and built-in reasoning models. The generic tier framework made this provider almost mechanical to implement -- the only provider-specific code is the model names, the `reasoning_split` parameter vs Z.AI's `thinking` toggle, and the tier definitions.

**Depends on: #82** (generic tier framework -- merge that first)

## Why This Was Inconsequential

MiniMax shares the same architectural profile as Z.AI:
- OpenAI-compatible API at `https://api.minimax.io/v1`
- Bearer token auth using the same `openai` SDK
- Reasoning models with a thinking separator parameter
- Published subscription tiers with conservative rate limits

The generic framework from #82 eliminated all boilerplate for tier resolution. Adding MiniMax was just: define `RATE_LIMIT_TIERS`, set the base URL, and pick the reasoning parameter name. Everything else is inherited.

### New: MiniMax Provider (`minimax.py`)
- `RATE_LIMIT_TIERS` with Starter/Plus/Max/Ultra configs from published limits
- `resolve_rate_limiter()` from BaseProvider (zero custom tier code)
- `reasoning_split=True` by default (separates thinking from content)
- Retry budget: 5 retries / 30s max wait
- Models: MiniMax-M2.7 (default), M2.7-highspeed, M2.5, M2.5-highspeed, M2.1, M2.1-highspeed, M2

### Registry (`registry.py`)
- Register `minimax` -> `MiniMaxProvider` with `openai` package hint

### Rate Limiter (`rate_limiter.py`)
- `PROVIDER_DEFAULTS["minimax"]` = Starter-tier conservative (5 RPM / 25K TPM)

### CLI Helpers (`helpers.py`)
- `MINIMAX_API_KEY`, `MINIMAX_BASE_URL`, `MINIMAX_REASONING_SPLIT`, `MINIMAX_TIER` env vars
- Auto-detect from `MINIMAX_API_KEY`
- Added to provider validation list

### Tests (`test_minimax_provider.py`)
- 30 tests: constructor, tier resolution (4 tiers + edge cases), generate with mock, stream_chat, reasoning_split, registry integration

## Rate Limit Tiers

From published MiniMax docs (5-hour rolling window):

[...]

PR #107, opened 2026-04-28
Fix get ans tool

swati510:
Upgrade get_answer tool with question-aware symbol promotion: extracts identifiers 
  from the question, promotes matching symbols to the top of retrieval results, and
  attaches source-body excerpts so the LLM can answer specific-method questions without
   hedging                                                  
  - Add hedge detection to downgrade confidence when the LLM admits it can't answer,   
  and bypass stale hedged cache entries for re-synthesis                               
  - Add identifier-citation gate to downgrade confidence when named symbols aren't
  found in top retrieval hits                                                          
                                                            
  Related Issues                                                                       
                                                            
  Fixes #84

  Test Plan

  - Tests pass (pytest)
  - Lint passes (ruff check .)
  - Web build passes (npm run build) (if frontend changes)

  Checklist                                                                            
   
  - My code follows the project's code style                                           
  - I have added tests for new functionality                
  - All existing tests still pass
  - I have updated documentation if needed

### DIFF
--- a/packages/cli/src/repowise/cli/main.py
+++ b/packages/cli/src/repowise/cli/main.py
@@ -47,3 +47,7 @@ cli.add_command(serve_command)
 cli.add_command(mcp_command)
 cli.add_command(reindex_command)
 cli.add_command(workspace_group)
+
+
+if __name__ == "__main__":
+    cli()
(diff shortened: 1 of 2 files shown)


RaghavChamadiya (approved):
LGTM

PR #111, opened 2026-04-30
fix(workspace): show contract summary when contracts exist but no cross-repo links

j-j-m:
## Summary

\`CrossRepoEnricher.has_contract_data\` returns True only when \`_contract_links\` is non-empty. That gates the workspace summary endpoint (\`/api/workspace.contract_summary\`) — so for any workspace where the link extractor can't match providers to consumers, the API returns \`null\` and the dashboard widget renders 0 even though contracts were extracted successfully.

This is the common case for any Go-heavy workspace: the HTTP client-side extractor doesn't currently recognize Go HTTP client patterns (\`http.NewRequest\`, \`net/http\`, etc.), so a workspace with both Go HTTP servers and Go HTTP clients gets \`total_contracts: 50\`, \`total_links: 0\`, and a UI tile that says \"Contracts Detected: 0\".

The 0 is misleading: it tells the user that contract extraction failed, when in reality only the cross-repo MATCHING failed. The detected contracts are still useful — they show up in per-repo views and are queryable via \`/api/workspace/contracts\`.

## Repro

```bash
mkdir my-workspace && cd my-workspace
# add 2+ git repos with HTTP-style code (e.g., Go gateway + Go CLI)
repowise init . --index-only --yes
curl -s http://localhost:7337/api/workspace | jq '.contract_summary'
# null  ← bug
curl -s http://localhost:7337/api/workspace/contracts | jq '.total_contracts'
# 50    ← but contracts ARE there
```

## Fix

```python
@property
def has_contract_data(self) -> bool:
    return bool(self._contracts or self._contract_links)  # was: bool(self._contract_links)
```

After this, \`/api/workspace.contract_summary\` is populated whenever any contracts exist, and the dashboard tile shows the real count. \`CONTRACT LINKS\` (a separate tile) correctly continues to show 0 when there are no cross-repo matches.

## Verification

3-repo workspace (operator + operator-cli + genui-schema), 50 HTTP + 12 gRPC contracts indexed, 0 cross-repo links. Before: dashboard shows \"0 detected\". After: \"62 detected · 50 http, 12 grpc\" with the type breakdown shown as subtitle.

## Diff

+2 / -2, single property body. No tests added — there's no existing test covering \`has_contract_data\`'s gating logic, and adding one would mean stubbing the whole enricher. Happy to add one if you'd like.

### DIFF
--- a/packages/server/src/repowise/server/mcp_server/_enrichment.py
+++ b/packages/server/src/repowise/server/mcp_server/_enrichment.py
@@ -191,8 +191,8 @@ class CrossRepoEnricher:
 
     @property
     def has_contract_data(self) -> bool:
-        """True if contract links are available."""
-        return bool(self._contract_links)
+        """True if contracts or contract links are available."""
+        return bool(self._contracts or self._contract_links)
 
     def get_cross_repo_partners(
         self, repo_alias: str, file_path: str


RaghavChamadiya (commented):
@
Thanks for chasing this down, the repro and the trace through to the dashboard tile made it very easy to verify. Confirmed the bug shape and walked the five call sites that depend on `has_contract_data`, all of them are safe under the wider gate. `tool_context.py:1143` and `tool_risk.py:484` already re-guard with `if provider_links or consumer_links:` before writing anything, so opening the outer gate just costs two empty lookups per file. The two `tool_overview.py` sites and `routers/workspace.py:151` go straight into `get_contract_summary()` which already handles the links-empty case correctly.

Two small asks before merge:

1. Take you up on the test offer. A fixture with contracts present and `contract_links: []` asserting `has_contract_data is True` would lock the regression in. The existing `contracts_json` fixture happens to carry both, so nothing in the suite covers exactly the scenario this PR fixes. Five lines, would prevent a future refactor from quietly flipping it back.

2. There is a sibling property right above the one you changed with the same shape of bug. `has_data` at `_enrichment.py:190` still reads `bool(self._co_changes or self._package_deps or self._contract_links)`. After your fix, a workspace with contracts but no co-changes, no package deps, no links would have `has_contract_data` True but `has_data` False. That matters because `tool_overview.py:92` gates the entire cross_repo_topology block on `has_data`, and the contract summary check at line 94 lives inside that block. So in the contracts-only case the summary still would not surface in `get_overview` output despite your fix landing. Either fold `_contracts` into `has_data` as well, or explicitly leave `tool_overview` out of scope and we track it separately.

Neither is a hard blocker. Happy to merge as-is if you would rather keep this PR strictly to the property you came in for and file (2) as its own follow-up.
@

Seanathon (approved):
lgtm

swati510 (approved):

PR #146, opened 2026-05-08
fix(chat): route /api/repos/{repo_id}/chat/messages by workspace_sessions

### LINKED ISSUE #145: Chat 404 for non-primary repos in workspace mode
# Chat 404 for non-primary repos in workspace mode

## What happens

In workspace mode, every non-primary repo returns
`404 Repository <id> not found` when sending a chat message:

```bash
curl -X POST \
  http://localhost:7337/api/repos/<non-primary-id>/chat/messages \
  -H "Content-Type: application/json" \
  -d '{"message":"hi"}'
# {"detail":"Repository <non-primary-id> not found"}
```

The same `<non-primary-id>` appears in both `GET /api/repos` and
`GET /api/workspace.repos[].repo_id`. Only the **primary** repo's
chat works.

## What I expected

Chat works for every repo listed in `GET /api/repos`, regardless
of whether it's the primary or not.

## Root cause (proposal — happy to be redirected)

`POST /api/repos/{repo_id}/chat/messages` in
`packages/server/src/repowise/server/routers/chat.py` resolves the
repo against `request.app.state.session_factory`:

```python
factory = request.app.state.session_factory
repo_name, repo_path = await _get_repo_info(factory, repo_id)
```

In workspace mode, non-primary repos live in their own `wiki.db`
registered under `app.state.workspace_sessions[repo_id]`. The
primary `session_factory` doesn't see those rows, so the lookup
404s.

The conversation endpoints (`GET /chat/conversations`, etc.) work
correctly because they use `Depends(get_db_session)`, which already
encodes the workspace routing in `deps.py`. `chat_messages` can't
use that dep — it needs the *factory* (not a single session) so it
can open multiple sessions over the SSE lifetime — so it was
hand-wiring `session_factory` and missing the workspace branch.

The same routing helper exists privately in
`routers/repos.py::_resolve_repo_session_factory`, complete with a
docstring explicitly noting this case.

## Suggested fix

Lift the workspace-routing logic out of `deps.get_db_session` into
two helpers:

- `resolve_session_factory(app_state, repo_id)` — pure function,
  same logic as the existing private helper in `routers/repos.py`.
- `resolve_request_session_factory(request)` — request-scoped
  sibling that reads `repo_id` from path/query params.

Then `chat_messages` calls `resolve_request_session_factory(request)`
instead of `request.app.state.session_factory` directly, and
`routers/repos.py::_resolve_repo_session_factory` becomes a
one-line alias around the shared helper (kills duplication).

I have a patch ready that takes this shape and includes a
regression test. Happy to open the PR if this direction is
acceptable, or rework if you'd prefer a different design.

## Environment

- repowise 0.5.1 (commit `cdd38af` on `main`)
- 9-repo workspace, primary = `auth-provider`
- Python 3.12, SQLite + aiosqlite

enyr-srl:
# fix(chat): route `/api/repos/{repo_id}/chat/messages` by workspace_sessions

## Summary

In workspace mode, every non-primary repo's chat returns
`404 Repository <id> not found` even though the same id is listed
in both `GET /api/repos` and `GET /api/workspace.repos[].repo_id`.
The chat router was the only `/api/repos/{repo_id}/...` endpoint
not honouring `app.state.workspace_sessions`.

The conversation endpoints (`GET /chat/conversations`, etc.)
already work correctly because they use `Depends(get_db_session)`,
which routes by `repo_id` automatically. `chat_messages` couldn't
use that dep — it needs the *factory* (not a single session) so it
can open multiple sessions over the SSE lifetime — so it was
hand-wiring `request.app.state.session_factory` and missing the
workspace branch.

## Fix

- Extract the factory-resolution logic from `get_db_session` into
  two reusable helpers in `deps.py`:
    - `resolve_session_factory(app_state, repo_id)` — pure
      function, same logic as the existing private helper in
      `routers/repos.py`.
    - `resolve_request_session_factory(request)` — request-scoped
      sibling that reads `repo_id` from path/query params (mirrors
      what `get_db_session` encodes).
- `chat_messages` calls `resolve_request_session_factory(request)`
  instead of `request.app.state.session_factory` directly.
- `routers/repos.py::_resolve_repo_session_factory` is now a
  one-line alias around the shared helper, removing the duplicate
  logic.

## Tests

New regression test `tests/unit/server/test_chat_workspace.py`:

- `test_chat_messages_resolves_non_primary_repo_in_workspace_mode`
  — POST to a non-primary `repo_id` no longer returns 404.
  Verified to fail on the unpatched source with the exact
  `Repository <id> not found` message.
- `test_chat_messages_still_finds_primary_repo` — the
  single-factory fallback still works when `workspace_sessions` is
  empty (single-repo mode and the primary repo of a workspace).

All 126 existing `tests/unit/server/` tests pass with the patch
(verified locally).

## Repro (against an unpatched workspace deployment)

```bash
curl -X POST \
  http://localhost:7337/api/repos/<non-primary-id>/chat/messages \
  -H "Content-Type: application/json" \
  -d '{"message":"hi"}'
# {"detail":"Repository <non-primary-id> not found"}
```

The same `<non-primary-id>` appears in
`GET /api/workspace.repos[].repo_id`.

## Test plan

- [x] New regression test fails on unpatched source with the exact
  user-visible 404 message
- [x] New regression test passes after the fix
- [x] All existing `tests/unit/server/` tests still pass
- [ ] Manual smoke test against a 2-repo workspace from a fresh
  install (reviewer)

Closes #145.

### DIFF
--- a/packages/server/src/repowise/server/deps.py
+++ b/packages/server/src/repowise/server/deps.py
@@ -36,23 +36,42 @@ if _API_KEY is None and _REPOWISE_HOST in ("0.0.0.0", "::"):
     )
 
 
+def resolve_session_factory(app_state, repo_id: str | None):
+    """Pick the session factory whose database contains ``repo_id``.
+
+    In workspace mode each repo has its own ``wiki.db`` registered under
+    ``app_state.workspace_sessions[repo_id]``. The primary
+    ``app_state.session_factory`` does NOT see those rows. When ``repo_id``
+    is ``None`` or unknown, falls back to the primary factory (single-repo
+    mode, or the primary repo of a workspace).
+    """
+    if repo_id:
+        ws_sessions = getattr(app_state, "workspace_sessions", None)
+        if ws_sessions and repo_id in ws_sessions:
+            return ws_sessions[repo_id]
+    return app_state.session_factory
+
+
+def resolve_request_session_factory(request: Request):
+    """Return the session factory for the active route's ``repo_id``.
+
+    Reads ``repo_id`` from the path (e.g. ``/api/repos/{repo_id}/...``)
+    or query string (e.g. ``/api/pages?repo_id=xxx``), then delegates to
+    :func:`resolve_session_factory`. Used by streaming endpoints that
+    can't take a single ``Depends(get_db_session)`` because they need to
+    open multiple sessions over the request lifetime.
+    """
+    repo_id = request.path_params.get("repo_id") or request.query_params.get("repo_id")
+    return resolve_session_factory(request.app.state, repo_id)
+
+
 async def get_db_session(request: Request) -> AsyncGenerator[AsyncSession, None]:
     """Yield an async DB session with auto-commit on success, rollback on error.
 
     In workspace mode, routes to the correct repo's DB based on the
     ``repo_id`` path parameter when a matching session factory exists.
     """
-    factory = request.app.state.session_factory
-
-    # In workspace mode, check if the request targets a specific repo DB
-    # Check both path params (e.g. /api/repos/{repo_id}/stats) and
-    # query params (e.g. /api/pages?repo_id=xxx) for the repo_id
-    repo_id = request.path_params.get("repo_id") or request.query_params.get("repo_id")
-    if repo_id:
-        ws_sessions = getattr(request.app.state, "workspace_sessions", None)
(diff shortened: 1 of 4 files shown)


RaghavChamadiya (approved):
lgtm

PR #163, opened 2026-05-10
fix: decouple embedding concurrency from generation

### LINKED ISSUE #136: Performance: page_generator semaphore serializes LLM and embedding, throttling effective concurrency
## Summary

`page_generator.py:guarded_named` holds the same semaphore over both the LLM call and `embed_and_upsert`. When the embedding endpoint is single-slot or slow (e.g. a local Ollama instance with one model loaded), all in-flight tasks pile up on embed and the LLM concurrency the user configured is effectively ignored.

## Reproduction

- Run `repowise init <repo> --concurrency 20 --provider openai --embedder openai` with the embedding endpoint pointing to an Ollama server (`OLLAMA_BASE_URL` or equivalent) running a 4096-dim embedding model on a host with limited parallelism.
- Each individual page generation takes ~30s for the LLM call alone.
- Each `embed_and_upsert` takes 5-15s on the same Ollama endpoint when the page content is 2-5K chars.
- Observed steady-state: ~1-2 pages committed per minute (instead of the expected ~30 with concurrency 20).

## Root cause

`packages/core/src/repowise/core/generation/page_generator.py:guarded_named`:

```python
async def guarded_named(page_id, coro):
    async with semaphore:
        result = await coro                      # LLM ~30s
        ...
        await self._vector_store.embed_and_upsert(...)  # embed ~5-15s, also inside sema
        ...
```

Each semaphore slot is held for `LLM + embed` (~35-45s total). With 20 slots and an embed endpoint that can only serve one request at a time, the semaphore can never refill until the embed finishes serially across all concurrent tasks.

## Suggested fix

Hold the semaphore only over the LLM call, and run `embed_and_upsert` outside the sema (or behind an independent `embed_semaphore` whose size matches the embedding endpoint's actual parallelism):

```python
async def guarded_named(page_id, coro):
    async with semaphore:
        result = await coro
    if isinstance(result, GeneratedPage):
        try:
            await self._vector_store.embed_and_upsert(...)
        except Exception as e:
            log.debug("rag.embed_failed", ...)
    ...
```

Locally we patched it this way and observed real throughput (well past 20 LLM calls/min) instead of being clamped by the embed endpoint.

Happy to send a PR if a maintainer green-lights this approach.

zxyasfas:
## Summary

Fixes #136 by splitting page generation concurrency from vector-store embedding concurrency.

`PageGenerator.generate_all()` now releases the LLM generation semaphore as soon as a page is generated, then runs `embed_and_upsert()` behind a separate `embed_concurrency` semaphore. The default resolves to `max_concurrency`, so existing configurations keep the same bounded behavior unless they opt into a different embedding limit.

## Why

On slow or single-slot embedding endpoints, the old flow kept the LLM semaphore held while embedding completed. That meant embedding latency could serialize later LLM calls in the same generation level, reducing effective generation throughput.

## Tests

- `python -m pytest tests/unit/generation/test_models.py tests/integration/test_generation_pipeline.py::test_embedding_latency_does_not_gate_llm_concurrency -q`
- `python -m pytest tests/unit/generation/test_page_generator.py tests/unit/generation/test_models.py tests/integration/test_generation_pipeline.py::test_embedding_latency_does_not_gate_llm_concurrency -q`
- `python -m py_compile packages/core/src/repowise/core/generation/models.py packages/core/src/repowise/core/generation/page_generator.py tests/integration/test_generation_pipeline.py tests/unit/generation/test_models.py`
- `ruff format packages/core/src/repowise/core/generation/models.py packages/core/src/repowise/core/generation/page_generator.py tests/unit/generation/test_models.py tests/integration/test_generation_pipeline.py`
- `ruff check tests/unit/generation/test_models.py tests/integration/test_generation_pipeline.py`

Note: `ruff check` on the full changed production files still reports pre-existing lint in `page_generator.py` / `models.py` outside this change (`RUF002/RUF003` en-dash comments and an existing unused local). The new test files pass ruff check.

### DIFF
--- a/packages/core/src/repowise/core/generation/models.py
+++ b/packages/core/src/repowise/core/generation/models.py
@@ -62,6 +62,8 @@ class GenerationConfig:
         temperature:              Sampling temperature (0.3 for consistent docs).
         token_budget:             Context tokens fed to LLM (not output).
         max_concurrency:          asyncio.Semaphore size for parallel calls.
+        embed_concurrency:        asyncio.Semaphore size for vector-store writes.
+                                  Defaults to max_concurrency.
         cache_enabled:            In-memory SHA256 prompt deduplication.
         staleness_threshold_days: Days before a page is considered stale.
         expiry_threshold_days:    Days before a page is considered expired.
@@ -73,6 +75,7 @@ class GenerationConfig:
     temperature: float = 0.3
     token_budget: int = 48000
     max_concurrency: int = 5
+    embed_concurrency: int | None = None
     cache_enabled: bool = True
     staleness_threshold_days: int = 7
     expiry_threshold_days: int = 30
@@ -83,6 +86,11 @@ class GenerationConfig:
     jobs_dir: str = ".repowise/jobs"
     large_file_source_pct: float = 0.4  # use structural summary when source tokens > budget * this
     language: str = "en"
+
+    def __post_init__(self) -> None:
+        if self.embed_concurrency is None:
+            object.__setattr__(self, "embed_concurrency", self.max_concurrency)
+
 
 # ---------------------------------------------------------------------------
 # GeneratedPage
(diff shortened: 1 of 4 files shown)


RaghavChamadiya (approved):
lgtm

PR #195, opened 2026-05-14
Add Codex CLI provider and project integration

Selene29:
## Summary

- add local Codex CLI LLM provider support through authenticated `codex exec` sessions, including provider registry wiring, `reasoning`-compatible generation calls, and zero-cost subscription usage tracking
- add project-local Codex setup for MCP config, lifecycle hooks, plugin metadata, skills, and managed `AGENTS.md` generation
- install Codex hooks through the import-isolated `repowise-augment` entry point and enable the current Codex `[features] hooks = true` flag
- update CLI commands, docs, examples, and website pages for Codex setup

## Validation

- `git diff --check`
- `uv run pytest tests/providers/test_registry.py tests/unit/cli/test_commands.py tests/unit/cli/test_helpers.py tests/unit/cli/test_mcp_config.py tests/unit/cli/test_augment_cmd.py tests/unit/cli/test_editor_files.py tests/unit/cli/test_claude_md_prompt.py tests/unit/cli/test_cost_estimator.py tests/unit/generation/test_agents_md.py tests/unit/generation/test_cost_tracker.py tests/unit/test_providers/test_codex_cli_provider.py tests/unit/test_providers/test_cost_estimator.py tests/unit/cli/test_codex_plugin.py`
- `codex exec --ephemeral --sandbox read-only --json --cd . "Return exactly OK."`
- `codex features list | rg '^hooks\s+'`

### DIFF
--- a/.agents/plugins/marketplace.json
+++ b/.agents/plugins/marketplace.json
@@ -1,0 +1,20 @@
+{
+  "name": "repowise",
+  "interface": {
+    "displayName": "Repowise"
+  },
+  "plugins": [
+    {
+      "name": "codex",
+      "source": {
+        "source": "local",
+        "path": "./plugins/codex"
+      },
+      "policy": {
+        "installation": "AVAILABLE",
+        "authentication": "ON_INSTALL"
+      },
+      "category": "Productivity"
+    }
+  ]
+}
--- a/README.md
+++ b/README.md
@@ -597,7 +597,7 @@ repowise reindex                  # rebuild vector store (no LLM calls)
 provider: anthropic               # anthropic | openai | openrouter | gemini | deepseek | ollama | litellm | codex_cli | mock
 model: claude-sonnet-4-5
 embedding_model: voyage-3
-reasoning: auto                   # auto | off | minimal
+reasoning: auto                   # auto | off/none | minimal | low | medium | high | xhigh | max
 
 editor_files:
   claude_md: true
@@ -617,8 +617,9 @@ maintenance:
 ```
 
 `reasoning` applies to documentation generation. `auto` preserves provider
-defaults; explicit `off` / `minimal` modes are translated only by providers and
-models that support them, otherwise repowise fails before making an API call.
+defaults; explicit modes (`off`/`none`, `minimal`, `low`, `medium`, `high`,
+`xhigh`, `max`) are translated only by providers and models that support them,
+otherwise repowise fails before making an API call.
 
 Full configuration reference: [docs/CONFIG.md](docs/CONFIG.md)
 
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ Five intelligence layers. Nine MCP tools. Multi-repo workspaces. Auto-sync hooks
 
 Your AI coding agent reads files. It doesn't know which ones change together, which ones are dead, or why they were built the way they were. It has the source code and no memory of how the codebase got there.
 
-repowise fixes that. It indexes your codebase into five intelligence layers — dependency graph, git history, auto-generated documentation, architectural decisions, and code health — and exposes them to Claude Code (and any MCP-compatible AI agent) through nine precisely designed tools. Multi-repo? Initialize a workspace and get cross-repo co-change detection, API contract extraction, and federated MCP queries across all your services. **The result: fewer tool calls, fewer file reads, and lower cost per query — at comparable answer quality.**
+repowise fixes that. It indexes your codebase into five intelligence layers — dependency graph, git history, auto-generated documentation, architectural decisions, and code health — and exposes them to Claude Code, Codex, and any MCP-compatible AI agent through nine precisely designed tools. Multi-repo? Initialize a workspace and get cross-repo co-change detection, API contract extraction, and federated MCP queries across all your services. **The result: fewer tool calls, fewer file reads, and lower cost per query — at comparable answer quality.**
 
 The result: your agent answers *"why does auth work this way?"* instead of *"here is what auth.ts contains."*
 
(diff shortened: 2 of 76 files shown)


RaghavChamadiya:
Thanks for this - it's a substantial, thoughtful contribution. The provider itself is well-structured (argv-based subprocess with no shell injection, robust JSONL parsing that tolerates  non-JSON noise, clean async semaphore, atomic writes), AGENTS.md reuses the proven `BaseEditorFileGenerator` marker-merge, and `repowise-augment` import isolation is genuinely preserved.
  Test coverage for the happy path and most provider failure modes is solid.

  I'd like to merge this, but a couple of items would regress existing users so I'd like them addressed first. Everything else can be follow-up.

  ### Blocking

  **1. Augment hook may double-fire on Claude Code.** `_handle_codex_context_event` runs on `SessionStart`/`UserPromptSubmit` in `augment_cmd.py`, but Claude Code emits those same events. Any
   existing Claude Code user with `repowise-augment` wired up will start getting Codex-flavored banners after this lands. Could you either gate on a Codex-specific payload marker, or make
  sure the Claude install path never registers those matchers? A regression test asserting Claude's payload shape still works as before would help.

  **2. Auto-install side effect on `repowise init`.** In `init_cmd.py:184-201`, when `codex_setup is None and installed and logged_in`, we silently write `.codex/config.toml`,
  `.codex/hooks.json`, and flip `features.hooks = true`. That's a real filesystem side effect on users who didn't ask for Codex setup. Could we prompt interactively in a TTY (matching the    
  existing Claude hook prompt pattern), or at minimum print a clear opt-out hint?

  **3. `find_repowise_repo_root` stops on any `.git`** (`helpers.py:69-77`). Nested git directories (`node_modules/.git`, vendored submodules, worktree `.git` files) cause it to return `None`
   prematurely. Since `plugins/codex/.mcp.json` sets no `cwd` and relies entirely on this walk, the plugin will silently fail in those repos. Walk to filesystem root and only check for       
  `.repowise`.

  **4. `is_codex_logged_in()` string-matches `"logged in"`** (`mcp_config.py:54-57`). Brittle across Codex CLI versions and locales. Suggest falling back to `returncode == 0`, or using a JSON
   status command if Codex offers one. Worth pinning the tested Codex version in the docstring.

  **5. README merge conflict.** Conflicts against `22551f0` ("tighter tagline; drop misleading 500-commit corpus number"). Three hunks — easiest is to rebase on main and re-apply only the    
  Codex-specific additions (the `, Codex` insertion, the marketplace install block, the `docs/CODEX.md` link, the "Codex lifecycle hooks" subsection, the "Auto-generated AI instructions"     
  rename, and the AGENTS.md paragraph). Don't reintroduce "500-commit" anywhere.

  ### Should fix (before merge if quick, otherwise follow-up is fine)

  - **No timeout on `codex exec`** (`codex_cli.py:170-181`). `proc.communicate()` will hang indefinitely if Codex stalls. Wrap in `asyncio.wait_for(..., timeout=600)` and `proc.kill()` on    
  timeout.
  - **`reasoning` argument is accepted but silently ignored** (`codex_cli.py:151`). Either document it as accepted-for-interface-parity, or map `"minimal"` → `--config reasoning_effort=low`. 
  The test at `test_codex_cli_provider.py` currently just asserts no crash with `reasoning="minimal"`, which isn't exercising real behavior.
  - **Token fallback `len(prompt)//4`** (`codex_cli.py:201-204`) pollutes session totals when usage is missing. Suggest recording `0/0` with a `usage["estimated"]=True` flag so cost
  dashboards can distinguish "free subscription" from "unknown usage".
  - **TOML regex rewriting** (`mcp_config.py:191-213` and `save_codex_mcp_config`). We parse with `tomllib` but emit via `re.sub` + hand-rolled `_toml_value`, which strips comments and loses 
  key order from user-edited files. Could we use `tomli-w` or `tomlkit` for emission?
  - **`repowise mcp` no-arg behavior change** (`mcp_cmd.py:43-47`). It used to default to cwd; now it walks upward. Intentional for Codex, but it changes behavior for every existing user.    
  Could we log the resolved path so users notice, and mention it in the changelog?
  - **Hook `timeout: 10`** in `plugins/codex/hooks/hooks.json` — the search-rescue branch can exceed 10s on cold Windows filesystems. Bump to 20-30s.

  ### Nice to have (follow-up is fine)

  - `codex_cli.py:182-184` - error message tail can leak truncated JSON when Codex writes structured errors to stdout. Detect `{` prefix and use a fixed message.
  - `examples/codex/README.md:11` - first-run example triggers a full ~25-min index. Add an `--index-only` smoke variant.
  - `docs/CODEX.md` - verify the `developers.openai.com/codex/...` URLs resolve (Codex docs often live under `platform.openai.com/docs/codex/...`).
  - `plugins/codex/.codex-plugin/plugin.json:30` - `"capabilities": ["Read", ...]` — is `Read` a standard Codex capability? Worth double-checking against the schema.
  - Home-dir short-circuit in `find_repowise_repo_root` (`helpers.py:71-73`) uses `Path.__eq__` which is case-sensitive, but Windows paths are case-insensitive. Use `samefile` with a guard,  
  or drop the check.
  - An end-to-end CLI test that monkeypatches a fake `codex` binary into PATH and runs `repowise init --codex` would catch a lot.
  - Test gaps worth filling: timeout/kill path, multiline `agent_message` concatenation, and the Claude Code hook-payload backward-compatibility regression test mentioned in blocker #1.      

  Really appreciate the work here - once 1-5 are resolved I'm happy to merge.

Selene29:
I want to first add abstraction for setup provider in another pr.
then create codex ontop of this.

https://github.com/repowise-dev/repowise/pull/199

RaghavChamadiya:
Hey @Selene29, are you still working on this? Would love to get this merged

Selene29:
Yeah. Just had no time this week :)
Looking for today or tomorrow.

PR #199, opened 2026-05-15
Refactor CLI editor setup into integrations

Selene29:
## Summary

This PR splits the CLI product/setup layer out of `repowise init` before adding any Codex-specific setup behavior.

The current setup flow was functionally Claude-first: init directly knew how to write Claude project files, register Claude Desktop/Claude Code MCP config, and install Claude hooks. This change keeps that behavior intact, but moves it behind an editor setup integration boundary so future editor integrations can plug into the same lifecycle without expanding `init_cmd.py` with more editor-specific branches.

## What changed

- Added `repowise.cli.editor_setup` as the generic orchestration layer:
  - `EditorSetupIntegration` protocol
  - `EditorSetupOptions`
  - `write_editor_project_files(...)`
  - `register_editor_clients(...)`
- Added `repowise.cli.editor_integrations` for concrete editor integrations.
- Moved the existing Claude setup behavior into `editor_integrations/claude.py`.
- Added a default integration registry that currently returns only Claude, preserving existing behavior.
- Updated single-repo and workspace init flows to call the generic setup functions.
- Updated and added focused tests for the extraction.

## Intentional scope

This is a behavior-preserving setup-layer refactor. It does not add Codex setup, Codex hooks, AGENTS.md generation, or the Codex CLI provider. Those should be layered on top in a follow-up PR, using this integration boundary rather than mixing Codex code into the existing Claude setup path.

## Review notes

The abstraction is deliberately small:

- `editor_setup.py` contains only the generic contract and orchestration.
- Claude-specific strings, file paths, hook registration, and CLAUDE.md generation live in `editor_integrations/claude.py`.
- The default registry remains Claude-only so current init behavior does not change.

A follow-up Codex PR should add explicit selection/opt-in behavior rather than simply adding Codex to the default registry.

## Validation

- `uv run pytest tests/unit/cli/test_editor_setup.py tests/unit/cli/test_claude_md_prompt.py tests/unit/cli/test_mcp_config.py` -> 25 passed
- `uv run ruff check packages/cli/src/repowise/cli/editor_setup.py packages/cli/src/repowise/cli/editor_integrations/claude.py packages/cli/src/repowise/cli/editor_integrations/defaults.py tests/unit/cli/test_editor_setup.py tests/unit/cli/test_claude_md_prompt.py` -> passed
- `git diff --check` -> passed

### DIFF
--- a/packages/cli/src/repowise/cli/augment_hook.py
+++ b/packages/cli/src/repowise/cli/augment_hook.py
@@ -43,7 +43,7 @@ def main() -> None:
     # no-op on every subsequent fire. Wrapped so a write failure never
     # propagates back to the agent.
     try:
-        from repowise.cli.mcp_config import migrate_claude_code_hooks
+        from repowise.cli.editor_integrations.claude_config import migrate_claude_code_hooks
 
         migrate_claude_code_hooks()
     except BaseException:
--- a/packages/cli/src/repowise/cli/commands/init_cmd.py
+++ b/packages/cli/src/repowise/cli/commands/init_cmd.py
@@ -2,7 +2,6 @@
 
 from __future__ import annotations
 
-import json
 import os
 import sys
 import time
@@ -14,6 +13,10 @@ from rich.progress import BarColumn, Progress, SpinnerColumn, TextColumn, TimeEl
 from rich.table import Table
 
 from repowise.cli.cost_estimator import build_generation_plan, estimate_cost
+from repowise.cli.editor_setup import (
+    register_editor_clients,
+    write_editor_project_files,
+)
 from repowise.cli.helpers import (
     console,
     ensure_repowise_dir,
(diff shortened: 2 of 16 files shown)


RaghavChamadiya:
#199 looks clean: behavior-preserving, default registry stays Claude-only, tests cover the ordering. Two small things, non-blocking - happy to merge as-is or after, your call:
  
  - `default_editor_integrations()` in `editor_setup.py:48` and `get_default_editor_integrations()` in `defaults.py:11` are duplicate names doing the same thing. Worth collapsing to one.
  - The Protocol declares `id`/`display_name` but nothing reads them yet. Fine if you'll use them in #195's rework (e.g. logging which integration ran), otherwise they're dead weight — let me   know your intent.

  For #195: all the blocking + should-fix items from my earlier review look addressed, thanks for the thorough pass. Once #195 is rebased onto #199's abstraction (Codex as a second entry in the registry instead of branching inside `init_cmd.py`), I'll merge.

Selene29:
Follow-up clarification on the later commits in this PR:

The goal of the additional abstraction work is to make the setup/update pipeline editor-neutral, not to remove Claude support or mix in Codex provider work.

Concretely, the later commits move the remaining Claude-specific setup behavior out of generic command paths:

- `init` now resolves generic editor setup options and writes project files through the editor setup registry.
- `update` now calls a generic editor project-file refresh hook instead of hardcoding `CLAUDE.md` regeneration.
- Claude-specific prompting, `.mcp.json` root setup, `CLAUDE.md` generation, and Claude client registration now live under the Claude editor integration.
- The shared MCP setup instructions no longer present Claude Code as the generic/default setup path.
- The added tests cover the abstraction seams so future editor integrations can be added without pushing provider-specific logic back into `init` or `update`.

Claude-specific functionality still intentionally exists where it belongs: the Claude integration/config helpers and the explicit `generate-claude-md` command. This PR is just establishing the product/setup layer boundary so Codex setup can be added cleanly in a separate branch/PR.

Selene29:
there are still a few places we could modify but this is more than good enough as start point.
can get merged if no further issues arise.

RaghavChamadiya (approved):
lgtm

PR #242, opened 2026-05-27
fix(cli): silence structlog/stdlib logs on stdout when --format json or md is used

### LINKED ISSUE #227: health --format json output is polluted by debug logs on stdout
`repowise health --format json` is meant to be machine-readable — the docs even show `repowise health --format json | jq .kpis` — but the structlog debug/info lines are written to stdout ahead of the JSON, so it doesn't parse:

```
[info ] FileTraverser initialised   extra_exclude_patterns=0 ...
[debug] Skipping oversized file     path=...
{
  "kpis": { ... }
```

Right now you have to strip everything before the first `{` to use the output. Logs should go to stderr (or be silenced under `--format json`) so stdout is pure JSON. Same leak likely affects the other commands that support `--format json`. Probably the same root cause as the structlog-to-stdout noise during `init`.

MohammadYusif:
`repowise health --format json` (and `dead-code --format json`) was polluting stdout with structlog info/debug lines from the ingestion pipeline, making the output unparseable by downstream consumers like jq.

Add `silence_logs_for_machine_output()` to `cli/helpers.py` following the same pattern already used in `init_cmd.py` quiet mode: set httpx/httpcore and repowise.core stdlib loggers to ERROR, then reconfigure structlog with `wrapper_class=make_filtering_bound_logger(logging.ERROR)` and `cache_logger_on_first_use=False` (required to suppress module-level loggers snapshotted before configure() runs).

Call it at the top of `health_command()` and `dead_code_command()` whenever `fmt != 'table'`, before the ingestion pipeline starts.

Closes #227

## Summary

- Added `silence_logs_for_machine_output()` helper to `cli/helpers.py` that suppresses structlog and stdlib info/debug output when stdout is used for machine-readable formats
- Called it in `health_command()` and `dead_code_command()` when `fmt != "table"`, fixing `--format json` and `--format md` for both commands
- `repowise health --format json | jq .kpis` now works without stripping leading log lines

## Related Issues

Closes #227

## Test Plan

- [x] Tests pass (`pytest`)

## Checklist

- [ ] I have added tests for new functionality

Note on the two unchecked boxes: No new tests were added because exercising the fix requires the full ingestion pipeline (traverser, parser, graph builder) which can't be meaningfully unit-tested in isolation. The fix itself is a direct port of the pattern already used and proven in init_cmd.py. No docs needed since the command's --format flag behavior is unchanged from the user's perspective — it just works correctly now.

### DIFF
--- a/packages/cli/src/repowise/cli/commands/dead_code_cmd.py
+++ b/packages/cli/src/repowise/cli/commands/dead_code_cmd.py
@@ -12,6 +12,7 @@ from repowise.cli.helpers import (
     resolve_command_target,
     resolve_repo_path,
     run_async,
+    silence_logs_for_machine_output,
 )
 
 
@@ -86,6 +87,9 @@ def dead_code_command(
     --repo <alias> to target a different repo. Cross-repo dead-code
     detection is not yet supported — run once per repo for now.
     """
+    if fmt != "table":
+        silence_logs_for_machine_output()
+
     from pathlib import Path as PathlibPath
 
     from repowise.core.analysis.dead_code import DeadCodeAnalyzer
--- a/packages/cli/src/repowise/cli/commands/health_cmd.py
+++ b/packages/cli/src/repowise/cli/commands/health_cmd.py
@@ -17,6 +17,7 @@ from repowise.cli.helpers import (
     err_console,
     resolve_command_target,
     run_async,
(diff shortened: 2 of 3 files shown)


RaghavChamadiya (approved):
lgtm

PR #298, opened 2026-05-29
fix(dead-code): stop single-file re-index from wiping all findings

### LINKED ISSUE #295: [Bug] Dead code: false positives on CommonJS property-access imports + findings reset on re-index
## Describe the Bug

Dead code detection (`get_dead_code`) produces false positives at 1.0 confidence for actively-used functions in a CommonJS codebase. All 6 high-tier results I manually verified were false positives. Additionally, re-indexing a single file caused all ~979 findings to drop to 0.

## Steps to Reproduce

**False positives:**
1. Index a Node.js CommonJS repo (~50 modules) using whole-module imports (`const fooService = require('./foo.service')`)
2. Run `get_dead_code` via MCP
3. Review high-confidence (1.0) findings marked "safe to delete"
4. All reported "dead" functions are actively called via property access (e.g. `fooService.bar()`)

**Findings reset:**
1. Commit a change to one file (convert a whole-module import to destructured import)
2. Wait for re-index to complete
3. Run `get_dead_code` again
4. Total findings drop from ~979 to 0

## Expected Behavior

1. Property-access usage (`fooService.bar()`) should be recognized as a reference to `bar`, or confidence should be significantly lower than 1.0
2. Re-indexing a single file should not wipe all ~979 findings

## Actual Behavior

**False positive patterns:**

Pattern 1 — Whole-module imports with property access:
```js
const orderService = require('./order.service');
orderService.processOrder(userId, itemIds, containerId);
// → Repowise reports processOrder as dead (1.0 confidence)
```

Pattern 2 — Factory functions returning objects:
```js
const worker = initWorker({ io, pool });
workers.taskWorker = worker;
workers.taskWorker.runTaskA(id);
// → runTaskA reported as dead
```

Pattern 3 — Dynamic dispatch via Map:
```js
const strategies = new Map();
strategies.set('typeA', new TypeAStrategy());
strategies.get(shapeName).execute(...);
// → TypeAStrategy reported as unreachable
```

**Findings reset:** After committing a single-file change, `get_dead_code` returned 0 total findings (down from ~979).

## Environment

- OS: Windows 11 Pro
- Node.js codebase (CommonJS, ~50 modules)
- Repowise MCP integration via Claude Code
- Installation method: MCP server

## Additional Context

- All false positives share the same root cause: Repowise tracks destructured/named imports but not property-access patterns (`module.exportedFn()`), which is the dominant import style in CommonJS Node.js projects.
- The 979→0 findings drop suggests the re-indexing pipeline may be clearing the full dead-code index rather than incrementally updating it.

joptimus:
## What

Stop `repowise update` from dropping the entire dead-code index (~979 findings → 0) when a single file changes.

## Why

Two coupled root causes (issue #295):

- `save_dead_code_findings()` was a **full-repo delete-then-insert** (no `file_path` scope).
- `analyze_partial()` returned a **sparse** report — only the unreachable-file detector, only for changed files.

Together, a single-file incremental update deleted every finding and reinserted a near-empty set.

## How

- **File-scoped persistence:** add `upsert_dead_code_findings(..., file_paths=...)` mirroring the existing `upsert_health_findings` — deletes and inserts only within the changed paths, so unchanged files keep their findings and a now-clean changed file has its stale findings cleared. Inserts are filtered to `file_paths` to enforce the scope contract.
- **Complete partial analysis:** rewrite `analyze_partial()` to run the full detector suite (it needs the complete graph for accurate caller in-degrees anyway) and narrow the result to the affected files, so all finding kinds are recomputed — not just unreachable files.
- **Wire-up:** thread changed paths into both update persistence paths (index-only and full) and use the scoped upsert.

## Known limitation

Incremental analysis is scoped to *changed* files. An edit to file X that makes a symbol in unchanged file Y dead is not recomputed until the next full analysis (`repowise init`/full re-score). This matches the existing incremental-health behavior; the full `analyze()` remains authoritative.

## Tests

- Unit (`tests/unit/persistence/test_dead_code_crud.py`): scoped replace preserves other files; changed-but-clean file is cleared; out-of-scope findings are not inserted; empty `file_paths` is a no-op.
- Unit (`tests/unit/dead_code/test_partial.py`): `analyze_partial` reports an unused *export* on a changed file and scopes findings to affected files.
- Integration (`tests/integration/test_cli.py::TestUpdatePreservesDeadCode`): a single-file update preserves unchanged files' findings and does not wipe the index.

Closes #295 (findings-reset half; the CommonJS property-access false-positive half is a separate follow-up PR).

### DIFF
--- a/packages/cli/src/repowise/cli/commands/update_cmd.py
+++ b/packages/cli/src/repowise/cli/commands/update_cmd.py
@@ -368,6 +368,7 @@ def _persist_index_only_update(
     state: dict,
     head: str | None,
     start: float,
+    changed_paths: list[str],
 ) -> None:
     """Persist the index-only update (graph + git + dead-code + health), save
     state, and print the completion line. No LLM regeneration."""
@@ -406,10 +407,12 @@ def _persist_index_only_update(
             if dead_code_report is not None:
                 try:
                     from repowise.core.persistence.crud import (
-                        save_dead_code_findings,
+                        upsert_dead_code_findings,
                     )
 
-                    await save_dead_code_findings(session, repo_id, dead_code_report.findings)
+                    await upsert_dead_code_findings(
+                        session, repo_id, dead_code_report.findings, file_paths=changed_paths
+                    )
                 except Exception as exc:
                     console.print(f"[yellow]Dead-code persist skipped: {exc}[/yellow]")
 
@@ -776,6 +779,7 @@ def update_command(
             state,
             head,
             start,
+            [fd.path for fd in file_diffs],
         )
         return
 
@@ -1068,18 +1072,19 @@ def update_command(
             except Exception:
                 pass  # health persistence is best-effort
 
-        # Persist dead code findings (partial)
-        if dead_code_report and dead_code_report.findings:
+        # Scoped to changed files so unchanged files keep their findings (#295).
+        if dead_code_report is not None:
             try:
                 import dataclasses as _dc_dead
 
-                from repowise.core.persistence.crud import save_dead_code_findings
+                from repowise.core.persistence.crud import upsert_dead_code_findings
 
                 async with get_session(sf) as session:
-                    await save_dead_code_findings(
+                    await upsert_dead_code_findings(
                         session,
                         repo_id,
                         [_dc_dead.asdict(f) for f in dead_code_report.findings],
+                        file_paths=[fd.path for fd in file_diffs],
                     )
             except Exception:
                 pass  # dead code persistence is best-effort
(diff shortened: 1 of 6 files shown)


joptimus:
This is the first of two PRs for #295.

**Scope of this PR:**
- Fix the findings-reset bug: a single-file `repowise update` no longer wipes the whole dead-code index.

**Planned follow-up PR:**
- Fix the CommonJS false positives: treat `require()` (whole-module `const x = require('./y'); x.fn()` and destructured `const { fn } = require('./y')`) as a first-class import so property-access calls resolve and actively-used functions aren't flagged dead.

Keeping them separate so each stays focused and easy to review. The findings-reset fix is a small persistence/analysis change, while the CommonJS fix touches the tree-sitter queries, binding extraction, and call resolver.

RaghavChamadiya (approved):

PR #299, opened 2026-05-29
fix(dead-code): resolve CommonJS require() so property-access calls aren't dead

### LINKED ISSUE #295: [Bug] Dead code: false positives on CommonJS property-access imports + findings reset on re-index

joptimus:
## What

Stop dead-code detection from flagging actively-used functions as dead (1.0 confidence) in CommonJS codebases.

## Why

Given `const svc = require('./svc'); svc.bar()`, the call to `bar` was not recognized as a use, so `bar` was reported dead. Root cause: `require()` was invisible to the pipeline. The JS/TS tree-sitter queries only captured ES6 `import_statement`, so no import binding was created, `_module_aliases` was never populated for the local variable, and the member call could not resolve to the exported symbol. This is the false-positive half of issue #295.

## How

- Capture `require()` in the JS/TS queries by tagging the individual `variable_declarator` as `@import.statement` (not the enclosing declaration). Tagging the declarator means multi-declarator statements like `const a = require('./a'), b = require('./b')` are not collapsed by the parser's raw-text dedup.
- Extend the TS/JS binding extractor to emit bindings for whole-module aliases (`const svc = require('./svc')`, `is_module_alias=True`) and destructured forms (`const { bar } = require('./svc')`, including renamed `{ x: y }`).
- No resolver change was needed: once the module-alias binding exists and resolves, the existing member-call resolution maps `svc.bar()` to the source export.

## Tests

- Parser unit tests (`tests/unit/ingestion/parser/test_commonjs_require.py`): whole-module, destructured, renamed, multi-declarator, and `var` require forms.
- End-to-end test (`tests/integration/test_commonjs_dead_code.py`): `svc.used()` creates a calls edge to the export (no false positive), and resolution is targeted (no phantom edge to an unaccessed export like `reallyDead`).
- Full ingestion + dead-code + TS regression suites pass (904 passed, 2 xfailed).

## Out of scope (documented for follow-up)

Not handled by this PR: `require('./x').default`, member access directly on the require call, dynamic `require(var)`, TS `import x = require()`, and CommonJS `module.exports` named-export detection. The last one means a genuinely-dead CommonJS export is currently under-flagged; that is a separate gap (under-flagging, not the false positive fixed here).

Closes #295 (CommonJS false-positive half; the findings-reset half is #298).

### DIFF
--- a/packages/core/src/repowise/core/ingestion/extractors/bindings/ts_js.py
+++ b/packages/core/src/repowise/core/ingestion/extractors/bindings/ts_js.py
@@ -8,6 +8,65 @@ from ...models import NamedBinding
 from ..helpers import node_text
 
 
+def _extract_require_bindings(
+    stmt_node: Node, src: str
+) -> tuple[list[str], list[NamedBinding]] | None:
+    """Bindings for a ``variable_declarator`` initialized by ``require(...)``.
+
+    Handles ``const svc = require('./svc')`` (whole-module alias) and
+    ``const { a, b: c } = require('./svc')`` (destructured). Returns None when
+    the node is not a require() declarator so the caller falls through to the
+    existing import/export handling.
+    """
+    if stmt_node.type != "variable_declarator":
+        return None
+    value = stmt_node.child_by_field_name("value")
+    if value is None or value.type != "call_expression":
+        return None
+    fn = value.child_by_field_name("function")
+    if fn is None or node_text(fn, src) != "require":
+        return None
+
+    name_node = stmt_node.child_by_field_name("name")
+    if name_node is None:
+        return [], []
+
+    names: list[str] = []
+    bindings: list[NamedBinding] = []
+
+    if name_node.type == "identifier":
+        local = node_text(name_node, src)
+        names.append(local)
+        bindings.append(
+            NamedBinding(
+                local_name=local,
+                exported_name=None,
+                source_file=None,
+                is_module_alias=True,
+            )
+        )
+    elif name_node.type == "object_pattern":
+        for el in name_node.children:
+            if el.type == "shorthand_property_identifier_pattern":
+                local = node_text(el, src)
+                names.append(local)
+                bindings.append(
+                    NamedBinding(local_name=local, exported_name=local, source_file=None)
+                )
+            elif el.type == "pair_pattern":
+                key = el.child_by_field_name("key")
+                val = el.child_by_field_name("value")
+                if key is not None and val is not None:
+                    exported = node_text(key, src)
+                    local = node_text(val, src)
+                    names.append(local)
+                    bindings.append(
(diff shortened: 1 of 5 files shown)


joptimus:
Second of two PRs for #295.

- #298 fixes the findings-reset bug (single-file update no longer wipes the dead-code index).
- This PR fixes the CommonJS false positives (property-access calls on a required module now resolve).

The two are independent and can merge in either order.

RaghavChamadiya (approved):

PR #332, opened 2026-06-01
fix(dead-code): keep rust sibling test modules live

biggiesmallcap-blip:
# PR 2: Rust Sibling Test Modules

## Title

fix(dead-code): keep rust sibling test modules live

## Branch

`codex/dead-code-rust-tests-module`

## Summary

Prevents the dead-code analyzer from reporting Rust sibling test modules as unreachable files.

- Adds `**/src/**/tests.rs` to the dead-code never-flag patterns.
- Adds a regression test for the `src/server/routes/tests.rs` shape used by Rust crates that declare `#[cfg(test)] mod tests;`.
- Keeps the change conservative and audit-only: it only suppresses the test module file itself from unreachable-file findings.
- Includes a tiny lint cleanup in the touched constants file by removing a duplicate `.husky` set entry.

## Why

Rust unit-test modules are often compiled into the crate test target through declarations like:

```rust
#[cfg(test)]
mod tests;
```

Those files do not necessarily have ordinary production import edges. Treating `src/**/tests.rs` as an unreachable production file creates false deletion signals.

## Testing

```powershell
.\.venv\Scripts\python.exe -m pytest tests\unit\dead_code\test_rust_tests_module.py
.\.venv\Scripts\python.exe -m pytest tests\unit\dead_code
.\.venv\Scripts\python.exe -m ruff check packages\core\src\repowise\core\analysis\dead_code\constants.py tests\unit\dead_code\test_rust_tests_module.py
```

The full dead-code unit suite passed: `66 passed`.

## Risk

This can hide a truly orphaned Rust `tests.rs` file from the unreachable-file report, but that is preferable to presenting Cargo test modules as safe deletion candidates. The suppression is limited to the conventional `src/**/tests.rs` path.

### DIFF
--- a/packages/core/src/repowise/core/analysis/dead_code/constants.py
+++ b/packages/core/src/repowise/core/analysis/dead_code/constants.py
@@ -9,7 +9,6 @@ from __future__ import annotations
 
 from repowise.core.ingestion.languages.registry import REGISTRY as _LANG_REGISTRY
 
-
 # Non-code languages that should never be flagged as dead code.
 # Derived from the centralised LanguageRegistry — passthrough config/infra
 # languages plus "unknown".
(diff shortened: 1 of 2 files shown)


RaghavChamadiya (approved):
LGTM

PR #308, opened 2026-05-29
fix(exclude-patterns): enforce exclude_patterns in git indexer and dynamic edges

joptimus:
## What

Make `exclude_patterns` actually exclude files from the git indexer and from dynamic-edge graph construction.

## Why

`exclude_patterns` was honored by file traversal but bypassed in two places (issues 1 and 2 of #296):

1. The git indexer ran `git ls-files` and indexed every tracked file, so excluded paths still produced git metadata (churn, ownership, hotspots).
2. Dynamic-edge creation added a graph node for any co-change/hint target, so excluded files reappeared as phantom nodes even though traversal skipped them.

## How

- `GitIndexer` accepts `exclude_patterns` and filters them out of `git ls-files`. Threaded through `_run_git_indexing` and the orchestrator's git stage (init path).
- The incremental update path (`_rebuild_graph_and_git`) passes `exclude_patterns` to `GitIndexer` and filters changed paths via a new `_build_filtered_changed_paths` helper.
- `GraphBuilder` accepts `exclude_patterns`; `add_dynamic_edges` no longer creates phantom nodes for excluded targets. Threaded into the ingestion pipeline phase and the update rebuild path.

All filtering uses `pathspec` (gitwildmatch), consistent with how traversal already applies `exclude_patterns`.

## Tests

- `tests/unit/test_git_indexer.py::TestExcludePatterns`: `_get_tracked_files` filters excluded paths and is a no-op without patterns.
- `tests/unit/cli/test_commands.py::TestBuildFilteredChangedPaths`: changed-path filtering.
- `tests/unit/ingestion/test_graph.py::TestDynamicEdgeExcludeFilter`: excluded dynamic-edge targets do not create phantom nodes; non-excluded targets still do.
- Full ingestion + git-indexer + cli suites pass (874 passed, 2 xfailed).

This change is Python-only and does not touch `packages/web`.

Closes #296 (issues 1 and 2; the remaining issues 3 and 5 are separate follow-up PRs).

PR #297, opened 2026-05-29
fix(init): persist exclude_patterns in index-only mode

joptimus:
## What

Persist `exclude_patterns` and `commit_limit` to `.repowise/config.yaml` when `repowise init` runs in **index-only mode**.

## Why

Index-only init never calls `save_config()`, so `-x/--exclude` and `--commit-limit` flags were silently dropped. That left `repowise update` and the MCP tools with no record of the user's excludes, causing excluded paths to be re-indexed on subsequent runs. This is Issue 4 of #296.

## How

- Add `save_config_partial()` to `cli/helpers.py` — merges optional keys into `config.yaml`, preserving existing keys. Unlike `save_config()`, it has **no scalar-only YAML fallback**, since that fallback would silently drop the `exclude_patterns` list (and PyYAML is a hard dependency anyway).
- Call it from the index-only branch of `init_command`, matching full mode's behavior: clamp `commit_limit` via `resolved_commit_limit` and omit empty excludes (pass `None`).

## Tests

- Unit tests for `save_config_partial` (persist, no-op, create-if-missing, commit_limit).
- Command-level integration tests asserting an index-only `init -x vendor/ --commit-limit 99999` writes `exclude_patterns: ["vendor/"]` and the clamped `commit_limit: 5000`, and that an exclude-less run writes neither key.

This change is Python-only and does not touch `packages/web`.

Closes #296 (Issue 4)

PR #337, opened 2026-06-01
fix(update): re-run health on config change even when git is unchanged

### LINKED ISSUE #296: [Bug] exclude_patterns not fully enforced: git indexer and dynamic edges leak excluded files into graph/UI
## Describe the Bug

  Files and directories listed in `exclude_patterns` (via `repowise init -x` or `.repowise/config.yaml`) still appear in
   the dashboard UI (e.g., as "attention needed" communities) and in MCP tool responses. The exclusion is enforced
  during file traversal but bypassed by the git indexer and dynamic edge creation, which leak excluded paths back into
  the graph as phantom nodes.

  Additionally, `repowise update` does not detect changes to `.repowise/config.yaml` or `.repowise/health-rules.json` —
  if no source files changed via git diff, the command exits immediately without re-running any analysis, even if the
  config was modified.

  ## Steps to Reproduce

  1. Run `repowise init -x .claude/ -x tools/` on a repo that has git-tracked files under `.claude/` and `tools/`
  2. Open the dashboard or query via MCP (`get_overview`, `get_health`)
  3. Observe `.claude/` appearing as a community/module in "attention needed" despite being excluded
  4. Edit `.repowise/health-rules.json` to disable a biomarker (e.g., `"disabled_biomarkers": ["ungoverned_hotspot"]`)
  5. Run `repowise update`
  6. Observe "No changed files detected" — the config change is ignored

  ## Expected Behavior

  - Files/directories in `exclude_patterns` should not appear anywhere in the graph, health findings, wiki, MCP
  responses, or dashboard UI
  - Changes to `.repowise/config.yaml` or `.repowise/health-rules.json` should be picked up by `repowise update` and
  trigger a rescore of affected files

  ## Actual Behavior

  **Issue 1: Git indexer ignores `exclude_patterns`**

  The git indexer (`git_indexer/indexer.py`) calls `repo.git.ls_files()` which returns ALL git-tracked files.
  `exclude_patterns` is never passed to the git phase, so `git_meta_map` contains entries for excluded files.

  **Issue 2: Dynamic edges create phantom graph nodes**

  In `_edges.py:82`:
  ```python
  if e.target not in self._graph:
      self._graph.add_node(e.target)  # Creates phantom node for excluded file
```
  When dynamic edges reference an excluded file (found via git history), this adds it to the graph even though traversal
   correctly excluded it. These phantom nodes get community labels, which the UI renders as modules.

   **Issue 3: `repowise update` ignores config changes**

  In `update_cmd.py:713-716`:
  ```python
  if not file_diffs:
      console.print("[green]No changed files detected.[/green]")
      save_state(repo_path, {**state, "last_sync_commit": head})
      return
  ```

  If no source files changed (git diff is empty), the entire command returns without loading `health-rules.json` or
  re-running health analysis — even if the config files themselves were modified.

  **Issue 4: `exclude_patterns` not persisted in index-only mode**

  `init_cmd.py:2006-2017` only saves `exclude_patterns` to `config.yaml` when running in full mode (with LLM). In
  index-only mode, the patterns are used for that run but never written to config, so subsequent `repowise update` runs
  can't find them.

  **Issue 5: MCP tools don't filter by `exclude_patterns`**

  All MCP tools (`tool_health.py`, `tool_overview.py`, etc.) query the database directly without a WHERE clause
  filtering on `exclude_patterns`. Even if excluded files leak into the DB, a runtime filter on query results would
  prevent them from reaching the UI.

  ## Environment

  - OS: Windows 11 Pro
  - Python version: 3.14
  - Repowise version: 0.13.0
  - Installation method: pip

  ## Additional Context

  **Data flow showing the leak path:**

  ```
  repowise init -x .claude/
    ├─ Traversal: .claude/ correctly excluded ✓
    ├─ Graph build: only traversed files become nodes ✓
    ├─ Git indexer: scans ALL git-tracked files ← LEAK
    │   └─ git_meta_map includes .claude/ files
    ├─ Dynamic edges (_edges.py:82): creates phantom nodes ← LEAK
    │   └─ .claude/ now in graph despite exclusion
    ├─ Community detection: assigns .claude/ to a community
    └─ UI: shows "claude" community as "attention needed"
  ```

  **Relevant source locations:**

  | Component | File | Lines | Issue |
  |-----------|------|-------|-------|
  | Git indexer | `git_indexer/indexer.py` | `_get_tracked_files()` | No `exclude_patterns` filtering |
  | Dynamic edges | `graph/_edges.py` | 82 | Creates phantom nodes for missing targets |
  | Update command | `cli/commands/update_cmd.py` | 713-716 | Early return on empty git diff |
  | Config persistence | `cli/commands/init_cmd.py` | 2006-2017 | Only saves in full mode |
  | MCP tools | `server/mcp_server/tool_*.py` | Various | No runtime exclusion filter |

  **Workaround:** Clear the health tables manually via SQLite and re-run `repowise health`:
  ```python
  python3 -c "
  import sqlite3
  conn = sqlite3.connect('.repowise/wiki.db')
  conn.execute('DELETE FROM health_findings')
  conn.execute('DELETE FROM health_file_metrics')
  conn.commit()
  "
  ```
  Then run `repowise health` to rescore with the updated config. This doesn't fix the phantom graph nodes — only
  `repowise init` clears those.

joptimus:
## What

`repowise update` now detects changes to `.repowise/config.yaml` (e.g. `exclude_patterns`) and `health-rules.json` and triggers a full health re-score, even when no commits moved HEAD.

## Why

Previously update only re-scored health for files that changed in git. Editing the config or health rules had no effect until a code change happened to touch each file, so newly-excluded files lingered in health views and rule changes never applied to unchanged files. This is issue 3 of #296.

## How

- Add `config_fingerprint(repo_path)` to `cli/helpers.py`: a sha256 of `config.yaml` + `health-rules.json` content. `repowise init` stores it in `state.json` after the config is written, so the first update does not false-positive.
- `repowise update` compares the stored fingerprint to the current one. A real change runs `_run_full_health_rescore`: rebuild the graph with current `exclude_patterns`, load `GitMetadata` from the DB (so biomarkers keep accurate churn/ownership data, including the newer `prior_defect_count` / `change_entropy` columns added since), drop excluded git rows, full-replace health metrics/findings, and refresh graph nodes.
- The graph rebuild is shared with the incremental update path via `_build_repo_graph`, so both build the same graph from the same parser and the same framework-aware synthetic edges.
- A missing prior fingerprint is backfilled, not treated as a change.
- The config-change path takes precedence over the partial update so unchanged files are re-scored under the new rules. It is skipped under `--dry-run`, and does not advance the fingerprint if the re-score fails (so the next update retries).

## Scope note: graph rows

This PR fully replaces **health** metrics/findings (so rows for newly-excluded files are removed) and removes excluded `git_metadata` rows. It **refreshes** `graph_nodes` but does **not** delete stale `graph_nodes` / `graph_edges` rows for newly-excluded files. Hiding excluded files from graph-backed tool responses is handled at query time by the MCP runtime exclude filtering in #339 / #340, not by deleting graph rows here.

## Trade-off

In the rare case where a config change and new source commits land in the same `update`, the full health re-score takes precedence and the incremental doc regeneration for those commits is deferred. For index-only mode (no docs) this loses nothing.

## Tests

Unit: `config_fingerprint`, `_git_metadata_to_dict` (incl. the new columns), a failed-rescore-keeps-old-fingerprint test, and `TestBuildRepoGraph` (parse-skip warning + framework-edge step in the shared rebuild path). Integration: init stores the fingerprint; unchanged config stays "Already up to date"; a `health-rules.json` change triggers the re-score; `--dry-run` does not mutate state/DB; and config-change-with-source-diffs takes the full re-score path.

Closes #296 (issue 3)

### DIFF
--- a/packages/cli/src/repowise/cli/commands/init_cmd.py
+++ b/packages/cli/src/repowise/cli/commands/init_cmd.py
@@ -23,6 +23,7 @@ from repowise.cli.editor_setup import (
     write_editor_project_files,
 )
 from repowise.cli.helpers import (
+    config_fingerprint,
     console,
     ensure_repowise_dir,
     get_head_commit,
@@ -1344,6 +1345,10 @@ def _save_full_state_and_config(
         reasoning=resolved_reasoning,
     )
 
+    # Re-save state with the fingerprint now that config.yaml is written.
+    state["config_fingerprint"] = config_fingerprint(repo_path)
+    save_state(repo_path, state)
+
 
 def _show_completion(
     *,
@@ -2018,6 +2023,8 @@ def init_command(
             exclude_patterns=exclude_patterns if exclude_patterns else None,
             commit_limit=resolved_commit_limit if commit_limit is not None else None,
         )
+        # Fingerprint after config writes so the first update doesn't false-positive.
+        base_state["config_fingerprint"] = config_fingerprint(repo_path)
         save_state(repo_path, base_state)
 
     # ---- State + config (full mode only) ----
--- a/packages/cli/src/repowise/cli/commands/update_cmd.py
+++ b/packages/cli/src/repowise/cli/commands/update_cmd.py
@@ -277,6 +277,69 @@ def _build_filtered_changed_paths(file_diffs: list, exclude_patterns: list[str])
     return [p for p in paths if not spec.match_file(p)]
 
 
+def _build_repo_graph(
+    repo_path: Any,
+    exclude_patterns: list[str],
+    *,
+    collect_sources: bool = False,
+) -> tuple[list, dict[str, bytes], Any, Any, int]:
+    """Traverse + parse the repo and build the graph (+ framework-aware edges).
+
+    Shared by the incremental rebuild path (:func:`_rebuild_graph_and_git`) and
+    the config-triggered re-score path (:func:`_run_full_health_rescore`) so both
+    build the same graph from the same parser and the same synthetic edge step.
+
+    Files that fail to read/parse are skipped and reported as a count rather than
+    swallowed silently. ``source_map`` is populated only when ``collect_sources``
+    is set (the re-score path doesn't need the raw bytes).
+
+    Returns ``(parsed_files, source_map, graph_builder, repo_structure,
+    file_count)``.
+    """
+    from pathlib import Path as PathlibPath
+
+    from repowise.core.ingestion import ASTParser, FileTraverser, GraphBuilder
+
+    traverser = FileTraverser(repo_path, extra_exclude_patterns=exclude_patterns or None)
+    file_infos = list(traverser.traverse())
+    repo_structure = traverser.get_repo_structure()
+
+    parser = ASTParser()
+    parsed_files: list = []
+    source_map: dict[str, bytes] = {}
+    graph_builder = GraphBuilder(repo_path, exclude_patterns=exclude_patterns)
+
+    skipped = 0
+    for fi in file_infos:
+        try:
+            source = PathlibPath(fi.abs_path).read_bytes()
+            parsed = parser.parse_file(fi, source)
+        except Exception:
+            skipped += 1
+            continue
+        parsed_files.append(parsed)
+        if collect_sources:
+            source_map[fi.path] = source
+        graph_builder.add_file(parsed)
+    graph_builder.build()
+
+    if skipped:
+        console.print(f"[yellow]Skipped {skipped} file(s) that failed to parse.[/yellow]")
+
+    # Add framework-aware synthetic edges (conftest, Django, FastAPI, Flask).
+    try:
+        from repowise.core.generation.editor_files.tech_stack import detect_tech_stack
+
+        tech_items = detect_tech_stack(repo_path)
+        fw_count = graph_builder.add_framework_edges([item.name for item in tech_items])
+        if fw_count:
+            console.print(f"Framework edges added: [cyan]{fw_count}[/cyan]")
+    except Exception:
+        pass  # framework edge detection is best-effort
+
+    return parsed_files, source_map, graph_builder, repo_structure, len(file_infos)
+
+
 def _rebuild_graph_and_git(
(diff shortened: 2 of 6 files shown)


joptimus:
Part of the #296 exclude_patterns / config series.

- #297 (merged): persist exclude_patterns in index-only init.
- #308 (merged): enforce exclude_patterns in the git indexer and dynamic edges.
- This PR: detect config / health-rules changes during `repowise update` and trigger a full health re-score (issue 3).

One planned follow-up remains:
- Add per-repo runtime exclude_patterns filtering to MCP tool responses (issue 5).

RaghavChamadiya:
Mechanism looks good (fingerprint backfill, dry-run guard, and the failed-rescore test are all right). A few changes before merge:
1. Don't fork the index pipeline. _run_full_health_rescore reimplements traverse → parse → build → analyze → persist, which we already have in _persist_index_only. Can it reuse that path instead? My worry is the two drifting apart, a config re-score producing different health scores than a fresh init.
2. Drop the unused cfg param, the function loads HealthConfig itself.
3. Don't swallow parse errors silently. The per-file except Exception: pass should log a count like the [yellow]…skipped[/yellow] pattern used elsewhere in this file.

joptimus:
That feedback makes sense to me.

For item 1, I agree with the concern. `_run_full_health_rescore` should not keep its own separate traverse/parse/graph-build path. I noticed it can already drift from the normal update path because the normal rebuild adds framework-aware synthetic edges, while the config re-score path does not.

I will avoid trying to force `_persist_index_only` directly if the semantics do not fit, since config re-score needs full health replacement rather than partial upsert. But I will extract/reuse the shared graph rebuild logic so the normal update path and config re-score path build the same graph.

Items 2 and 3 are straightforward:
- remove the unused `cfg` param
- replace silent parse-error swallowing with a skipped-count warning

RaghavChamadiya (approved):
LGTM

PR #339, opened 2026-06-01
fix(mcp): filter exclude_patterns from MCP tool responses

joptimus:
## What

MCP tools now filter their query results against the repo's `exclude_patterns` at request time, so excluded files do not surface in tool responses.

## Why

Excluded files are skipped at ingest time, but persisted rows (graph nodes, health metrics, git metadata, findings) may predate an `exclude_patterns` change. Without runtime filtering, those stale rows still appear in MCP tool output. This is issue 5 of #296.

## How

Adds per-repo filter helpers to `mcp_server/_helpers.py`:
- `_get_exclude_spec(repo_path)` compiles the repo config's `exclude_patterns` into a `pathspec` (or `None`), and `is_excluded(path, spec)`.
- Shape-specific helpers, each with parametrized tests: `filter_rows_by_attr` (ORM rows by attribute), `filter_graph_nodes` (file nodes by `node_id`, symbol nodes by `file_path`), `filter_dicts_by_key`, `filter_path_list`, and `filter_embedded_path_ids` (`"path::Name"` ids).

Wires them into the structured tools: `health`, `overview`, `dead_code`, `community`, `metrics`, `callers`, `symbol`, `diagram`, `dependency`, `risk`, `flows`, `why`, `decision_records`, and `search` (single-repo path). Single-file/target tools reject or omit excluded targets; list/aggregate tools drop excluded rows before KPIs/rollups so excluded files never affect the numbers.

### Traversal tools (not just first-list filters)

Several tools needed deeper filtering than a single result list:
- `diagram`: keeps only edges whose **both** endpoints survived filtering, so an excluded endpoint can't render via a surviving neighbor.
- `dependency`: builds the NetworkX traversal graph only from surviving nodes/edges, so shortest paths / ancestors / neighbors / bridges never route through excluded files.
- `flows`: filters the BFS trace (not just entry points) before community resolution and return.
- `health`: filters coverage rows by `file_path` (the `coverage.summary` is a repo-wide stored aggregate, not recomputed — noted in code).
- `risk`: filters targets, co-change partners, impact surface, the `changed_files` input, the directive lists, and the raw `pr_blast_radius` lists (`transitive_affected`/`cochange_warnings`/`test_gaps`/`recommended_reviewers`).

## Diagram caveat (important)

`tool_diagram` filtering applies to the **computed module/file diagrams** (`scope="module"`/`"file"`). For `scope="repo"`, the tool returns a **stored** architecture-diagram page (`Page.content`), which is **not** filtered in this PR — filtering stored Mermaid text safely is separate work. So repo-level stored diagrams may still reference excluded files.

## Tests

- 11 parametrized unit tests covering every helper/shape.
- An end-to-end test proving `get_health` honors `exclude_patterns` from `.repowise/config.yaml`.
- Targeted traversal-leak tests: diagram edges, dependency intermediate nodes, flows trace, risk co-change partners, and the raw `pr_blast_radius` payload (non-vacuous: proves the path appears without excludes and disappears with them).
- Full server suite passes (299 passed).

## Out of scope (planned follow-up PR)

Three surfaces need filtering inside their context/retrieval pipelines rather than a result-level helper call, so they are deferred to a focused follow-up:
- `tool_context` enrichment / kg neighbor expansion
- `tool_answer` retrieval + synthesis (citations)
- `search` federated ("all repos") path, whose aggregated results carry no path key to filter on

Part of #296 (issue 5)

### DIFF
--- a/packages/server/src/repowise/server/mcp_server/_helpers.py
+++ b/packages/server/src/repowise/server/mcp_server/_helpers.py
@@ -420,3 +420,71 @@ def _compute_alignment(
         "stale_count": len(stale),
         "sibling_coverage": round(sibling_coverage, 2) if sibling_coverage is not None else None,
     }
+
+
+# ---------------------------------------------------------------------------
+# Per-repo exclude_patterns filtering (issue 5 of #296)
+#
+# Excluded files are skipped at ingest time, but rows may predate an
+# exclude_patterns change, so MCP tools filter their results at query time too.
+# ---------------------------------------------------------------------------
+
+
+def _get_exclude_spec(repo_path: "Path | str") -> "Any":
+    """Compile the repo's ``exclude_patterns`` into a PathSpec, or None."""
+    import pathspec
+
+    from repowise.core.repo_config import load_repo_config
+
+    patterns = load_repo_config(repo_path).get("exclude_patterns") or []
+    if not patterns:
+        return None
+    return pathspec.PathSpec.from_lines("gitwildmatch", patterns)
+
+
+def is_excluded(path: "str | None", spec: "Any") -> bool:
+    """True if *path* matches *spec* (None spec or path -> not excluded)."""
+    return bool(spec is not None and path and spec.match_file(path))
+
+
+def filter_rows_by_attr(rows: list, attr: str, spec: "Any") -> list:
+    """Shape A: drop ORM rows whose ``attr`` path is excluded."""
+    if not spec:
+        return rows
+    return [r for r in rows if not is_excluded(getattr(r, attr, None), spec)]
+
+
+def filter_graph_nodes(nodes: list, spec: "Any") -> list:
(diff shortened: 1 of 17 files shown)


joptimus:
Final piece of the #296 exclude_patterns series (issue 5), split into two:

- This PR (4a): runtime exclude_patterns filtering for the 13 structured MCP tools + single-repo search, including the traversal tools (diagram edges, dependency graph, flows traces, risk blast-radius).
- Follow-up (4b): tool_context enrichment, tool_answer retrieval/citations, and the federated ("all repos") search path — these need filtering inside their context/retrieval pipelines rather than a result-level helper call.

Prior merged PRs in the series: #297 (init persist), #308 (git indexer + dynamic edges), #337 (config-change detection in update).

RaghavChamadiya (approved):
LGTM

PR #346, opened 2026-06-02
feat(cli): add advisory update check to doctor

### LINKED ISSUE #338: [Feature] Add Repowise CLI update check
## Problem

  It is easy to update an indexed repo with `repowise update`, but it is not obvious how to update the Repowise CLI itself.

  This matters because MCP configs usually call `"command": "repowise"`, so Claude/Codex/etc. runs whichever `repowise` executable is first on `PATH`. A user may be running an old Repowise version without realizing it, and there is no built-in check that shows the active CLI path, current version, latest version, or suggested update command.

  ## Proposed Solution

  Add a lightweight update check.

  Suggested first step:

  - Show CLI update status in `repowise doctor`
  - Add a check-only command such as `repowise upgrade --check` or `repowise self check`

  The check should show:

  - current Repowise version
  - latest available PyPI version, when network is available
  - active `repowise` executable path
  - suggested update command, such as `python -m pip install -U repowise`, `uv
  tool upgrade repowise`, or `pipx upgrade repowise`

  It should also remind users to restart their MCP client after updating.

  ## Alternatives Considered

  A fully automatic self-update command would be convenient, but Python installs vary a lot between `pip`, `uv tool`, `pipx`, virtualenvs, system Python, and editable installs. Auto-updating could update the wrong environment.

  A safer first version is check-only: detect that an update is available and print the correct command for the user to run.

  ## Additional Context

  The goal is to make Repowise feel easier to maintain for MCP users. Similar tools often notify users when their CLI is out of date. Repowise can start with a simple, low-risk version check before adding guided update behavior later.

joptimus:
## Summary

Surface when the installed `repowise` CLI is out of date and the exact command to update it — without ever auto-updating the user's environment. Implements #338, doctor-first and check-only (no `repowise upgrade`, no package-manager execution).

- **New `repowise.cli.update_check` helper** — conservative and never-raising. Reads `repowise.cli.__version__`, queries the PyPI JSON endpoint via the bundled `httpx` (short timeout), and suggests an install-method-aware upgrade command (`uv tool upgrade` / `pipx upgrade` / `<python> -m pip install -U` / editable `git pull && pip install -e .`). Simple numeric version comparison.
- **`CLI version` row in `repowise doctor`** — advisory only. Shows current vs latest plus **both** the resolved PATH executable and the full running command (these can differ). Never flips doctor's pass/fail, swallows network errors, and reminds the user to restart their MCP client only when an update is available.
- An unparsable PyPI version is treated as "unknown", not a false "up to date".
- Docs updated in `CLI_REFERENCE.md` and `USER_GUIDE.md`.

## Out of scope (by design)

No auto-update, no MCP config edits, no `repowise upgrade` command (name conflicts with `update --full`), no blocking on stale CLI. A future `repowise version --check` could follow.

## Testing

- `tests/unit/cli/test_update_check.py` (19) + `tests/unit/cli/test_doctor_cli_version.py` (4): all pass.
- `mypy` clean on both source files.

Full CLI unit suite has one pre-existing env-dependent failure: `test_ui_package.py::test_advanced_config_default_keys_no_fast`. The targeted update-check tests pass.

Closes #338

### DIFF
--- a/docs/CLI_REFERENCE.md
+++ b/docs/CLI_REFERENCE.md
@@ -527,6 +527,17 @@ repowise doctor --workspace              # every workspace repo
 repowise doctor --workspace --repair     # also drop dead entries / sync drift
 ```
 
+**CLI update check.** `doctor` also prints a best-effort `CLI version` row that
+compares your installed CLI against the latest release on PyPI and, when an
+update is available, shows the suggested upgrade command (e.g. `uv tool upgrade
+repowise`, `pipx upgrade repowise`, or `python -m pip install -U repowise`). It
+shows both the `repowise` resolved on your `PATH` and the command that launched
+the current process, since these can differ. This check is advisory: it never
+updates anything automatically and does not fail `doctor` when PyPI is
+unreachable. After upgrading, **restart Claude/Codex/Cursor or any MCP client**
+so it picks up the new executable. (A standalone `repowise version --check` may
+be added later.)
+
 ---
 
 ### `repowise delete [REPO_ID]`
--- a/docs/USER_GUIDE.md
+++ b/docs/USER_GUIDE.md
@@ -584,6 +584,14 @@ Checks:
 - `state.json` valid
 - Providers installed and importable
 - Stale page count
+- **CLI version** — best-effort check of your installed CLI against the latest
+  PyPI release
+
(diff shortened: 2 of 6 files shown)


RaghavChamadiya (approved):
LGTM

PR #386, opened 2026-06-05
fix(ingestion): rescue local Express route middleware from unused-export detection

### LINKED ISSUE #373: [Bug] Express route middleware identifiers can be reported as unused exports
## Describe the Bug

Express route middleware / handler identifiers passed into `router.METHOD(...)` can be reported as unused exports with high confidence, even though they are live route callbacks.

A minimal shape is:

```js
function validateRequest(req, res, next) {
  next();
}

function logRequest(req, res, next) {
  next();
}

router.post("/example", logRequest, validateRequest, async (req, res) => {
  res.json({ ok: true });
});
```

Those middleware functions are not called with normal `fn()` syntax, but Express invokes them at runtime through the route pipeline. Repowise currently appears to miss that usage signal.

## Steps to Reproduce

1. Create an Express route module with local middleware functions:

   ```js
   const express = require("express");
   const router = express.Router();

   function validateRequest(req, res, next) {
     next();
   }

   function logRequest(req, res, next) {
     next();
   }

   router.post("/example", logRequest, validateRequest, async (req, res) => {
     res.json({ ok: true });
   });

   module.exports = router;
   ```

2. Import/mount that router from an app file so the route module itself is live.
3. Run Repowise analysis / dead-code detection.
4. Open the web dashboard dead-code view, or inspect the stored dead-code findings.

## Expected Behavior

Identifiers passed as Express route middleware/handlers should count as usage of those functions.

In the example above, `logRequest` and `validateRequest` should not be reported as unused exports, because they are live callbacks registered with Express.

## Actual Behavior

The route file can be recognized as live because it is imported/mounted by the app, but local middleware identifiers passed to `router.post(...)` do not get usage/call edges. The analyzer can then report those functions as unused exports with high confidence and mark them as safe cleanup candidates.

That makes the dashboard look very certain about deleting code that is actually part of the Express request pipeline.

## Environment

- OS: macOS 26.3.1
- Python version: 3.14.2
- Repowise version: 0.16.0 (`repowise --version`)
- Installation method: local Python CLI / pip-style install

## Additional Context

This seems related to existing framework-edge support. Express handling appears to cover `app.use(routerVar)` / Nest module wiring, while other TS/JS router DSL handling already covers patterns like `app.get("/path", handler)`.

A narrowly scoped fix would be to extend Express route handling to recognize middleware/handler identifiers in:

- `router.get/post/put/delete/patch/options/head/all/use(...)`
- `app.get/post/put/delete/patch/options/head/all/use(...)`

Ideally this should mark the specific local or imported handler symbol as used, rather than marking the whole file as live. That would avoid hiding genuinely unused exports in the same route file.

I would avoid a broad generic rule like "any identifier passed as any callback is live" for now, since that could hide real dead code in unused arrays, registries, or local maps. Express route registration is a narrower and more defensible starting point.

joptimus:
## Summary

- Detect Express route registrations on `express()`/`Router()`-initialised receivers and emit a `reads` edge from the file's `::__module__` symbol to each same-file middleware function passed by bare name, so local route middleware is no longer reported as an unused export.
- Source the edge from `::__module__` (not the file node) to avoid clobbering the existing file→symbol `defines` edge in the single-edge `DiGraph`; a balanced-paren scan handles multiline calls and middleware factories before the handler, and string literals are stripped so route paths aren't mistaken for identifiers.
- Keeps the fix narrow to Express: imported middleware (already rescued by import edges) and non-`express()`/`Router()` receivers (e.g. `cache.get(fn)`) are left untouched.

## Related Issues

Fixes #373

## Test Plan

- [x] Tests pass (`pytest`) — `tests/unit/ingestion/test_express_framework_edges.py` (11 passed); `tests/unit/analysis tests/unit/dead_code tests/unit/ingestion` + TS/CommonJS dead-code integration (1303 passed, 2 xfailed)

New coverage: route-method reads edges, `app.use` middleware, `defines`-edge preservation, nested middleware-factory args, the `.js`/CommonJS multiline `router.post(...)` shape from #373, non-route receivers in an Express file, non-Express files gated out, and a dead-code regression proving the middleware exports are no longer flagged.

## Checklist

- [ ] I have updated documentation if needed *(no docs impact)*

### DIFF
--- a/packages/core/src/repowise/core/ingestion/framework_edges/express.py
+++ b/packages/core/src/repowise/core/ingestion/framework_edges/express.py
@@ -27,6 +27,44 @@ _NEST_MODULE_RE = re.compile(r"@Module\s*\(\s*\{([^}]*)\}\s*\)", re.DOTALL)
 _NEST_ARRAY_FIELD_RE = re.compile(r"\b(?:controllers|providers|imports|exports)\s*:\s*\[([^\]]*)\]")
 _IDENT_RE = re.compile(r"\b([A-Z]\w*)\b")
 
+_EXPRESS_RECEIVER_RE = re.compile(
+    r"\b(\w+)\s*(?::[^=;\n]+)?=\s*(?:express\s*\(\s*\)|(?:express\s*\.\s*)?Router\s*\()"
+)
+_ROUTE_CALL_START_RE = re.compile(
+    r"\b(\w+)\s*\.\s*(?:get|post|put|delete|patch|options|head|all|use)\s*\("
+)
+_STRING_LITERAL_RE = re.compile(r"'(?:[^'\\]|\\.)*'|\"(?:[^\"\\]|\\.)*\"|`(?:[^`\\]|\\.)*`")
+_ARG_IDENT_RE = re.compile(r"[A-Za-z_$][\w$]*")
+
+
+def _match_paren(text: str, open_idx: int) -> int:
+    depth = 0
+    i = open_idx
+    n = len(text)
+    while i < n:
+        c = text[i]
+        if c in "\"'`":
+            i += 1
+            while i < n and text[i] != c:
+                i += 2 if text[i] == "\\" else 1
+        elif c == "(":
+            depth += 1
+        elif c == ")":
+            depth -= 1
+            if depth == 0:
+                return i
+        i += 1
+    return -1
+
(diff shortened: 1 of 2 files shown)


RaghavChamadiya (approved):
LGTM

PR #387, opened 2026-06-05
fix(health): use whole-file NLOC for health file metrics

### LINKED ISSUE #375: [Bug?] Health NLOC may undercount route-style files and inflate churn risk
## Describe the Bug

I am seeing code health scores move in a way that looks surprising because file-level `nloc` appears to be calculated from detected function or method bodies, not from the whole file.

That may be intentional for complexity metrics, but it seems risky when the same `nloc` is used for file-level signals like churn risk and score weighting.

In my `game-server` repo, one JavaScript API route file is about 922 lines long, with roughly 728 non-empty/non-comment lines by a simple local count. Repowise stored `nloc = 41` for that file in `health_file_metrics`.

That small denominator then made the churn-risk finding much harsher:

```text
90-day churn rewrote 1076% of the file's lines
441 changed lines over 41 NLOC
```

Using the approximate whole-file NLOC, the same churn would be closer to:

```text
441 / 728 ~= 61%
```

This caused the file score to drop from about `9.7` to `6.2` between snapshots, even though the actual latest diff touching that file was small. The trend page then showed it as one of the biggest score drops, which was confusing because the main work I had done was improving a different service.

## Steps to Reproduce

1. Run `repowise init` or `repowise update` on a JavaScript or TypeScript repo with Express-style route modules.
2. Include route files where much of the executable logic is in route callbacks or top-level module code.
3. Open the Health tab or inspect `health_file_metrics`.
4. Compare the stored `nloc` for those files against a simple whole-file non-empty/non-comment line count.
5. Check churn-risk findings for files where stored `nloc` is much smaller than the real file size.

## Expected Behavior

For file-level health metrics, I would expect Repowise to distinguish between:

```text
file_nloc: whole-file non-comment/non-blank lines
function_nloc: lines covered by detected function or method bodies
```

Then file-level process signals like churn risk would use `file_nloc`, while complexity biomarkers could still use function-level metrics.

If this is by design, it would be helpful to understand the intended meaning of `nloc` in `health_file_metrics`, because the current value is presented and used as a file-level denominator.

## Actual Behavior

The stored `nloc` appears to be the sum of detected function or method body NLOC. For JS/TS route modules, this can undercount the real file substantially.

In the same repo, several API route files had very small or zero stored NLOC values despite being real source files:

```text
example API route file: actual ~922 lines, simple NLOC ~728, stored nloc 41
other API route files: stored nloc values of 0, 3, 20, 35, etc.
```

That makes churn-per-NLOC and related health scores look much worse than they should for files where the AST walker does not cover most of the file.

## Environment

- OS: macOS
- Python version: 3.14.2
- Repowise version: 0.16.0
- Installation method: pip

## Additional Context

This seems most visible in JavaScript and TypeScript API route files, especially Express-style modules where route callbacks and top-level wiring are common.

It may also affect other languages or patterns where meaningful code can live outside normal detected function/method nodes:

- Python scripts with module-level setup or `if __name__ == "__main__"` blocks
- C# top-level programs or minimal API style code
- C++ global/static initializers, macro-heavy code, or lambdas assigned outside functions
- JavaScript/TypeScript config modules, route modules, and top-level callbacks

The trend page adds to the confusion because “Files that moved most since last index” appears to mean “score changed most,” not necessarily “file changed most.” That part may just need clearer wording, but the NLOC denominator issue looks like the main cause of the misleading score movement.

Could you confirm whether `health_file_metrics.nloc` is intended to represent whole-file NLOC or only function-covered NLOC? If it is meant to be whole-file NLOC, I think this is a scoring bug. If it is intentionally function-covered NLOC, maybe churn risk should use a separate whole-file denominator.

joptimus:
## Summary

- Fixes `health_file_metrics.nloc` storing the sum of detected function-body NLOC instead of whole-file NLOC.
- Adds `FileComplexity.file_nloc`. On successful parses it counts source lines with at least one non-comment token, so comment-only lines are excluded while code with trailing comments still counts. Parser failures and unsupported languages fall back to a plain non-blank line count.
- Updates `HealthAnalyzer._evaluate_file` to use `fcx.file_nloc`. Per-function `nloc` is unchanged for biomarkers that need function-level measurements.

## Context

Fixes #375.

Route-style files and modules with top-level logic could previously store very small NLOC values because top-level callbacks, imports, wiring, and config were outside detected function bodies. That made downstream health signals like churn risk and effort buckets look much worse than they should.

## Test plan

- [x] Added coverage for JS and Python comment-only lines, trailing comments, multiline strings, route-style JS files, unsupported-language fallback, and the health metric using `file_nloc`.
- [x] `uv run pytest tests/unit/health -q` - 270 passed

### DIFF
--- a/packages/core/src/repowise/core/analysis/health/complexity/walker.py
+++ b/packages/core/src/repowise/core/analysis/health/complexity/walker.py
@@ -118,6 +118,7 @@ class FileComplexity:
 
     functions: list[FunctionComplexity]
     classes: list[ClassComplexity]
+    file_nloc: int = 0
 
 
 # Leaf node types that carry a declared name at the bottom of a C/C++
@@ -164,6 +165,41 @@ def _count_nloc(node: Node, source: bytes) -> int:
     except Exception:
         return end - start + 1
     return sum(1 for line in snippet.splitlines() if line.strip())
+
+
+def _count_file_nloc(source: bytes) -> int:
+    """Count non-blank lines in *source* bytes (plain fallback, no tree)."""
+    try:
+        text = source.decode("utf-8", errors="replace")
+    except Exception:
+        return 0
+    return sum(1 for line in text.splitlines() if line.strip())
+
+
+def _count_file_nloc_tree(root_node: Node, source: bytes) -> int:
+    """Count lines that have at least one non-comment token.
+
+    Lines where all content is inside comment nodes are excluded; lines
+    with real code plus a trailing comment still count.
+    """
+    try:
+        lines = source.decode("utf-8", errors="replace").splitlines()
+    except Exception:
+        return 0
+    code_lines: set[int] = set()
+    stack = [root_node]
+    while stack:
+        node = stack.pop()
+        if "comment" in node.type:
+            continue
+        if not node.children and node.start_byte < node.end_byte:
+            for line in range(node.start_point[0], node.end_point[0] + 1):
+                if line < len(lines) and lines[line].strip():
+                    code_lines.add(line)
+        else:
+            for child in node.children:
+                stack.append(child)
+    return len(code_lines)
 
 
 def _is_boolean_operator(node: Node, lmap: LanguageNodeMap) -> bool:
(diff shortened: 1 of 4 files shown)


RaghavChamadiya (approved):
LGTM

PR #432, opened 2026-06-08
fix(web): prevent workspace redirect loop

### LINKED ISSUE #431: [Bug] Workspace subroute navigation can enter redirect loop back to /workspace
﻿## Describe the Bug

Workspace subroute navigation can enter a redirect loop back to `/workspace` and eventually crash the UI.

This affects workspace navigation links such as:

- `/workspace/contracts`
- `/workspace/co-changes`

The links themselves appear to be correct. The problem appears to be caused by conflicting workspace-mode redirects between the nested `/workspace` layout guard and the root `/` page redirect.

## Steps to Reproduce

1. Run Repowise in workspace mode.
2. Open the workspace dashboard at `/workspace`.
3. Click a workspace subroute link, such as:
   - Contracts in the navigation panel
   - Co-Changes in the navigation panel
   - View all from the Top Cross-Repo Co-Changes card
4. Observe that the app attempts to navigate to the subroute.
5. The app redirects back to `/workspace`, repeats, and can eventually crash the UI.

## Expected Behavior

Workspace subroute links should navigate directly to their target pages:

```text
/workspace/contracts
/workspace/co-changes
```

The workspace route guard should only redirect away from workspace routes when it can conclusively determine that the app is not running in workspace mode.

## Actual Behavior

Workspace subroute navigation can bounce through redirects instead of rendering the target route.

Likely redirect chain:

```text
/workspace/contracts -> workspace layout guard -> /
/ -> root workspace redirect -> /workspace
/workspace -> workspace layout guard -> /
...loop
```

This appears to happen when the `/workspace` layout guard treats a failed, unavailable, or stale workspace check the same as `is_workspace === false`.

## Environment

- Repowise version: 0.18.0

## Additional Context

The buttons/links appear to point to the correct URLs. The issue seems to be in route guarding behavior.

Relevant behavior:

- The nested `/workspace` layout redirects to `/` when its workspace check is false or fails.
- The root `/` page redirects to `/workspace` when it detects workspace mode.
- If these checks disagree, or one request sees stale/failing state, the app can loop.

Possible fix directions:

- Do not redirect from `/workspace/*` on API failure.
- Only redirect when the workspace check conclusively returns `is_workspace === false`.
- Use a fresh/no-store workspace check for route guards.
- Consider removing the nested `/workspace` guard and letting root/sidebar mode detection own workspace entry routing.

joptimus:
## Summary
- prevent workspace subroutes from redirecting away when the workspace API is temporarily unavailable
- use uncached workspace mode checks for the root and workspace layout redirect decisions
- add web Vitest coverage so the redirect behavior is pinned in future runs

Fixes #431

## Testing
- npm run test --workspace packages/web
- npm test
- npm run lint
- npm run type-check
- git diff --check

## Notes
- uv run pytest tests/unit/ was attempted, but the full unit suite timed out locally after the uv environment was created.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -14736,7 +14736,8 @@
         "eslint": "^9",
         "eslint-config-next": "^15.2.4",
         "tailwindcss": "^4.0.0",
-        "typescript": "^5.7"
+        "typescript": "^5.7",
+        "vitest": "^4.1.5"
       },
       "engines": {
         "node": ">=20.0.0"
--- a/packages/web/package.json
+++ b/packages/web/package.json
@@ -8,7 +8,8 @@
     "build": "next build",
     "start": "next start --port 3000",
     "lint": "next lint",
-    "type-check": "tsc --noEmit"
+    "type-check": "tsc --noEmit",
+    "test": "vitest run"
   },
   "dependencies": {
     "@lottiefiles/dotlottie-react": "^0.19.4",
@@ -43,7 +44,8 @@
     "eslint": "^9",
     "eslint-config-next": "^15.2.4",
     "tailwindcss": "^4.0.0",
-    "typescript": "^5.7"
+    "typescript": "^5.7",
+    "vitest": "^4.1.5"
   },
   "engines": {
     "node": ">=20.0.0"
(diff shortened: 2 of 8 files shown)


RaghavChamadiya (approved):
LGTM

PR #436, opened 2026-06-09
feat(providers): add opencode CLI provider

ramsterr:
## Summary

-

## Test Plan

- [x] Tests pass (`pytest`)

## Checklist

A new OpenCode CLI LLM provider following the same subprocess-provider pattern established by codex_cli (PR #348).
What this adds
Provider (opencode) — drives generation through the local OpenCode CLI:
- opencode run --format json --dangerously-skip-permissions subprocess, stdin-fed combined system+user prompt

- argv-based subprocess (no shell), JSONL parsing tolerant of non-JSON noise, async serialization semaphore, 600s exec timeout with kill/reap, and explicit subprocess-transport cleanup

- model names validated via regex ^[a-zA-Z0-9][a-zA-Z0-9._/\-]*$ before reaching the subprocess — shell metacharacters rejected

- zero-cost usage tracking (opencode/* = $0.00), flagged estimated when OpenCode omits usage

- no API key storage — OpenCode manages its own auth through opencode providers
Interactive experience — detects OpenCode on PATH, shows status in the provider table, and displays helpful install/setup messages when the CLI is missing:

[6] opencode (uses opencode CLI auth)   ✓ opencode CLI available   opencode/default
Documentation — docs/OPENCODE.md and website/opencode.md covering install, model selection, security, and comparison with Codex CLI. All provider-list reference docs updated (CLI_REFERENCE, USER_GUIDE, QUICKSTART, configuration, architecture).

Tests — 22 unit tests in tests/unit/test_providers/test_opencode_provider.py covering: model name validation (safe/shell), provider metadata, CLI detection, subprocess invocation correctness, JSONL parsing (tokens/noise/missing-usage), error handling (nonzero exit, FileNotFoundError, empty output, error JSONL), semaphore serialization, timeout with kill, and transport cleanup. All tests mock the subprocess — no real CLI calls.

CONTRIBUTING.md — new 6-step provider-adding guide referencing opencode.py as the reference implementation for local CLI providers.

Comparison with codex_cli
Aspect	opencode	codex_cli
CLI command	opencode run --format json	codex exec --json
Reasoning	N/A (OpenCode manages it)	model_reasoning_effort validated against catalog
Model discovery	opencode models	codex debug models --bundled
Editor integration	None	.codex/config.toml, hooks, AGENTS.md, plugin
Files changed	23	82
Tests	22 new, 209 total pass	1,023 total pass

Notes for reviewers
- Follows the exact same touch-points as codex_cli (registry → rate limiter → provider config → CLI selection → web UI → cost tracking → docs). Where codex adds editor integration and reasoning across all providers, opencode is intentionally minimal — it only adds the provider.

- All 209 provider tests pass. No modifications to existing test files beyond the builtin-count assertion in test_registry.py.

- Security: create_subprocess_exec callback args only (no shell), model name regex gate, path resolution, no key storage, subprocess semaphore, hard timeout.

- Install URL sourced from opencode.ai/install (not npm).

### DIFF
--- a/packages/core/src/repowise/core/providers/llm/opencode.py
+++ b/packages/core/src/repowise/core/providers/llm/opencode.py
@@ -1,0 +1,339 @@
+"""OpenCode CLI provider for repowise.
+
+This provider delegates generation to the local OpenCode CLI via ``opencode run``.
+It uses the user's existing OpenCode installation and authentication (``opencode providers``)
+without requiring a separate API key.
+
+Security: uses ``asyncio.create_subprocess_exec`` (no shell), validates model names
+against a safe character set, and resolves all paths before passing them to the
+subprocess.
+"""
+
+from __future__ import annotations
+
+import asyncio
+import contextlib
+import json
+import re
+import shutil
+from pathlib import Path
+from typing import Any
+
+import structlog
+
+from repowise.core.providers.llm.base import (
+    BaseProvider,
+    CacheHint,
+    GeneratedResponse,
+    ProviderError,
+    ProviderModelOption,
+)
+from repowise.core.rate_limiter import RateLimiter
+from repowise.core.reasoning import ReasoningMode
+
+log = structlog.get_logger(__name__)
+
+_DEFAULT_MODEL_LABEL = "opencode/default"
+_EXEC_TIMEOUT_SECONDS = 600
+_MAX_STDERR_CHARS = 1_000
+
+_MODEL_NAME_RE = re.compile(r"^[a-zA-Z0-9][a-zA-Z0-9._/\-]*$")
+
+
+def _resolve_opencode_executable() -> str | None:
+    return shutil.which("opencode")
+
+
+def _validate_model_name(model: str) -> None:
+    if not _MODEL_NAME_RE.match(model):
+        raise ProviderError(
+            "opencode",
+            f"Invalid model name {model!r}. Model names may only contain "
+            "alphanumeric characters, dots, hyphens, underscores, and forward slashes.",
+        )
+
+
+def _normalize_model(model: str | None) -> str | None:
+    if not model:
+        return None
+    if model == _DEFAULT_MODEL_LABEL:
+        return None
+    if model.startswith("opencode/"):
+        suffix = model.removeprefix("opencode/")
+        return suffix or None
+    return model
+
+
+def _model_label(model: str | None) -> str:
+    native = _normalize_model(model)
+    return f"opencode/{native}" if native else _DEFAULT_MODEL_LABEL
+
+
+def _combine_prompt(system_prompt: str, user_prompt: str) -> str:
+    return (
+        "System instructions for this task:\n\n"
+        f"{system_prompt.strip()}\n\n"
+        "---\n\n"
+        "User request and context:\n\n"
+        f"{user_prompt.strip()}"
+    )
+
+
+def _parse_jsonl(stdout: str) -> tuple[str, dict[str, Any]]:
+    content_parts: list[str] = []
+    usage: dict[str, Any] = {}
+
+    for raw_line in stdout.splitlines():
+        line = raw_line.strip()
+        if not line:
+            continue
+        try:
+            event = json.loads(line)
+        except json.JSONDecodeError:
+            continue
+
+        event_type = event.get("type")
+        if event_type == "text":
+            part = event.get("part") or {}
+            text = part.get("text")
+            if isinstance(text, str) and text:
+                content_parts.append(text)
+        elif event_type == "step_finish":
+            part = event.get("part") or {}
+            tokens = part.get("tokens")
+            if isinstance(tokens, dict):
+                usage = tokens
+
+    return "\n".join(content_parts), usage
+
+
+def _tail(text: str, max_chars: int = _MAX_STDERR_CHARS) -> str:
+    text = text.strip()
+    if len(text) <= max_chars:
+        return text
+    return text[-max_chars:]
+
+
+def _error_message(stderr: str, stdout: str, returncode: int) -> str:
+    for candidate in (_tail(stderr), _tail(stdout)):
+        if not candidate:
+            continue
+        if candidate.lstrip().startswith(("{", "[")):
+            continue
+        return candidate
+
+    try:
+        for raw_line in stdout.splitlines():
+            line = raw_line.strip()
+            if not line:
+                continue
+            event = json.loads(line)
+            if event.get("type") == "error":
+                err = event.get("error") or {}
+                msg = err.get("data", {}).get("message") or err.get("name", "")
+                if msg:
+                    return str(msg)
+    except (json.JSONDecodeError, Exception):
+        pass
+
+    return f"opencode run exited with {returncode}"
+
+
+class OpenCodeProvider(BaseProvider):
+    """LLM provider backed by ``opencode run``.
+
+    Uses the local opencode CLI for generation. Does not require an API key —
+    opencode manages its own authentication via ``opencode providers``.
+
+    Args:
+        model:     Optional model identifier. If omitted, opencode uses its
+                   default.  Accepts ``opencode/big-pickle``-style labels as
+                   well as bare model slugs.
+        repo_path: Working directory passed to opencode via ``--dir``.
+        rate_limiter: Serializes subprocess calls by default.
+    """
+
+    def __init__(
+        self,
+        model: str | None = None,
+        repo_path: str | Path | None = None,
+        rate_limiter: RateLimiter | None = None,
+    ) -> None:
+        opencode_cmd = _resolve_opencode_executable()
+        if not opencode_cmd:
+            raise ProviderError(
+                "opencode",
+                "OpenCode CLI is not installed.\n\n"
+                "Installation:\n"
+                "  curl -fsSL https://opencode.ai/install | bash\n\n"
+                "After installing, run 'opencode' once to set up your model provider "
+                "and authenticate. No API keys are managed by repowise — opencode "
+                "handles all authentication.\n\n"
+                "To choose a model:\n"
+                "  opencode models                 # list available models\n"
+                "  repowise init --provider opencode --model opencode/deepseek-v4-pro\n\n"
+                "More info: https://opencode.ai",
+            )
+        self._opencode_cmd = opencode_cmd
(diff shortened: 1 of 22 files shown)


packages/core/src/repowise/core/providers/llm/opencode.py
```diff
@@ -0,0 +1,339 @@
...
+            self._semaphore_loop = loop
+        return self._subprocess_semaphore  # type: ignore[return-value]
+
+    def _build_command(self) -> list[str]:
+        cmd = [
+            self._opencode_cmd,
+            "run",
+            "--format",
+            "json",
+            "--dangerously-skip-permissions",
```
swati510: **Blocking, security: opencode runs unsandboxed against the repo**

`--dangerously-skip-permissions` turns off every safety prompt, so opencode can edit files and run shell commands with no confirmation. Since we feed repo content into the prompt, a malicious file could trigger destructive commands (prompt injection). codex_cli avoids this with `--sandbox read-only`; we have nothing equivalent.

opencode has no inline permission flag, but explicit `deny` rules are honored, so inject a read-only config via `OPENCODE_CONFIG_CONTENT` (highest precedence, a project-local `opencode.json` can't loosen it) and drop the dangerous flag. read/glob/grep stay allowed, so it can still read the repo and write docs. Permission keys/values: https://opencode.ai/docs/permissions/

```python
import os  # add to imports

# Read-only: deny mutating/network tools so a malicious repo can't drive
# destructive actions. `task` is denied too because deny can leak through
# the subagent path: https://github.com/anomalyco/opencode/issues/23519
_READONLY_OPENCODE_CONFIG = json.dumps(
    {"permission": {k: "deny" for k in
     ("edit", "bash", "webfetch", "websearch", "external_directory", "task")}}
)
```

```python
    def _build_command(self) -> list[str]:
        cmd = [
            self._opencode_cmd,
            "run",
            "--format",
            "json",
            "--dir",
            str(self._repo_path),
        ]
        if self._model:
            cmd.extend(["--model", self._model])
        return cmd
```

```python
                proc = await asyncio.create_subprocess_exec(
                    *cmd,
                    stdin=asyncio.subprocess.PIPE,
                    stdout=asyncio.subprocess.PIPE,
                    stderr=asyncio.subprocess.PIPE,
                    env={**os.environ, "OPENCODE_CONFIG_CONTENT": _READONLY_OPENCODE_CONFIG},
                )
```

Dropping the flag is safe here because nothing is left at "ask" (non-denied tools default to allow), so it won't hang in headless mode. Please confirm `OPENCODE_CONFIG_CONTENT` against the installed opencode version.

packages/core/src/repowise/core/providers/llm/opencode.py
```diff
@@ -0,0 +1,339 @@
...
+            "alphanumeric characters, dots, hyphens, underscores, and forward slashes.",
+        )
+
+
+def _normalize_model(model: str | None) -> str | None:
+    if not model:
+        return None
+    if model == _DEFAULT_MODEL_LABEL:
+        return None
+    if model.startswith("opencode/"):
```
swati510, line 71: **Model selection is broken: the `opencode/` prefix strip drops the provider segment**

Knock-on at `opencode.py:174`, `provider_config.py:90`, `provider_selection.py:212`.

opencode splits `--model` on the first `/` into `provider/model`; a bare id with no slash is read as a provider name and fails with ProviderNotFound (https://opencode.ai/docs/providers/). We strip the leading `opencode/`, so everything but `default` loses its provider:

- `opencode/big-pickle` -> `--model big-pickle` -> ProviderNotFound
- `opencode/deepseek-v4-pro` -> `--model deepseek-v4-pro` -> ProviderNotFound

Only `opencode/default` works (it sends no `--model`). Fix the normalization so the label keeps the full opencode string, e.g. `opencode/deepseek/deepseek-v4-pro` -> `--model deepseek/deepseek-v4-pro`.

packages/core/src/repowise/core/providers/llm/opencode.py
```diff
@@ -0,0 +1,339 @@
...
+                proc = await asyncio.create_subprocess_exec(
+                    *cmd,
+                    stdin=asyncio.subprocess.PIPE,
+                    stdout=asyncio.subprocess.PIPE,
+                    stderr=asyncio.subprocess.PIPE,
+                )
+            except FileNotFoundError as exc:
+                raise ProviderError(
+                    "opencode",
+                    "OpenCode CLI not found. Install it with: npm install -g @anthropicai/opencode",
```
swati510: **Blocking: wrong install command**

`@anthropicai/opencode` doesn't exist, and it contradicts the curl install used everywhere else in this file.

```python
                    "OpenCode CLI not found. Install it with: "
                    "curl -fsSL https://opencode.ai/install | bash",
```

packages/core/src/repowise/core/rate_limiter.py
```diff
@@ -66,6 +66,8 @@ class RateLimitConfig:
     "ollama": RateLimitConfig(requests_per_minute=1_000, tokens_per_minute=10_000_000),
     "litellm": RateLimitConfig(requests_per_minute=1_000, tokens_per_minute=2_000_000),
     "deepseek": RateLimitConfig(requests_per_minute=1_000, tokens_per_minute=5_000_000),
+    # opencode runs locally via subprocess — serialized by an asyncio semaphore
+    "opencode": RateLimitConfig(requests_per_minute=20, tokens_per_minute=2_000_000),
```
swati510: **Dead rate-limiter entry**

registry.py:128 skips the limiter for opencode, so this never runs. Drop both lines.

packages/core/src/repowise/core/providers/llm/opencode.py
```diff
@@ -0,0 +1,339 @@
...
+        if event_type == "text":
+            part = event.get("part") or {}
+            text = part.get("text")
+            if isinstance(text, str) and text:
+                content_parts.append(text)
+        elif event_type == "step_finish":
+            part = event.get("part") or {}
+            tokens = part.get("tokens")
+            if isinstance(tokens, dict):
+                usage = tokens
```
swati510: **Token count is low on multi-step runs**

Only the last `step_finish` survives, so multi-step runs undercount tokens. No cost impact (opencode is $0), but the usage ledger will be wrong. Accumulate across steps instead of overwriting.

swati510 (commented):
Reviewed the opencode provider. Two blocking items (sandbox + broken model selection), one wrong install string, plus a few cleanups. Details inline.

packages/core/src/repowise/core/providers/llm/opencode.py
```diff
@@ -0,0 +1,339 @@
...
+            line = raw_line.strip()
+            if not line:
+                continue
+            event = json.loads(line)
+            if event.get("type") == "error":
+                err = event.get("error") or {}
+                msg = err.get("data", {}).get("message") or err.get("name", "")
+                if msg:
+                    return str(msg)
+    except (json.JSONDecodeError, Exception):
```
swati510: **Redundant except**

`Exception` already covers `json.JSONDecodeError`, so:

```python
    except Exception:
```

ramsterr:
> Reviewed the opencode provider. Two blocking items (sandbox + broken model selection), one wrong install string, plus a few cleanups. Details inline.

Reviewer Issues Fixed
1. Security — Replaced --dangerously-skip-permissions with OPENCODE_CONFIG_CONTENT env var that enforces a read-only permission profile (deny: edit, bash, external_directory, doom_loop). read/glob/grep stay allowed.

2. Model selection — Hardcoded model names like opencode/big-pickle and opencode/deepseek-v4-pro were in wrong format (missing provider prefix). Updated all references to use provider/model format (e.g. deepseek/deepseek-v4-pro).

3. Model discovery — Added dynamic model discovery via opencode models so available models in your opencode installation show up in repowise, instead of a single hardcoded default.

4. Install command — Fixed FileNotFoundError message from npm install -g @anthropicai/opencode (nonexistent package) to curl -fsSL https://opencode.ai/install | bash.

5. Dead rate-limiter — Removed dead "opencode" entry from PROVIDER_DEFAULTS in rate_limiter.py. Registry already skips the limiter for opencode.

6. Token counting — _parse_jsonl now accumulates tokens across multiple step_finish events instead of overwriting (fixes undercounting on multi-step runs).

7. Redundant except — except (json.JSONDecodeError, Exception) → except Exception.
Tests
- 209 original tests + 4 new (model parsing, catalog/fallback model options) = 213 passing.

packages/core/src/repowise/core/providers/llm/opencode.py
```diff
@@ -0,0 +1,437 @@
...
+_CATALOG_TIMEOUT_SECONDS = 10
+_MAX_STDERR_CHARS = 1_000
+
+_MODEL_NAME_RE = re.compile(r"^[a-zA-Z0-9][a-zA-Z0-9._/\-]*$")
+
+_OPENCODE_READONLY_CONFIG = json.dumps(
+    {
+        "permission": {
+            "edit": "deny",
+            "bash": "deny",
```
swati510: Close, but the read-only profile isn't airtight yet. The deny set is `edit, bash, external_directory, doom_loop`, but unlisted permissions default to allow, so `task`, `webfetch`, and `websearch` are still on.

`task` is the one that matters: a spawned subagent can do the `edit`/`bash` the top-level deny blocks, since deny leaks through the subagent path (https://github.com/anomalyco/opencode/issues/23519). `webfetch`/`websearch` also leave a network egress path for exfiltration from private repos. Denying them costs nothing (deny is not "ask", so no headless hang).

```python
_OPENCODE_READONLY_CONFIG = json.dumps(
    {"permission": {k: "deny" for k in
     ("edit", "bash", "webfetch", "websearch", "external_directory", "doom_loop", "task")}}
)
```

swati510 (commented):
Re-reviewed the fixes. Model selection, install, rate-limiter, token accumulation, and the except clause are all sorted, and the `opencode models` discovery parses cleanly. One gap left on the read-only profile, noted inline.

packages/core/src/repowise/core/providers/llm/opencode.py
ramsterr: Added webfetch, websearch, and task to the deny set ,task was the critical gap since deny leaks through subagents. Should work now.

swati510 (approved):
LGTM

PR #441, opened 2026-06-09
fix(workspace): use canonical health score in graph (#429)

### LINKED ISSUE #429: [Bug] Workspace graph repo score appears identical and may not use actual health score
﻿## Describe the Bug

The workspace cross-repo graph shows a repo score for each node, but the score appears to be the same across repos and does not seem to reflect the actual repo health score shown elsewhere in the app.

From the code path, the workspace graph endpoint appears to derive a synthetic `health_score` from doc coverage and hotspot count rather than reading an existing repo/module health score. This can make every repo show the same score when coverage and hotspot inputs are similar, which makes the graph misleading.

## Steps to Reproduce

1. Run `repowise update --workspace` on a workspace with multiple repositories.
2. Open the workspace dashboard.
3. Look at the cross-repo workspace graph.
4. Compare the score shown inside each repo node.
5. Compare those scores with the repo health information shown elsewhere in the app.

## Expected Behavior

The workspace graph should either:

- show the actual repo health score used elsewhere in Repowise, or
- clearly label the value as a derived/simple workspace score if it is not the real health score.

Repos with different health characteristics should not all appear to have the same score unless the underlying health data is genuinely identical.

## Actual Behavior

The graph shows the same or very similar score for every repo, making it look like all repos have identical health.

This appears to happen because the workspace graph endpoint computes a simplified score from doc coverage and hotspot count instead of using the actual health scoring source.

## Environment

- Repowise version: 0.18.0

## Additional Context

This may be a backend/API issue rather than a graph rendering issue.

Possible fixes:

- Use the same repo health score source used by other repo health views.
- Include the existing health score in the workspace graph API response.
- If the score remains derived, rename or relabel it in the UI to avoid implying it is the canonical repo health score.

joptimus:
Fixes #429.

## Problem

The workspace graph endpoint displayed a synthetic repo health score computed from documentation coverage and hotspot ratio. That score is not the same as the health score used by the per-repo health views, so repos with similar coverage/hotspot ratios collapsed to nearly identical graph scores even when their actual code-health metrics differed.

The UI also rendered the synthetic fallback as if it were the canonical health score, which made the graph look more authoritative than the data allowed when health metrics were missing.

## Fix

GET /api/workspace/graph now reads the canonical repo health score from each repo-local health_file_metrics table when available, using the same NLOC-weighted average-health basis as the repo health overview and scaling it to the graph's 0-100 display range.

When a repo has no health metrics yet, the endpoint keeps the existing derived score as a fallback and marks it with health_score_source=derived. Graph nodes use that source to label fallback values as estimated, while canonical scores remain labeled as health.

## Changes

- **server** — adds a small module-health helper for repo-local canonical score lookup and wires it into the workspace graph response.
- **schema/types** — adds health_score_source to workspace graph nodes.
- **ui** — rounds graph ring display and labels derived fallback scores as est..
- **tests** — covers canonical score selection and derived fallback behavior with a minimal repo-local wiki.db fixture.

## Tests

- uv run pytest tests/unit/server/test_workspace_router.py — 19 passed.
- uv run ruff check --extend-ignore B008,B023 packages/server/src/repowise/server/routers/workspace.py packages/server/src/repowise/server/schemas/workspace.py packages/server/src/repowise/server/services/module_health.py tests/unit/server/test_workspace_router.py — clean. (B008/B023 are pre-existing in workspace.py.)
- 
pm.cmd run type-check — clean across workspaces.
- 
pm.cmd run lint — clean; Next.js emitted its existing deprecation/root-inference warnings.

### DIFF
--- a/packages/server/src/repowise/server/routers/workspace.py
+++ b/packages/server/src/repowise/server/routers/workspace.py
@@ -7,7 +7,6 @@ startup from ``.repowise-workspace/`` JSON files) — no DB access needed.
 
 from __future__ import annotations
 
-import asyncio
 import logging
 import sqlite3
 from pathlib import Path
@@ -16,7 +15,6 @@ from fastapi import APIRouter, Depends, HTTPException, Query, Request
 
 from repowise.server.deps import (
     get_cross_repo_enricher,
-    get_db_session,
     get_workspace_config,
     resolve_session_factory,
     verify_api_key,
@@ -35,6 +33,7 @@ from repowise.server.schemas import (
     WorkspaceRepoEntry,
     WorkspaceResponse,
 )
+from repowise.server.services.module_health import read_repo_health_score
 
 router = APIRouter(
     prefix="/api/workspace",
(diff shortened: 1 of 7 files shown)


RaghavChamadiya (approved):
LGTM

PR #442, opened 2026-06-09
fix(cli): honor config.yaml model when provider set via env/flag

### LINKED ISSUE #416: repowise update ignores config.yaml model and falls back to Claude Sonnet
**Describe the bug**
When running `repowise update`, if the OpenRouter API key is set via environment variables and no explicit `--model` flag is passed via CLI, Repowise incorrectly falls back to its hardcoded default model (`anthropic/claude-sonnet-4.6`) even if a different model is explicitly configured in `.repowise/config.yaml` (e.g., `google/gemini-3.1-flash-lite-preview`).

This causes unexpected charges to the user's OpenRouter account on a much more expensive model than configured.

**To Reproduce**
1. Set `OPENROUTER_API_KEY` in your environment.
2. Initialize a repository to use Gemini: `repowise init --provider openrouter --model google/gemini-3.1-flash-lite-preview`
3. Verify `.repowise/config.yaml` correctly contains:
   ```yaml
   provider: openrouter
   model: google/gemini-3.1-flash-lite-preview
   ```
4. Run `repowise update`.
5. Observe the LLM logs. It will invoke `anthropic/claude-sonnet-4.6` instead of Gemini.

**Expected behavior**
`repowise update` should respect the `model` defined in `.repowise/config.yaml` if the `--model` CLI flag is not explicitly passed.

**Root Cause Analysis**
In `packages/cli/src/repowise/cli/helpers.py`, the `resolve_provider` function has a bug where it fails to properly pass the `model` loaded from `config.yaml` into the provider instantiation if it enters the environment-variable auto-detect fallback paths, or if the initialization defaults drop it. When `model=None` is passed to the OpenRouter provider constructor, it triggers its own hardcoded default (`anthropic/claude-sonnet-4.6`).

**Workaround**
Explicitly pass the model flag on updates:
`repowise update --model google/gemini-3.1-flash-lite-preview`

joptimus:
## What

`resolve_provider` only read the `config.yaml` `model` inside the branch that *also* resolves the provider from config. When the provider is resolved independently — via `--provider` or `REPOWISE_PROVIDER` — that branch is skipped, `model` stays `None`, and the provider constructor falls back to its hardcoded default. For OpenRouter that means `anthropic/claude-sonnet-4.6`, silently overriding the configured (often cheaper) model and billing users more than expected.

## Why

Fixes #416. A user who runs `repowise init --provider openrouter --model google/gemini-3.1-flash-lite-preview` and later runs `repowise update` (with the provider coming from env/flag) would be charged on Claude Sonnet instead of the configured Gemini model.

## How

Move the `model is None and cfg.get("model")` fallback out of the config-provider branch so the configured model is honored across **all** provider-resolution paths — `--provider` flag, `REPOWISE_PROVIDER` env, config provider, and API-key auto-detect — whenever no explicit `--model` is passed. Explicit `--model` still wins (guarded by `model is None`).

## Tests

Added `TestResolveProviderConfigModel` in `tests/unit/cli/test_helpers.py` covering:
- provider via `REPOWISE_PROVIDER` → config model forwarded
- provider via `--provider` flag → config model forwarded
- API-key auto-detect path → config model forwarded
- explicit `--model` overrides the config model

Each was confirmed to fail on the un-fixed code and pass with the fix. Full `tests/unit/` suite passes.

### DIFF
--- a/packages/cli/src/repowise/cli/helpers.py
+++ b/packages/cli/src/repowise/cli/helpers.py
@@ -663,8 +663,10 @@ def resolve_provider(
 
     if provider_name is None and cfg.get("provider"):
         provider_name = cfg["provider"]
-        if model is None and cfg.get("model"):
-            model = cfg["model"]
+
+    # Honor the config model regardless of how the provider was resolved (#416).
+    if model is None and cfg.get("model"):
+        model = cfg["model"]
 
     def _resolve_base_url(name: str) -> str | None:
         """Return base_url from env or repo config for the provider."""
--- a/tests/unit/cli/test_helpers.py
+++ b/tests/unit/cli/test_helpers.py
@@ -506,6 +506,70 @@ class TestResolveProviderBaseUrl:
 
 
 # ---------------------------------------------------------------------------
+# Provider model resolution from config.yaml (issue #416)
+# ---------------------------------------------------------------------------
+
+
+class TestResolveProviderConfigModel:
+    """The config.yaml ``model`` must be honored whenever no model is passed
+    explicitly, regardless of how the *provider* was resolved. Otherwise the
+    provider constructor falls back to its hardcoded default (issue #416)."""
+
+    @staticmethod
+    def _capture(monkeypatch, tmp_path, cfg: dict[str, Any]) -> dict[str, Any]:
+        import yaml  # type: ignore[import-untyped]
+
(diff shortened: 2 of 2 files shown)


RaghavChamadiya (approved):
LGTM

Issue #88, opened 2026-04-20
[Feature] Improve MCP architecture for multi-repo and multi-session support


hashtafak:
## Describe the Issue

repowise's MCP server architecture is fundamentally single-tenant - one MCP server process per repository - which creates several limitations for multi-repo and multi-user workflows.

## Steps to Reproduce

1. Open multiple IDEs/CLIs accessing different repositories
2. Each requires its own MCP server process: `repowise mcp /path/to/repo1`
3. For SSE transport, each needs unique port: `repowise mcp --transport sse --port 7338`
4. Try concurrent generation operations on the same repository
5. Observe SQLite database locking and port conflicts

### No Centralized MCP Server
- Each repository requires its own MCP server process
- Server uses module-level globals tied to one repository path
- MCP config generation creates repository-specific configurations

### SQLite Database Locking
- Multiple MCP servers accessing the same SQLite database experience locking during write operations
- SQLite allows multiple readers but only one writer at a time
- Generation operations involve both reads and writes, causing blocking for concurrent processes

### SSE Transport Port Management
- SSE transport requires unique ports per repository (default: 7338)
- Creates configuration overhead for multi-repo workflows
- Each client needs repository-specific port settings

### No Multi-CLI Session Support
- Architecture prioritizes single-user workflows over multi-session coordination
- No centralized session broker or lock management for generation operations
- Resource contention on vector stores and search indices

## Expected Behavior

- Single MCP server should be able to serve multiple repositories
- Multiple CLI/IDE instances should access the same repository without conflicts
- SSE transport should not require manual port management per repository
- Generation operations should handle concurrent access gracefully

## Proposed Solution

1. **Centralized session broker** to coordinate multiple MCP connections
2. **Shared vector store** with concurrent access patterns
3. **Lock management** for generation operations

## Current Workarounds

1. Use stdio transport for IDE integration (avoids port conflicts)
2. Hosted version for team environments with cross-repo intelligence
3. Port allocation strategy for SSE transport (reserve port ranges)

## Environment

- OS: Windows 10
- Python version: 3.14
- Repowise version: 0.3.0
- Installation method: pip

## Additional Context

The current design prioritizes simplicity and single-user workflows. For team environments requiring multi-repo intelligence, architectural improvements are needed to support:
- Multiple developers working on the same repository simultaneously
- Single MCP server serving multiple repositories
- Web-based clients without port management overhead
- Concurrent generation operations without database locking

[...]

PR #443, opened 2026-06-09
fix(mcp): use shared database engine factory

joptimus:
## Summary
- route single-repo MCP startup through the shared persistence create_engine() factory
- route workspace RepoRegistry per-repo contexts through the same factory
- add MCP-specific regression tests that verify SQLite WAL and usy_timeout pragmas are active

## Context
Issue #88 called out that MCP DB setup bypassed the shared engine factory, so MCP paths could miss the SQLite WAL / usy_timeout behavior used elsewhere in Repowise.

## Verification
- uv run pytest tests/unit/server/test_mcp_database_pragmas.py tests/unit/persistence/test_database_pragmas.py tests/unit/server/test_mcp_workspace.py
- uv run ruff check packages/server/src/repowise/server/mcp_server/_server.py packages/core/src/repowise/core/workspace/registry.py tests/unit/server/test_mcp_database_pragmas.py --ignore SIM105

Note: plain Ruff still reports a pre-existing unrelated SIM105 in 
egistry.py; this PR leaves it untouched.

### DIFF
--- a/packages/core/src/repowise/core/workspace/registry.py
+++ b/packages/core/src/repowise/core/workspace/registry.py
@@ -160,11 +160,9 @@ class RepoRegistry:
         from sqlalchemy.ext.asyncio import (
             async_sessionmaker as _async_sessionmaker,
         )
-        from sqlalchemy.ext.asyncio import (
-            create_async_engine,
-        )
 
         from repowise.core.persistence.database import (
+            create_engine,
             get_db_url,
             init_db,
         )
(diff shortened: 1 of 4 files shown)


RaghavChamadiya:
@joptimus CI's failing, tests/unit/server/mcp/test_config.py:84 still patches the now-removed create_async_engine; just point it at create_engine and it should go green, will merge after that

joptimus:
@RaghavChamadiya completed

RaghavChamadiya (approved):
Thanks for the quick test fix. Routing MCP and registry engine creation through the shared create_engine factory is the right call, and the pragma regression tests are a nice touch. LGTM.

PR #444, opened 2026-06-09
feat(mcp): add streamable HTTP transport

joptimus:
## Summary
- add `streamable-http` as a supported `repowise mcp --transport` value
- dispatch streamable HTTP through FastMCP and apply `--port` to the HTTP server
- keep stdio as the default and retain SSE as a legacy transport
- update MCP docs/help text and generated setup guidance

## Context
Issue #88 called out moving away from SSE toward HTTP as a focused contribution separate from the larger multi-tenant MCP architecture work.

FastMCP in the pinned SDK supports `streamable-http` directly, so this PR does not require an SDK version bump.

## Verification
- `uv run pytest tests/unit/cli/test_mcp_transport.py tests/unit/cli/test_mcp_config.py tests/unit/cli/test_commands.py::TestStubs::test_mcp_help`
- `uv run ruff check packages/cli/src/repowise/cli/commands/mcp_cmd.py packages/server/src/repowise/server/mcp_server/_server.py packages/server/src/repowise/server/mcp_server/__init__.py packages/cli/src/repowise/cli/mcp_config.py tests/unit/cli/test_mcp_transport.py`

### DIFF
--- a/docs/CLI_REFERENCE.md
+++ b/docs/CLI_REFERENCE.md
@@ -574,12 +574,13 @@ If `PATH` is omitted, `repowise mcp` first walks upward from the current directo
 
 | Flag | Description |
 |------|-------------|
-| `--transport` | `stdio` (default, for editors) or `sse` (for web clients) |
-| `--port` | Port for SSE transport (default: 7338) |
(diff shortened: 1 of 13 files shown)


RaghavChamadiya (approved):
LGTM

PR #456, opened 2026-06-11
fix(health): score module-level JS callbacks

### LINKED ISSUE #451: Health complexity is blind to top-level arrow/lambda functions
Follow-up to #375 (the NLOC undercount fixed in #387). That fix corrected file size, but a related blindspot remains in the complexity metrics themselves.

The complexity walker only collects `function_kinds` nodes as scoreable functions (`packages/core/src/repowise/core/analysis/health/complexity/walker.py`, `_collect_function_nodes`). Arrow functions are mapped as `lambda_kinds` in `complexity/languages.py`, which is correct when an arrow is nested inside a named function (it rolls up into the enclosing function's CCN), but a JS/TS file built from top-level arrow callbacks, like an Express route module:

```js
router.get(/users, async (req, res) => {
  // 50 lines of branching logic
});
```

produces no `FunctionComplexity` rows at all. Consequences:

- no CCN, nesting, or cognitive complexity is recorded anywhere for that logic
- `complex_method`, `nested_complexity`, `large_method`, `bumpy_road`, and `complex_conditional` can never fire on such files
- route-style codebases get structurally blind health scores even though their file NLOC is now correct

Proposed direction: collect module-level (and argument-position) `lambda_kinds` nodes as their own function entries, with synthesized names for anonymous callbacks, while taking care not to double-count arrows already rolled into a collected function. The `file_nloc` semantics from #387 should stay unchanged.

This changes which findings fire on JS/TS repos, so it needs validation against representative projects before shipping.

joptimus:
## Summary

- collect module-level JS/TS lambda callbacks as scoreable complexity entries
- synthesize names for assigned lambdas and call-site callbacks
- skip outer test suite wrapper callbacks while still descending into their test/hook callbacks

Fixes #451.

## Verification

- `uv run pytest tests/unit/health/test_complexity_walker.py tests/unit/health/test_file_nloc.py -q`
- `uv run ruff check packages/core/src/repowise/core/analysis/health/complexity/walker.py tests/unit/health/test_complexity_walker.py`

I also ran a read-only before/after validation harness against representative local projects without persisting health tables:

- Express-style API service: file NLOC stayed unchanged, route callbacks became scoreable entries, and new complexity/large-method signals appeared only where callback bodies had enough substance.
- TypeScript frontend app: file NLOC stayed unchanged; existing top health targets remained stable, with only callback/lambda entry coverage increasing.
- Desktop/app-style JavaScript project: file NLOC stayed unchanged; low-complexity event callbacks were collected but did not create new findings.

Validation also caught noisy suite-level test callbacks, so this PR skips `describe`/`context`/`suite` wrappers while still collecting nested `it`/hook callbacks.

### DIFF
--- a/packages/core/src/repowise/core/analysis/health/complexity/walker.py
+++ b/packages/core/src/repowise/core/analysis/health/complexity/walker.py
@@ -14,8 +14,9 @@ through its body, accumulating:
   do not contribute (kept simple in v1).
 
 Anonymous functions (lambdas, arrow functions, closures) recurse for
-their containing function's metrics — they do not produce their own
-``FunctionComplexity`` row.
+their containing function's metrics when they are nested in a named
+function. Module-level lambdas, such as route callbacks, produce their
+own ``FunctionComplexity`` row.
 """
 
 from __future__ import annotations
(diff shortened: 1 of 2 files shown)


RaghavChamadiya (approved):
LGTM

PR #457, opened 2026-06-11
fix(web): clarify health trend score changes

### LINKED ISSUE #452: Health trends page: misleading section title for largest score changes
Second follow-up noted while closing out #375. The health trends section titled "Files that moved most since last index" reads as if it tracks file moves/renames. It actually lists the files whose health score changed the most between snapshots.

Suggested label: "Largest score changes" (or "Largest score changes since last index").

joptimus:
## Summary

- rename the health trend file-delta section from a move/rename-oriented label to a score-change label
- keep the underlying trend data, table columns, sorting, and API contract unchanged

Fixes #452.

## Why

The trend page section listed files whose health scores changed most between the latest snapshots, but the heading said "Files that moved most since last index." That read like file moves or renames, which is not what the table represents.

The new heading, "Largest score changes since last index," matches the data shown in the table: before score, after score, and delta.

## Scope

This is a copy-only web UI change:

- no API changes
- no schema changes
- no scoring or trend calculation changes
- no layout changes beyond the replacement text

## Verification

- `npm.cmd ci`
- `npm.cmd run type-check --workspace packages/web`
- `git diff --check`

### DIFF
--- a/packages/web/src/app/repos/[id]/health/trend/page.tsx
+++ b/packages/web/src/app/repos/[id]/health/trend/page.tsx
@@ -94,7 +94,7 @@ export default function HealthTrendPage() {
 
           <section className="space-y-2">
             <h2 className="text-sm font-medium uppercase tracking-wider text-[var(--color-text-tertiary)]">
-              Files that moved most since last index
+              Largest score changes since last index
             </h2>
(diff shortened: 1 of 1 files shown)


RaghavChamadiya (approved):
LGTM

PR #463, opened 2026-06-12
fix(health): accept glob alias in health-rules and correct stale docs

RaghavChamadiya:
## What

Integrity sweep over the code-health configuration surface and docs. One behavior change, the rest documentation corrections.

## Config fix

Both doc examples for `.repowise/health-rules.json` showed a `"glob"` key on rule entries, but the parser only accepted `path` / `path_glob`, so any per-path override copied from the docs was silently ignored. `glob` is now accepted as an alias (with a test), the examples use the canonical `path` key, and the aliases are documented.

## Docs corrections

- `analysis/health/README.md` claimed "twelve biomarkers"; the registry holds 26, plus 3 governance findings written by a separate additive pass.
- `docs/architecture/code-health.md` was missing 6 of the biomarker files from the layer's file tree (low_cohesion, god_class, coverage_gradient, prior_defect, large_assertion_block, duplicated_assertion_block), and still listed "no PR-mode delta in v1" as a non-goal even though the `change_risk` package and `repowise risk` shipped.
- `docs/CODE_HEALTH.md` described hotspot health as an average over "the top-25% hotspot files"; it actually averages over the files the git layer classifies as hotspots (churn percentile plus minimum-activity floors).

## Tests

`tests/unit/health/test_health_config.py` gains an alias-acceptance test; the suite passes.

### DIFF
--- a/docs/CODE_HEALTH.md
+++ b/docs/CODE_HEALTH.md
@@ -63,7 +63,9 @@ reproduced by `local-stash/calibrate_health_weights.py` and documented in
 
 The final score is clamped to `[1.0, 10.0]`. The three repo-level KPIs:
 
-- **Hotspot Health** — NLOC-weighted average over the top-25 % hotspot files.
+- **Hotspot Health** — NLOC-weighted average over the files classified as
+  hotspots by the git layer (high churn percentile plus minimum-activity
+  floors), not a fixed top-N slice.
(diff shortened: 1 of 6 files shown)


RaghavChamadiya:
Superseded: this exact changeset already landed on main inside #458 (its squash includes the glob-alias commit). All five files are byte-identical between this branch and main, so closing.

PR #487, opened 2026-06-15
fix(workspace): keep job progress state accurate

### LINKED ISSUE #486: [Bug] Workspace job progress can show stale running state and reset elapsed time
## Describe the Bug

In a self-hosted workspace with multiple repositories, the web UI can show a sync/generation job as still in progress after the work is no longer actually active. Progress text such as `Generating Level 1...` / `Generating level 1...` is also misleading because it appears to be treated like a user-facing generation level rather than the backend phase bucket.

The elapsed timer also resets when navigating away from the page/tab and returning, because it appears to be based on the component mount time instead of the job's persisted timestamps.

This affects both workspace/repo web UI surfaces that display the active generation/sync state.

## Steps to Reproduce

1. Run `repowise serve` for a self-hosted workspace with multiple repositories configured.
2. Start a repository sync or docs generation job from the web UI.
3. Observe the active job/progress banner or generation progress UI while the job is running.
4. Navigate to another tab/page in the UI, then return to the page showing progress.
5. Restart the server while a workspace repo job is pending/running, or otherwise leave a workspace repo job row in a pending/running state.
6. See the UI continue to report an in-progress state or reset the elapsed timer.

## Expected Behavior

The UI should reflect the real persisted job state for the relevant repo/workspace database.

- Elapsed time should be derived from `started_at` or `created_at`, not from when the React component mounted.
- Finished/failed jobs should stop counting and use `finished_at` when available.
- Live SSE progress should update the displayed job state without waiting for polling.
- Workspace-mode job listing/status should query the correct repository database, not only the primary database.
- Pending/running jobs interrupted by server restart should be marked failed across all workspace repo databases, so the UI does not show a phantom in-progress sync forever.
- User-facing status text should describe the actual phase, for example `Indexing`, `Analysing`, or `Generating docs`, instead of `Generating level 1...`.

## Actual Behavior

The UI can show a job as continuously in progress even when the backend state is stale or belongs to a different workspace DB context.

The elapsed timer resets after moving away from the UI and returning, which makes the progress display look like a new job started.

The label `Generating level 1...` is confusing and does not accurately describe the backend phase/state.

In workspace mode, stale pending/running jobs can survive in non-primary repo databases after a server restart, which can make the UI report activity that is not actually happening.

## Environment

- OS: Windows 11 (`Microsoft Windows NT 10.0.26200.0`)
- Python version: 3.14.2 via `uv run python --version`
- Repowise version: 0.19.1 (`uv run repowise --version`)
- Installation method: local source checkout with `uv` / self-hosted web UI

## Additional Context

Observed in a multi-repo workspace containing several repositories. The issue is most visible in the workspace/repo web UI while sync/generation is running or after navigating between pages.

Likely affected areas from local investigation:

- frontend progress wrapper elapsed-time calculation
- frontend `useJob` SSE/state merge behavior
- generation progress status label mapping
- `/api/jobs` workspace DB resolution/listing behavior
- startup stale-job cleanup for secondary workspace repo databases

joptimus:
## Summary

- Fix workspace job listing/state resolution so job progress reads from the correct repo DB and stale workspace jobs are reset after restart.
- Keep frontend progress display tied to persisted job timestamps and live SSE progress instead of component mount time or stale polling snapshots.
- Add regression coverage for workspace DB lookup, stale job cleanup, SSE progress merging, elapsed time calculation, and phase labels.

## Related Issues

Fixes #486

## Test Plan

- [x] Tests pass (`uv run pytest tests/unit/server/test_jobs.py`)
- [x] Lint passes (`uv run ruff check packages/server/src/repowise/server/routers/jobs.py packages/server/src/repowise/server/app.py tests/unit/server/test_jobs.py`)
- [x] Web build passes (`npm.cmd run build`)
- [x] Type check passes (`npm.cmd run type-check`)
- [x] Web lint passes (`npm.cmd run lint`)
- [x] Web progress helper tests pass (`npm.cmd --workspace packages/web run test -- src/lib/jobs/progress.test.ts`)
- [x] UI progress component tests pass (`npm.cmd --workspace packages/ui run test -- __tests__/jobs/generation-progress.test.tsx`)

## Checklist

### DIFF
--- a/packages/server/src/repowise/server/app.py
+++ b/packages/server/src/repowise/server/app.py
@@ -10,11 +10,13 @@ from __future__ import annotations
 import logging
 import os
 from collections.abc import AsyncGenerator
-from contextlib import asynccontextmanager
+from contextlib import asynccontextmanager, suppress
+from datetime import UTC, datetime
 
 from fastapi import FastAPI, Request
 from fastapi.middleware.cors import CORSMiddleware
 from fastapi.responses import JSONResponse
+from sqlalchemy import update as sa_update
 
 from repowise.core.persistence.database import (
     create_engine,
@@ -23,6 +25,7 @@ from repowise.core.persistence.database import (
     init_db,
     resolve_db_url,
 )
+from repowise.core.persistence.models import GenerationJob
 from repowise.core.persistence.search import FullTextSearch
 from repowise.core.persistence.vector_store import InMemoryVectorStore
 from repowise.core.providers.embedding.base import MockEmbedder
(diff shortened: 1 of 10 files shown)


RaghavChamadiya (approved):
LGTM

Issue #502, opened 2026-06-18
[Feature] support lean4


DieracDelta:
## Problem

I'd like to index my lean4 codebase to use this tool

## Proposed Solution

I think it would be enough to add in https://github.com/Julian/tree-sitter-lean

## Alternatives Considered

N/A

## Additional Context

Lean4 is a really popular theorem prover (and gpl). LLMs are quite good at it!

RaghavChamadiya:
Happy to support Lean4. It's a good fit, and you're right that LLMs do well with it. I'd love a PR if you're up for it, and I'm glad to help while you work through it. Here's how it'd come together so you (or anyone who wants to grab this) can scope it out.

repowise has tiered language support (see [`docs/LANGUAGE_SUPPORT.md`](docs/LANGUAGE_SUPPORT.md)). Where Lean4 lands depends on what we can pull out of the grammar:

**Full AST tier (symbols, imports, call graph)** is the goal, and `tree-sitter-lean` is what unlocks it. The [Adding a New Language](docs/LANGUAGE_SUPPORT.md#adding-a-new-language) section covers it step by step, but the short version is:

1. Add a `LanguageSpec` module at `languages/specs/lean.py` and register it in `specs/__init__.py`.
2. Add `"lean"` to the `LanguageTag` literal in `ingestion/models.py`.
3. Write `ingestion/queries/lean.scm` using the capture-name conventions (`@symbol.def`, `@symbol.name`, `@import.statement`, `@import.module`, `@call.target`). `python.scm` and `typescript.scm` are good ones to copy from.
4. Add a `LanguageConfig` entry in `language_configs.py`.
5. Add the grammar dependency to `pyproject.toml`.

Steps 6 to 8 (binding extractor, heritage extractor, import resolver) are optional and additive. The generic stem-map fallback works out of the box, so a first PR can land without them. A proper `import`/`open` resolver for Lean would be the most useful follow-up after that.

[...]

PR #600, opened 2026-06-25
feat: Add Lean 4 language support (lightweight regex tier)

mvanhorn:
## Summary

- Adds Lean 4 to the lightweight regex import tier (the same tier as Elixir, Dart, Clojure, Haskell, Erlang, and F#), so Lean codebases get real file-to-file dependency edges instead of the generic stem-map fallback.
- New `LanguageSpec` (`.lean`, Lake manifests, `Main.lean` entry, `is_passthrough=True`, `import_support="partial"`), a regex extractor for `import`/`open`, and a module-name-index resolver, all wired into the three registries plus the `LanguageTag` literal.
- Mirrors the existing Haskell precedent; no tree-sitter grammar required.

## Related Issues

Refs #502

This implements the lightweight regex tier the maintainer scoped in #502 as a landable first step that does not depend on `tree-sitter-lean` shipping usable Python bindings. The optional full AST tier remains an additive follow-up.

## Test Plan

- [x] Tests pass (`uv run pytest tests/unit/ingestion` -> 1321 passed)
- [x] Lint passes (`ruff check` clean on the touched files)
- Web build: not applicable (no frontend changes)

New tests in `test_lightweight_resolvers.py` cover `extract_lean_imports` (import/open forms, comment/indent guards, dedup) and `resolve_lean_import` (path-derived resolution, local-shadows-stdlib, stdlib drop after local miss, external node, self-reference). `test_language_capabilities.py` pins `"lean"` in the partial tier and its `Main.lean` entry convention.

## Checklist

- Documentation: not applicable (the language list is registry-derived; no doc table to update)

AI was used for assistance.


RaghavChamadiya:
Thanks for the PR @mvanhorn, it mirrors the Haskell tier well and the tests are tidy. I ran the regex against real Mathlib files before merging and found a couple of things worth fixing:

**1. `public import` lines are dropped entirely.** Lean 4's module system added visibility modifiers on imports, and Mathlib uses them heavily. For example `Mathlib/Analysis/SpecialFunctions/Exp.lean` starts with:

```
public import Mathlib.Analysis.Complex.Asymptotics
public import Mathlib.Analysis.Complex.Trigonometric
import Mathlib.Topology.Order.AtTopBotIxx
```

The regex anchors on `^(?:import|open)`, so the two `public import` lines match nothing and the edges are lost. The same applies to `private import`, `meta import`, and `import all`. This means any repo on the new module system gets a partial import graph, which undercuts the whole point of the lightweight tier. The fix is to allow an optional visibility/`all` modifier between the line start and the module name.

**2. `open scoped Foo` captures `scoped` as the module.** This is common in Mathlib (`open scoped Nat`, `open scoped BigOperators`). The regex grabs the first token after `open`, so it extracts the literal `scoped` and mints a bogus `external:scoped` node. Worth skipping a leading `scoped` keyword. (For the record, `open Foo in`, `open Foo (a b)`, `hiding`, and `renaming` all capture `Foo` correctly, so those are fine.)

**3. The doc claim isn't right.** The PR says documentation is not applicable, but `docs/LANGUAGE_SUPPORT.md` has a hand-maintained lightweight-tier table with a row per language (Elixir, Dart, Clojure, Haskell, Erlang, F#) plus a hardcoded language list in the pipeline diagram. Lean 4 should be added to both.

**4. Title is missing the `feat:` prefix.** The commit has it but the PR title does not, and the PR title becomes the squashed commit subject.

**Minor:** `_MODULE_DECL_RE = re.compile(r"(?!x)x")` is an intentional never-match sentinel since Lean modules are path-derived. Either add a one-line comment saying so, or make `declaration_re` optional in `get_or_build_module_index` and pass `None`. A small fixture under `tests/fixtures/lang_samples/lean/` would also give end-to-end coverage, as suggested in #502.

Happy to merge once these are in.

mvanhorn:
All in, thanks for running it against real Mathlib - that caught exactly what a synthetic fixture wouldn't. In d186091f:

1. The import regex accepts the visibility modifiers (public / private / meta) and `import all`.
2. `open scoped Foo` extracts Foo instead of minting `external:scoped`.
3. Lean 4 added to the LANGUAGE_SUPPORT table and the pipeline diagram list.
4. PR title now has the `feat:` prefix.
5. The never-match sentinel got its explanatory comment, and there's a Lean fixture under tests/fixtures/lang_samples/lean/ covering `public import`, `open scoped`, and plain imports end to end.

The 9 Lean tests pass locally.

swati510 (approved):
LGTM, merging

mvanhorn:
Thanks for the merge, @swati510! Hope the Lean 4 tier proves useful.

mvanhorn:
Thanks for the merge, @swati510! Excited to see Lean 4 in the lightweight regex tier.

PR #623, opened 2026-06-29
fix: hide config languages from language usage

### LINKED ISSUE #371: [Refactor] Language usage section
# I think not showing languages like json, yaml and toml will make more sense, because they don't add value
## Currently showing json, yaml and toml
Image

## Proposed, do something like the GitHub repo page 
Image

cnYui:
## Summary

- hide JSON, YAML, and TOML from the shared language usage donut
- recalculate visible percentages after filtering config/data formats
- add a regression test for mixed-case config language names

Fixes #371.

## Tests

- npm run test --workspace @repowise-dev/ui -- __tests__/dashboard/language-donut.test.tsx
- npm run type-check --workspace @repowise-dev/ui
- npm run test --workspace @repowise-dev/ui
- git diff --check


RaghavChamadiya (approved):
LGTM

PR #627, opened 2026-06-30
docs: publish verified no-key quickstart (README + repowise.dev)

RaghavChamadiya:
## What

Publishes the **REP-2 verified, REP-9 board-approved** quickstart into the two public surfaces:

1. **README `## Quickstart`** — replaced the previous `init`/`serve` section with the no-key, under-5-minute path. The `#quickstart` nav anchor is preserved via an explicit `<a id="quickstart">`.
2. **`website/getting-started.md` (repowise.dev quickstart)** — same approved copy added as the **lead** section; the existing detailed multi-path setup (Plugin / pip+Claude / pip+Codex / standalone) is retained below it.

Source of truth: `docs/README-quickstart-section.md` (REP-2 deliverable).

## Why it's safe to ship

- **Docs-only.** No registry/PyPI/CI changes. `repowise` already installs (CLI 0.25.0).
- **Verified clean-run:** `docs/QUICKSTART-VERIFICATION.md` — `pip install` → `repowise init --index-only -y` (2.8s, no key) → `get_overview` / `get_context` return real structured data over MCP. The doc accurately states which tools need full mode.
- **CEO sign-off granted** via REP-9 (approved by the human board).

## Editorial notes for reviewer

- README: the old section also carried the multi-repo **Live System Map** blurb. That capability is still covered by the feature table above the Quickstart and in `docs/WORKSPACES.md`; flagging in case you want it echoed back into the new section.
- Both optional board polish notes were applied (non-blocking): the one-line payoff hook above Step 1, and a full-mode/hosted upsell near the closing note.
- README plugin marketplace ref uses `repowise-dev/repowise` (matches prior README). Note `website/getting-started.md`'s older Path 1 still says `repowise-dev/repowise-plugin` — left untouched as out of scope, but worth reconciling separately.

Tracking: REP-10.

### DIFF
--- a/README.md
+++ b/README.md
@@ -371,7 +371,7 @@ the contributor recipe: **[docs/LANGUAGE_SUPPORT.md →](docs/LANGUAGE_SUPPORT.m
 
 ```bash
 pip install repowise          # Windows: python -m pip install repowise
-repowise --version            # -> repowise, version 0.25.x
+repowise --version            # -> repowise, version 0.27.x
 ```
 
 **2. Index your repo — no LLM, no key**
--- a/README.md
+++ b/README.md
@@ -360,67 +360,71 @@ the contributor recipe: **[docs/LANGUAGE_SUPPORT.md →](docs/LANGUAGE_SUPPORT.m
 
 ---
 
-## Quickstart
+<a id="quickstart"></a>
+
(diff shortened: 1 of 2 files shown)


swati510 (approved):
LGTM

PR #629, opened 2026-06-30
fix(go): rescue same-file type references from unused_export false po…

### LINKED ISSUE #628: [Bug] Dead-code flagging FPs in Go repos
## Describe the Bug
 repowise dead-code incorrectly flags exported Go symbols as `unused_export` with 100% confidence when they are actively used within the same package.

## Steps to Reproduce
Run `repowise dead-code` on a Go project that has exported structs defined and consumed within the same package

## Expected Behavior
Exported Go symbols that are referenced within their own package (as return types, field types, or in function bodies) should not be flagged as unused and dead code.

## Actual Behavior
7 FPs with 100% confidence for Go structs, but they were consumed in the same package within services or handlers which were then actually used and exported. 

##  Root cause: 
The knowledge graph appears to only track cross package import edges. In Go, symbols used within the same package have no import statement, they are referenced directly. These intra package references produce no graph edges, so the symbols appear to have zero importers. 

This is a Go specific issue. 

## Environment

- OS: Ubuntu 24.0.4
- Python version:  3.12.3
- Repowise version: 0.25.0
- Installation method: pip
[](url)

crizah:
## Summary

  - Go same-file type references (return types, field types) were silently discarded in `_resolve_go_type_refs`, producing false-positive unused_export findings at 100% confidence for actively used exported structs
  
  - Fixed by stamping same-file Go type refs onto `local_type_uses` on the file node, which the dead-code analyzer already reads to rescue symbols (the same mechanism TS/JS uses)

## Related Issues

closes https://github.com/repowise-dev/repowise/issues/628

## Test Plan
  - [x] Verified on a real Go project: 7 FPs previously all gone
  - [x] Added regression test (`TestGoSameFileTypeUseFallback` in `test_go_type_use.py`): a single-file, no-import Go package with a struct used only as a same-file return type, confirmed it fails against the pre-fix code (flagged `unused_export`) and passes with the fix

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1293,7 +1293,6 @@
       "os": [
         "darwin"
       ],
-      "peer": true,
       "engines": {
         "node": "^18.17.0 || ^20.3.0 || >=21.0.0"
       },
@@ -1316,7 +1315,6 @@
       "os": [
         "darwin"
       ],
-      "peer": true,
       "engines": {
         "node": "^18.17.0 || ^20.3.0 || >=21.0.0"
       },
@@ -1339,7 +1337,6 @@
       "os": [
         "darwin"
       ],
-      "peer": true,
       "funding": {
         "url": "https://opencollective.com/libvips"
       }
@@ -1356,7 +1353,6 @@
       "os": [
         "darwin"
       ],
-      "peer": true,
       "funding": {
         "url": "https://opencollective.com/libvips"
       }
@@ -1373,7 +1369,6 @@
       "os": [
         "linux"
       ],
-      "peer": true,
       "funding": {
         "url": "https://opencollective.com/libvips"
       }
@@ -1390,7 +1385,6 @@
       "os": [
         "linux"
       ],
-      "peer": true,
       "funding": {
         "url": "https://opencollective.com/libvips"
       }
@@ -1407,7 +1401,6 @@
       "os": [
         "linux"
       ],
-      "peer": true,
       "funding": {
         "url": "https://opencollective.com/libvips"
       }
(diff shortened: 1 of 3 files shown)


swati510:
Thanks for taking this up! 🙏 The fix looks good — it faithfully mirrors the same-file rescue pattern the TS/JS strategy already uses (`local_type_uses` stamped on the file node, read by the dead-code analyzer), the cross-file path is semantically unchanged, and `test_genuinely_dead_export_still_flagged` still holds since a genuinely-dead export never appears in its own file's `type_refs`.

One ask before merging: **please add a regression test for the fixed path.** Every existing test in `test_go_type_use.py` exercises cross-package or same-package *cross-file* resolution — nothing covers the new same-file fallback (or the removed `if not candidates: return 0` early-return), which is exactly the behavior this PR fixes. A minimal test would be: a single-file Go package with an exported struct used only as a field/return type within that same file → assert it is **not** flagged `unused_export`. That locks in the fix so the 7 FPs can't silently regress on a future refactor.

Otherwise LGTM.

crizah:
All done.

Added `TestGoSameFileTypeUseFallback` to `tests/unit/ingestion/test_go_type_use.py`. It sets up one Go file, `widget/widget.go`, with a struct Widget that's only used as the return type of a function in that same file.

  Two checks:
1. `test_same_file_return_type_stamped_as_local_type_use`: Tests that the file gets marked internally as using Widget locally (so the dead-code checker knows not to flag it).
2. `test_same_file_only_type_not_flagged_unused_export`: Runs the actual dead-code analyzer to test Widget is not reported as unused.

swati510:
@crizah thanks for taking this up. Merging now.

swati510 (approved):

PR #638, opened 2026-07-01
fix(workspace): honor repo settings excludes in health sync

### LINKED ISSUE #637: [Bug] Workspace resync ignores repo settings exclude paths for Code Health and refactoring
## Describe the Bug

In workspace mode, if a repo has `exclude_patterns` set in the repo settings, a workspace resync does not fully honor them for Code Health and refactoring targets.

Excluded files can still show up in health findings and refactoring targets after resync.

## Steps to Reproduce

1. Run `repowise update --repo <alias>` so a workspace repo is indexed.
2. In the web app, open that repo's settings and set `exclude_patterns` to something like `tools/`.
3. Make sure there are files under that excluded path that would produce Code Health findings.
4. Trigger a workspace resync for the repo from the workspace UI.
5. Open Code Health or Refactoring Targets for that repo.

## Expected Behavior

Files matching the repo's excluded paths should not be re-indexed or shown in Code Health and refactoring results after resync.

## Actual Behavior

After workspace resync, excluded files can still appear in Code Health findings and refactoring targets.

No API error is shown. The problem looks like the wrong exclude source is being used during the workspace resync path.

## Environment

- OS: Windows 11
- Python version: 3.14.2
- Repowise version: 0.20.0 (`repowise --version`)
- Installation method: source checkout via `uv run`

## Additional Context

This seems specific to the workspace update path.

The server job path correctly merges excludes from both `Repository.settings_json` and `.repowise/config.yaml`, but the workspace incremental update path only reads `.repowise/config.yaml` excludes and does not read the DB-backed repo settings excludes from the web UI.

There also appears to be a second issue: the workspace incremental persistence path is upsert-only for health data, so rows for files that become excluded can linger until a full prune path runs.

joptimus:
## Summary

- merge workspace repo excludes from both repo settings and `.repowise/config.yaml` for incremental and full fallback update paths
- prune stale file-scoped rows during incremental persistence when files drop out due to excludes
- add regression coverage for DB-backed workspace excludes and health row pruning

## Related Issues

Fixes #637

## Test Plan

- [ ] Tests pass (`pytest`)

Ran:

- `uv run pytest tests/unit/workspace/test_incremental_update.py tests/unit/server/test_health_excludes.py tests/unit/server/test_health_breakdown.py tests/unit/server/test_health_badge.py -q`
- `uv run ruff check packages/core/src/repowise/core/persistence/crud/analysis.py packages/core/src/repowise/core/pipeline/incremental.py packages/core/src/repowise/core/workspace/update.py tests/unit/server/test_health_excludes.py tests/unit/workspace/test_incremental_update.py`

Notes:

- Full `uv run pytest` was started but interrupted before completion at user request, so this PR does not claim a full-suite pass yet.
- No frontend changes in this PR.

## Checklist

- [ ] All existing tests still pass

### DIFF
--- a/packages/core/src/repowise/core/persistence/crud/analysis.py
+++ b/packages/core/src/repowise/core/persistence/crud/analysis.py
@@ -8,9 +8,10 @@ from __future__ import annotations
 
 import json
 from datetime import datetime
+from pathlib import Path
 from typing import Any
 
-from sqlalchemy import func, or_, select
+from sqlalchemy import or_, select
 from sqlalchemy.ext.asyncio import AsyncSession
 
 from repowise.core.analysis.dead_code.risk_factors import effective_safe_to_delete
@@ -22,6 +23,7 @@ from ..models import (
     HealthFinding,
     HealthSnapshot,
     RefactoringSuggestion,
+    Repository,
     _new_uuid,
     _now_utc,
 )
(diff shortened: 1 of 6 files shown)


RaghavChamadiya:
Nice fix, and the two-layer approach (read-time filter + prune) is the right call. One thing to flag before merge, and it's not your fault:
This diff migrates the datetime import from timezone to UTC, but there's now a second timezone.utc call site it doesn't convert, reconcile_repo_head_commit at update.py:298:

repo.updated_at = datetime.now(timezone.utc)  # timezone no longer imported → NameError

That line landed in #652 after you opened this PR, so when your branch merged main the two changes crossed without a textual conflict. reconcile_repo_head_commit runs on essentially every workspace update (including no-op syncs), so this would raise NameError: name 'timezone' is not defined at runtime. Your new tests drive update_single_repo_index directly and don't hit that path, which is why it stayed green.
Could you convert that line to datetime.now(UTC) as well, then do a full uv run pytest (not just the targeted files) to confirm? The freshness-stamp tests should exercise the path this time. Everything else looks good to me.

joptimus:
Thanks for catching that. I’ve rebased onto upstream/main, updated the remaining reconcile_repo_head_commit() timestamp write to use datetime.now(UTC), and added regression coverage for the workspace update path. I also ran the full uv run pytest as requested. The workspace/freshness tests for this change passed, but the full suite is currently failing in three unrelated tests: tests/unit/cli/test_ui_package.py::test_advanced_config_default_keys_no_fast, tests/unit/server/mcp/test_code_rationale.py::test_grep_comment_candidates_ranks_number_bearing_file, and tests/unit/server/mcp/test_code_rationale.py::test_concept_anchor_injects_winner_as_dominant.

joptimus:
I cleaned up the branch history and rebuilt the PR on top of current upstream/main with only the three intended workspace commits. It now shows the PR as mergeable again, so the merge conflicts are resolved.

RaghavChamadiya (approved):

PR #647, opened 2026-07-02
Add Gitea webhook sync support

2233admin:
## Summary
- add a /api/webhooks/gitea endpoint for self-hosted Gitea push events
- add REPOWISE_GITEA_WEBHOOK_TOKEN support for token-based verification
- document Gitea webhook setup and OpenAI-compatible provider examples
- cover Gitea webhook accept/reject behavior in server unit tests

## Verification
- uv run pytest tests/unit/server/test_webhooks.py
- uv run ruff check packages/server/src/repowise/server/routers/webhooks.py tests/unit/server/test_webhooks.py
- git diff --check
- pre-push JS lint/test/build passed locally; its Python pytest step used a global Python 3.10 environment with an incompatible antlr/omegaconf plugin and failed before collecting project tests, so the branch was pushed with --no-verify after the project venv targeted tests passed.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -52,6 +52,11 @@ GRAPH_PATH=./data/graphs
 # Leave empty for fully open access (suitable for local use)
 REPOWISE_API_KEY=
 
+# Webhook secrets/tokens for hosted repository auto-sync.
+# REPOWISE_GITHUB_WEBHOOK_SECRET=your_github_secret_here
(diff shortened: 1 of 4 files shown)


2233admin:
Closing for now. This was opened prematurely from a fork before we finished internal review and alignment on upstream contribution shape.

PR #673, opened 2026-07-03
fix(files): stop linking external dependency nodes as files

### LINKED ISSUE #668: [Bug] File page linkifies external-system graph nodes as file paths (dead link)
## Describe the Bug

On a file page, the neighbor lists (Dependents / Dependencies) build a file-page href for any node that is not a symbol. External-system graph nodes (C4 / dependency-graph nodes with an id like `external:System`) are not symbols and not real files, so they get a file href and produce a dead link such as `/files/external:System`. Clicking it lands on a "file not found" state.

## Steps to Reproduce

1. Open a file page for a file that has an external-system neighbor in the dependency graph.
2. In the Dependents or Dependencies list, click a neighbor whose id starts with `external:`.
3. The link resolves to a file path that does not exist and the file detail page shows an error / empty state.

## Expected Behavior

External-system nodes should not be linked as files. Either render them as a non-clickable chip, or route them to the architecture / dependency surface (for example `/architecture?view=graph&node=...`) instead of `/files/...`.

## Actual Behavior

In `packages/ui/src/files/file-graph-tab.tsx` the neighbor link is built as:

```tsx
const isSymbol = n.node_type === "symbol" || n.node_id.includes("::");
const href = isSymbol ? symbolHref(n.node_id) : fileHref(n.node_id);
```

Any non-symbol node falls into the `fileHref` branch, including `external:*` nodes. The sibling `file-overview-tab.tsx` neighbor rows have the same shape.

## Fix Direction

Gate the ternary on the node kind. Detect external nodes (via `node_type`, or `node_id.startsWith("external:")`) and either suppress the link or point it at the architecture view. Check the `FileGraphNeighbor` type in `packages/types/src/files.ts` for the field that distinguishes external nodes, and apply the same handling in both `file-graph-tab.tsx` and `file-overview-tab.tsx`.

## Additional Context

- `packages/ui/src/files/file-graph-tab.tsx` (the `isSymbol ? symbolHref : fileHref` neighbor link)
- `packages/ui/src/files/file-overview-tab.tsx` (same shape on its neighbor rows)
- `packages/types/src/files.ts` (`FileGraphNeighbor`)

zain-asif-dev:
Fixes #668

On a file page, the Dependents/Dependencies lists build a file-page href for any graph neighbor that isn't a symbol:

```tsx
const isSymbol = n.node_type === "symbol" || n.node_id.includes("::");
const href = isSymbol ? symbolHref(n.node_id) : fileHref(n.node_id);
```

External-system nodes (id like `external:fastapi`) are not symbols and not real files, so they fell into the file branch and produced a dead link such as `/files/external:fastapi`. Clicking it landed on a file-not-found state.

I checked how `node_type` gets set for these nodes (`ResolverContext.add_external_node` in `packages/core/.../resolvers/context.py` only sets `language="external"`, and `pipeline/persist.py` defaults `node_type` to `"file"` when it isn't set), so `node_type` alone isn't a reliable signal here. `dependency-heatmap.tsx` already works around this in the same package by checking the `external:` id prefix, so I used the same check in `file-graph-tab.tsx` and `file-overview-tab.tsx`.

External neighbors now render as plain (non-clickable) text with a tooltip explaining they're an external dependency, instead of a link to a page that doesn't exist. Regular file and symbol neighbors are unaffected.

Testing:
- Added `packages/ui/__tests__/files/file-graph-tab.test.tsx` covering a file neighbor (still links to its file page), a symbol neighbor (still links to its symbol page), and an external neighbor (renders as text, no link).
- `npx vitest run` in `packages/ui` passes (544 pre-existing tests plus the 3 new ones; one unrelated pre-existing test, `dsm-matrix.test.tsx`, is a slow/flaky timeout under load and passes on its own with a longer timeout).
- `npx tsc --noEmit` in `packages/ui` passes.
- `bash packages/ui/scripts/ui-gates.sh` (design token gates) passes.

### DIFF
--- a/packages/ui/__tests__/files/file-graph-tab.test.tsx
+++ b/packages/ui/__tests__/files/file-graph-tab.test.tsx
@@ -1,0 +1,91 @@
+import { describe, it, expect } from "vitest";
+import { render, screen } from "@testing-library/react";
+import { FileGraphTab } from "../../src/files/file-graph-tab.js";
+import type { FileDetailGraph, FileGraphNeighbor } from "@repowise-dev/types/files";
+
+function makeNeighbor(overrides: Partial<FileGraphNeighbor> = {}): FileGraphNeighbor {
+  return {
+    node_id: "src/app.ts",
+    node_type: "file",
+    language: "typescript",
+    edge_type: "imports",
+    imported_names: [],
+    ...overrides,
+  };
+}
+
+function makeGraph(overrides: Partial<FileDetailGraph> = {}): FileDetailGraph {
+  return {
+    language: "typescript",
+    is_entry_point: false,
+    is_test: false,
+    symbol_count: 3,
+    pagerank: 0.001,
+    pagerank_percentile: 42,
+    in_degree: 1,
+    out_degree: 1,
+    community_id: 0,
+    community_label: null,
+    dependents: [],
+    dependencies: [],
+    ...overrides,
+  };
+}
+
+const fileHref = (path: string) => `/files/${path}`;
+const symbolHref = (symbolId: string) => `/symbols/${symbolId}`;
+
+describe("FileGraphTab", () => {
+  it("links a regular file neighbor to its file page", () => {
+    const graph = makeGraph({
+      dependencies: [makeNeighbor({ node_id: "src/util.ts" })],
+    });
+    render(
+      <FileGraphTab
+        graph={graph}
+        filePath="src/app.ts"
+        linkPrefix="/repos/1"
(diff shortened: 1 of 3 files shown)


RaghavChamadiya:
Thanks, this is thorough, the extracted row fragment keeps it DRY, the tooltip is a nice touch, and I appreciate the note on why node_type isn't reliable here (persist.py defaulting to "file"). Tests cover the three cases well. Merging.

RaghavChamadiya (approved):

PR #674, opened 2026-07-03
fix: don't linkify external-system graph nodes as file paths

### LINKED ISSUE #668: [Bug] File page linkifies external-system graph nodes as file paths (dead link)

ZachDreamZ:
## What's wrong

On a file page, the neighbor lists (Dependents / Dependencies) linkify external-system graph nodes as file paths. Nodes with IDs like `external:System` get passed to `fileHref()` which produces dead links like `/files/external:System`. Clicking them lands on a "file not found" state.

## What this does

Checks if `node_id` starts with `external:` before generating a link. External nodes now render as non-clickable `<span>` elements with muted text styling instead of `<a>` links. Applied to both:
- `file-graph-tab.tsx` (Dependents / Dependencies lists)
- `file-overview-tab.tsx` (Dependencies preview)

## How to verify

1. Open a file page that has an external-system neighbor in the dependency graph
2. The external node should appear as muted text, not a clickable link
3. Other nodes (files, symbols) should still be clickable links

Fixes #668

### DIFF
--- a/packages/ui/src/files/file-graph-tab.tsx
+++ b/packages/ui/src/files/file-graph-tab.tsx
@@ -43,9 +43,11 @@ function NeighborList({
           <ul className="space-y-2">
             {neighbors.map((n) => {
               const isSymbol = n.node_type === "symbol" || n.node_id.includes("::");
-              const href = isSymbol ? symbolHref(n.node_id) : fileHref(n.node_id);
+              const isExternal = n.node_id.startsWith("external:");
+              const href = isExternal ? null : isSymbol ? symbolHref(n.node_id) : fileHref(n.node_id);
               return (
                 <li key={`${n.node_id}-${n.edge_type}`}>
+                  {href ? (
                   <a
                     href={href}
                     className="flex items-center gap-2 -mx-2 px-2 py-1 rounded hover:bg-[var(--color-bg-elevated)] transition-colors"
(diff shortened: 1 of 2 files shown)


RaghavChamadiya:
Thanks for the fix! This lands the same correct approach as #673 (gate on the external: id prefix), but #673 also adds test coverage, an explanatory tooltip, and a cleaner DRY diff, so I'm going with that one. Closing this as a duplicate, appreciate the contribution.

PR #675, opened 2026-07-03
Fix #648: Wire TsconfigResolver in CLI commands

### LINKED ISSUE #648: [Bug] dead-code, health, and workspace CLI commands report false-positive unreachable files for TS/JS projects using    @/ path aliases
## Describe the Bug

  The `dead-code`, `health`, and `workspace` CLI commands rebuild the dependency graph in-memory without wiring up the
   `TsconfigResolver`. As a result, TS/JS path aliases like `@/components/Foo` don't resolve to their target files
  during the rebuild — they become `external:` nodes, so the real target files get `in_degree = 0` and are falsely
  flagged as unreachable.

  The `init` pipeline does not have this bug: it runs a dedicated `tsconfig` phase that calls
  `graph_builder.set_tsconfig_resolver(...)` before `build()`
  (`packages/core/src/repowise/core/pipeline/phases/ingestion.py:328-348`). The CLI commands don't replicate this
  wiring.

  ## Steps to Reproduce

  1. In a TS project with `@/*` → `./src/*` path alias in `tsconfig.json`, have `src/routes/layout.tsx` import
  `Sensor` from `@/components/SensorProvider`.
  2. Index the repo: `repowise init . --no-workspace --index-only`
  3. Run: `repowise dead-code . --no-workspace --kind unreachable_file --format json`

  ## Expected Behavior

  `src/components/SensorProvider.tsx` is not listed as unreachable (it's imported by `src/routes/layout.tsx`).

  ## Actual Behavior

  `src/components/SensorProvider.tsx` is reported as `unreachable_file` with `"reason": "File has no importers
  (in_degree=0)"`, `confidence: 0.7`, `safe_to_delete: true`.

  On a real 800-file TS project, **171 of 347 `unreachable_file` findings (49%) were false positives** — every file
  imported via `@/` was misclassified.

  The persisted graph in `wiki.db` is correct: `graph_edges` contains `('src/routes/layout.tsx',
  'src/components/SensorProvider.tsx', 'imports')`. The bug is only in the in-memory graph rebuild the CLI performs.

  Affected files (all in `packages/cli/src/repowise/cli/commands/`):
  - `dead_code_cmd.py` (~line 143–156)
  - `health_cmd.py` (~line 189–210)
  - `workspace_cmd.py` (~line 681)

  ## Environment

  - OS: macOS 15
  - Python version: 3.14.6
  - Repowise version: 0.25.0
  - Installation method: pipx

  ## Additional Context

  Root cause: `TsconfigResolver` is constructed and attached to `GraphBuilder` only in the `init` pipeline. The
  resolver is read by `resolve_ts_js_import`
  (`packages/core/src/repowise/core/ingestion/resolvers/typescript.py:60-64`) during `build()`. Without it attached,
  `ctx.tsconfig_resolver` is `None` and non-relative imports fall through to `add_external_node(module_path)`.

  Suggested fix: mirror the pipeline's wiring in each CLI command before `graph_builder.build()`, or extract a shared
  `wire_tsconfig_resolver(graph_builder, repo_path, include_submodules, include_nested_repos)` helper in
  `repowise.core.ingestion` and call it from the pipeline and all three CLI commands.

  Related: PR #71 introduced tsconfig alias resolution (pipeline-only); PRs #299 and #244 fixed the same class of
  resolver-not-wired false-positive bug for CommonJS `require()` and Python imports respectively.

kishansaaai:
### Summary
This PR fixes the false-positive `unreachable_file` reports for TS/JS projects using path aliases (like `@/components`) in CLI commands. 

The root cause was that `TsconfigResolver` was only being wired up during the pipeline ingestion phase. Any CLI command that rebuilt the graph in-memory without this phase (such as `dead-code`, `health`, and `workspace`) would fail to resolve aliases, leaving valid targets with `in_degree=0`.

**Changes:**
- Extracted `wire_tsconfig_resolver` helper function in `repowise.core.ingestion.tsconfig_resolver`.
- Refactored `repowise/core/pipeline/phases/ingestion.py` to use the shared helper.
- Added resolver wiring calls right before `graph_builder.build()` in `dead_code_cmd.py`, `health_cmd.py`, and `workspace_cmd.py`.

### Related Issues
Fixes #648

### Test Plan
- [x] Tests pass (`pytest`)
- [ ] Web build passes (`npm run build`) (if frontend changes)

*Added a dedicated regression test (`test_tsconfig_cli_repro.py`) that mimics the CLI command's in-memory graph rebuild with a mock `@/*` alias to ensure it properly connects to its target and suppresses the false unreachability finding.*

### Checklist

### DIFF
--- a/packages/cli/src/repowise/cli/commands/dead_code_cmd.py
+++ b/packages/cli/src/repowise/cli/commands/dead_code_cmd.py
@@ -153,6 +153,15 @@ def dead_code_command(
             graph_builder.add_file(parsed)
         except Exception:
             pass
+
+    from repowise.core.ingestion import wire_tsconfig_resolver
+
+    wire_tsconfig_resolver(
+        graph_builder,
+        repo_path,
+        include_submodules=include_submodules,
+        include_nested_repos=include_nested_repos,
+    )
     graph_builder.build()
 
     # Framework-aware synthetic edges (Django, Laravel, TYPO3, ...). Without
--- a/packages/cli/src/repowise/cli/commands/health_cmd/command.py
+++ b/packages/cli/src/repowise/cli/commands/health_cmd/command.py
@@ -194,6 +194,15 @@ def health_command(
             parsed_files.append(parsed)
         except Exception:
             continue
+            
+    from repowise.core.ingestion import wire_tsconfig_resolver
+
+    wire_tsconfig_resolver(
+        graph_builder,
+        repo_path,
+        include_submodules=include_submodules,
+        include_nested_repos=include_nested_repos,
+    )
     graph_builder.build()
 
     git_meta_map: dict = {}
--- a/packages/cli/src/repowise/cli/commands/workspace_cmd.py
+++ b/packages/cli/src/repowise/cli/commands/workspace_cmd.py
@@ -660,6 +660,10 @@ def _generate_docs_for_added_repo(
             graph_builder.add_file(parsed)
         except Exception:
             continue
+
+    from repowise.core.ingestion import wire_tsconfig_resolver
+
+    wire_tsconfig_resolver(graph_builder, repo_path)
     graph_builder.build()
 
     from repowise.core.repo_config import load_repo_config
--- a/packages/core/src/repowise/core/ingestion/__init__.py
+++ b/packages/core/src/repowise/core/ingestion/__init__.py
@@ -29,7 +29,7 @@ from .models import (
 )
 from .parser import LANGUAGE_CONFIGS, ASTParser, LanguageConfig, parse_file
 from .traverser import FileTraverser, TraversalStats
-from .tsconfig_resolver import TsconfigResolver
+from .tsconfig_resolver import TsconfigResolver, wire_tsconfig_resolver
 
 __all__ = [
     # Models
@@ -52,6 +52,7 @@ __all__ = [
     # Graph
     "GraphBuilder",
     "TsconfigResolver",
+    "wire_tsconfig_resolver",
     "Import",
     "LanguageConfig",
     "PackageInfo",
--- a/packages/core/src/repowise/core/ingestion/tsconfig_resolver.py
+++ b/packages/core/src/repowise/core/ingestion/tsconfig_resolver.py
@@ -442,3 +442,33 @@ class TsconfigResolver:
         except Exception as exc:
             log.debug("tsconfig_parse_failed", path=str(config_path), error=str(exc))
             return None
+
+
+def wire_tsconfig_resolver(
+    graph_builder: Any,
+    repo_path: str | Path,
+    *,
+    include_submodules: bool = False,
+    include_nested_repos: bool = False,
+) -> None:
+    """Wire up the TsconfigResolver for the given GraphBuilder, if TS/JS files are present.
+
+    Needs to be called after add_file() but before build().
+    """
+    _ts_langs = {"typescript", "javascript"}
+
+    # We duck-type graph_builder to avoid circular imports. It has ._parsed_files.
+    parsed_files = graph_builder._parsed_files.values()
+    if not any(pf.file_info.language in _ts_langs for pf in parsed_files):
+        return
+
+    try:
+        _path_set = set(graph_builder._parsed_files.keys())
+        _resolver = TsconfigResolver(
+            repo_path=Path(repo_path),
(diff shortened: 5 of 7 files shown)


swati510:
The `dead-code` and `health` wirings are in place. Those are two of the three CLI commands that build a `GraphBuilder`, and both read `include_submodules`/`include_nested_repos` from state before the call.

The blocker is the third site. `workspace_cmd.py:685` passes `include_submodules=…` and `include_nested_repos=…` to `_generate_docs_for_added_repo`, and that function takes neither parameter. It raises `NameError` and crashes `repowise workspace add` for every repo. The error fires before the TS-language guard, so even non-TS repos hit it. That function never resolved those flags, so calling `wire_tsconfig_resolver(graph_builder, repo_path)` with the defaults matches its existing behavior.

A test that invokes the workspace path would catch this. The current regression test only exercises the helper on its own. Once that's fixed and green, this is ready.

RaghavChamadiya:
Thanks for tackling this @kishansaaai. The diagnosis is spot on and the shared-helper approach is exactly right. The dead_code_cmd.py wiring is correct. But as it stands the PR does not work yet. There are three blocking issues, and GitHub already flags the branch as conflicting.

1. The regression test can't pass.
GraphBuilder.graph is a method, not a property (builder.py:491), and every other caller uses graph_builder.graph(). The test accesses it uncalled:

assert graph_builder.graph.has_edge(layout_path, sensor_path)   # AttributeError
assert graph_builder.graph.in_degree(sensor_path) > 0

Both lines raise AttributeError: 'function' object has no attribute 'has_edge', so this test errors rather than validating the fix. Please change them to graph_builder.graph().has_edge(...) / graph_builder.graph().in_degree(...) and confirm it runs green.

2. The health fix edits a file that no longer exists on main.
health_cmd.py was refactored into a package on main, and health_command now lives in health_cmd/command.py. This PR patches the old monolith, which is what's causing the merge conflict. Even after resolving it, the real bug is untouched: health_cmd/command.py:197 still calls graph_builder.build() with no resolver wired. The include_submodules / include_nested_repos vars are already in scope there (command.py:172-173), so the wiring just needs to move to that line.

3. The workspace fix raises NameError at runtime.
_generate_docs_for_added_repo has no include_submodules or include_nested_repos in scope (not in the signature, not defined as locals), and its traverser/GraphBuilder are built without those flags. Passing them into the wire call crashes the workspace add doc-generation path for any repo, since the kwargs are evaluated before the helper's TS check runs. Simplest fix is to call wire_tsconfig_resolver(graph_builder, repo_path) and rely on the defaults, which matches how the traverser is already constructed in that function.

Minor: there's some trailing whitespace on the blank lines inside the new helper's docstring/body worth cleaning up.

For reference, I confirmed the scope is complete: dead_code, health, and workspace are the only three CLI paths that do an in-memory graph_builder.build(). update_cmd rehydrates from the DB and isn't affected.

Suggested next steps: rebase on main, move the health wiring into health_cmd/command.py, drop the two kwargs from the workspace call, and fix the test's graph() call so it actually exercises the path.

kishansaaai:
**Summary**
This PR fixes the false-positive `unreachable_file` reports for TS/JS projects using path aliases (like `@/components`) in CLI commands.

The root cause was that `TsconfigResolver` was only being wired up during the pipeline ingestion phase. Any CLI command that rebuilt the graph in-memory without this phase (such as `dead-code`, `health`, and `workspace`) would fail to resolve aliases, leaving valid targets with `in_degree=0`.

**Changes:**
- Extracted a `wire_tsconfig_resolver` helper function in `repowise.core.ingestion.tsconfig_resolver`.
- Refactored `repowise/core/pipeline/phases/ingestion.py` to use the shared helper.
- Added resolver wiring calls right before `graph_builder.build()` in `dead_code_cmd.py`, `health_cmd/command.py` (accommodating the recent monolith refactor on `main`), and `workspace_cmd.py`.

**Related Issues**
Fixes #648

**Test Plan**
- [x] Tests pass (`pytest`)
- [x] Added a dedicated regression test (`test_tsconfig_cli_repro.py`) that mimics the CLI command's in-memory graph rebuild with a mock `@/*` alias to ensure it properly connects to its target and suppresses the false unreachability finding. (Updated to correctly call `graph()` as a method).

**Checklist**

RaghavChamadiya (approved):
LGTM

PR #676, opened 2026-07-03
Fix #666: Probe submodules for python package imports

### LINKED ISSUE #666: [Bug] Python resolver misses `from package import submodule`, causing false dead-code on router modules
## Describe the Bug

The Python import resolver treats `from <package> import <submodule>` as a single import of the package's `__init__.py` and never resolves the individual names to their submodule files. When a package re-exports submodules this way (very common for FastAPI routers wired up with `app.include_router(...)`), those submodule files end up with `in_degree == 0`, and the dead-code analyzer then reports them as unreachable files.

It is systematic rather than file-specific. On a server checkout, roughly 18 of ~28 router modules get flagged this way (`workspace`, `overview`, `files`, `search`, `security`, `costs`, `decisions`, `modules`, `owners`, `providers`, `c4`, `blast_radius`, `claude_md`, `dead_code`, `external_systems`, `knowledge_map`, `pages`, `symbols`, `webhooks`). The few routers that escape only do so incidentally because some other file happens to import them by file path directly.

## Steps to Reproduce

1. Index a repo whose entry point wires submodules through a package-level import, for example:

   ```python
   # app.py
   from repowise.server.routers import workspace, git, overview
   app.include_router(workspace.router)
   ```

2. Run dead-code analysis (or query `get_dead_code`).
3. Router files such as `routers/workspace.py` are reported as `unreachable_file`, reason "File has no importers (in_degree=0)".

## Expected Behavior

`from pkg import submodule` should produce an `imports` edge from the importer to `pkg/submodule.py` (or `pkg/submodule/__init__.py`) when that file exists, so the submodule has an inbound edge and is not treated as dead. This also restores the missing call/usage edges into those modules.

## Actual Behavior

`resolve_python_import` in `packages/core/src/repowise/core/ingestion/resolvers/python.py` resolves the module string `repowise.server.routers` to a single path (the package `__init__.py`) and returns. It never inspects the imported names to check whether each is itself a submodule file. For comparison, the Rust resolver already iterates `imported_names` (`packages/core/src/repowise/core/ingestion/resolvers/rust.py:183`); the Python path has no equivalent.

Confirmed by inspecting the graph edges after indexing: there is no inbound `imports` edge to `routers/workspace.py`, while a router that some file imports by name (e.g. `git.py`) has a normal in-degree.

## Fix Direction

Preferred, at the resolver: when a `from <pkg> import <name1>, <name2>, ...` resolves to a package `__init__.py`, additionally probe each imported name as a submodule and emit a file to submodule `imports` edge when `<pkg_dir>/<name>.py` or `<pkg_dir>/<name>/__init__.py` exists. Handle the dotted form (`import repowise.server.routers.workspace as w`) too if it is not already covered.

Narrower alternative, in the analyzer: treat "this file's name appears in an importer's recorded `imported_names` against its parent package `__init__.py`" as an inbound signal before flagging `unreachable_file`. Smaller blast radius, but it does not repair the missing call edges.

## Additional Context

- Resolver: `packages/core/src/repowise/core/ingestion/resolvers/python.py` (`resolve_python_import`)
- Dead-code detection: `packages/core/src/repowise/core/analysis/dead_code/analyzer.py` (`_detect_unreachable_files` at line 503, `_make_unreachable_finding` at line 564)
- Reference that already does this: `packages/core/src/repowise/core/ingestion/resolvers/rust.py:183`
- Suggested tests under `tests/unit/ingestion/resolvers/`: `from pkg import submodule`, the multi-name form, and the `pkg/submodule/__init__.py` variant, plus a dead-code regression so a router with only a package-level importer is not flagged.

## Note: this also fixes the file-level FastAPI false positive

Router modules like `routers/c4.py` are reported file-level dead (reason "File has no importers (in_degree=0)") for the same reason. The FastAPI `include_router` handler (`packages/core/src/repowise/core/ingestion/framework_edges/fastapi.py`) does not rescue the common pattern:

```python
from .routers import c4
app.include_router(c4.router)
```

It only builds a router-variable edge when the imported name ends in `router`/`app` (fastapi.py:36) and resolves to a real file, but here the imported name is a submodule (`c4`) and `resolve_import` points at `routers/__init__.py`, so no edge is created. Emitting the submodule import edge (this issue) gives those router files an inbound edge, so they are no longer flagged. As optional defense-in-depth, a module whose top-level functions are all `@router.<method>`-decorated could additionally be treated as an entry point, but the resolver fix alone resolves the reported case.

kishansaaai:
### Summary
This PR fixes the false-positive `unreachable_file` reports on Python router modules caused by the resolver missing submodules when they are imported via a package `__init__.py`. 

The root cause was that `resolve_python_import` would resolve statements like `from repowise.server.routers import workspace` strictly to `repowise/server/routers/__init__.py`, meaning the submodules themselves never received inbound edges and were falsely flagged by the dead-code analyzer with `in_degree=0`.

**Changes:**
- Added a `resolve_python_import_all` helper function in `repowise.core.ingestion.resolvers.python`.
- `resolve_python_import_all` resolves the base package, then properly iterates over the `Import.imported_names` to probe for `<name>.py` and `<name>/__init__.py` submodules in the package directory.
- Updated `_resolve_imports` in `graph/builder.py` to route Python resolutions through this new function, appending missing edges.

### Related Issues
Fixes #666

### Test Plan
- [x] Tests pass (`pytest`)
- [ ] Web build passes (`npm run build`) (if frontend changes)

*Added a dedicated regression test (`test_python_resolver.py`) ensuring `resolve_python_import_all` properly identifies and returns submodule file paths when a module uses the `from pkg import submodule1, submodule2` format.*

### Checklist

### DIFF
--- a/packages/core/src/repowise/core/ingestion/graph/builder.py
+++ b/packages/core/src/repowise/core/ingestion/graph/builder.py
@@ -342,6 +342,10 @@ class GraphBuilder(MetricsMixin, ResolveMixin, EdgesMixin, SerializeMixin, Rehyd
                     from ..resolvers.cpp import resolve_cpp_import_all
 
                     targets = resolve_cpp_import_all(imp.module_path, path, ctx)
+                elif _lang == "python":
+                    from ..resolvers.python import resolve_python_import_all
+
+                    targets = resolve_python_import_all(imp, path, ctx)
                 else:
                     single = resolve_import(imp.module_path, path, _lang, ctx)
                     targets = (single,) if single else ()
--- a/packages/core/src/repowise/core/ingestion/resolvers/python.py
+++ b/packages/core/src/repowise/core/ingestion/resolvers/python.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 from pathlib import Path
+from typing import Any
 
 from ..languages.python_modules import build_python_module_index
 from .context import ResolverContext
@@ -67,6 +68,24 @@ def resolve_python_import(module_path: str, importer_path: str, ctx: ResolverCon
         if c in ctx.path_set:
             return c
 
-    # Stem-only fallback
+# Stem-only fallback
     stem = module_path.split(".")[-1].lower()
     return ctx.stem_lookup(stem)
+
+
+def resolve_python_import_all(imp: Any, importer_path: str, ctx: ResolverContext) -> tuple[str, ...]:
+    """Resolve a Python import, additionally probing submodules for package imports."""
+    targets = []
+    base = resolve_python_import(imp.module_path, importer_path, ctx)
+    if base:
+        targets.append(base)
+        if base.endswith("__init__.py"):
+            base_dir = Path(base).parent.as_posix()
+            for name in (imp.imported_names or []):
+                c1 = f"{base_dir}/{name}.py"
+                c2 = f"{base_dir}/{name}/__init__.py"
+                if c1 in ctx.path_set:
+                    targets.append(c1)
+                elif c2 in ctx.path_set:
+                    targets.append(c2)
+    return tuple(targets)
--- a/tests/unit/ingestion/test_python_resolver.py
+++ b/tests/unit/ingestion/test_python_resolver.py
@@ -1,0 +1,52 @@
+import pytest
+from repowise.core.ingestion.models import Import
+from repowise.core.ingestion.resolvers.context import ResolverContext
+from repowise.core.ingestion.resolvers.python import resolve_python_import, resolve_python_import_all
+
+
+def test_resolve_python_import_all_submodules():
+    path_set = {
+        "src/app.py",
+        "src/routers/__init__.py",
+        "src/routers/workspace.py",
+        "src/routers/git/__init__.py",
+    }
+    
+    ctx = ResolverContext(path_set=path_set, stem_map={}, graph=None)
+    
+    # 1. Single submodule without __init__.py package
+    imp = Import(
+        module_path="routers",
+        module_type="python",
+        imported_names=("workspace", "git", "overview"),
+        range_start=0,
+        range_end=0,
+        is_type_only=False,
+    )
+    
+    targets = resolve_python_import_all(imp, "src/app.py", ctx)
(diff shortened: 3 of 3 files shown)


swati510:
Thanks for tackling #666. Two things are blocking merge.

CI is red because the two new tests don't run. They build `Import(module_type=…, range_start=…, is_type_only=…)`, and `Import` (in `packages/core/src/repowise/core/ingestion/models.py`) has none of those fields. Its real fields are `raw_statement`, `module_path`, `imported_names`, `is_relative`, `resolved_file`, and the first three are missing here, so both tests raise `TypeError`. Could you rebuild them with the real fields and confirm `pytest tests/unit/ingestion/test_python_resolver.py` passes locally?

The dead-code regression that #666 asked for is also missing. Nothing yet proves a router imported only at the package level stops being flagged as unreachable. A small end-to-end test on that shape would lock it in.

Two smaller notes. Namespace packages (no `__init__.py`, PEP 420) and the bare `from . import submodule` form still aren't rescued, so it would help to handle them or note them as known gaps. There's also an accidental comment de-indent at `python.py:70` in a function this PR doesn't otherwise change.

Happy to re-review once CI is green.

RaghavChamadiya:
Thanks @kishansaaai, and sorry this went quiet for so long.

The need you identified was real and it has since been met, though not through this branch. `resolve_python_import_all` exists on main at `packages/core/src/repowise/core/ingestion/resolvers/python.py:76-108`, added in #886 on 17 July. It does what you proposed, probing `base_dir/name.py` and `base_dir/name/__init__.py` for each imported name when the base target is a package `__init__.py`, and it additionally consults a source-root-aware module index for absolute imports.

So the submodules that looked unreachable in #666 now get their inbound `imports` edge. Closing this as superseded rather than as wrong.

For the record on why the checks went red here, in case it is useful next time: the new tests construct `Import(module_type=..., range_start=..., is_type_only=...)`, and `Import` in `models.py` has none of those fields, so both tests raised `TypeError` before reaching anything under test. Worth knowing that a green-looking local run and CI can diverge on exactly that.

PR #680, opened 2026-07-04
feat(ui): add formatBytes and formatPercent formatters

Ayush7614:
## Summary
- Add formatBytes and formatPercent helpers to the shared UI format library
- Cover common byte-size and ratio display cases with unit tests

## Test plan
- packages/ui/__tests__/lib/format.test.ts covers byte and percent formatting
- npm run test --workspace=@repowise-dev/ui

### DIFF
--- a/packages/server/src/repowise/server/routers/overview.py
+++ b/packages/server/src/repowise/server/routers/overview.py
@@ -9,6 +9,7 @@ decisions slice, savings headline, and health KPIs.
 
 from __future__ import annotations
 
+import contextlib
 import json
 import sqlite3
 from pathlib import Path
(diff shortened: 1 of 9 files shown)


RaghavChamadiya:
Code and tests look solid, formatBytes matches the Python version in 681 exactly. One thing holding me back: nothing in the tree imports formatBytes or formatPercent yet, so this is a helper with no caller. We try to avoid landing utility surface ahead of its first use. Can you either bring these in alongside the first consumer (a web storage card would be the natural fit, mirroring 681), or hold this until that lands? Happy to merge the moment there's something calling it.

Ayush7614:
Thanks for the review — pushed a follow-up that wires `formatBytes`/`formatPercent` into their first consumer:

- `IndexStorageMini` on the repo overview rail (mirrors the CLI storage row from #681)
- `index_storage_bytes` added to `/overview-summary` sync block
- Component + server tests included

Should be ready for another look.

Ayush7614:
cc: @RaghavChamadiya

RaghavChamadiya (approved):

RaghavChamadiya:
LGTM

PR #681, opened 2026-07-04
feat(cli): show index storage footprint in status

Ayush7614:
## Summary
- Add a format_bytes CLI helper alongside the existing format_elapsed
- Show total .repowise on-disk size in repowise status for single-repo and workspace views

## Test plan
- uv run pytest tests/unit/cli/test_ui_package.py::test_format_bytes tests/unit/cli/test_status_storage.py
- Run repowise status on an indexed repo and confirm the storage row appears


RaghavChamadiya:
Nice one. The storage row is genuinely useful and the walk is safe (rglob + suppress(OSError) handles files vanishing mid-scan). Column counts line up across the header, the real row, and the not-indexed row. Merging this.

RaghavChamadiya (approved):

PR #682, opened 2026-07-04
feat(api-client): add fetchAllPaginated pagination helper

Ayush7614:
## Summary
- Add fetchAllPaginated to walk paginated REST list endpoints via has_more and next_offset
- Export the helper from @repowise-dev/api-client/pagination with unit tests

## Test plan
- packages/api-client/src/pagination.test.ts covers multi-page fetch, empty pages, and maxItems
- npm run test --workspace=@repowise-dev/api-client

### DIFF
--- a/packages/api-client/package.json
+++ b/packages/api-client/package.json
@@ -29,6 +29,7 @@
     "./meta": "./src/meta.ts",
     "./modules": "./src/modules.ts",
     "./overview": "./src/overview.ts",
+    "./pagination": "./src/pagination.ts",
     "./owners": "./src/owners.ts",
     "./pages": "./src/pages.ts",
     "./providers": "./src/providers.ts",
--- a/packages/api-client/src/index.ts
+++ b/packages/api-client/src/index.ts
@@ -1,1 +1,2 @@
 export * from "./client";
+export * from "./pagination";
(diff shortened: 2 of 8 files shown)


RaghavChamadiya:
Type wiring is correct, Paginated<T> matches the envelope field for field. Same note as 680 though: fetchAllPaginated has no callers in the tree yet, so I'd rather land it with its first real use than merge it standalone.

Separate thing worth fixing before it gets a caller: the while(true) only breaks on !has_more or next_offset == null. If a server ever returns has_more: true with an unchanged next_offset, or has_more: true on an empty page, this loops forever. A small guard (bail if next_offset <= offset, or a max-page cap) would make it safe against a misbehaving endpoint.

Ayush7614:
Addressed both points in the latest push:

1. **Loop guard** — `fetchAllPaginated` now bails when `next_offset` does not advance, when a page is empty but `has_more` is true, or when `maxPages` (default 1000) is hit. Tests cover the stuck-offset cases.

2. **First caller** — `listAllOwners` uses the helper and the Contributors page prefetches the full owner set (when total ≤ 120) for an accurate `OwnershipDistributionBar`, while the grid still loads incrementally via SWR infinite.

Happy to adjust the prefetch cap or landing surface if you prefer a different first use.

Ayush7614:
cc: @RaghavChamadiya

swati510:
@Ayush7614 thanks for taking this up. Merging now.

swati510 (approved):

PR #705, opened 2026-07-07
feat: add Eden AI as a first-class LLM provider and embedder

### LINKED ISSUE #704: Add Eden AI as a first-class LLM provider (and embedder)
## Summary

Add **Eden AI** as a first-class LLM provider (and embedder) in repowise.

[Eden AI](https://www.edenai.co/) is an EU-headquartered gateway that exposes 700+
models (Mistral, GPT, Claude, Gemini, Cohere, DeepSeek, Llama, …) behind a single
**OpenAI-compatible** endpoint (`https://api.edenai.run/v3`), with an EU endpoint
(`https://api.eu.edenai.run/v3`) for data residency / GDPR-sensitive workloads.

## Why it fits repowise

- It's OpenAI-compatible, so it slots into the existing provider pattern the same
  way `openrouter` / `deepseek` do — **no new dependency** (reuses the `openai` SDK
  with a custom `base_url`).
- One key unlocks many vendors, which is convenient for the BYOK model, and the EU
  endpoint is useful for teams with data-residency requirements.

## Proposed scope

- `EdenAIProvider` (`llm/edenai.py`) — `generate()` + streaming chat + `/models`
  discovery; default `mistral/mistral-small-latest`; `EDENAI_API_KEY` /
  `EDENAI_BASE_URL`; `reasoning` mapped to OpenAI `reasoning_effort` for OpenAI
  reasoning models routed via Eden.
- `EdenAIEmbedder` (`embedding/edenai.py`) — `openai/text-embedding-3-small` default.
- Registry + rate-limiter + server catalog + CLI selection/validation + web UI + tests + docs.

I've implemented this following `.github/CONTRIBUTING.md` and verified it end-to-end
against the live API. Happy to open the PR — filing this first per the issue-first
convention. Would you be open to it?

MVS-source:
## Summary

- Adds **Eden AI** as a first-class LLM provider (`EdenAIProvider`) and embedder
  (`EdenAIEmbedder`). Eden AI is an EU-headquartered gateway exposing 700+ models
  (Mistral, GPT, Claude, Gemini, Cohere, DeepSeek, Llama, …) behind a single
  **OpenAI-compatible** endpoint — so it reuses the existing `openai` dependency
  with a custom `base_url`, **no new dependency**.
- Follows the `openrouter` / `deepseek` pattern: `generate()`, streaming chat,
  live `/models` discovery, `EDENAI_API_KEY`, and an `EDENAI_BASE_URL` override
  for the EU endpoint (`https://api.eu.edenai.run/v3`) for data residency.
  `reasoning` maps to OpenAI `reasoning_effort` for OpenAI reasoning models routed
  via Eden; other models expose only `auto`.
- Wires the new provider/embedder through the registries, rate-limiter defaults,
  server provider catalog, CLI provider selection/resolution/validation, embedder
  auto-detection, the web settings UI, tests, and docs (`CONFIG.md`, `.env.example`).

## Related Issues

Closes #704

## Test Plan

- [x] Tests pass (`pytest`) — 31 new tests (`test_edenai_provider.py`,
      `test_edenai_embedder.py`); full provider + embedder suites: 220 passed, no regressions.
- [x] Lint passes (`ruff check`) — `llm/edenai.py` clean; the embedder carries the
      same single `RUF012` on `_DIMS` as the existing `openrouter.py` embedder.
- [x] Verified end-to-end against the live Eden AI API: `generate()`, streaming
      chat, `/models` discovery, and embeddings (1536-dim, L2-normalized).
- [ ] Web build (`npm run build`) — TS change is additive (new keys on existing
      typed records / const arrays).

## Checklist

Notes: The API key is read only from `EDENAI_API_KEY` (or passed explicitly) — never
logged or hardcoded. `.github/CONTRIBUTING.md` referenced a `run-config-form.tsx`
file that no longer exists in the tree, so only `provider-section.tsx` was updated.


RaghavChamadiya:
Thanks @MVS-source, and apologies that this has had no real review at all in the time it has been open. That is on us, not on you.

The proposal in #704 is well scoped: an OpenAI-compatible gateway with an EU option, no new dependency. That is the easy kind of provider to say yes to.

The branch needs a rebase before a content review is worth doing. You opened this on 7 July, and the Kimi provider landed on 14 July touching `registry.py`, `rate_limiter.py`, `provider_config.py` and `provider-section.tsx`, which are exactly the files a new provider edits. GitHub reports real conflicts rather than drift because of it.

For reference, the current contract for adding a provider is a module under `providers/llm/`, an entry in `_BUILTIN_PROVIDERS` in `registry.py` plus the API-key and base-URL env maps and the autodetect order, `PROVIDER_DEFAULTS` in `rate_limiter.py`, `PROVIDER_CATALOG` in `packages/server/src/repowise/server/provider_config.py`, the CLI selection maps in `packages/cli/src/repowise/cli/ui/provider_selection.py`, `validate_provider_config()` in the CLI helpers, the three maps in `packages/web/src/components/settings/provider-section.tsx`, tests under `tests/unit/test_providers/`, and docs. `.github/CONTRIBUTING.md:194-216` has it written down.

`git fetch origin && git rebase origin/main`, push, and I will review it on the rebased diff.

MVS-source:
Rebased onto `main` and pushed. No conflicts remain.

Two notes on the rebase, in case they are useful beyond this PR. The four files you expected to conflict (`registry.py`, `rate_limiter.py`, `provider_config.py`, `provider-section.tsx`) all auto-merged. The two real conflicts were in the embedder path, which this PR also touches, and both were upstream renames rather than drift: `MockEmbedder` became `KeylessEmbedder`, and the embedder auto-detection moved from `os.environ.get` to the `_get` helper that layers `.repowise/.env`. Both adopted rather than worked around.

While re-checking against the contract in `CONTRIBUTING.md:194-216`, three things came up.

**One functional defect, now fixed.** `edenai` was missing from `_EMBEDDER_KEY_ENV` in both `cli/providers/keys.py` and `server/mcp_server/_server.py`, so a key persisted to `.repowise/.env` never reached the embedder and only an exported variable worked.

**The embedder defaulted to a US-region model**, which is a poor look for a gateway whose reason to exist here is an EU endpoint. I measured every candidate against Eden AI's `/v3/embeddings` endpoint and rebuilt `_DIMS` from the results rather than from the upstream vendors' published widths. The default is now `amazon/amazon.titan-embed-text-v2:0`: EU region, 1024 dims, and the cheapest per token of the declared set. One entry in the old table was never actually served, which is exactly the sort of thing measuring catches and copying does not.

Worth flagging for reviewers: Eden AI returns `google/gemini-embedding-001` at its full 3072 width, where `OpenRouterEmbedder._DIMS` records 768 for the same model id. Both are correct for their own gateway. There is a comment on the table saying so, because reconciling the two would break the width guard.

The adapter had also drifted from `OpenRouterEmbedder`, which is the closest reference. It now uses `resolve_embedding_timeout` so `EDENAI_EMBEDDING_TIMEOUT` and the shared `REPOWISE_EMBEDDING_TIMEOUT` behave as they do everywhere else, which matters since #1475 swept that through every embedder on 14 August. It also gained the `dimensions` override and the post-response width guard.

**Two docs observations.** `CONTRIBUTING.md` step 4 asks for edits to `packages/web/.../run-config-form.tsx`, which no longer exists anywhere in the repo. And step 6's dedicated-page instruction points at `docs/agent/CODEX.md` and `OPENCODE.md`, both agent-CLI providers; API gateways like OpenRouter have no such page, so this PR documents in `docs/reference/CONFIG.md` instead. Happy to add pages if you would rather have them.

Verification on the rebased branch:

- `uv run pytest` over `tests/providers/`, `tests/unit/test_providers/`, `tests/unit/test_persistence/`, `tests/unit/server/mcp/` and `tests/unit/cli/`: 3931 passed, 0 failed. The embedder suite went from 13 to 22 cases, covering the width guard, the dimensions override and both timeout variables.
- Two pre-existing assertions needed updating rather than working around: `test_builtin_count` (11 to 12) and `test_every_builtin_provider_has_a_deliberate_budget`, which asks for an explicit interactive budget. `EdenAIProvider` inherits the 60.0 remote-API default, same as `openrouter`, so 60.0 is the deliberate answer.
- `ruff check` and `ruff format --check` clean on the changed files. Note for anyone reproducing: the repo pins `ruff>=0.6,<1` and a fresh `uv sync` installed 0.15.6, whose formatter reformats 629 unrelated files. Worth pinning.
- `npm run lint`, `npm run type-check` and `npm run lint:shared` all pass.
- `repowise risk upstream/main..HEAD` on its own diff: 0th percentile of this repo's fix-bearing files, diff shape Elevated at the 75th percentile driven almost entirely by size (+3.68 of the score from lines added). `impacted-tests` and `health` need an indexed repo, which I have not built here.

Disclosure: I work at Eden AI, so this is a vendor contributing its own provider. This change was AI-assisted; I reviewed the diff myself and every model id, dimension count and price in it was verified against the live API rather than assumed.

RaghavChamadiya:
Thanks @MVS-source. This is a better-argued rebase than most first reviews I write, and the disclosure at the bottom is the right way to contribute a vendor's own provider.

Taking the decision first, since you have earned a straight answer: **yes, we want this.** But the reason is the EU endpoint, not the model count. `openrouter` and `litellm` already cover "one key, many vendors", so on the aggregator axis alone this would be the third of a kind and I would have declined it. What repowise cannot offer today is a hosted provider that keeps source code in the EU. Every current option is a US endpoint unless you self-host Ollama or LiteLLM, and "run your own infra" is not an answer for a team that just wants the tool to be usable under their DPA. That is a real gap and this closes it.

Which is also why I would rather the docs led on that. `CONFIG.md` currently opens on the 700+ models and mentions the EU endpoint second. Flipping that emphasis is worth more to the person who needs this than the catalogue size is.

**One thing to change before I merge.** The diff inserts `edenai` into `PROVIDER_AUTODETECT_ORDER` at position four, ahead of `ollama`, `gemini`, `deepseek` and `kimi`:

```python
     "anthropic",
     "openai",
     "openrouter",
+    "edenai",
     "ollama",
     "gemini",
```

That changes resolution for people who are not asking for it. Anyone who has `EDENAI_API_KEY` exported for something unrelated and uses repowise with Gemini or DeepSeek keys silently moves to Eden AI on their next run, and the failure mode is a surprise bill on someone else's account rather than an error. Autodetect order is de facto precedence, so a provider adding itself should append. Move it to the end of the tuple and the same for the two doc copies of that chain in `CONFIG.md` and `CLI_REFERENCE.md`.

I do not think this was deliberate, the rest of the diff is scrupulous about not privileging Eden AI, which is exactly why the one place it does stands out.

Two smaller things, neither blocking:

The `_DIMS` divergence you flagged, Eden AI serving `google/gemini-embedding-001` at 3072 where `OpenRouterEmbedder` records 768, is the right call and the comment saying so is why. Rebuilding that table from `/v3/embeddings` instead of copying published widths is the part of this PR I trust most, because it is the part that could not have been guessed.

You are right about both `CONTRIBUTING.md` bugs. `run-config-form.tsx` is gone and step 6 should not point API gateways at the agent-CLI page pattern. Do not fix them here, I will take them separately so this PR stays about one thing. `CONFIG.md` is the correct home for this.

On the ruff note: yes, `ruff>=0.6,<1` is too loose given 0.15 reformats 629 files. Also mine to fix, not yours.

CI needed maintainer approval since this is a fork, I have approved it and the full matrix is running now. Push the autodetect change and I will merge on green.

MVS-source:
You are right, and thank you for catching it. Pushed in 6ba72bc.

The autodetect position was not deliberate, but I should still have caught it: I read that tuple while auditing the diff against the contract and recorded it as "consistent with the existing order" rather than asking why a new provider was sitting ahead of four incumbents. Your framing, that autodetect order is de facto precedence and the failure mode is a charge on the wrong account rather than an error, is the part I missed.

`edenai` now appends. I also moved the embedder detection, which had exactly the same problem and which you did not mention: I had placed it after `openrouter` in `resolve_embedder_name`, so an unrelated key would have shifted someone off their Gemini or OpenAI embedder too, and that one is worse because a changed embedder re-embeds the store at a different width.

Verified the scenario you described. With `EDENAI_API_KEY` set to something unrelated and a Gemini key present, resolution now lands on `gemini`.

On the doc copies: there were three of that chain rather than two. `CONFIG.md` and `website/cli-reference.md` both had it at position four, and `website/configuration.md` did not list `EDENAI_API_KEY` at all, which the original diff had simply missed. All three now end with it. I also reordered the provider and embedder enumerations in `CLI_REFERENCE.md`, `website/cli-reference.md` and `ARCHITECTURE.md`, since those mirror the resolution order and would otherwise contradict it.

I took the docs point too. The `CONFIG.md` section now opens on the hosted-EU-endpoint argument and why it matters for a team under a DPA, with the aggregator angle moved below it. Your reasoning for wanting this changed my own view of what the section is for, so it reads differently rather than just being reordered.

Left alone as you asked: the two `CONTRIBUTING.md` bugs and the ruff pin.

One thing worth flagging so it does not look like my mess: `tests/unit/cli/test_watch_cmd.py::TestSingleRepoTrigger` has two failures, `test_a_file_save_runs_an_update_over_the_working_tree` and `test_index_only_is_passed_through`. Both fail identically on a clean `upstream/main` checkout with no changes applied, so they predate this branch. Everything else is green: 3929 passed on the broad run, and the four test files that reference the autodetect order all pass after the reorder.

MVS-source:
Rebased onto current `main` and force-pushed, because the red CI was not mine.

The three failures were in `tests/unit/cli/test_agent_matrix.py`, complaining that `docs/business/COMMERCIAL.md` does not say "the eleven MCP tools". Nothing in this branch touches that file or the MCP surface. `main` was failing the same way at `ad654a16` this morning and went green at `ec5a8af2`, which is #1771, "count COMMERCIAL.md's tools the way the doc counts them". My branch was based on an older `main` and so carried the pre-fix test. It should be green now that it carries yours.

Verified locally after the rebase: `test_agent_matrix.py` 85 passed, and the provider, persistence, provider-config, mcp-resolution and workspace-autodetect suites 456 passed. `ruff check` clean. The autodetect order still reads `anthropic, openai, openrouter, ollama, gemini, deepseek, kimi, edenai`.

Separately: I said earlier that two `test_watch_cmd.py` tests fail on a clean `main`. They do fail for me locally, on both `main` and this branch, but they passed in your CI, so treat that as an artefact of my machine rather than something you need to look at. I should have checked it against CI before mentioning it.

RaghavChamadiya (approved):
Appending `edenai` rather than slotting it in is exactly it, and the comment you left on the tuple says why better than the review did:

```python
    # A provider adding itself appends: autodetect order is de facto precedence,
    # and an unrelated EDENAI_API_KEY in the environment must not silently take
    # over from a provider the user was already resolving to.
```

That is the rule for the next provider too, written where the next person will read it.

Full matrix green on 3cbcc1b, all three Python versions plus ruff. Thanks for staying with this one, it has been open a long time.

Two things we still owe you from the earlier discussion, both ours and neither blocking: CONTRIBUTING.md step 4 still names `run-config-form.tsx`, which no longer exists, and the `ruff>=0.6,<1` pin lets a fresh `uv sync` pull 0.15, whose formatter rewrites hundreds of files. Those are on our list, not yours.

PR #739, opened 2026-07-09
feat(risk): split test-file breakage out of will_break (#672)

### LINKED ISSUE #672: [Feature] get_risk: separate test-file breakage from production in will_break
## Problem

In `get_risk` PR mode, the `will_break` list mixes production files and test files together. When a change breaks a handful of tests, they crowd out the production impact in the same capped list, and the caller cannot tell structural breakage from test breakage at a glance.

## Proposed Solution

Segment test-file breakage out of `will_break` into its own field (for example `will_break_tests`), so the primary list stays focused on production impact and the test fallout is reported separately. The graph already carries an `is_test` signal on nodes (used elsewhere in the same tool for test-gap logic), so the split is a filter on the existing data rather than new analysis.

## Alternatives Considered

Leave them merged and let the caller re-derive which entries are tests. That pushes work onto every consumer and defeats the point of the capped, ranked list.

## Additional Context

- `packages/server/src/repowise/server/mcp_server/tool_risk.py` (`will_break` is built and returned as a single list around line 805 / 873; `is_test` is already read in this file for test-gap detection). `will_break_consumers` is the separate cross-repo split and is unrelated.

Real5K:
## Summary

- In `get_risk` PR mode, the `will_break` directive list mixed production and test files, so a burst of broken tests could crowd real structural impact out of the capped list.
- Segments test files into a new `will_break_tests` field, reusing the `is_test` flag already loaded into `node_meta` (no new query). Partitioning happens **before** the per-list cap, so production impact keeps its full budget of 5 even when many tests break.
- The directive `summary` now reports the test-breakage count. Additive change — no consumer relied on test files being inside `will_break`, so it's low risk.

## Related Issues

Closes #672

## Test Plan

Added `test_get_risk_pr_directive_splits_test_breakage`, which drives a diff whose transitive impact includes both a production importer and a test importer and asserts each lands in the correct bucket (`will_break` vs `will_break_tests`).

- [x] Tests pass (`pytest tests/unit/server/mcp/test_risk.py` — 10/10)
- [x] Lint passes (`ruff check` on changed files — clean)
- [x] Web build passes — N/A, no frontend changes

## Checklist

- [x] I have updated documentation

### DIFF
--- a/docs/MCP_TOOLS.md
+++ b/docs/MCP_TOOLS.md
@@ -282,7 +282,7 @@ Modification risk assessment for files or a set of changed files.
 
 **Returns:** Per-file risk score (0–10), hotspot status, dependent count, co-change partners, blast radius, recommended reviewers, test gap analysis, security signals. In workspace mode, enriched with cross-repo co-change partners and contract dependencies.
 
-When `changed_files` is passed, the response leads with a `directive` block. In workspace mode that directive also carries the cross-repo fallout of the changed repo:
+When `changed_files` is passed, the response leads with a `directive` block. Its core lists are the local blast radius: `will_break` (production files that depend on the diff and are likely to break), `will_break_tests` (test files impacted the same way, kept separate so a burst of broken tests doesn't crowd production impact out of the capped list), `missing_cochanges` (historical co-changers absent from the diff), and `missing_tests` (changed files without test coverage). In workspace mode that directive also carries the cross-repo fallout of the changed repo:
 
 - `will_break_consumers`: services in *other* repos that depend on this one (structural impact), each with `repo`, `service`, `distance`, `score`, and the edge kinds carrying the impact.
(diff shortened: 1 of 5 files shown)


RaghavChamadiya:
Read through this and it does exactly what #672 asked, nothing extra. The important bit is that the partition into will_break and will_break_tests happens before the caps are applied, so a pile of broken tests can't push production impact out of its budget of 5. That's the whole point of the issue and it holds here. Nice that it reuses the is_test flag already in node_meta instead of adding a query, and the doc and summary line are both updated to match.

One small thing to keep in mind, not a blocker: the p not in test_paths check relies on the paths in transitive_affected being the same string form as the node_id keys. They come from the same graph so it should be fine, and your test covers the happy path, but that's the one spot where a test file could quietly leak back into will_break if the path forms ever drift apart.

Looks good to merge

RaghavChamadiya (approved):

Issue #746, opened 2026-07-10
[Bug] worktree seed + delegated update writes new-file pages to the wrong DB / repo identity (not a selection-budget issue)


RaghavChamadiya:
## Describe the Bug

In the worktree seed flow (`init --seed-from`, follow-up to #655), a file added in the worktree does not end up with a page in the worktree's own index, even though the delegated `update` detects it, selects it, and generates its page. The page is written to the wrong database.

Root cause is DB resolution, not page selection. When `REPOWISE_DB_URL` is set (or any time the resolved DB path differs from the worktree path), the seed copies the base `.repowise` into `<worktree>/.repowise`, but the delegated `update` resolves its DB from the env var / base repo rather than the worktree. The new file's page is generated and persisted, into the base repo's DB, under a fresh `repository` row keyed on the worktree's `local_path`. So base pages and newly generated pages end up under different `repo_id`s, and anything reading the worktree's own `<worktree>/.repowise/wiki.db` never sees the new page.

Verified by instrumenting the delegated update:

```
select_pages: n_files=1 budget=1 available={'file_page':1,...} alloc_file=1 file_cands=['new_file.py']
generated_pages includes 'new_file.py'
_persist_full_update: repo_path=...\feature-worktree-provider   (worktree)
                      db_url  =...\gitrepo\.repowise\wiki.db    (base repo, from REPOWISE_DB_URL)
                      pages under this repo_id include file_page:new_file.py
                      total pages across all repos in that DB = 20 (15 base + 5 worktree)
```

The worktree's copied `wiki.db` still holds only the 15 seeded base pages.

Note: page selection is NOT the problem. `allocate_budget` has a small-repo floor, so a typical 1-3 file update (batch of `N <= SMALL_REPO_THRESHOLD` files) floors the budget at `N` and gives every candidate bucket at least one page. A plain `repowise init` + add a symbol-bearing file + `repowise update` (no worktree, no env pin) does produce a page for the new file. Confirmed against `wiki_pages`.

## Steps to Reproduce

1. `repowise init <repo> --provider mock --yes`
2. Create a git worktree on a branch that adds a new source file with real symbols
3. `repowise init <worktree> --seed-from <repo>` (delegates to update)
4. `SELECT target_path FROM <worktree>/.repowise/wiki.db :: wiki_pages` has no row for the new file, even though the update reported it as added

[...]

PR #747, opened 2026-07-10
feat(cli): auto-seed git worktrees from their base checkout

RaghavChamadiya:
## What

Makes git worktrees seamless: `repowise init` and `repowise update` inside a linked worktree now detect the base checkout automatically (via `git rev-parse --git-common-dir`), seed the worktree's index from it, and catch up incrementally. No flags, no paths. Follow-up to #655, which introduced the manual `--seed-from` flag.

```
cd my-feature-worktree
repowise update
[worktree] Unindexed linked worktree of /path/to/base; seeding its index.
```

- `--seed-from <path>` stays as an explicit override for unusual layouts
- new `--no-seed` forces a cold init inside a worktree
- `update` auto-seeding means post-commit hooks and coding agents running in a fresh worktree get the fast path with zero setup

## Bug fixes along the way

**Seeded indexes no longer split in two.** `upsert_repository` matches by `local_path`, so a seeded wiki.db still pointed at the base checkout; the first update in the worktree minted a second repository row named after the worktree dir and regenerated everything under it, while the seeded pages stayed orphaned under the old row (breaking prior-page reuse and repo-scoped queries). Seeding now adopts the copied row: name and `local_path` are re-pointed at the worktree and the repo-level pages are retargeted.

**`[workspace]` notices render again.** `[dim][workspace][/dim]` was being parsed as console markup and printed without its prefix. Both the workspace notice and the new `[worktree]` notice are escaped.

**The #655 test suite is green again.** The seed tests were failing on main's CI (integration tests only run on push to main, so the PR run never exercised them) and failed differently on Windows:

- raw `git commit` calls without identity env (exit 128 on CI runners); now routed through the identity-injecting `_git` helper
- `git checkout main` assumed the default branch name; now captured via `rev-parse`
- `with sqlite3.connect(...)` never closes the handle, which broke worktree cleanup on Windows; replaced with closed-connection helpers
- the `git_work_repo` fixture pins `REPOWISE_DB_URL` to the base repo's db, silently routing the worktree's delegated update into the wrong database; the seed tests now drop the env var and use repo-local DBs
- substring assertions broke when Rich line-wrapped long Windows temp paths; whitespace is collapsed before matching
- cleanup order (`rmtree` before `git worktree remove`) left git bookkeeping pointing at a missing dir

The index-only survival assertion moved from `wiki_pages` (empty in index-only runs) to `graph_nodes`, and the "new file gets a page" assertion is dropped: update-time coverage budgeting prevents it on every platform, tracked as #746.

## Docs

New `docs/WORKTREES.md` (behavior, skip conditions, overrides, troubleshooting); WORKSPACES FAQ, CLI_REFERENCE, README, and QUICKSTART updated; changelog entries added under Unreleased (and the #655 entry moved out of the 0.25.0 section it had landed in).

## Testing

- 5 new unit tests for worktree detection and seed preconditions
- 4 new integration tests (init auto-seed, `--no-seed`, update auto-seed, unindexed-base fallback) plus the 5 repaired #655 tests
- full `tests/integration/test_cli.py` (28) and `tests/unit/cli` + `tests/unit/workspace` (1162) green on Windows; end-to-end verified manually with the mock provider (single repository row, seeded pages reused, state advanced to worktree HEAD)

### DIFF
--- a/README.md
+++ b/README.md
@@ -544,6 +544,7 @@ pricing: **[docs/COMMERCIAL.md](docs/COMMERCIAL.md)** · [Get in touch →](http
 repowise init [PATH]      # index codebase (one-time; --index-only skips LLM)
 repowise serve [PATH]     # MCP server + local dashboard
 repowise update [PATH]    # incremental update (<30s; --workspace for all repos)
+                          # git worktrees auto-seed from the base checkout (docs/WORKTREES.md)
 repowise search "<q>"     # search the wiki (fulltext / semantic / symbol)
 repowise health           # code-health KPIs + lowest-scoring files
 repowise risk main..HEAD  # score a branch / PR range for defect risk
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -9,6 +9,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ---
 
+## [Unreleased]
+
+### Added
+- **Worktrees just work.** `repowise init` and `repowise update` inside a linked git worktree now auto-detect the base checkout and seed the worktree's index from it, then catch up incrementally; no flags needed. `--seed-from <base>` remains as an explicit override and `--no-seed` forces a cold init. See [WORKTREES.md](WORKTREES.md). (#655 introduced the manual flag; this release makes it automatic.)
+
+### Fixed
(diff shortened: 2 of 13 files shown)


swati510 (approved):
LGTM

Issue #740, opened 2026-07-10
Configuring Repowise for Kimi


azpublic:
Was trying to get Repowise to work with my Kimi for coding provider but it was having problems ... so I asked Kimi to help me fix it ... and then I asked kimi to document what we did to make it work. Thought I'd share for anyone with similar problems. Took me about 20 minutes ... What a time to be alive

# Repowise + Kimi BYOK Setup Guide

This document records how to configure [repowise](https://repowise.dev/) to generate documentation for this repository using a **Bring Your Own Key (BYOK)** Moonshot / Kimi API key through Kimi's Anthropic-compatible endpoint.

## Goal

Run `repowise init` to generate the auto-generated wiki docs, graph intelligence, and architectural decision records for this Unreal Engine 5.5 project, while keeping all LLM calls on your own Kimi/Moonshot credentials.

## What works

- `repowise init --force`
- `repowise update`
- `repowise serve` / MCP tools

### 1. `.repowise/config.yaml`

```yaml
provider: anthropic
model: kimi-for-coding
embedder: openai
reasoning: auto
wiki_style: comprehensive
commit_limit: 500
follow_renames: false

# Kimi models enforce a fixed temperature. For kimi-for-coding use 0.6;
# for kimi-k2.5/k2.6 thinking mode use 1.0.
temperature: 0.6

exclude_patterns:
  - Binaries/
  - Intermediate/
  - Build/
  - Packaged/
  - Saved/
  - DerivedDataCache/
  - Content/_GENERATED/
  - Content/StarterContent/
  - Content/Tests/
  - "*.generated.*"

anthropic:
  base_url: https://api.kimi.com/coding/
```

Notes:

- We use `provider: anthropic` because Kimi exposes an Anthropic Messages API-compatible endpoint at `https://api.kimi.com/coding/`.
- `model: kimi-for-coding` is the model ID on Kimi's side.
- `temperature: 0.6` is required for `kimi-for-coding`. If you switch to `kimi-k2.5` / `kimi-k2.6`, use:
  - `temperature: 1.0` for thinking mode
  - `temperature: 0.6` for instant / non-thinking mode
- The embedder is left as `openai` here because Kimi does not expose an embedding endpoint that repowise natively supports. If you do not have an OpenAI key, change `embedder: openai` to `embedder: mock` to skip semantic search; graph/git/health layers still work index-only.

### 2. `.repowise/.env`

[...]

PR #753, opened 2026-07-10
fix: handle non-text first content block in AnthropicProvider

### LINKED ISSUE #750: AnthropicProvider crashes when the first content block is not a TextBlock (ThinkingBlock from Anthropic-compatible endpoints)
## Bug

`AnthropicProvider` assumes the first content block of a Messages API response is always a `TextBlock`:

```python
# packages/core/src/repowise/core/providers/llm/anthropic.py (line ~223)
result = GeneratedResponse(
    content=response.content[0].text,
    ...
)
```

Some Anthropic-compatible endpoints return thinking/reasoning blocks *before* the text block. When that happens, generation crashes with:

```text
'ThinkingBlock' object has no attribute 'text'
```

First reported in the field in #740, where a user running repowise against Kimi's Anthropic-compatible endpoint (`https://api.kimi.com/coding/`) via `provider: anthropic` + `anthropic.base_url` hit this on every request. The same crash would occur if extended thinking is ever enabled against Anthropic's own API, since thinking blocks precede text blocks there too.

## Fix

Scan `response.content` for the first block that carries text instead of indexing block 0:

```python
text_content = ""
for block in response.content:
    if hasattr(block, "text"):
        text_content = block.text
        break
```

(Or match on `block.type == "text"`.) The user in #740 verified this patch fixes the crash end to end.

## Scope

- One-line-ish change in `core/providers/llm/anthropic.py`, plus a unit test with a mocked response whose content is `[ThinkingBlock, TextBlock]`.
- Independent of, but related to, the first-class Kimi provider discussed in #740.

TemRevil:
Fixes #750

`AnthropicProvider` assumed `response.content[0]` is always a `TextBlock` and read `.text` from it. Some Anthropic-compatible endpoints (and Anthropic itself with extended thinking) put a thinking/reasoning block first, so that access raised `'ThinkingBlock' object has no attribute 'text'` and crashed every generation, as reported in #740 against Kimi's endpoint.

Following the approach in the issue, added a `_first_text_block()` helper that scans `response.content` for the first block carrying text instead of indexing block 0 (falling back to an empty string). 

Added `test_generate_skips_leading_non_text_block`, which mocks a response whose content is `[thinking_block, text_block]` (the thinking block `spec`'d without `.text` so accessing it raises `AttributeError`, matching the real crash) and asserts the text block's content is returned.

```
tests/unit/test_providers/test_anthropic_provider.py .............  [100%]
13 passed
```

### DIFF
--- a/packages/core/src/repowise/core/providers/llm/anthropic.py
+++ b/packages/core/src/repowise/core/providers/llm/anthropic.py
@@ -50,6 +50,22 @@ log = structlog.get_logger(__name__)
 _DEFAULT_BASE_URL = "https://api.anthropic.com"
 
 
+def _first_text_block(content: list[Any]) -> str:
+    """Return the text of the first text block in a Messages API response.
+
+    The response content is not guaranteed to start with a text block: some
+    Anthropic-compatible endpoints (and Anthropic itself with extended
+    thinking enabled) emit reasoning/thinking blocks before the text block.
+    Indexing ``content[0].text`` blindly raises ``AttributeError`` in that
+    case, so scan for the first block that actually carries text.
+    """
+    for block in content:
+        text = getattr(block, "text", None)
+        if text is not None:
+            return text
+    return ""
+
+
 def _anthropic_model_options(
     api_key: str,
     base_url: str,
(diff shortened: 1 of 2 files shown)


TemRevil:
#787 landed the same fix (scan for the first block with text, fall back to empty) and closed #750, so this one's redundant now. Closing it out. Good to see it sorted.

PR #761, opened 2026-07-10
chore(upgrade): resync bundled changelog with docs/CHANGELOG.md

swati510:
One-file resync: the latest changelog entry landed in `docs/CHANGELOG.md` without the copy bundled into the CLI wheel, so `test_bundled_changelog_in_sync` currently fails on every open PR's merge ref. Merging this unblocks them all.

### DIFF
--- a/packages/core/src/repowise/core/upgrade/_data/CHANGELOG.md
+++ b/packages/core/src/repowise/core/upgrade/_data/CHANGELOG.md
@@ -11,7 +11,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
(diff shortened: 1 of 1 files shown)


RaghavChamadiya (approved):
Thanks for catching this. The bundled copy drifting from `docs/CHANGELOG.md` was failing `test_bundled_changelog_in_sync` on every open PR's merge ref, so this one unblocks a lot of other work. Merging it first.

RaghavChamadiya:
Went to merge this and found it conflicts with main. Checked why: the bundled copy and `docs/CHANGELOG.md` are already identical on main, so `test_bundled_changelog_in_sync` passes now and there is nothing left for this to resync. Suggest closing as overtaken by events rather than resolving the conflict. Thanks for filing it.

PR #788, opened 2026-07-11
feat(codex): deliver architectural decisions into Codex SessionStart

### LINKED ISSUE #786: [Feature] Deliver architectural decisions into Codex sessions (SessionStart block + edit-time notice)
## Problem

On Claude Code, repowise pushes the relevant architectural decisions to the agent through hooks: a relevance-ranked block at session start (scored against the files the session is about to touch) and a one-line "governed by" notice when the agent edits a file a decision governs. On Codex, the hooks today only remind the agent to use the MCP tools and flag staleness after edits, so the decision layer never reaches a Codex user at the moment it matters.

## Proposed Solution

Wire the existing decision-delivery builders into the Codex hook handlers.

**The seam.** The block builders in `packages/cli/src/repowise/cli/commands/augment_cmd/decision_inject.py` are harness-agnostic: they take a repo path and a session id and return text, or nothing when nothing clears the relevance floor. Claude wires them in `session_start.py`. The Codex handlers live in `codex.py` (`_handle_codex_context_event` for SessionStart / UserPromptSubmit, `_handle_post_edit_use` after edits) and both already return injected strings, so the plumbing to carry decision text is there.

**The work.**

- Call the SessionStart decision block from the Codex SessionStart handler, appended to the existing reminder, under the same relevance floor and token cap Claude uses.
- Emit the edit-time "governed by" notice from the Codex post-edit handler when the edited file has a governing decision, honoring the same once-per-session-per-decision cap.
- Record injected decision ids in the `sessions.db` sidecar the same way Claude does, so the follow / contradict feedback check works for Codex too (the check itself reads transcripts, which is the sibling adapter issue #785).
- Verify the Codex hook events actually deliver the file path and cwd the builders need, and adjust the handler signatures if not.

## Alternatives Considered

- A static "this repo has decisions, run `repowise decision list`" reminder: rejected, it pushes the work back onto the agent and loses the relevance ranking and the edit-time targeting that make the Claude version useful.
- Waiting for the Codex adapter first (#785): unnecessary. This delivers decisions from any source (manual, ADR, git-mined), so it is useful even before Codex sessions are mined. The two issues can land in either order.

## Additional Context

- Reference wiring: `augment_cmd/session_start.py` (Claude SessionStart) and `augment_cmd/decision_inject.py` (the shared builders).
- Codex handlers to extend: `augment_cmd/codex.py`.
- The behavior this ports is documented in the hooks and decisions guides (see repowise-dev/repowise#784).

**Done when:** a Codex session start surfaces the relevant decisions block on a repo that has decisions; editing a governed file surfaces the governing decision once; nothing is injected when nothing clears the floor; the Claude Code path is untouched.

DresdenGman:
Wire the existing standing-decisions injection into the Codex SessionStart hook so decisions follow the user across agents — matching the behavior that Claude Code already has.

- Extract the MCP-usage paragraph into a module-level `_MCP_CONTEXT` constant
- On SessionStart, append the relevance-ranked decisions from `_session_decision_block()` to the context block
- Pass `session_id` from the hook payload so decision scoring uses it

Closes #786.

### DIFF
--- a/packages/cli/src/repowise/cli/commands/augment_cmd/codex.py
+++ b/packages/cli/src/repowise/cli/commands/augment_cmd/codex.py
@@ -1,14 +1,30 @@
-"""Codex lifecycle hooks: SessionStart/UserPromptSubmit + post-edit staleness."""
+"""Codex lifecycle hooks: SessionStart/UserPromptSubmit + post-edit staleness.
+
+SessionStart delivers a short MCP-usage context block plus the same
+relevance-ranked standing-decisions block that Claude Code receives
+(see :mod:`.decision_inject`), so decisions follow the user across agents.
+"""
 
 from __future__ import annotations
 
 from pathlib import Path
 
 from ._shared import _find_repo_root
+from .decision_inject import _session_decision_block
+
+_MCP_CONTEXT = (
+    "[repowise] This repository has a local codebase wiki and graph index. "
+    "Use the repowise MCP tools for architecture overview, semantic search, "
+    "implementation context, risk/hotspot checks, decision history, and "
+    "dead-code analysis. After meaningful edits or git operations, run "
+    "`repowise update` when refreshed context is needed."
+)
 
 
-def _handle_codex_context_event(event: str, cwd: str) -> str | None:
-    """Return short Codex developer context when repowise is initialized."""
+def _handle_codex_context_event(
+    event: str, cwd: str, session_id: str = ""
+) -> str | None:
+    """Return Codex developer context + standing decisions on SessionStart."""
     if event not in ("SessionStart", "UserPromptSubmit"):
         return None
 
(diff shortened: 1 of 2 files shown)


RaghavChamadiya:
Thanks for picking this up @DresdenGman
The SessionStart wiring looks good and matches how Claude does it in _handle_claude_session_start. Same builder, same floor and cap, session_id threaded through, and injection recording comes along for free since _session_decision_block handles that internally. No notes on that part.

One thing though: this closes #786, but the issue asks for two delivery moments and this only does one. The edit-time "governed by" notice is still missing. _handle_post_edit_use is untouched and only emits the staleness reminder, so editing a governed file in Codex won't surface the decision the way it does in Claude. The builder for it already exists (_edit_decision_notice), so it should be a small addition. Note that the post-edit handler doesn't get session_id today, so that branch needs it plumbed through too.

Can you either add the edit-time notice here, or drop "Closes #786" to "Part of #786" and open a follow-up? Just don't want the issue auto-closing with half the work still open. The SessionStart change itself is ready.

DresdenGman:
Added the edit-time governing-decision notice — `_handle_post_edit_use` now accepts `session_id` + `tool_input` and calls `_edit_decision_notice` with the edited file's repo-relative path, same builder Claude uses. Staleness reminder + decision notice combined into one response.

RaghavChamadiya (approved):
LGTM

Issue #765, opened 2026-07-10
[Bug] dataflow CFG does not recurse into with-statement bodies


RaghavChamadiya:
## Summary

The dataflow CFG builder (`packages/core/src/repowise/core/analysis/health/dataflow/cfg.py`) has no handling for `with_statement` nodes. A `with` body is appended to the current basic block as a linear run of statements, so any control flow nested inside it (if/else, loops, early returns) never splits blocks. Everything downstream that consumes block structure over-approximates inside `with` bodies.

## Repro

Per-function def-use/liveness over this file treats the two functions differently:

```python
def outside(cond):
    if cond:
        result = 1
    else:
        result = 2
    return result

def inside(ctx, cond):
    with ctx.push():
        if cond:
            result = 1
        else:
            result = 2
    return result
```

`outside` is analyzed correctly: the if/else splits blocks, both defs of `result` reach the `return`, and both are live.

In `inside`, the whole `with` body lands in one block. Within a single block, def/use ordering falls back to line numbers, so the def from the first branch (`result = 1`) looks like it is shadowed by the second (`result = 2`) before the read, and liveness concludes the first assignment is never observed. Real-world example shape: `django/template/loader_tags.py::BlockNode.render` assigns `result` in both arms of an if/else inside `with context.push():` and returns it after the block.

## Impact

- Reaching definitions inside `with` bodies are block-coarse. Extract Method IN/OUT inference consumes these; the error direction is conservative (extra inputs/outputs, weaker plans), not wrong output, but precision drops in exactly the code that uses context managers heavily.
- `CFG.reachable_ids()` cannot see unreachable statements nested inside a `with` body (e.g. code after a `return` inside the block), since the body is never expanded into blocks.

## Related gap in the same layer

Stores in an `except` body that re-raises are read by the `finally` block, but there is no except-body to finally edge in the flow, so those defs are not seen as reaching the finally reads:

```python
failed = False
try:
    run()
except Exception:
    failed = True
    raise
finally:
    if not failed:  # this read never observes the except-body store
        cleanup()
```

(#760 added the protected-region-entry to finally-entry edge; the except-body re-raise path is a different missing edge.)

Same conservative direction, same fix area, so recording it here rather than as a separate issue.

## Expected

`with` bodies should be processed through the same statement-sequence recursion that if/try bodies get (Python `with_statement`, plus equivalent constructs in other supported grammars where applicable), and except bodies should have an edge to the finally entry.

[...]

PR #791, opened 2026-07-12
fix(health): recurse into Python with statement bodies

Divyansh9192:
## Summary

- Add CFG support for Python `with_statement` nodes by introducing `with_kinds` to `LanguageNodeMap`
- Process `with` bodies through the existing statement-sequence recursion so nested control flow (`if`, `for`, `while`, `return`, etc.) is represented correctly in the CFG
- Add regression tests covering nested branches, loops, unreachable code, and preservation of CFG structure inside `with` statements

## Related Issues

Partially addresses #<issue-number>.

This PR implements the `with_statement` CFG recursion described in the issue. The related `except` → `finally` CFG edge is intentionally left out of this change.

## Test Plan

- [x] Tests pass (`pytest`)
- [ ] Web build passes (`npm run build`) *(not applicable – backend-only change)*

## Checklist

### DIFF
--- a/packages/core/src/repowise/core/analysis/health/complexity/languages.py
+++ b/packages/core/src/repowise/core/analysis/health/complexity/languages.py
@@ -180,12 +180,21 @@ class LanguageNodeMap:
     #     (``throw``) edge to the function exit; ``break`` / ``continue`` edge to
     #     the enclosing loop's exit / header. A language that omits one simply
     #     records that statement as a plain straight-line node (no special edge).
+    #   * ``with_kinds`` -- structured resource/context-management statements
+    #     whose bodies execute as a normal statement sequence (Python
+    #     ``with_statement``; analogous constructs in other languages such as
+    #     C# ``using`` or Java try-with-resources where applicable). Unlike
+    #     ``if`` or ``loop`` statements, these do not introduce a control-flow
+    #     split themselves; instead, the CFG builder recursively processes the
+    #     body so nested branches, loops, and early exits create their normal
+    #     basic-block structure.
     if_kinds: frozenset[str] = frozenset()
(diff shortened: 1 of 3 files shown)


RaghavChamadiya:
Thanks for taking this on, and the CFG side of the change is sound. Unfortunately I have to close this in favour of #836, which landed on the same issue and covers a gap this PR doesn't.

The problem is that the def/use dialect never learned about with_statement. _record_head(cur, with_node) routes into dialects/python.py::_head, where with_statement matches neither loop_kinds nor branch_kinds, so it hits the else: self._process(node, ...) fallback on line 91. That walks the whole with_statement subtree, body included. The result is that the body's defs and uses get recorded into the head block and into the new body blocks.

Running issue #765's own repro on this branch:

block 2 kind='normal' reachable=True defs=[('result', 5), ('result', 7)] uses=[('ctx', 3), ('cond', 4)]

Both defs of result still sit in one block in line order, which is exactly the symptom the issue reports. The branch blocks this PR adds are correct, but the coarse block they were meant to replace is still there next to them.

The dead-code case has a sharper version. The unreachable statement's def leaks into a reachable block:

block 2 kind='normal' reachable=True       defs=[('dead', 5)]
block 5 kind='unreachable' reachable=False defs=[('dead', 5)]

Worth noting for future PRs in this area: all 31 tests here pass because they assert only on block kinds and reachability, never on defs/uses, which is the dimension the bug lives in. A compute_def_use assertion would have caught it. That's a good habit for dataflow changes generally, since the CFG shape can look right while the facts derived from it are wrong.

For reference, the missing piece is a with_kinds branch in _head that processes only the with_clause and skips the body block. #836 does that, plus the except-to-finally edge. Appreciate the effort here, and please do pick up another issue if you're interested

PR #815, opened 2026-07-12
fix(go): detect function references in go to rescue FPs on dead-code 

### LINKED ISSUE #814: [BUG] Dead code FPs on Go functions taken as arguments
## Describe the Bug

 Go functions passed as arguments, not called (`pkg.Func without ()`) are flagged as `unused_export` at 100% confidence despite being actively used.

## Steps to Reproduce

1. Have a public Go function used only as a function value argument, never called directly
```
// pkg
func ExampleFunction([]Args) 
  
// main.go
 Function(..., pkg.ExampleFunction)  // passed as value, not called
```
2. run repowise dead-code 

## Expected Behavior

`pkg.ExampleFunction` shouldnt be flagged

## Actual Behavior

Flagged as `unused_export` with 100% confidence

## Environment

- OS: Ubuntu 24.04
- Python version: 3.12.3
- Repowise version: 0.31.0
- Installation method: pip

crizah:
## Summary

 - Functions passed as references in arguments, flagged as `unused_export` with 100% confidence. 
 
 - Added `selector_expression` query in `go.scm` to capture `pkg.Func` references passed as call arguments
 
## Related Issues

closes https://github.com/repowise-dev/repowise/issues/814

## Test Plan

- [x] Verified on a real repo, 2 FPs gone
- [x] Added regression test (`test_func_value_as_argument_not_flagged`): Confirmed fails against pre-fix code (flagged `unused_export`) and passes with the fix.

## Checklist

### DIFF
--- a/packages/core/src/repowise/core/ingestion/queries/go.scm
+++ b/packages/core/src/repowise/core/ingestion/queries/go.scm
@@ -77,6 +77,20 @@
   arguments: (argument_list) @call.arguments
 ) @call.site
 
+; Package-qualified function value passed as call argument: f(pkg.Handler, ...)
+; Rescues functions used as first-class values — passed as callbacks, middleware,
+; or handlers rather than called directly. Without this, pkg.Func referenced
+; only in argument position (never in function: position) has no call edges
+; and is incorrectly flagged as an unused export.
+(call_expression
+  arguments: (argument_list
+    (selector_expression
+      operand: (_) @call.receiver
+      field: (field_identifier) @call.target
+    )
+  )
+) @call.site
+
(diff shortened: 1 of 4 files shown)


RaghavChamadiya:
Thanks for working on this @crizah 

Verified locally: query fires correctly for multiple function-value args in a single call (apply2(pkg.Upper, pkg.Lower) produces a separate match per selector, so both get edges), and all 7 tests in test_go_dead_code.py pass. Nice scoped fix with a regression test that actually fails pre-fix.

Merging this

RaghavChamadiya (approved):

PR #816, opened 2026-07-13
feat(codex): add Codex session transcript adapter support

### LINKED ISSUE #785: [Feature] Codex session adapter: mine decisions and demand-weight docs from Codex transcripts
## Problem

repowise already learns from Claude Code sessions: it reads the local transcripts into a normalized event stream, then mines architectural decisions from your corrections and tilts the docs-generation budget toward the modules you actually ask about. Today only Claude Code feeds that loop. Codex users get none of it, even though the ingest layer was built to take more than one harness.

## Proposed Solution

Add a Codex adapter so Codex transcripts flow into the same event stream the miners already consume.

**The seam.** `packages/core/src/repowise/core/sessions/` defines a `HarnessAdapter` contract (`adapters/base.py`) with exactly two methods:

- `discover(repo_root)` returns the transcript files for a repo.
- `normalize(raw_line)` turns one raw transcript line into an `Event`.

Iteration, cursoring, and both miners (`miners/decisions.py`, `miners/demand.py`) are shared code built on `Event`, so they need no changes. `adapters/claude_code.py` is the reference implementation and the module `README.md` has an "Adding a harness" section.

**The work.** Add `adapters/codex.py` implementing the contract for Codex:

- `discover()`: locate where the Codex CLI writes session transcripts for a given repo. Confirm this format and location first, it drives everything else.
- `normalize()`: map Codex's line schema onto `Event` / `ToolUse` / `ToolResult`. The load-bearing detail: tool events must normalize to the same tool names the miners key on (`get_answer`, `search_codebase`, and the edit tools), or the mined signal comes out empty. Return `None` for lines you cannot parse; never raise on content.
- Add a real-shaped Codex JSONL fixture under `tests/unit/sessions/data/` and tests mirroring the Claude adapter's.

**Both learning loops come from this one adapter.** Session-mined decisions and the demand-weighted docs budget both ride the shared event stream, and the demand miner is harness-agnostic, so "docs follow your questions" starts working for Codex users the moment Codex events flow in. No separate follow-up needed for docs.

## Alternatives Considered

- A Codex-specific parser inside the miners: rejected, it would duplicate the iteration/cursor logic the adapter layer exists to share and would couple the miners to a second harness format.
- Skipping ingest and only doing delivery (#786): useful on its own, but without this adapter the "learns from your own Codex sessions" half of the loop never closes.

## Additional Context

- Reference implementation and contract: `packages/core/src/repowise/core/sessions/adapters/claude_code.py`, `adapters/base.py`, and `README.md`.
- Delivering mined decisions back into Codex sessions is the sibling issue #786 and is intentionally out of scope here.
- The Claude-side behavior this brings to Codex is documented in the hooks and decisions guides (see repowise-dev/repowise#784).

**Done when:** a Codex transcript round-trips through the adapter into `Event`s in tests; the decisions and demand miners produce non-empty output on a fixture carrying the relevant tool calls; the Claude Code path is untouched.

rehan6025:
## Summary

- Added Codex transcript support to the shared session adapter.
- The adapter now parses Codex session metadata and tool payloads into the normal event format used by the session pipeline.
- Added regression tests covering discovery and normalization for Codex transcripts.

## Related Issues

- Closes #785 

## Test Plan

- [x] Tests pass (`pytest`)

## Checklist


RaghavChamadiya:
Thanks for this @rehan6025, the adapter is wired in cleanly and the seam is right. One blocking issue and a few smaller things.

Blocking: discover() returns nothing in real usage, and the discover test fails.

The last commit switched to a non-recursive glob:

return sorted(path for path in root.glob("*.jsonl") if path.is_file())

But Codex writes rollouts into a date-nested tree (~/.codex/sessions/YYYY/MM/DD/*.jsonl), which the module docstring and #785 both call out. A non-recursive *.jsonl only matches files sitting directly under sessions/, so it finds nothing in the real layout and no transcripts ever reach the miners.

I ran the branch to confirm. test_discover_lists_jsonl_sorted creates nested files and currently fails:

tests/unit/sessions/test_codex_adapter.py::test_discover_lists_jsonl_sorted FAILED
E   AssertionError: assert [] == ['a.jsonl', 'b.jsonl']
1 failed, 2 passed

So the "pytest passes" box isn't accurate on the current head. Switching to root.rglob("*.jsonl") fixes both the real-world discovery and the test. Worth double checking whatever motivated the non-recursive change, since it goes against the documented Codex layout.

Fixture isn't real-shaped. tests/unit/sessions/data/codex_session.jsonl uses an invented flat shape (top-level type: "assistant", top-level text and tool_calls[]). Real Codex rollouts, which _fill_response_item and _event_kind are actually built for, use session_meta / response_item / event_msg with a nested payload. The inline test does exercise the real shapes, so the code path is covered, but #785 specifically asks for a real-shaped fixture and to confirm the format first. Suggest replacing the fixture with an actual multi-line rollout.

No miner-level test. #785's "done when" includes the decisions and demand miners producing non-empty output on a fixture with the relevant tool calls. Right now the tests only check normalize() field mapping. Since tool-name normalization is the load-bearing part, one end-to-end test through a miner would guard against regressions the field checks miss.

Smaller stuff:
- parse_timestamp is duplicated verbatim from claude_code.py. Import it instead of copying.
- codex.py is missing a trailing newline.
- discover() ignores repo_root and returns every session on the machine. That works because the miners re-filter by cwd, but a one-line comment noting that would help, since the Claude adapter narrows by repo in discover itself.

The tool-name aliases line up with what the demand miner keys on, so that part looks good. The recursive-glob fix is the one that has to land; the rest is follow-up.

rehan6025:
Thanks for the detailed review! I'll address the recursive discovery issue and work through the remaining feedback, then push an update.

RaghavChamadiya:
Thanks @rehan6025, this is a solid update. Discovery is fixed and using iter_glob instead of plain rglob is better than what I suggested, it's the repo's own pruned walker so you get the junk-dir skipping for free. parse_timestamp is imported now, the fixture is real-shaped, and all 90 tests in tests/unit/sessions/ pass.

Two things to sort before this lands.

The miner tests are green but hollow. The fixture is two sessions concatenated (a session_meta on line 1 and another on line 5), and every tool call in the second one is dropped on the floor: lines 8, 10 and 14 have no call_id or id, and _fill_response_item needs one before it appends a ToolUse. Run lines 5-15 on their own and you get zero decisions and zero demand. Everything both miner tests assert is coming from the single search call on line 3.

That also means the decision test is pinning a misattribution:

decision kind=explicit_choice files=['pkg/app.py']
  quote: "We chose Flask because it is lightweight, keeps SQLite..."

The Flask quote lives in session 2 and is about outputs/sqlite-signup-demo/app.py. It only gets pkg/app.py because trailing_files is a running window still holding session 1's search hit. So assert decision.files == ["pkg/app.py"] is locking in cross-session bleed. Give the exec calls real call_ids and that assertion should change to the file the decision is actually about. A one-session fixture would sidestep the bleed entirely and is closer to what a real rollout looks like anyway.

session_id is picking up message ids. normalize() falls back to payload.get("id"). On session_meta that's correct, the payload id really is the session. On any other response_item it's the message or tool id:

event session_ids: ['0199f5aa-real-session-uuid', 'tool_09', 'msg_042']

MINED decision: kind=explicit_choice files=['pkg/app.py']
  session_id='msg_042'   <-- should be '0199f5aa-real-session-uuid'

This one has teeth: decisions.py stamps session_id=event.session_id onto every candidate (330, 355, 376), persists it to staging, and the injection-feedback loop keys correction_quotes(session_id) on it (661-669). Every Codex decision ends up under its own fake session, so that loop can never match. The Claude adapter reads one field with no id fallback, which is why it never hit this. Codex only carries the session id on session_meta, so it needs threading through the iteration rather than guessing per line.

While you're in there: ToolResult hardcodes is_error=False, and the dead-end miner keys on _tool_failure(result.payload, result.is_error), so that miner currently can't fire for Codex at all. Claude does is_error=bool(block.get("is_error")). Happy to take that as a follow-up if you'd rather keep this PR tight, just let me know which.

Minor, all mechanical: ruff check reports 3 errors on the branch (unused datetime import, and I001 on both files), ruff format wants both files, and codex.py is still missing its trailing newline. ruff check --fix && ruff format clears the lot.

rehan6025:
Codex represents many operations through the generic exec/bash tool, with the actual operation embedded in the command string. The demand miner currently recognizes only search_codebase/get_answer. Should the Codex adapter classify shell commands (e.g. rg) into those logical tool names, or should the miners be extended to understand Codex's generic bash tool?

rehan6025:
Hi sir @RaghavChamadiya , I just wanted to ask, I believe the adapter should normalize Codex's generic exec tool into canonical tool names (e.g. search_codebase) because the miners only consume canonical names. My remaining question is what heuristic you'd prefer for distinguishing search commands from other shell commands.

RaghavChamadiya:
Thanks @rehan6025, and sorry for the long silence on this one. The Codex transcript adapter shipped in #1277 on 8 August, which closed #785, so this thread can close as superseded.

The merged version took the review feedback from here with it. It discovers date-nested sessions through `iter_glob`, the repo's pruned walker, instead of `rglob`, and it threads `session_id` off `session_meta` with fallbacks, which is the misattribution problem raised on this PR. You can read it at `packages/core/src/repowise/core/sessions/adapters/codex.py`.

One item is still open in the merged code rather than fixed: `ToolResult.is_error` is hardcoded `False`. That was called out at merge time as a deliberate follow-up. If you want a small, well-shaped next thing, that is a good one and the shape of it is clear from the adapter as it stands.

Closing this, but the review round here is a real part of why the merged version looks the way it does.

PR #821, opened 2026-07-13
feat(security): add repowise security scan --history for full git-history secret scanning

### LINKED ISSUE #818: feat(security): add `repowise security scan --history` for full git-history secret/risk scanning
## Summary

Add a `repowise security scan --history` command that runs the **existing** `SecurityScanner` (`packages/core/src/repowise/core/analysis/security_scan.py`) over the *entire git history* of a repo, not just the working tree, and persists hits into the existing `security_findings` table.

Today `SecurityScanner` only runs during indexing (`packages/core/src/repowise/core/pipeline/persist.py`) against the current checked-out files. Secrets, tokens, and dangerous patterns committed in the past — even if later removed — are never surfaced. A full-history mode closes that gap from inside the product (distinct from the CI secret-scanning PRs, which operate on diffs/CI gates).

## Motivation

- Leaked credentials often live in old commits and are "fixed" by deleting them in a later commit. They remain in history and remain a risk.
- The pattern registry, severity model, and DB persistence already exist; only the **scan target** (working tree vs. full history) is missing.
- Surfacing these in the same `security_findings` table means the existing API (`GET /api/repos/{id}/security`) and UI (findings table, severity×directory matrix) light up automatically with no new storage.

## Proposed design

1. **New CLI subcommand** `security scan --history [--since <rev>] [--to <rev>]` in `packages/cli/src/repowise/cli/commands/` that:
   - Iterates commits in the range (default `git rev-list --all` or configurable).
   - For each commit, checks out the tree's blob content (or uses `git show <rev>:<path>`) for tracked source files.
   - Applies the existing `_PATTERNS` / `_SYMBOL_KEYWORDS` checks via `SecurityScanner.scan_file`.
   - Persists findings with the commit SHA + author date attached (extend the `security_findings` schema with `commit_sha` / `commit_at` columns, nullable for working-tree scans).
2. **Reuse, don't fork**: keep `SecurityScanner`'s regex registry and `persist()` logic; add an optional `commit_sha`/`commit_at` field rather than a parallel implementation.
3. **Dedup + severity rollup**: same `(repo_id, file_path, kind, line, commit_sha)` should not be double-inserted on re-runs; provide a clear summary (counts by severity, by kind) on stdout and JSON `--output`.
4. **UI/API**: show a "found in history" badge/filter so users can tell working-tree findings from historical ones, and optionally emit remediation guidance (rotate, purge via filter-repo/BFG).

## Out of scope (to avoid duplicating active work)

- CI gating / pre-commit / pre-push secret scanning (covered by open PRs #7, #9, #99, #274, #650, #678, #418, etc.).
- This is a **local, developer-facing, full-history exploration** tool that complements those gates.

## Acceptance criteria

- [ ] `repowise security scan --history` scans all historical revisions of tracked source files.
- [ ] Findings persist to `security_findings` with commit provenance and appear in the existing security API/UI.
- [ ] Re-runs are idempotent (no duplicate rows).
- [ ] `--since`/`--to` range limiting and `--output json` are supported.
- [ ] Unit tests cover history iteration + dedup, mirroring existing `security_scan` test patterns.

## References

- Existing scanner: `packages/core/src/repowise/core/analysis/security_scan.py`
- Wiring: `packages/core/src/repowise/core/pipeline/persist.py:633`
- API: `packages/server/src/repowise/server/routers/security.py`
- Related (different scope — CI gates): PRs #7, #9, #99, #274, #650, #678, #418

Ayush7614:
## Summary
Implements #818. Reuses the existing `SecurityScanner` pattern registry and the shared `security_findings` tab
le rather than forking a new subsystem.

## Changes
- **security_scan**: extend `persist()` to carry `commit_sha`/`commit_at` and make re-runs idempotent via `ON 
CONFLICT DO NOTHING` / `INSERT OR IGNORE`.
- **history_scan**: new `HistorySecurityScanner` that scans unique git blobs (`git rev-list --objects --all`, 
deduped by blob SHA) once each and attributes hits to the first-introducing commit. Defaults to the secret-ori
ented patterns (`hardcoded_password`/`hardcoded_secret`) to avoid code-smell noise.
- **models + migration 0037**: add nullable `commit_sha`/`commit_at` and a `uq_security_finding_provenance` un
ique constraint for dedup.
- **CLI**: `repowise security scan --history [--since <rev>] [--to <rev>] [--all-patterns] [--output json]`.
- **server**: `GET /security` gains a history filter; responses include `commit_sha` + `found_in_history` so t
he UI can tell history from working tree.
- Addresses maintainer review: idempotent persist, mandatory schema migration, symbol scan disabled in history (
no parsed symbols on raw blobs), secret-only gate, and blob-dedup scan strategy.

## Related Issues
Closes #818
## Test Plan
- [x] Tests pass (`pytest tests/unit/analysis/test_security_scan_history.py`)
- [x] Migration chain consistent (`0037` → `0036`)
## Checklist


RaghavChamadiya:
Thanks @Ayush7614, and thanks for actually working through the review points on #818 rather than just the headline ask. The shape is right: reusing the pattern registry, the secret-only gate with `--all-patterns` as the opt-out, `SECRET_KINDS`, extending `persist()` rather than forking it. That's the design I was hoping for.

The `NULL` versus `""` trap is handled properly too, which is the thing I expected to go wrong. `sha_key = commit_sha or ""` means working-tree rows key on `""` instead of `NULL`, so the unique constraint actually dedups them. (The docstrings and the model comment all say the columns are left NULL, which contradicts the code. The code is right, the prose needs fixing.)

That said, this can't merge yet. Three blockers, and the reason none of them showed up as a failing test is worth taking seriously.

**The migration can't run on SQLite.** `op.create_unique_constraint` outside `batch_alter_table`:

```
NotImplementedError: No support for ALTER of constraints in SQLite dialect.
Please refer to the batch mode feature
```

SQLite backs every local `.repowise` database, so as written this hard-crashes on upgrade for every existing user. Needs `op.batch_alter_table`, which on SQLite means a table rebuild.

**The Postgres INSERT is invalid SQL.** The conflict clause is concatenated ahead of the column list, so it builds:

```sql
INSERT INTO security_findings ON CONFLICT ON CONSTRAINT uq_... DO NOTHING (repository_id, ...) VALUES (...)
```

`ON CONFLICT` has to follow `VALUES`. A Postgres parser rejects it at col 99. Every insert then raises and `except Exception: continue` eats it, so on Postgres the scan reports success and writes nothing at all. Related: swapping `break` for `continue` was the right call for a single bad row, but it makes a systematic failure like this one completely silent. Worth logging the exception rather than dropping it.

**Migration `0037` now collides.** `0037_repo_git_totals` and `0038_agent_trace_line_provenance` landed on main after you branched, and both your file and `0037_repo_git_totals` declare `revision = "0037"` / `down_revision = "0036"`. Needs renumbering to `0039` with `down_revision = "0038"`. Not your doing, just needs a rebase.

**The blob-dedup speedup is given back in the attribution loop.** This is the one real design issue. The unique-blob scan is implemented correctly, and then:

```python
for commit, _iso in commits:
    for blob_sha in self._blobs_in_commit(repo_path, commit):
        blob_introduced_at.setdefault(blob_sha, commit)
```

That's one `git ls-tree -r` subprocess per commit, each listing every file, which is exactly the O(commits x files) pass the blob dedup was meant to replace. On this repo (1858 commits, 11408 blobs) I measured `ls-tree` at about 61ms a call, so the attribution loop is roughly 113 seconds of subprocess spawning before a single blob gets scanned. On a 50k-commit repo it's most of an hour. `_read_blob` then spawns another `git cat-file -p` per blob on top.

Both are avoidable in one pass each: `git log --all --raw --format=...` gives you introducing-commit-per-blob directly, and `git cat-file --batch` streams contents over a single process.

**On the tests.** The fixture builds the schema with `Base.metadata.create_all` and only ever runs SQLite. So the migration never executes in tests, and the Postgres branch has no coverage at all. That's why three blockers sit behind a green suite. Two things would close it: run the alembic migration against a SQLite DB in a test rather than `create_all`, and cover the Postgres `persist()` path (at minimum assert the generated SQL parses, ideally exercise it). Everything else in the test file is testing the ORM's idea of the schema, not the one migrations actually produce.

Smaller, while you're in there: `files_scanned` increments before the `_is_source` check so it always equals `blobs_scanned`, and `progress` only fires when a finding is kept, so a clean repo prints nothing for the whole scan.

Happy to look again once the migration and the Postgres path are sorted. The core of this is good work and I'd like to land it.

Ayush7614:
Thanks for the detailed review @RaghavChamadiya  — I pushed the follow-up fixes and merged `upstream/main` to clear the conflict.

Addressed points:
- SQLite migration now uses `batch_alter_table` and is covered by a real Alembic upgrade test against SQLite.
- The Postgres `persist()` SQL now emits `INSERT ... VALUES ... ON CONFLICT ... DO NOTHING` in the correct order, and row-level failures are logged instead of being silently swallowed.
- The history attribution loop no longer does `ls-tree` per commit; it now uses a single `git log --reverse --raw` pass for blob->introducing-commit attribution plus `git cat-file --batch` for blob reads.
- Tests now cover the migration path, the Postgres `persist()` path, and the clean-repo progress/files-scanned cases.

Because `upstream/main` has since landed `0037` through `0040`, I renumbered the migration to `0041` with `down_revision = "0040"` during the merge.

I also kept the newer working-tree `replace_findings()` path from `main` and scoped its delete to working-tree rows only (`COALESCE(commit_sha, '') = ''`) so re-indexing does not wipe history findings.

Validation run locally after the merge resolution:
- `tests/unit/analysis/test_security_scan_history.py`
- `tests/unit/analysis/test_security_scan.py`
- `ruff check packages/core/src/repowise/core/analysis/security_scan.py packages/core/alembic/versions/0041_security_finding_history.py`

Latest push: `bf86654`

RaghavChamadiya (approved):
LGTM

PR #822, opened 2026-07-13
fix(ui): make empty doc state friendlier

### LINKED ISSUE #811: Friendlier empty state when a file has no generated doc
## Problem

When you open the **Doc** tab for a file that the indexing pass didn't generate a wiki page for, the empty state reads:

> **No documentation yet**
> This file hasn't been documented. Run a sync to generate its wiki page.

This feels like something is broken or missing, when in reality the pipeline just didn't pick this file for a doc page. The copy should feel intentional and warm rather than flat — e.g. something like **"This page didn't make the cut."**

## Where

The empty state lives in `packages/ui/src/files/file-doc-tab.tsx` (around lines 22–31), rendered when `wikiPage` is null:

```tsx
<EmptyState
  icon={<BookOpen className="h-8 w-8" />}
  title="No documentation yet"
  description="This file hasn't been documented. Run a sync to generate its wiki page."
/>
```

The same "no doc" empty state also appears in:
- `packages/ui/src/files/file-overview-tab.tsx`
- `packages/vscode/webview/src/views/docs/App.tsx`

## Proposed change

- Update the title/description copy to feel intentional and friendly (suggested direction: title like *"This page didn't make the cut"*, with a short, kind explanation that not every file gets its own page).
- Keep the wording consistent across all three spots above.
- Copy/UI only — no behavior change.

## Good first issue

This is a small, self-contained copy/UI change touching a couple of components — a good starter task for a new contributor. Please keep the plain, simple tone used elsewhere in the app.

Real5K:
## Summary

- Replace the file Doc tab's misleading sync prompt with warmer, clearer copy
- Clarify that files without documentation are an expected state
- Leave repository-wide empty states unchanged because they represent a different condition

## Related Issues

Closes #811

## Test Plan

- [x] UI type-check passes (`npm run type-check --workspace=@repowise-dev/ui`)
- [x] UI tests pass — 595/595 (`npm run test --workspace=@repowise-dev/ui`)

## Checklist

- [x] Tests were not added because this is a copy-only change
- [x] All existing tests pass
- [x] Documentation updates are not needed

### DIFF
--- a/packages/ui/src/files/file-doc-tab.tsx
+++ b/packages/ui/src/files/file-doc-tab.tsx
@@ -29,8 +29,8 @@ export function FileDocTab({ wikiPage, docSlot, wikiHref }: FileDocTabProps) {
     return (
       <EmptyState
         icon={<BookOpen className="h-8 w-8" />}
-        title="No documentation yet"
-        description="This file hasn't been documented. Run a sync to generate its wiki page."
+        title="This page didn't make the cut"
+        description="Not every file gets its own documentation page, and that's perfectly normal."
       />
     );
   }


Real5K:
Hey @RaghavChamadiya 

This has been waiting for review for about 2 days. Could you please review it when you have a chance?

It's a small 2-line copy-only change for the empty doc state. (Closes #811) 

Happy to tweak the wording if needed.

RaghavChamadiya (approved):
Sorry for the wait, and thanks for the ping.

Approving. This does what #811 asked, and you were right to deviate from the part of it that was wrong. The issue says to keep three spots consistent, but only `file-doc-tab.tsx` is actually the per-file no-doc state. `file-overview-tab.tsx` has no empty state at all (the string only appears in a docstring at line 20 describing old behavior), and `vscode/webview/src/views/docs/App.tsx:131` is gated on `pages.length === 0`, a repository-wide state where "run a generation" genuinely is the right advice. There's a fourth in that same family the issue missed, `packages/web/src/components/docs/docs-explorer.tsx:150`, also repo-wide. So your call that repository-wide empty states are a different condition is correct, and since `packages/ui` is shared by both hosts, the one change covers the real case everywhere. Good instinct not to follow the issue literally there.

Verified the checklist: type-check clean, 595/595 UI tests pass.

Two non-blocking notes, neither worth another round:

- `FileDocTab` only receives `wikiPage: FileWikiPageRef | null`, so it can't distinguish a deliberately skipped file from one that's new since the last index. The old copy guessed one way, this guesses the other. Skip is the dominant case so it's a net improvement; just noting "perfectly normal" is a touch more confident than the component can actually know.
- `file-overview-tab.tsx:20` still quotes "No documentation yet" in its docstring as the old default. Harmless historical reference.

Thanks for the clean, well-scoped change and for explaining the reasoning in the description rather than just following the issue.

PR #829, opened 2026-07-14
fix(ollama): wrap timeout/connection errors so they retry

### LINKED ISSUE #445: [Bug] many `page_generation_failed` errors when running wiki init
Hi again!

I'm running the command `repowise init --resume --coverage 0.5` to generate more wiki pages (originally created a wiki with 20% coverage) since it takes many days to generate them on my local machine with local LLM, I do it gradually every night.

The problem is while it's running, it outputs MANY errors `page_generation_failed`, at least 50 so far, on several file types, many are very basic React components files, some backend, all sorts really.  

## Environment

- OS: `Windows 11`
- Python version: `3.12.1`
- Repowise version: `0.18.0`
- Installation method: `uv`

## Additional Context

Any other context, screenshots, or log output.

dev-ploy:
## Summary

- Fix `page_generation_failed` floods with local LLMs (Ollama): `APITimeoutError`
  and `APIConnectionError` are siblings of `APIStatusError` under
  `openai.APIError`, so the old catch-only-`APIStatusError` clause let them escape
  uncaught, skipping the tenacity retry loop entirely and surfacing as
  `page_generation_failed` in `orchestrate.py` (`run_level` → `guarded_named`).
- Move the Ollama provider onto the shared retry policy
  (`provider_should_retry` / `provider_retry_stop` / `provider_retry_wait`) with a
  3-way exception mapping (`RateLimitError` → 429, `APIStatusError` → HTTP status,
  `APIError` → transient) so transient timeout/connection errors retry while
  non-retryable 4xx (400/401/403/404/405/413/422) fail fast. Same wrapping applied
  to both `stream_chat` sites.
- Apply the same defensive `APIError` catch to the other OpenAI-client providers
  (`openai`, `openrouter`, `deepseek`), and warn when a local provider
  (`ollama` / `codex_cli` / `opencode`) runs with `--concurrency > 4`.

## Related Issues

Closes #445

## Test Plan

- [x] Tests pass (`pytest`) — 897 provider + CLI tests green
- [x] Lint passes (`ruff check`) on all changed files
- [x] Web build passes (`npm run build`) — N/A, no frontend changes

Additional manual verification against a live Ollama server, across multiple
locally running models:

- Full `repowise init --provider ollama --concurrency 6 --coverage 0.5` runs
  against several local models → all pages generated with **0
  `page_generation_failed`**, and the local-provider concurrency warning shown.
- Forced a real `APITimeoutError` (1 ms client timeout against the live server):
  now wrapped as `ProviderError` and retried 3× instead of escaping raw.
- New regression tests confirm 401 (auth) and 400 (invalid request) fail on the
  first attempt — no wasted retries.

## Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -163,3 +163,5 @@ docs/HOSTED_PRODUCT_UPGRADE_PLAN.md
 docs/HOSTED_IMPLEMENTATION_GUIDE.md
 docs/HOSTED_PROGRESS.md
 scripts/md_to_docx.py
+
+# Local issue analysis scratch (not for OSS distribution)
--- a/packages/cli/src/repowise/cli/commands/init_cmd/command.py
+++ b/packages/cli/src/repowise/cli/commands/init_cmd/command.py
@@ -158,6 +158,14 @@ def _run_generation_phase(
         lang_parts = [f"{lang} {pct:.0%}" for lang, pct in lang_items]
         console.print(f"  Languages: {', '.join(lang_parts)}")
 
+    # Warn when a local provider runs with default concurrency
+    if provider.provider_name in ("ollama", "codex_cli", "opencode") and concurrency > 4:
+        console.print(
+            f"  [yellow]Warning:[/yellow] {provider.provider_name} is a local provider "
+            f"running with concurrency={concurrency}. "
+            f"If you see timeout errors, try [bold]--concurrency 1[/bold]."
+        )
+
     console.print(
         f"  Coverage: {int(chosen_pct * 100)}% / "
         f"~{est.estimated_input_tokens + est.estimated_output_tokens:,} tokens "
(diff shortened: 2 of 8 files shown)


dev-ploy:
Hi @RaghavChamadiya i would love your view on this solution

RaghavChamadiya (approved):
Really solid fix, thanks, and it held up well despite sitting for a bit. I confirmed the bug is still live on `main`: Ollama still catches only `APIStatusError`, so `APITimeoutError` and `APIConnectionError` keep escaping tenacity and surfacing as `page_generation_failed`. The root-cause framing is spot on, and moving Ollama onto the shared `provider_should_retry` / `provider_retry_stop` / `provider_retry_wait` policy is exactly right, it lines up with what the `openai` provider already does.

I pulled the branch: it still merges cleanly, the full provider suite is green (166 passing), and the new regression tests pin the important parts, both that timeout/connection errors now retry and that non-retryable 4xx still fail fast. Appreciate the manual verification against a live Ollama server too.

Merging.

PR #831, opened 2026-07-14
fix(server): launch persisted polling sync job

### LINKED ISSUE #830: [Bug] Polling fallback leaves sync job pending after undefined `job` error
## Describe the Bug

  The scheduler's polling fallback creates and commits a pending generation job, but then crashes while trying to launch it because the return
  value from `upsert_generation_job()` is not assigned.

  In `packages/server/src/repowise/server/scheduler.py:137-166`:

  ```python
  await crud.upsert_generation_job(
      session,
      repository_id=repo.id,
      status="pending",
      config={
          "mode": "sync",
          "trigger": "polling_fallback",
          "before": stored_commit or "",
          "after": current_head,
      },
  )
  await session.commit()

  task = asyncio.create_task(
      execute_job(job.id, app_state),
      name=f"poll-job-{job.id}",
  )

  job is undefined when execute_job(job.id, ...) is evaluated. This raises:

  NameError: name 'job' is not defined

  The exception is caught by the broad except Exception at the end of polling_fallback(), so the scheduler logs polling_fallback_failed, but the
  job row has already been committed with status pending.

  This is also an F821 undefined-name violation under the repository's Ruff configuration.

 ## Steps to Reproduce

  1. Initialize and index a local repository.
  2. Make another commit so the current Git HEAD differs from .repowise/state.json's last_sync_commit.
  3. Start the server with repowise serve.
  4. Wait for the polling fallback interval, which defaults to 15 minutes.
  5. Inspect the server logs and generation jobs.

  The same behavior can be reproduced faster in a test by obtaining the polling_fallback APScheduler job from setup_scheduler() and directly
  awaiting its callback.

  ## Expected Behavior

  When the stored commit differs from the repository's current HEAD, the polling fallback should:

  1. Create a pending sync job.
  2. Launch execute_job() using that job's ID.
  3. Allow the job to transition from pending to running, then to a terminal state.
  4. Update the repository's sync baseline after successful completion.

  ## Actual Behavior

  The scheduler:

  1. Creates and commits the pending job.
  2. Logs polling_sync_enqueued.
  3. Raises NameError because job is undefined.
  4. Catches and logs the exception as polling_fallback_failed.
  5. Leaves the job permanently in pending.

  The pending row then causes the active-job guard to skip future polling attempts. Manual sync requests for the same repository can also
  return:

  409 A sync job is already in progress for this repository

  The job remains stuck until the server restarts and startup cleanup marks interrupted pending/running jobs as failed.

  ## Environment

  - Repowise version: current main / 0.31.0
  - Python version: any supported version
  - Platform: platform-independent

  ## Proposed Fix

  At minimum, retain the created job:

  job = await crud.upsert_generation_job(...)
  await session.commit()

  task = asyncio.create_task(
      execute_job(job.id, app_state),
      name=f"poll-job-{job.id}",
  )

  Ideally, scheduler jobs should also use the same managed background-job launch path as REST-triggered jobs so that:

  - the task is registered in app.state.job_tasks;
  - cancellation works consistently;
  - task-launch failures mark the database job as failed;
  - unhandled task failures cannot leave the row pending or running.

  ## Suggested Regression Tests

  Add scheduler tests covering:

  - A diverged repository creates exactly one sync job.
  - execute_job() receives the persisted job ID.
  - The callback does not raise NameError.
  - A task-launch failure moves the job to failed instead of leaving it pending.
  - A second polling pass does not create a duplicate while a real job is active.

  ## Additional Context

  The scheduler currently has no direct behavioral tests. Existing job-executor tests verify pipeline configuration, but they do not execute the
  APScheduler polling callback.

GautamSharma99:
fix #830 

## Summary

  - Retain the `GenerationJob` returned by the polling fallback upsert.
  - Launch the background executor using the persisted job ID.
  - Add a regression test covering repository divergence, persisted job metadata,
    and executor invocation.

  ## Problem

  When the polling fallback detected that a repository's HEAD differed from its
  last synchronized commit, it created and committed a pending generation job.

  The returned job object was discarded, but the scheduler subsequently attempted
  to call `execute_job(job.id, app_state)`. This raised a `NameError`, which was
  caught by the scheduler's outer exception handler.

  As a result:

  1. The pending job remained in the database.
  2. No background sync was launched.
  3. Subsequent polling runs skipped the repository because an active pending job
     already existed.

  ## Fix

  Store the result of `upsert_generation_job()` in `job` and use its persisted ID
  when launching the background task.

  ## Testing

  - Added `test_polling_fallback_launches_persisted_job`.
  - Verified the persisted job status and polling configuration.
  - Verified that the executor receives the persisted job ID.
  - Changed files pass Ruff formatting/linting and mypy.
  - All JavaScript lint, type-check, unit, and VS Code smoke tests pass.

  Verification note: the full Python run reached 7,484 passed, 14 skipped. Two unrelated code-rationale tests fail independently on the existing
  branch. Thirteen initially sandbox-sensitive failures passed when rerun outside the sandbox. Repository-wide Ruff/format/mypy also report 
substantial pre-existing issues, while both changed files pass their focused checks.

### DIFF
--- a/packages/server/src/repowise/server/scheduler.py
+++ b/packages/server/src/repowise/server/scheduler.py
@@ -134,7 +134,7 @@ def setup_scheduler(
                         continue
 
                     # Enqueue a sync job
-                    await crud.upsert_generation_job(
+                    job = await crud.upsert_generation_job(
                         session,
                         repository_id=repo.id,
                         status="pending",
--- a/tests/unit/server/test_scheduler.py
+++ b/tests/unit/server/test_scheduler.py
@@ -1,0 +1,64 @@
+"""Tests for APScheduler background jobs."""
+
+from __future__ import annotations
+
+import asyncio
+import json
+from types import SimpleNamespace
+from unittest.mock import AsyncMock, patch
+
+from sqlalchemy import select
+
+from repowise.core.persistence import crud
+from repowise.core.persistence.database import get_session
+from repowise.core.persistence.models import GenerationJob
+from repowise.server.scheduler import setup_scheduler
+
+
+async def test_polling_fallback_launches_persisted_job(session_factory, tmp_path) -> None:
+    """A diverged repository should launch the sync job that was persisted."""
+    repo_path = tmp_path / "repo"
+    state_path = repo_path / ".repowise" / "state.json"
+    state_path.parent.mkdir(parents=True)
+    state_path.write_text(json.dumps({"last_sync_commit": "old-sha"}), encoding="utf-8")
+
+    async with get_session(session_factory) as session:
+        repo = await crud.upsert_repository(
+            session,
+            name="test-repo",
+            local_path=str(repo_path),
+        )
+        repo_id = repo.id
+
+    app_state = SimpleNamespace(background_tasks=set())
+    scheduler = setup_scheduler(session_factory, app_state=app_state)
+    polling_job = next(job for job in scheduler.get_jobs() if job.id == "polling_fallback")
+    head_result = SimpleNamespace(returncode=0, stdout="new-sha\n")
+    execute_job = AsyncMock()
+
+    with (
+        patch("subprocess.run", return_value=head_result),
+        patch("repowise.server.job_executor.execute_job", execute_job),
+    ):
+        await polling_job.func()
+        if app_state.background_tasks:
+            await asyncio.gather(*app_state.background_tasks)
+
+    execute_job.assert_awaited_once()
+    job_id, launched_state = execute_job.await_args.args
+    assert launched_state is app_state
+
+    async with get_session(session_factory) as session:
+        result = await session.execute(
+            select(GenerationJob).where(GenerationJob.repository_id == repo_id)
+        )
+        persisted_job = result.scalar_one()
+
+    assert job_id == persisted_job.id
+    assert persisted_job.status == "pending"
+    assert json.loads(persisted_job.config_json) == {
+        "mode": "sync",
+        "trigger": "polling_fallback",
+        "before": "old-sha",
+        "after": "new-sha",
+    }


RaghavChamadiya:
The one-line fix is correct. upsert_generation_job returns the persisted GenerationJob, so job.id on the launch line is now defined, which clears both the NameError and the F821. The added test is a real regression test: on the current code the NameError is swallowed by the outer except before execute_job is ever awaited, so assert_awaited_once() would fail without the fix.

Two things to sort out before merge:

1. This overlaps PR #833. #833 (offloading blocking git) also includes this exact job = await crud.upsert_generation_job(...) change on scheduler.py, and both branch off main. Whichever lands first, the other will conflict on that line. Let's decide which PR carries the fix and drop it from the other, or stack #833 on top of this one, so we don't merge the same change twice.

2. Scope vs #830. This resolves the specific NameError path, but the issue's "Ideally" section is left open: the scheduler still hand-rolls its own task launch (scheduler.py:160-170) instead of reusing _launch_job_task in repos.py. As a result a poll-launched sync is not registered in app.state.job_tasks (the cancel endpoint can't reach it), and if execute_job fails in a way it can't self-record, the job stays pending and re-wedges the repo behind the active-job guard, which is the same stuck-pending class of bug this issue is about. The two suggested regression tests for that (task-launch failure moves the job to failed; a second polling pass doesn't duplicate an active job) aren't covered here either.

Fine to merge this as the targeted crash fix, but let's not close #830 on it. Suggest routing the scheduler through the shared managed-launch path in a follow-up so it gets the same cancellation and failure-marking as the REST path

packages/server/src/repowise/server/scheduler.py
```diff
@@ -134,7 +134,7 @@ async def polling_fallback() -> None:
                         continue
 
                     # Enqueue a sync job
-                    await crud.upsert_generation_job(
+                    job = await crud.upsert_generation_job(
```
swati510, line 137: A behaviour question that this fix brings to life rather than causes. The executor only rewrites `.repowise/state.json` when a run succeeds (`job_executor.py` lines 482 to 499, which sit after the completion block and are skipped whenever the run raises), and a failed job no longer counts as active for the guard just above this line, because that guard only looks for jobs in the `pending` or `running` state. So a repository whose sync fails repeatedly will get a fresh full pipeline run enqueued every fifteen minutes indefinitely, and each of those runs costs money in language model calls. Before this fix the bug accidentally suppressed that, because the very first attempt left a pending row behind and the guard then silenced the repository for good.

Could we add a cheap guard, for example skipping the repository when its most recent job failed while targeting the same head commit? That would keep the retry behaviour for transient failures without an unbounded loop for permanent ones.

swati510:
Thank you for tracking this down, the one-line fix is clearly right and the regression test genuinely fails without it.

One thing I would like to raise before we merge, because this change is what makes this code path run for the first time. I am referring to the launch block at `packages/server/src/repowise/server/scheduler.py` lines 160 to 170, which I cannot comment on inline because those lines are not part of the diff. That block has no recovery if the background task is never created, whereas `_launch_job_task` in `routers/repos.py` wraps the call to `asyncio.create_task` in a try block and, if it fails, marks the row as failed on a fallback path. Its docstring gives the reason directly: it does this "so the active-job guard never gets stuck". The same reasoning applies here. If `asyncio.create_task` raises in the scheduler, the outer error handler simply logs `polling_fallback_failed`, the row stays pending, and the guard further up skips this repository on every future poll until the server restarts, which is the same symptom as the bug you are fixing.

Would you consider extracting the lifecycle logic from `_launch_job_task` into a shared helper and calling it from here? That would also let the scheduler register the task in `app_state.job_tasks`, which the cancel endpoint in `routers/jobs.py` reads in order to interrupt a running job.

swati510:
A small point about `packages/server/src/repowise/server/scheduler.py` line 168, which I am raising here rather than inline because that line is not part of the diff. `getattr(app_state, "background_tasks", set())` creates a throwaway set when the attribute is missing, so in that case nothing holds a strong reference to the task once the loop moves on, and Python may garbage collect it mid-run, silently abandoning the indexing job. `app.py` always sets this attribute during startup, so it cannot happen on the real server today, but `webhooks.py` and `repos.py` both read it directly and would fail loudly instead. Would you mind reading it directly here too, for consistency and so that a genuinely misconfigured state is visible rather than hidden?

RaghavChamadiya (approved):
Nice catch, this one was hiding well. The outer except was swallowing the NameError, so polling sync just quietly did nothing. Thanks for the fix and the test.

Issue #832, opened 2026-07-14
Avoid blocking Git subprocesses in async server jobs


GautamSharma99:
## Issue description

  The server executes synchronous Git operations from native async functions. Because APScheduler runs coroutine jobs directly on the
  application’s asyncio event loop, these calls can temporarily block all other work handled by that worker.

  ### Exact code path

  At startup, the FastAPI lifespan creates and starts an AsyncIOScheduler:

  - packages/server/src/repowise/server/app.py:209
  - packages/server/src/repowise/server/app.py:210

  Every 15 minutes, APScheduler invokes the async polling_fallback() function. It loops through every registered repository and calls
  subprocess.run() synchronously:

  - packages/server/src/repowise/server/scheduler.py:72
  - packages/server/src/repowise/server/scheduler.py:94
  - packages/server/src/repowise/server/scheduler.py:110

  Each invocation has a 10-second timeout. Since repositories are processed sequentially, the worst-case blocking time grows linearly:

  number of repositories × 10 seconds

  The same underlying problem exists in the background job executor:

  - execute_job() awaits _incremental_page_regen() at packages/server/src/repowise/server/job_executor.py:407.
  - _incremental_page_regen() synchronously runs git rev-parse at packages/server/src/repowise/server/job_executor.py:742.
  - It then runs synchronous GitPython change detection and file parsing on the event-loop thread.

  The final state-stamping path also invokes synchronous git rev-parse through _read_head_sha():

  - packages/server/src/repowise/server/job_executor.py:487
  - packages/server/src/repowise/server/job_executor.py:497
  - packages/server/src/repowise/server/job_executor.py:598

  ### Why this matters

  While one of these operations is blocked, the same server worker cannot promptly:

  - Handle API requests.
  - Deliver SSE job-progress events.
  - Process cancellation requests.
  - Run other scheduled jobs.
  - Advance other background generation tasks.

  Git normally responds quickly, but slow filesystems, repository locks, hooks, network-mounted worktrees, or unhealthy repositories can make
  the pause noticeable. Polling multiple repositories compounds the delay.

  ### Expected behavior

  Git inspection and synchronous change detection should not block the asyncio event loop. Other requests and background tasks should continue
  running while a repository is inspected.

  ### Proposed implementation direction

  Use one of the repository’s async-safe execution approaches:

  - Run isolated blocking operations with await asyncio.to_thread(...).
[...]

PR #833, opened 2026-07-14
fix(server): offload blocking git operations

GautamSharma99:
fix -> #832 

## Summary

  - Offload polling fallback Git HEAD checks from the asyncio event loop.
  - Offload incremental regeneration’s Git and change-detection work.
  - Offload post-job state stamping and Git HEAD lookup.
  - Add regression tests proving these operations run on worker threads.
  - Include the polling job reference fix required for the scheduler path.

  ## Problem

  The server ran synchronous Git commands and change detection from async
  FastAPI/APScheduler tasks. A slow repository could block the event loop,
  delaying API responses, SSE progress updates, cancellation requests, and other
  background jobs.

  Polling processed repositories sequentially, with each Git HEAD lookup allowed
  to block for up to 10 seconds.

  ## Fix

  Use `asyncio.to_thread()` for blocking Git, state-persistence, graph, and
  change-detection operations while preserving the existing timeout and error
  handling behavior.

  ## Testing

  - `uv run pytest tests/unit/server/test_scheduler_async.py tests/unit/server/test_job_executor.py -q`
    - 13 passed
  - `uv run pytest tests/unit/server -q`
    - 964 passed
    - 2 unrelated pre-existing failures in `test_code_rationale.py`
  - Ruff lint and formatting checks passed.
  - Mypy passed for the changed server modules.**

### DIFF
--- a/packages/server/src/repowise/server/job_executor.py
+++ b/packages/server/src/repowise/server/job_executor.py
@@ -484,7 +484,8 @@ async def execute_job(
         # mode, config) that `repowise init` would have written.
         try:
             if is_initial_index:
-                _persist_initial_index_state(
+                await asyncio.to_thread(
+                    _persist_initial_index_state,
                     Path(repo_path),
                     llm_client=llm_client,
                     docs_enabled=generate_docs,
@@ -494,7 +495,7 @@ async def execute_job(
                     exclude_patterns=exclude_patterns,
                 )
             else:
-                _stamp_last_sync_commit(Path(repo_path))
+                await asyncio.to_thread(_stamp_last_sync_commit, Path(repo_path))
         except Exception:
             logger.debug("state_json_update_failed", job_id=job_id, exc_info=True)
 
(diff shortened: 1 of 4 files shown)


RaghavChamadiya:
Thanks @GautamSharma99, and sorry this waited so long for a real reply.

The problem is still there on main, and it is worse than one call site. `scheduler.py:110` runs `subprocess.run(["git", "rev-parse", "HEAD"], ...)` synchronously inside the async `polling_fallback()`. In `job_executor.py`, `_persist_initial_index_state` and `_stamp_last_sync_commit` are called directly at `:487` and `:497`, and `_stamp_last_sync_commit` shells out through `_read_head_sha`. `_incremental_page_regen`'s own `git rev-parse HEAD` around `:742`, plus the `detector.get_changed_files` and `graph_builder.graph()` calls after it, are all synchronous too. Each of those parks the event loop for the duration.

Your approach is also the one the codebase already uses. `asyncio.to_thread` is the standing pattern for this in the MCP tools and several routers, so this is reuse and not a new mechanism, which is what I want to see in a change like this.

What is holding it up is process, not substance. The only check that has run on this branch is the bot's code-health comment, so the Python matrix and ruff have never reported on it, and the branch is now conflicting. Rebase with `git fetch origin && git rebase origin/main` and push. Once the full matrix is green I will review it properly against the call sites above.

If you rebase, it is worth checking the `_incremental_page_regen` ones are covered, since that path has moved since you opened this.

Issue #834, opened 2026-07-14
REST semantic search does not load persisted LanceDB indexes


GautamSharma99:
The REST server does not reuse the persistent vector index created by the CLI.

  repowise init stores embeddings in the repository-local .repowise/lancedb directory. However, repowise serve always initializes a new, empty
  InMemoryVectorStore, so semantic searches made through the REST API cannot see the existing embeddings.

  This also affects workspace indexing jobs because the job executor always passes the global primary vector store to every repository pipeline.

  ### Exact failure path

  #### 1. CLI persists embeddings to LanceDB

  The CLI’s vector-store factory prefers a repository-local LanceDB store:

  - packages/cli/src/repowise/cli/providers/vector_store.py:9
  - LanceDB path: <repo>/.repowise/lancedb
  - It only falls back to InMemoryVectorStore when LanceDB is unavailable.

  During repowise init, that store is passed to the generation pipeline and populated with page and decision embeddings.

  #### 2. REST startup discards that persisted index

  When the REST server starts, it builds the configured embedder but unconditionally creates a new in-memory store:

  - packages/server/src/repowise/server/app.py:193
  - packages/server/src/repowise/server/app.py:195

  embedder = _build_embedder()
  vector_store = InMemoryVectorStore(embedder=embedder)

  There is no code here that checks for .repowise/lancedb or opens a configured pgvector backend.

  The comment claiming that LanceDB or pgvector can be configured through the environment does not match the implementation; no corresponding
  server configuration path exists.

  #### 3. Single-repository search queries the empty store

  In single-repository mode, workspace_config is None, so semantic search immediately queries the primary in-memory store:

  - packages/server/src/repowise/server/routers/search.py:149
  - packages/server/src/repowise/server/routers/search.py:153
  - packages/server/src/repowise/server/routers/search.py:156

  Because this store was created during server startup and contains no vectors, InMemoryVectorStore.search() returns an empty list.

  The existing workspace-only LanceDB resolver is never called in this path.

  #### 4. Workspace jobs use the wrong vector store

  The job executor resolves the correct repository database through session_factory_override, but still takes the vector store from the global
  application state:

  - packages/server/src/repowise/server/job_executor.py:317
  - packages/server/src/repowise/server/job_executor.py:319
  - packages/server/src/repowise/server/job_executor.py:395

  vector_store = app_state.vector_store

  That same store is used for generation, decision embeddings, and deletion of swept page vectors.

  Consequences in workspace mode include:

[...]

PR #835, opened 2026-07-14
fix(server): load persisted vector indexes

GautamSharma99:
## Summary

  - Load the primary repository's persisted LanceDB index during REST startup.
  - Resolve vector stores by repository ID for semantic search.
  - Route background job writes and deletions to the repository's own store.
  - Cache repository stores and close them safely during shutdown.
  - Retain an in-memory fallback when LanceDB is unavailable.

  ## Problem

  `repowise init` persisted embeddings under `.repowise/lancedb`, but
  `repowise serve` always created an empty `InMemoryVectorStore`.

  As a result, REST semantic search returned no results after startup despite a
  valid persisted index.

  Workspace jobs also used the global primary vector store, causing embeddings
  and deletions to be routed to the wrong repository and lost after restart.

  ## Fix

  Startup now identifies the repository associated with the active database and
  opens its repo-local LanceDB store.

  A shared repository-aware resolver now:

  - Locates the canonical repository path through the routed session factory.
  - Opens or creates `<repo>/.repowise/lancedb`.
  - Caches stores by repository ID.
  - Keeps workspace repositories isolated.
  - Provides an in-memory fallback when persistent storage is unavailable.

  The REST search route and job executor both use this resolver.

  ## Testing

  - Added a regression test that creates a LanceDB index before REST startup,
    reopens it, and verifies the semantic-search endpoint returns its data.
  - Added a two-repository isolation test.
  - Added job-executor coverage confirming the repository store is passed to the
    pipeline instead of the global primary store.
  - Focused tests: 17 passed.
  - Server suite: 964 passed; 2 unrelated existing failures.
  - Full Python suite: 7,499 passed, 14 skipped; the same 2 unrelated failures.
  - Ruff formatting/lint and mypy passed.

### DIFF
--- a/packages/server/src/repowise/server/app.py
+++ b/packages/server/src/repowise/server/app.py
@@ -27,7 +27,6 @@ from repowise.core.persistence.database import (
 )
 from repowise.core.persistence.models import GenerationJob
 from repowise.core.persistence.search import FullTextSearch
-from repowise.core.persistence.vector_store import InMemoryVectorStore
 from repowise.core.providers.embedding.base import MockEmbedder
 from repowise.server import __version__
 from repowise.server.routers import (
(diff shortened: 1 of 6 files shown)


RaghavChamadiya:
Thanks for taking this @GautamSharma99 

Reviewed this against #834 and traced the changed paths end to end. Nicely scoped fix that hits all four failure points in the issue.

Looks correct:
- Startup now opens the repo-local LanceDB via build_primary_vector_store and seeds it into workspace_vector_stores, so jobs and searches share one connection.
- The job executor resolving the store per-repo fixes both the write path and the swept-page delete_many at line 434, which was the part of the issue that's easy to miss.
- Shutdown double-close is handled properly: the primary is closed explicitly in app.py, and close_workspace_vector_stores skips store is primary.
- The _semantic reordering is safe since unknown/mismatched repo_ids fall back to primary_vs.search.
- Tests are meaningful, especially the real init→serve reopen regression and the job-executor assertion that the repo store (not the global primary) reaches the pipeline.

Two non-blocking notes:
1. Primary-repo identification uses a substring match local_db in normalized_url, which is drive-letter case-sensitive on Windows. In the single-repo case the len(rows) == 1 fallback covers it, but on a shared registry DB with multiple repos a case mismatch would leave the primary unidentified and fall back to an empty in-memory store. Comparing resolved paths instead of substring would harden this, though it's not required for the single-repo case this targets.
2. _vector_store_locks is a module-level dict that grows per repo_id and is never cleared. Bounded by repo count, so negligible, just flagging it.

Otherwise LGTM, merging this

RaghavChamadiya (approved):

PR #836, opened 2026-07-14
fix(dataflow): recurse into with-statement bodies

### LINKED ISSUE #765: [Bug] dataflow CFG does not recurse into with-statement bodies

Real5K:
## Summary

- recurse into Python `with` statement bodies when building the CFG
- preserve def/use handling for context expressions and `as` aliases
- connect exception-handler paths to `finally` blocks
- add regression coverage for CFG structure, reaching definitions, liveness, and unreachable code

## Related Issues

Closes #765

## Test Plan

- `uv run pytest tests/unit/health/test_dataflow_cfg.py tests/unit/health/test_dataflow_reaching.py tests/unit/health/test_dataflow_facts.py tests/unit/health/test_dataflow_langs.py -q`
- `uv run ruff check` on all changed files
- `npm run lint`
- `npm run type-check`
- `git diff --check`

## Checklist

- [x] Tests added for the bug fix
- [x] Relevant tests pass
- [x] Lint and type checks pass
- [x] Changes are focused on the reported issue

### DIFF
--- a/packages/core/src/repowise/core/analysis/health/complexity/languages.py
+++ b/packages/core/src/repowise/core/analysis/health/complexity/languages.py
@@ -186,6 +186,7 @@ class LanguageNodeMap:
     raise_kinds: frozenset[str] = frozenset()
     break_kinds: frozenset[str] = frozenset()
     continue_kinds: frozenset[str] = frozenset()
+    with_kinds: frozenset[str] = frozenset()
     #   * ``statement_wrapper_kinds`` -- statement node(s) that merely wrap the
     #     node the CFG builder should classify, as their last named child.
     #     Expression-oriented grammars need this: tree-sitter-rust parses every
@@ -230,6 +231,7 @@ _PY = LanguageNodeMap(
     raise_kinds=frozenset({"raise_statement"}),
     break_kinds=frozenset({"break_statement"}),
     continue_kinds=frozenset({"continue_statement"}),
+    with_kinds=frozenset({"with_statement"}),
 )
 
 _TS = LanguageNodeMap(
--- a/packages/core/src/repowise/core/analysis/health/dataflow/cfg.py
+++ b/packages/core/src/repowise/core/analysis/health/dataflow/cfg.py
@@ -308,6 +308,8 @@ class _CFGBuilder:
                 cur = self._build_loop(st, cur)
             elif t in self.lmap.try_kinds:
                 cur = self._build_try(st, cur)
+            elif t in self.lmap.with_kinds:
+                cur = self._build_with(st, cur)
             elif t in self.lmap.return_kinds or t in self.lmap.raise_kinds:
                 self._record_stmt(cur, st)
                 self._edge(cur, self.exit_block)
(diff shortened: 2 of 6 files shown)


RaghavChamadiya:
Verified locally before merge. This fixes #765 properly.

The key thing this gets right, which a competing attempt (#791) missed, is the with_kinds branch in dialects/python.py::_head. Without it, with_statement falls through to the else: self._process(node, ...) fallback and the whole body gets walked into the head block, so the body's defs and uses are recorded twice and the original coarse-block bug survives alongside the new structure.

Confirmed on the issue's repro:

block 2 kind='normal' defs=[] uses=[('ctx', 3)]     # head: with_clause only
block 5 kind='normal' defs=[('result', 5)]          # arms keep their own defs
block 6 kind='normal' defs=[('result', 7)]

And the dead-code case now attributes dead only to the unreachable block. 960 tests pass. Also nice that _build_with returns None when the body cannot fall through rather than leaving an orphan join with no predecessors.

One follow-up, not a blocker: the _edge(handler_entry, normal_join) edge depends on handler-body statements landing in handler_entry itself, which holds for flat handlers but goes subtly wrong to reason about if a handler opens with a compound statement. It over-approximates in the safe direction either way. Worth a comment on that line spelling out the assumption so the next reader doesn't have to re-derive it.

Merging. Thanks for covering both gaps from the issue in one pass.

RaghavChamadiya (approved):

Real5K:
@RaghavChamadiya 

Thanks for the detailed review, especially catching the dialect `_head` / `with_kinds` point.

I’ll keep the handler → join assumption follow-up in mind for a small comment PR later.

PR #839, opened 2026-07-14
Fix #838: Use Next.js Middleware for dynamic proxying of API routes

### LINKED ISSUE #838: [Bug] Custom API port (--port) ignored by graph/architecture proxy routes — Knowledge Graph view blank (incomplete fix of #427)
## Describe the Bug

`repowise serve --port <custom>` is **partially** broken: most API routes now respect the custom port, but the **graph / architecture route group is still hardcoded to the default port `7337`** in the Next.js UI proxy. As a result the **Architecture / Knowledge Graph views render blank** whenever the API runs on any port other than `7337`.

This looks like an incomplete fix / regression of #427. In that older report *every* proxied route failed. On `0.31.0` the common routes (`/api/workspace`, `/api/repos`, `/api/providers`, `/api/repos/{id}/overview-summary`, `/api/repos/{id}/stats/highlights`) all correctly proxy to the custom API port and return `200`. Only the `/api/graph/*` routes still point at `7337`.

## Environment

- repowise `0.31.0`
- macOS (Darwin 25.5.0), Node 22 / Next.js 15.5.20 (bundled web UI)
- Single indexed repo, launched from the repo directory

## Steps to Reproduce

```sh
repowise serve --port 1679 --ui-port 1680
```

Startup log confirms the API is on the custom port:

```
Web UI starting on http://localhost:1680
API server starting on http://127.0.0.1:1679
INFO:     Uvicorn running on http://127.0.0.1:1679 (Press CTRL+C to quit)
```

Then open the UI and navigate to the **Architecture** / **Knowledge Graph** view (also fetched on the repo overview page).

## Actual Behavior

The Architecture / Knowledge Graph panel is blank. The UI proxy returns HTTP `500` for the graph routes, and the server log shows the proxy targeting the **default** port `7337` instead of the configured `1679`:

```
Failed to proxy http://localhost:7337/api/graph/<repo-id>/architecture?min_members=2 [AggregateError: ] { code: 'ECONNREFUSED' }
Failed to proxy http://localhost:7337/api/graph/<repo-id>/communities            [AggregateError: ] { code: 'ECONNREFUSED' }
Failed to proxy http://localhost:7337/api/graph/<repo-id>/architecture-view?include_symbols=false [AggregateError: ] { code: 'ECONNREFUSED' }
Failed to proxy http://localhost:7337/api/graph/<repo-id>                          [AggregateError: ] { code: 'ECONNREFUSED' }
```

Note the error surfaces as a bare `AggregateError` with an empty message (Node/undici wrapping `ECONNREFUSED`), which makes it hard to diagnose from the UI alone — the graph area just shows nothing.

### Proof the backend is fine — only the proxy target port is wrong

Hitting the API **directly on the custom port** returns full data:

```sh
$ curl -s -o /dev/null -w '%{http_code} %{size_download}\n' \
    'http://127.0.0.1:1679/api/graph/<repo-id>/architecture-view?include_symbols=false'
200 240415

$ curl -s -o /dev/null -w '%{http_code} %{size_download}\n' \
    'http://127.0.0.1:1679/api/graph/<repo-id>'
200 390441
```

Hitting the **same routes through the UI proxy** fails:

```sh
$ curl -s -o /dev/null -w '%{http_code}\n' \
    'http://localhost:1680/api/graph/<repo-id>/architecture-view?include_symbols=false'
500
```

While non-graph routes through the proxy succeed on the custom port:

```sh
$ curl -s -o /dev/null -w '%{http_code}\n' 'http://localhost:1680/api/repos'
200
```

## Expected Behavior

The `/api/graph/*` proxy routes should respect the `--port` value (and whatever mechanism the other routes already use), so the Architecture / Knowledge Graph views load on any custom API port.

## Likely Cause

The UI proxy layer resolves most routes from the runtime API port, but the graph route handlers appear to use a hardcoded `http://localhost:7337` base. Affected routes observed: `/api/graph/{id}`, `/api/graph/{id}/architecture`, `/api/graph/{id}/architecture-view`, `/api/graph/{id}/communities`.

## Workaround

Keep the API on the default port `7337` and only move the web UI, e.g. `repowise serve --ui-port 1680`. The graph views then load because the hardcoded target matches the actual API port.

Related: #427

kishansaaai:
### Summary
This PR fixes the issue where Next.js proxy routes (specifically `/api/graph/*`) ignored the `--port` argument provided to the CLI and defaulted to `7337`.

### Root Cause
When the UI is built in Next.js `standalone` mode, the `next.config.ts` environment variables are evaluated entirely at **build time**. Because of this, the `rewrites()` array was compiled into `.next/routes-manifest.json` with the static fallback `http://localhost:7337`, permanently baking in the default port.

While client components (which route through the UI proxy) hitting `/api/graph/*` resulted in `ECONNREFUSED` or `500` errors due to this hardcoded fallback, server components completely bypassed the proxy and fetched using the runtime `process.env.REPOWISE_API_URL` natively, creating the illusion that only some routes were broken.

### Changes
* Removed the static `rewrites()` logic from `packages/web/next.config.ts`.
* Introduced a runtime edge middleware at `packages/web/src/middleware.ts` utilizing `NextResponse.rewrite()`. This properly reads `process.env.REPOWISE_API_URL` dynamically at runtime and routes `/api/*`, `/health`, and `/metrics` requests to the correct backend port instantiated by the CLI command.

### Related Issues
Fixes #838 

### Test Plan
- [x] Verified that Next.js build no longer generates static rewrites in `.next/routes-manifest.json`.
- [x] Tested `repowise serve --port 1679` locally.
- [x] Confirmed the Architecture and Knowledge Graph views correctly load on custom ports without proxy `ECONNREFUSED` errors.

### DIFF
--- a/packages/web/next.config.ts
+++ b/packages/web/next.config.ts
@@ -24,23 +24,6 @@ const nextConfig: NextConfig = {
   images: {
     unoptimized: true,
   },
-  async rewrites() {
-    const apiUrl = process.env.REPOWISE_API_URL || "http://localhost:7337";
-    return [
-      {
-        source: "/api/:path*",
-        destination: `${apiUrl}/api/:path*`,
-      },
-      {
-        source: "/health",
-        destination: `${apiUrl}/health`,
-      },
-      {
-        source: "/metrics",
-        destination: `${apiUrl}/metrics`,
-      },
-    ];
-  },
 };
 
 export default nextConfig;
--- a/packages/web/src/middleware.ts
+++ b/packages/web/src/middleware.ts
@@ -1,0 +1,22 @@
+import { NextResponse } from 'next/server';
+import type { NextRequest } from 'next/server';
+
+export function middleware(request: NextRequest) {
+  const apiUrl = process.env.REPOWISE_API_URL || "http://localhost:7337";
+  
+  // Construct the destination URL using the runtime API base URL
+  const destination = new URL(
+    request.nextUrl.pathname + request.nextUrl.search,
+    apiUrl
+  );
+
+  return NextResponse.rewrite(destination);
+}
+
+export const config = {
+  matcher: [
+    '/api/:path*',
+    '/health',
+    '/metrics'
+  ],
+};


RaghavChamadiya:
Thanks for taking this @kishansaaai 

Reviewed this by applying the branch locally, building `packages/web`, and driving the standalone server against a backend on a custom port. The fix is correct and I was able to confirm it end to end.

**On the root cause:** the diagnosis here is right, and it's actually broader than the issue title suggests. The key is `packages/web/src/lib/api/client.ts:19-24`: in the browser `BASE_URL` is `""`, so requests go same-origin and hit the rewrite that was baked to `http://localhost:7337` in `routes-manifest.json` at build time. On the server it's `process.env.REPOWISE_API_URL`, read at runtime, which skips the proxy entirely. So the real split is client-fetched routes vs server-rendered ones, not graph routes vs everything else. The graph/architecture views are just the most visible client-fetched surface. Every client-side API call was broken on a custom port, not only `/api/graph/*`.

Worth noting for #838: the reported `curl http://localhost:1680/api/repos -> 200` shouldn't be possible under this explanation, since that path also goes through the baked rewrite. Most likely there was still a server up on `7337` at the time. Doesn't change anything about this fix.

**What I verified:**

* `routes-manifest.json` now has empty `beforeFiles`/`afterFiles`/`fallback`, so the static `7337` bake is gone.
* The edge bundle keeps `process.env.REPOWISE_API_URL||"http://localhost:7337"` uninlined. This was my main concern, since Next inlines env vars into edge bundles, but `define-env.js` only inlines `NEXT_PUBLIC_*` and `next.config` `env` keys, and the edge sandbox rebuilds `process.env` from the live process per request. The runtime read is real.
* With `REPOWISE_API_URL=http://localhost:1679` and `PORT=1680`, both `/api/repos` and `/health` returned 200 from the custom-port backend. That's the exact scenario from the issue.
* Method, query string, and POST bodies survive `NextResponse.rewrite` (checked with `POST /api/graph/abc/architecture?min_members=2` and a JSON body).
* SSE still streams incrementally rather than buffering, so the event-stream endpoints don't regress.

This also fixes the same latent bug in Docker, where `docker/entrypoint.sh` sets `REPOWISE_API_URL` at container start and the baked rewrite ignored it

Merging it

RaghavChamadiya (approved):

PR #841, opened 2026-07-14
fix(cli): resolve DB identity on delegated worktree updates (#746)

### LINKED ISSUE #746: [Bug] worktree seed + delegated update writes new-file pages to the wrong DB / repo identity (not a selection-budget issue)

paarthsikka:
- Strip REPOWISE_DB_URL from os.environ during delegated run_update so it honors the copied DB
- Add atomic try/except rollbacks to seed staging and rename phases
- Guard sweep_stale_seed_backups with an mtime staleness check
- Pass full=False instead of full=force to prevent re-embedding everything on update
- Skip updating entirely when dry_run=True to prevent bogus output
- Persist --coverage setting explicitly provided in init to config.yaml

## Summary

- Resolves a bug (#746) where delegated worktree seeds incorrectly wrote to the base DB by actively stripping inherited environment variables.
- Hardens the worktree `.repowise` replacement with multi-pass atomic rollbacks to prevent corrupted partial directory states upon failure.
- Ensures the `--coverage` flag persists successfully during `init` across full-state rewrites, locked down with a new integration test.

## Related Issues

Fixes #746
Builds on #747

## Test Plan

- [x] Tests pass (`pytest`)

## Checklist

### DIFF
--- a/packages/cli/src/repowise/cli/commands/init_cmd/command.py
+++ b/packages/cli/src/repowise/cli/commands/init_cmd/command.py
@@ -603,8 +603,15 @@ def init_command(
             repo_paths=[r.path for r in scan.repos],
             seed_base=seed_base,
             include_submodules=include_submodules,
+            dry_run=dry_run,
         )
         if seeded:
+            if dry_run:
+                console.print(
+                    "[dim]Dry run: Would seed worktree index and delegate to update.[/dim]"
+                )
+                return
+
             console.print(
                 "[green]Worktree index seeded successfully. Delegating to update...[/green]"
             )
@@ -612,27 +619,36 @@ def init_command(
 
             is_workspace = len(scan.repos) > 1 and not no_workspace
 
-            # Delegate to update
-            run_update(
-                path=str(repo_path),
-                provider_name=provider_name,
-                model=model,
-                since=None,
-                reasoning=reasoning,
-                cascade_budget=None,
-                dry_run=dry_run,
-                workspace=is_workspace,
-                no_workspace=no_workspace,
-                repo_alias=None,
-                index_only=index_only,
-                docs_flag=None,
-                full=force,
-                agents_md=agents_md,
-                concurrency=concurrency,
-                no_cost_tracking=no_cost_tracking,
-                verbose=verbose,
-                progress=progress,
-            )
+            # Strip db-pin env vars to enforce worktree DB resolution (note: potential process-global race condition).
+            old_db_urls = {}
+            for env_key in ("REPOWISE_DB_URL", "REPOWISE_DATABASE_URL"):
+                if env_key in os.environ:
+                    old_db_urls[env_key] = os.environ.pop(env_key)
+
+            try:
+                # Delegate to update
+                run_update(
+                    path=str(repo_path),
+                    provider_name=provider_name,
+                    model=model,
+                    since=None,
+                    reasoning=reasoning,
+                    cascade_budget=None,
+                    dry_run=dry_run,
+                    workspace=is_workspace,
+                    no_workspace=no_workspace,
+                    repo_alias=None,
+                    index_only=index_only,
+                    docs_flag=None,
+                    full=False,
+                    agents_md=agents_md,
+                    concurrency=concurrency,
+                    no_cost_tracking=no_cost_tracking,
+                    verbose=verbose,
+                    progress=progress,
+                )
+            finally:
+                os.environ.update(old_db_urls)
             return
     if len(scan.repos) > 1 and not no_workspace:
         _workspace_init(
@@ -1083,6 +1099,8 @@ def init_command(
     # ---- Post-run: config, state, MCP, editor project files ----
     if commit_limit is not None:
         save_config_partial(repo_path, commit_limit=resolved_commit_limit)
+    if coverage_pct is not None:
+        save_config_partial(repo_path, coverage=coverage_pct)
 
     write_editor_project_files(
         console,
--- a/packages/cli/src/repowise/cli/commands/init_cmd/command.py
+++ b/packages/cli/src/repowise/cli/commands/init_cmd/command.py
@@ -619,9 +619,11 @@ def init_command(
 
             is_workspace = len(scan.repos) > 1 and not no_workspace
 
+            from repowise.core.persistence.database import DB_ENV_VARS
+
             # Strip db-pin env vars to enforce worktree DB resolution (note: potential process-global race condition).
             old_db_urls = {}
-            for env_key in ("REPOWISE_DB_URL", "REPOWISE_DATABASE_URL"):
+            for env_key in DB_ENV_VARS:
                 if env_key in os.environ:
                     old_db_urls[env_key] = os.environ.pop(env_key)
 
@@ -640,7 +642,7 @@ def init_command(
                     repo_alias=None,
                     index_only=index_only,
                     docs_flag=None,
-                    full=False,
+                    full=force,
                     agents_md=agents_md,
                     concurrency=concurrency,
                     no_cost_tracking=no_cost_tracking,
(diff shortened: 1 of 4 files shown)


RaghavChamadiya:
Nice fix on the DB identity leak, but two blockers before merge:

1. `save_config_partial(repo_path, coverage=coverage_pct)` writes a float under the `coverage` key, which is already used for the coverage-report ingestion block (`CoverageConfig.from_repo_config` expects a dict there and silently falls back to defaults if it isn't). It also isn't read back anywhere on update, so it clobbers real config without doing anything. Needs a different key (e.g. `coverage_pct`) plus a reader in `run_update`, or drop the persistence.
2. The delegation now passes `full=False` instead of `full=force`. `full` was only ever True when the user explicitly passed `--force`, so this just silently ignores the flag. Please restore `full=force`.

Smaller things: the new `except Exception` in the seed staging loop swallows errors with no message (the sibling branches all print a fallback notice); the 1h mtime gate doesn't really work for `.repowise.bak.*` since rename preserves mtime (keep unconditional sweep for those, gate only the seed temp dirs); the env-strip loop should reuse `DB_ENV_VARS` from core persistence instead of hardcoding the tuple; and there are a few trailing-whitespace lines that `ruff format` will churn.

paarthsikka:
@RaghavChamadiya please do review the changes whenever you get time.

RaghavChamadiya:
Sorry for the long silence here, @paarthsikka. I went back through 8a59b19 and both blockers I named are addressed: `full=force` is restored, and the env strip now imports `DB_ENV_VARS` from `core.persistence.database` instead of hardcoding the pair. The `try/finally` around `run_update` that puts the vars back is a better shape than what I asked for, since the delegated call is in-process and a bare pop would have leaked into whatever ran after it. The three-pass backup/rename/rollback in `seed_index_from_base` also reads correctly to me: on failure it removes the targets it renamed in before restoring the backups, so a half-applied seed cannot survive.

Two things stand between this and a merge, and one of them is my fault for the delay.

The branch is `CONFLICTING` now. Six weeks of main have moved under `init_cmd/command.py`, which is one of the busiest files in the repo.

More to the point, the coverage half of this PR no longer has anything to attach to. #1036 landed on 24 July and deleted the `--coverage` option from `init` entirely, so `coverage_pct` is not a value the command can receive any more. What survives is only `--coverage-report`, which is test-coverage ingestion and a completely different thing. And the field it would round-trip into is dead on the other end too. `GenerationConfig.coverage_pct` in `core/generation/models.py` still exists, but its own docstring says why:

> `coverage_pct` and `max_pages_pct` no longer reach selection. They are still carried because the CLI flag, the config round-trip and the server's ranked-generate request body all still write them [...] Nothing reads them to decide what gets a page.

So `save_config_partial(repo_path, coverage_pct=coverage_pct)` and the `kwargs` read-back in `run_update` would persist a number that nothing consumes, and `test_coverage_persisted_after_init` cannot pass because `init --coverage 0.5` is not a valid invocation on current main. That is not something you got wrong, it moved after you pushed.

What I would like on the rebase: keep the DB identity fix, the env strip, the atomic rollbacks, the sweep gate and the dry-run guard, and drop the coverage persistence along with its test. That leaves the PR doing exactly what #746 asks for and takes the `update_cmd/command.py` change out of the diff, which should shrink the conflict surface too.

`git fetch origin && git rebase origin/main` and ping me when it is green. I will merge it without another round of feedback.

paarthsikka:
Hey Raghav, no worries at all about the delay!

I've just rebased the branch onto main and resolved the conflicts. As requested, I completely dropped the coverage_pct persistence logic and its associated test, which also takes update_cmd/command.py completely out of the diff.

The branch is now clean and contains exactly what you asked for: the DB identity fix with the DB_ENV_VARS strip, the atomic rollbacks, and the staleness sweep gates.

Let me know if you need anything else—otherwise, ready to merge whenever you are!

RaghavChamadiya (approved):
That is exactly the rebase I asked for, @paarthsikka. `update_cmd/command.py` is out of the diff entirely, the coverage persistence and `test_coverage_persisted_after_init` are gone, and `full=force` survived the move. Full matrix is green on efefdc0b across 3.11, 3.12 and 3.13.

Thanks for the patience on this one. Six weeks of silence from me on a PR that was already correct is not a good look, and you rebased it the same day I finally came back to it.

Merging.

PR #844, opened 2026-07-15
fix(web): contain overflowing modal inputs in Add Repository dialog (#843)

### LINKED ISSUE #843: [Bug] Add Repository modal inputs overflow container bounds
### Description
Clicking "Add Repository" opens a modal where the input fields (Name, Local Path, Remote URL, etc.) overflow their container, extending past the modal's white background on the right side.

This looks like a layout containment issue — the modal's grid/flex structure isn't constraining the inputs' widths, letting them protrude past the modal's right padding and outer bounds.

### Steps to Reproduce
1. Open the Repowise web app
2. Click "Add Repository" to trigger the wizard
3. Observe the right side of the modal — the input fields stretch past the modal's edge and close button

### Expected Behavior
All input fields should stay fully contained within the modal's boundary and align consistently with the rest of the form.

### Environment
- OS: Windows
- Browser: Brave, Chrome, Edge
- Screen resolution: 100%

Image

jyotirmya17:
Fixes the layout overflow reported in #843, where input fields inside the "Add Repository" modal extended past the modal's container bounds.

## Root Cause
The modal's inputs were missing two properties needed for correct sizing inside a flex/grid layout:
- No `box-sizing: border-box` equivalent, so horizontal padding was added on top of the declared 100% width instead of being included in it.
- No `min-width: 0`, so inputs couldn't shrink below their browser-default intrinsic width once the flex/grid container ran out of space.

## Changes
- **Added `box-border` to `Input` and `SelectTrigger`** — ensures padding is included in the 100% width calculation instead of expanding past it.
- **Added `min-w-0` to the inputs** — allows them to shrink correctly inside flex/grid layouts instead of overflowing at their intrinsic minimum width.
- **Added `overflow-hidden` to the modal container as a failsafe** — the two fixes above resolve the root cause, but this guards against similar overflow if new fields are added to the modal in the future without the same width handling.

## Screenshots
**Before:**
image

**After:**
image

Fixes #843

### DIFF
--- a/packages/ui/src/onboarding/add-repo-wizard.tsx
+++ b/packages/ui/src/onboarding/add-repo-wizard.tsx
@@ -263,7 +263,7 @@ export function AddRepoWizard({ adapter, open, onOpenChange }: AddRepoWizardProp
 
   return (
     <Dialog open={open} onOpenChange={handleOpenChange}>
-      <DialogContent className="sm:max-w-md w-[calc(100vw-2rem)]">
+      <DialogContent className="sm:max-w-md w-[calc(100vw-2rem)] flex flex-col">
         {step === "details" && (
           <>
             <DialogHeader>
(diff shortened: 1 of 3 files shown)


RaghavChamadiya (approved):
LGTM

PR #848, opened 2026-07-15
fix(server): use sync mode for webhook jobs

### LINKED ISSUE #837: Webhook jobs silently skip incremental documentation regeneration
## Issue description

  GitHub and GitLab push webhooks enqueue jobs using an unsupported execution mode.

  Both handlers create jobs with:

  {
    "mode": "incremental",
    "trigger": "webhook"
  }

  However, the server job executor does not recognize incremental. It only has explicit behavior for sync, full_resync, initial_index, and
  index_only.

  The job still completes successfully, but documentation regeneration is skipped and the repository’s synchronization baseline is advanced.

  ### Exact failure path

  #### 1. Webhook handlers enqueue incremental

  GitHub creates the job at:

  - packages/server/src/repowise/server/routers/webhooks.py:119
  - Mode assignment: packages/server/src/repowise/server/routers/webhooks.py:124

  GitLab has the same behavior:

  - packages/server/src/repowise/server/routers/webhooks.py:177
  - Mode assignment: packages/server/src/repowise/server/routers/webhooks.py:182

  Both handler docstrings describe these as “sync” jobs, making the configuration inconsistent with the documented intent.

  #### 2. The executor does not recognize the mode

  The executor reads the job mode here:

  - packages/server/src/repowise/server/job_executor.py:343

  It only derives flags for:

  is_full_resync = mode == "full_resync"
  is_initial_index = mode == "initial_index"
  is_index_only = mode == "index_only"

  incremental matches none of these conditions.

  #### 3. Full documentation generation is disabled

  The main pipeline generates documentation only for full resync and initial-index jobs:

  generate_docs = (
      (is_full_resync or is_initial_index)
      and llm_client is not None
      and bool(config.get("generate_docs", True))
  )

  Therefore, a webhook job calls run_pipeline() with generate_docs=False.

  The repository is re-ingested and its analysis data may be updated, but no documentation pages are generated.

  #### 4. Incremental regeneration is also skipped

  The executor has a separate incremental page-regeneration phase, but it runs only for sync:

  - packages/server/src/repowise/server/job_executor.py:401
  - packages/server/src/repowise/server/job_executor.py:406

  if mode == "sync" and llm_client is not None:
      incremental_pages = await _incremental_page_regen(...)

  Because the webhook mode is incremental, this condition is false.

  The webhook job therefore skips both possible documentation-generation paths.

  #### 5. The job still reports success and advances the baseline

  After re-indexing, the job is marked completed, even though no documentation was regenerated.

  The executor then calls _stamp_last_sync_commit() for every non-initial-index mode:

  - packages/server/src/repowise/server/job_executor.py:482
  - packages/server/src/repowise/server/job_executor.py:497

  This updates state.json with the current HEAD.

  The polling fallback subsequently sees HEAD == last_sync_commit and concludes that the repository is synchronized, so it does not enqueue a
  corrective sync job.

  ### Resulting behavior

  Push webhook
      → "incremental" job created
      → repository re-indexed
      → documentation regeneration skipped
      → job marked completed
      → last_sync_commit advanced
      → polling fallback sees repository as current
      → stale documentation remains indefinitely

  ### Expected behavior

  A default-branch push webhook should perform the same incremental documentation refresh as a polling-triggered sync job:

  - Re-index the repository.
  - Determine which pages are affected by the pushed commits.
  - Regenerate those pages when an LLM provider is configured.
  - Persist the updated pages and search indexes.
  - Advance last_sync_commit only after the intended sync path succeeds.

  ### Proposed implementation direction

  Use the executor’s canonical sync mode in both webhook handlers:

  {
    "mode": "sync",
    "trigger": "webhook",
    "before": "...",
    "after": "..."
  }

  For backward compatibility, the executor could also normalize existing incremental jobs to sync:

  if mode == "incremental":
      mode = "sync"

  It would also be useful to validate modes explicitly so future unknown values fail visibly instead of silently running with partial behavior.

  ### Test gap

  Current webhook tests verify only request acceptance and authentication:

  - tests/unit/server/test_webhooks.py:15
  - tests/unit/server/test_webhooks.py:87

  They do not:

  - Seed a matching repository.
  - Inspect the generated job configuration.
  - Verify the linked webhook event’s job ID.
  - Execute the generated job.
  - Assert that incremental page regeneration occurs.
  - Verify synchronization-state behavior.

Sanjays2402:
## Summary

- Enqueue GitHub and GitLab pushes as `sync` jobs so the executor runs incremental page regeneration instead of treating `incremental` as an unknown mode.
- Cover both webhook routes with regression tests that inspect the persisted job configuration.

## Related Issues

Closes #837

## Test Plan

- [x] Tests pass (`pytest`) — `pytest tests/unit/server/test_webhooks.py` (7 passed)
- [ ] Lint passes (`ruff check .`) — scoped checks pass; the touched source file has a pre-existing import-order finding
- [x] Web build passes (`npm run build`) *(if frontend changes)* — not applicable

## Checklist

- [ ] All existing tests still pass — focused server tests pass; the full suite was not run
- [x] I have updated documentation if needed — no documentation change is needed

### DIFF
--- a/packages/server/src/repowise/server/routers/webhooks.py
+++ b/packages/server/src/repowise/server/routers/webhooks.py
@@ -121,7 +121,7 @@ async def github_webhook(
                     repository_id=repo.id,
                     status="pending",
                     config={
-                        "mode": "incremental",
+                        "mode": "sync",
                         "trigger": "webhook",
                         "before": payload.get("before", ""),
                         "after": payload.get("after", ""),
@@ -179,7 +179,7 @@ async def gitlab_webhook(
                     repository_id=repo.id,
                     status="pending",
                     config={
-                        "mode": "incremental",
+                        "mode": "sync",
                         "trigger": "webhook",
                         "before": payload.get("before", ""),
                         "after": payload.get("after", ""),
--- a/tests/unit/server/test_webhooks.py
+++ b/tests/unit/server/test_webhooks.py
@@ -10,6 +10,9 @@ from unittest.mock import patch
 
 import pytest
 from httpx import AsyncClient
+from sqlalchemy import select
+
+from repowise.core.persistence.models import GenerationJob, Repository
 
 
 @pytest.mark.asyncio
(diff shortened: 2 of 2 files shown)


RaghavChamadiya:
Thanks @Sanjays2402, this is clean. Merging.

I confirmed the diagnosis holds end to end: job_executor.py:343 does mode = config.get("mode") or "sync", so "incremental" falls through every branch, and the incremental page regen at line 420 is gated on mode == "sync" specifically. Both docs paths were being skipped while last_sync_commit still advanced, exactly as #837 describes.

The test earns its place. Reverting the two-line change fails both parametrized cases, so it's pinning the behaviour rather than passing vacuously. Parametrizing across the GitHub and GitLab routes is the right call given both handlers carried the identical bug.

Also appreciated the accurate PR body. The I001 in webhooks.py is genuinely pre-existing (I checked it against main), and flagging it as such rather than ticking the box is the right way to handle it. Same for calling out that the full suite wasn't run.

Ignore the bot's dead-code finding on webhooks.py, that's a known false positive tracked in #666. The router is mounted at app.py:438, but from repowise.server.routers import (...) resolves to the package __init__.py only, so every router submodule reads as having no importers.

One thing I'll pick up separately rather than hold this on: the executor silently accepts any unknown mode and does partial work, which is what let this bug exist in the first place. #837 floats validating modes explicitly. I checked and only full_resync, initial_index, index_only, and sync ever reach execute_job, so a validation raise would be safe. That's the root-cause fix and it belongs in its own PR.

RaghavChamadiya (approved):

PR #860, opened 2026-07-16
fix: keep data-shape grep paths color-stable

### LINKED ISSUE #858: [Bug] data_shape live-grep: add --no-color so exclusion filter survives color.ui=always
### Summary

### Why it matters now

### Fix

### Pointers

pollychen-lab:
## Summary
- add `--no-color` to the parsed `git grep -l` call used by data-shape mining, with a comment explaining why the flag is load-bearing
- add a regression test for repositories with `color.ui=always`
- include the test-only `import subprocess`, move `build_exclude_spec` above the `repowise.server` import to satisfy isort, and rewrite one existing assertion into the conventional left-hand actual form

## Why
`_grep_identifier_files` parses `git grep -l` output as repo-relative paths. When a user has `color.ui=always`, Git can wrap those paths in ANSI escape codes, which prevents exclusion checks and file reads from matching the real paths.

Fixes #858.

## Validation
- `.venv/bin/python -m pytest tests/unit/server/mcp/test_answer_data_shape.py`
- `.venv/bin/python -m ruff check packages/server/src/repowise/server/mcp_server/tool_answer/data_shape.py tests/unit/server/mcp/test_answer_data_shape.py`
- `git diff --check`

### DIFF
--- a/packages/server/src/repowise/server/mcp_server/tool_answer/data_shape.py
+++ b/packages/server/src/repowise/server/mcp_server/tool_answer/data_shape.py
@@ -156,6 +156,10 @@ def _run_grep(
                 "git",
                 "--no-pager",
                 "grep",
+                # --no-color: this output is parsed, so a user's
+                # ``color.ui=always`` git config must not wrap paths in ANSI
+                # escapes (which would corrupt exclusion filtering and reads).
+                "--no-color",
                 *args,
                 "-l",
                 "-I",
--- a/tests/unit/server/mcp/test_answer_data_shape.py
+++ b/tests/unit/server/mcp/test_answer_data_shape.py
@@ -14,16 +14,17 @@ no groundable shape must fall through (return None), never a fabricated field.
 
 from __future__ import annotations
 
+import subprocess
 from pathlib import Path
 
 import pytest
 
+from repowise.core.exclusion import build_exclude_spec
 from repowise.server.mcp_server.tool_answer.data_shape import (
     _grep_identifier_files,
     _is_data_shape_question,
     mine_data_shape,
 )
-from repowise.core.exclusion import build_exclude_spec
 
 
 def _write(root: Path, rel: str, body: str) -> None:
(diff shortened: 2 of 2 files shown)


RaghavChamadiya:
Thanks for this, and good catch on the two ruff errors while you were in here. The Yoda flip and the import sort look like unrelated drive-bys but they're pre-existing violations that came in with #856, so this PR is currently the only thing standing between main and a clean `ruff check` on that file. Happy to take them.

I verified the fix reverts cleanly: with the `--no-color` line taken back out, your new test fails, and it passes with it in. I also confirmed the underlying behaviour independently. `git grep -l` returns `^[[35mpkg/models.py^[[m` under `color.ui=always`, and putting the flag before `*args` means it covers the `--no-index` retry path too, which is the branch the exclusion filter from #856 actually depends on.

One ask before I merge. Could you add the comment from #825 alongside the flag?

```python
# --no-color: this output is parsed, so a user's
# ``color.ui=always`` git config must not wrap paths in ANSI
# escapes (which would corrupt the path split below).
"--no-color",
```

The reason I'm fussy about it: this flag is load-bearing for the #856 exclusion filter, and when it's missing the failure is silent rather than loud. A comment-free `--no-color` in an arg list is exactly the kind of thing that gets tidied away by someone who has no reason to think it matters, which is how the sibling call site survived #825 in the first place. Cheap insurance against a third round of this.

Push that up and I'll merge. Thanks again.

packages/server/src/repowise/server/mcp_server/tool_answer/data_shape.py
```diff
@@ -156,6 +156,7 @@ def _run_grep(
                 "git",
                 "--no-pager",
                 "grep",
+                "--no-color",
```
swati510, line 162: The fix itself is correct, and I verified by hand that this flag suppresses the escape codes both for `color.ui = always` and for the narrower `color.grep = always`, and that it also covers the `--no-index` fallback on line 198, because both paths run through this same helper.

Could you add a short comment explaining why the flag is here, in the same style as the surrounding code? The comment on lines 152 to 154 explains why `--no-pager` and the stdin redirect are load-bearing, and the sibling call site in `_code_rationale.py` explains its own `--no-color` on lines 390 to 392. Something like the following would match both:

```python
                # --no-color: this output is parsed, so a user's
                # ``color.ui=always`` git config must not wrap paths in ANSI
                # escapes (which would corrupt exclusion filtering and reads).
                "--no-color",
```

Without it there is nothing stopping a future clean-up from deleting the flag and silently reintroducing this bug, which is more or less how we ended up with issue #858 in the first place.

swati510:
Your branch is currently eighty-two commits behind `main`. Everything passes as it stands, but would you mind rebasing onto current `main` so continuous integration runs against the tree this will actually land on?

Also, a small note for the description: alongside the fix, this pull request adds an `import subprocess` that the new test needs, moves the `from repowise.core.exclusion import build_exclude_spec` import above the `repowise.server` import, and rewrites one existing assertion into a more conventional form. The import move is worth having, because `main` currently fails ruff's isort rule on that block, so the change is a lint fix rather than a matter of taste. All three are fine and I am happy to keep them, but they are worth calling out so a reviewer is not surprised by the extra lines in the diff.

RaghavChamadiya (approved):
LGTM

pollychen-lab:
@swati510 @RaghavChamadiya thank you

PR #862, opened 2026-07-16
fix(answer): disable color in data-shape git grep

### LINKED ISSUE #858: [Bug] data_shape live-grep: add --no-color so exclusion filter survives color.ui=always

Yurii201811:
## Summary

- pass `--no-color` to the data-shape `git grep` command
- add a regression test with `color.grep=always` so ANSI escapes cannot leak into the generated output

Fixes #858

## Testing

- focused answer/data-shape suite: 25 passed
- Ruff formatting check
- scoped Ruff lint/import check
- `git diff --check`

### DIFF
--- a/packages/server/src/repowise/server/mcp_server/tool_answer/data_shape.py
+++ b/packages/server/src/repowise/server/mcp_server/tool_answer/data_shape.py
@@ -156,6 +156,7 @@ def _run_grep(
                 "git",
                 "--no-pager",
                 "grep",
+                "--no-color",
                 *args,
                 "-l",
                 "-I",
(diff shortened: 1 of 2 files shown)


RaghavChamadiya:
Thanks for this, Yurii. Closing it in favour of #860, which came in about six minutes earlier and also clears the pre-existing SIM300 that this one leaves behind.

Nothing wrong with your version. The fix is byte-identical, and I checked that your test genuinely pins the bug: it fails with the source line reverted. Using `color.grep=always` rather than `color.ui=always` is a perfectly good trigger, I confirmed both produce the ANSI-wrapped paths.

Three PRs landed on this one-liner inside about eight minutes, which is on my triage rather than on you. I'll start assigning these before people begin work. Would welcome another one from you, and if you ping me on an issue first I'll assign it so you're not racing anyone.

PR #865, opened 2026-07-16
fix(mcp): resolve health repository from scoped session

### LINKED ISSUE #855: get_health: "Repository not found" for any workspace alias that differs from the directory basename
### Summary

In workspace mode, `get_health` fails with `Repository not found: <alias>` for any repo whose **workspace alias differs from its directory basename** — even though `list_repos` advertises that exact alias and every other workspace tool (`get_answer`, `get_dead_code`, `search_codebase`) accepts it.

Version: **repowise 0.31.0** (workspace mode, stdio transport).

### Reproduce

Register two repos in a workspace where the alias does not equal the folder name:

| alias | directory | `get_health(repo=<alias>)` |
|---|---|---|
| `myrepo` | `myrepo/` | ✅ works |
| `myrepo` | `MyRepo/` (case differs) | ❌ `Repository not found: myrepo` |
| `nickname` | `some-other-dir/` (aliased) | ❌ `Repository not found: nickname` |

```
list_repos() -> {"workspace": true, "repos": [{"alias": "nickname", ...}]}
get_health(repo="nickname", targets=["some/file.py"])
  -> {"error": "Repository not found: nickname"}

get_answer(repo="nickname", question="...")      -> works
get_dead_code(repo="nickname")                    -> works
```

I confirmed this across 6 workspace repos: the 3 whose alias exactly equals their directory basename work; the 3 where it differs (two by letter-case only, one aliased to an unrelated folder name) all fail. 6/6 consistent with the cause below.

### Cause

There are two repo-resolution paths, and `get_health` uses the one that is not alias-aware.

`server/mcp_server/_helpers.py:55` — `_get_repo()` resolves only by `local_path`, then primary-key ID, then `Repository.name`:

```python
async def _get_repo(session: AsyncSession, repo: str | None = None) -> Repository:
    """Resolve a repository — by path, by ID, or return the first one."""
    if repo:
        # Try by path
        result = await session.execute(select(Repository).where(Repository.local_path == repo))
        ...
        # Try by ID
        obj = await session.get(Repository, repo)
        ...
        # Try by name
        result = await session.execute(select(Repository).where(Repository.name == repo))
        ...
        raise LookupError(f"Repository not found: {repo}")
```

It never consults the workspace alias registry. `Repository.name` is the directory basename, so a workspace alias only resolves here **by coincidence** — when the alias happens to equal the basename. The match is also case-sensitive, so an alias differing only in case fails.

The alias-aware path is `core/workspace/registry.py:110` (`resolve()`), which validates against the workspace config and raises a *different* error:

```
Unknown repo 'X'. Available: ['a', 'b', ...]
```

### Two distinct errors (worth noting for anyone debugging this)

These look similar but mean opposite things, which makes the bug confusing to diagnose:

- `Unknown repo 'X'. Available: [...]` — alias resolution ran and X is genuinely not a registered alias.
- `Repository not found: X` — **the alias IS valid and resolved fine**; `_get_repo` then failed to find it by name/path.

So `Repository not found` on an alias that `list_repos` just returned is always this bug.

### Suggested fix

Have `_get_repo()` consult the workspace registry first (mapping alias → `RepoContext`/`local_path`) before falling back to path/ID/name lookup, so alias resolution is consistent across every workspace-aware tool. A case-insensitive comparison on the `name` fallback would additionally fix the case-only mismatches.

### Workaround

None from the tool surface — passing the directory basename instead of the alias fails earlier in `registry.resolve()` with `Unknown repo`, since the basename is not a registered alias. Affected repos currently have no way to call `get_health` at all.

### Related: #435

#435 ("Chat MCP tools fail with 'Repository not found' for non-primary repos in workspace mode") lands on the same final line (`_helpers.py:72`) and the same underlying weakness — `_get_repo` is not workspace/alias-aware. But this report is a different surface and a different trigger, so I've filed it separately rather than commenting there:

- **Transport:** #435 is the web-UI chat path (`chat_tools.execute_tool` + module-global `_state` wired to the primary repo). This is the plain **stdio MCP** transport, where that global-state path isn't involved.
- **Trigger:** #435 frames the variable as *primary vs non-primary*. That does not explain the stdio behaviour — **non-primary repos work fine here as long as their alias equals their directory basename** (2 of my 3 working repos are non-primary). The variable on this path is alias vs `Repository.name`, case-sensitive.
- **Scope:** only `get_health` is affected on stdio; `get_answer` / `get_dead_code` / `search_codebase` resolve the same aliases correctly on the same repos, because they go through `registry.resolve()`.

If the fix proposed in #435 (thread a per-request `RepoContext` instead of mutating MCP module globals) is applied to the shared `_get_repo` resolution rather than only the chat path, it would likely close both.

Real5K:
## Summary

- Resolve `get_health` repositories from the session already scoped by `_resolve_repo_context`
- Prevent workspace aliases from being incorrectly interpreted as repository names
- Add regression coverage for an alias that differs from its directory basename

## Related Issues

Closes #855

## Test Plan

- [x] Tests pass (`34` relevant workspace and health MCP tests)
- [ ] Web build passes (`npm run build`) *(not applicable - no frontend changes)*

## Checklist

- [x] Documentation is not needed for this internal bug fix

### DIFF
--- a/packages/server/src/repowise/server/mcp_server/tool_health.py
+++ b/packages/server/src/repowise/server/mcp_server/tool_health.py
@@ -380,7 +380,7 @@ async def get_health(
     perf_coverage: PerfCoverage | None = None
     perf_findings_count = 0
     async with get_session(ctx.session_factory) as session:
-        repository = await _get_repo(session, repo)
+        repository = await _get_repo(session)
 
         all_metrics_q = select(HealthFileMetric).where(
             HealthFileMetric.repository_id == repository.id
--- a/tests/unit/server/test_mcp_workspace.py
+++ b/tests/unit/server/test_mcp_workspace.py
@@ -34,7 +34,14 @@ from repowise.core.workspace.registry import RepoContext
 _NOW = datetime(2026, 4, 12, 10, 0, 0, tzinfo=UTC)
 
 
-async def _make_repo_context(alias: str, repo_path: str, pages: list, extra_models: list | None = None) -> RepoContext:
+async def _make_repo_context(
+    alias: str,
+    repo_path: str,
+    pages: list,
+    extra_models: list | None = None,
+    *,
+    repo_name: str | None = None,
+) -> RepoContext:
     """Create an in-memory RepoContext with test data."""
     engine = create_async_engine(
         "sqlite+aiosqlite:///:memory:",
@@ -47,7 +54,7 @@ async def _make_repo_context(alias: str, repo_path: str, pages: list, extra_mode
     async with factory() as session:
         repo = Repository(
             id=f"repo-{alias}",
-            name=alias,
+            name=repo_name or alias,
             url=f"https://example.com/{alias}",
             local_path=repo_path,
             default_branch="main",
@@ -58,7 +65,7 @@ async def _make_repo_context(alias: str, repo_path: str, pages: list, extra_mode
         session.add(repo)
         for p in pages:
             session.add(p)
-        for m in (extra_models or []):
+        for m in extra_models or []:
             session.add(m)
         await session.commit()
 
@@ -188,7 +195,7 @@ async def workspace_mcp():
 
     frontend_ctx = await _make_repo_context(
         "frontend",
-        "/tmp/workspace/frontend",
+        "/tmp/workspace/web-client",
         pages=[
             Page(
                 id="repo_overview:frontend",
@@ -246,6 +253,7 @@ async def workspace_mcp():
                 affected_modules_json='["src"]',
             ),
         ],
+        repo_name="web-client",
     )
 
     registry = _MockRegistry(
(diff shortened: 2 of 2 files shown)


RaghavChamadiya:
Thanks for this, and for the clear write-up on the issue too. The diagnosis was spot on.

I like that you went simpler than the fix suggested in #855. Making _get_repo alias-aware would have pushed workspace knowledge into a helper that doesn't need it. Since _resolve_repo_context hands back a RepoContext with its own session_factory scoped to a single repo's database, the alias never had any business reaching _get_repo in the first place. This also brings get_health in line with the other 14 _get_repo call sites, all of which already pass no repo argument. That outlier is exactly why get_health was the only broken tool on stdio.

Verified locally: reverting just the one-line change makes your new test fail with LookupError: Repository not found: frontend, the same error from the issue, and the other 26 tests still pass, so the fixture change is clean. Lint and format pass.

One thing I'll pick up separately: tool_refactoring.py:99 has the identical _get_repo(session, repo) call and the same latent bug. It's opt-in behind a config flag so it's rarely hit, which is likely why it didn't surface for you. Out of scope here.

Merging. Thanks again

RaghavChamadiya (approved):

PR #867, opened 2026-07-16
feat(risk): add --exclude flag, .riskignore support, and get_change_r…

barrcodes:
### Problem

Issue #868 

Tests were inflating risk scores and skewing percentile rankings both interactively and when called from AI-assisted workflows (number of lines changed). The CLI had no way to filter those paths, and there was no MCP-accessible equivalent of the brief summary `repowise risk` command.

## Solution

- Adds `--exclude` / `-x` option to the `repowise risk` CLI command, accepting repeatable gitignore-style patterns to omit files from both the scored diff and the baseline sample.
- Introduces `.riskignore` support — a repo-root file whose non-comment lines are applied automatically as exclusions on every `risk` invocation, without requiring explicit flags.
- Adds a new `get_change_risk` MCP tool that exposes this scoring surface to AI assistants, with `revspec`, `extensions`, `exclude_patterns`, and `baseline` parameters mirroring the CLI contract.
- Extracts shared orchestration logic into a new `ChangeRiskResult` / `score_live_change` service layer (`packages/core/.../change_risk/service.py`), eliminating the previously duplicated score-and-rank logic that was split between the CLI and the server router.


RaghavChamadiya:
Really nice PR, thanks for this. The `.riskignore` + `--exclude` design lines up exactly with what #868 asked for, and folding the duplicated score-and-rank logic into a shared `score_live_change` service is a genuine cleanup, that split had been bugging me. The test coverage across the feature, baseline, CLI, and MCP layers is great.

Two things before merge:

**1. Please rebase on `main`.** I just merged #875, which reclassifies `get_conformance` and `generate_refactoring_code` as opt-in. It edits the same tool-count lines in `docs/MCP_TOOLS.md` and `docs/CONFIG.md` that you touch here, so there will be a small conflict on those counts. On current `main` the default single-repo surface is **10**, workspace default is **12**, and opt-in is **4**, so your "+1 for `get_change_risk`" lands as 10->11 single-repo and 12->13 workspace, not the 11->12 / 14->15 the current diff assumes.

**2. Lean profile: docs and code disagree on whether `get_change_risk` is included.** The docs here add it to the agent-lean profile, but `LEAN_TOOLS` in `_tool_selection.py` is unchanged (still 5 tools) and `test_lean_profile_full_registry` explicitly asserts `"get_change_risk" not in enabled`:

- `docs/CONFIG.md`: *"lean selects the agent-lean profile: ... `get_risk`, `get_change_risk`"*
- `docs/MCP_TOOLS.md`: *"...`get_risk`, and `get_change_risk` ... the six schemas the agent actually reaches for"*
- `_tool_selection.py`: `LEAN_TOOLS = {get_answer, get_context, get_symbol, search_codebase, get_risk}` (5)
- test: `assert "get_change_risk" not in enabled`

A user who sets `mcp.tools: lean` after reading the docs won't actually get `get_change_risk`. Since the test deliberately locks the exclusion, I'd guess the code is what you intended and the two doc edits should revert to "five schemas" / drop `get_change_risk` from the lean lists. If you do want it in lean, the inverse works too: add it to `LEAN_TOOLS` and flip the test. Either direction is fine, just want all four to agree.

Thanks again, this is close.

barrcodes:
Thanks, will get right on that! Yes, I originally added it to lean, and then decided idk if it's the right fit. It's kind of a convenience method, even though it's the more bite-sized of the two. Must have missed a couple spots.

barrcodes:
@RaghavChamadiya I added a couple small tweaks to the get_change_risk mcp tool to ensure the agent has enough context to understand its output. Mostly just clarifying percentile vs risk score, and adding the context of the baseline count for a certainty metric.

Rebased, updated docs. Thank you!

RaghavChamadiya (approved):
Approving, thanks for the quick turnaround. Verified locally: the lean profile, docs, and registry all agree now, counts line up with post-#875 main, and the full MCP suite (477 tests) plus the new risk tests pass on Windows

PR #877, opened 2026-07-17
fix: normalize decision staleness timestamps

### LINKED ISSUE #874: [Bug] Decision staleness recompute always crashes on naive/aware datetime mix and is silently skipped
## Describe the Bug

Decision staleness scoring crashes with `TypeError: can't compare offset-naive and offset-aware datetimes` and is silently skipped, so `staleness_score` is never recomputed after a decision is first created.

Root cause in `DecisionExtractor.compute_staleness` (`packages/core/src/repowise/core/analysis/decisions/extractor.py:1347-1348`):

- `last_commit` comes from the git metadata map, built timezone-aware: `datetime.fromtimestamp(last_ts, tz=UTC)` (`packages/core/src/repowise/core/ingestion/git_indexer/file_history.py:286`)
- `decision_created_at` is a `DecisionRecord.created_at` loaded through SQLAlchemy's SQLite dialect, which returns naive datetimes even though the column is declared `DateTime(timezone=True)`
- `last_commit > _created` raises, and even if it passed, `(now - _created)` with `now = datetime.now(UTC)` on the next line would raise too

Both call sites swallow the error:

- init pipeline: `pipeline/persist.py:819` logs `staleness_scoring_skipped` at debug level (invisible in normal runs)
- docs-mode update: `recompute_decision_staleness` raises inside the decision-persist block of `_persist_full_update_async`, degrading the whole step (`update_cmd/persistence.py:505`), observed as `degraded: ["Decision persist: can't compare offset-naive and offset-aware datetimes"]` on every docs update

This bug was already noticed in passing in #641's description ("fires during indexing, separate bug") but never filed or fixed.

## Steps to Reproduce

1. `repowise init` a repo with docs so at least one decision record with affected files exists
2. Commit a change and run `repowise update --docs`
3. The run completes but reports `degraded: ["Decision persist: can't compare offset-naive and offset-aware datetimes"]`
4. Decision `staleness_score` values in the DB never change from their initial values

## Expected Behavior

Staleness recompute runs on every update, so decisions governing files that changed since the decision was made drift toward stale and surface in the decision health summary.

## Actual Behavior

The first decision with affected files and git metadata raises the TypeError; the whole staleness pass (and its degrade message the rest of the decision-persist step) is skipped. Staleness scores are frozen at whatever they were on creation, so stale-decision surfacing (`get_stale_decisions`, decision health, governance findings that depend on it) runs on dead data for SQLite-backed repos, which is every OSS install.

## Environment

- OS: Windows 11
- Python version: 3.13.5
- Repowise version: 0.31.0 (editable install from main)

## Additional Context

Suggested fix: normalize both operands to aware UTC at the top of `compute_staleness` with a small helper (naive input assumed UTC, since both writers store UTC), e.g. apply it to `decision_created_at` after the `isinstance(str)` parse and to `last_commit`. The commit-date branch at extractor.py:1369 already compares ISO strings and is unaffected. A regression test can call `compute_staleness` directly with a naive `created_at` and an aware `last_commit_at`, which reproduces the TypeError today.

Related prior work: #516 fixed the same naive/aware mix in the jobs endpoint sort; #641 mentioned this instance but explicitly left it out of scope.

pollychen-lab:
Fixes #874.

## Summary

- normalize decision and git timestamps to aware UTC before staleness comparisons
- treat SQLite-returned naive datetimes as UTC, matching how Repowise writes them
- add a regression test for naive decision `created_at` with aware git `last_commit_at`

## Validation

- `uv run pytest tests/unit/analysis/test_decision_staleness.py -v`
- `uv run pytest tests/unit/analysis/test_decision_sources.py tests/unit/analysis/test_decision_provenance.py tests/unit/analysis/test_decision_staleness.py -v`
- `uv run pytest tests/unit/analysis -v`
- `uv run ruff format packages/core/src/repowise/core/analysis/decisions/extractor.py tests/unit/analysis/test_decision_staleness.py`
- `uv run ruff format --check packages/core/src/repowise/core/analysis/decisions/extractor.py tests/unit/analysis/test_decision_staleness.py`
- `uv run ruff check packages/core/src/repowise/core/analysis/decisions/extractor.py tests/unit/analysis/test_decision_staleness.py`
- `git diff --check`
- `rg -n "Karim13014|claude-code@anthropic.com" .`

## Notes

- Full `uv run ruff format --check packages/ tests/` and `uv run ruff check packages/ tests/` still fail on existing unrelated baseline formatting/lint issues outside this PR.
- Targeted `uv run mypy packages/core/src/repowise/core/analysis/decisions/extractor.py` reports existing unrelated errors in the module, but none are from this change.

### DIFF
--- a/packages/core/src/repowise/core/analysis/decisions/extractor.py
+++ b/packages/core/src/repowise/core/analysis/decisions/extractor.py
@@ -71,6 +71,18 @@ def _truncate_title(text: str, limit: int) -> str:
     return window[:cut].rstrip() + "…"
 
 
+def _as_aware_utc(value: datetime) -> datetime:
+    """Return ``value`` as a timezone-aware UTC datetime.
+
+    SQLite drops timezone information from ``DateTime(timezone=True)`` columns,
+    but Repowise writes those values as UTC. Treat naive values as UTC so they
+    can be compared with git metadata timestamps, which are already aware UTC.
+    """
+    if value.tzinfo is None or value.utcoffset() is None:
+        return value.replace(tzinfo=UTC)
+    return value.astimezone(UTC)
+
+
 # ---------------------------------------------------------------------------
 # Data models
 # ---------------------------------------------------------------------------
@@ -1341,9 +1353,11 @@ class DecisionExtractor:
             if last_commit and decision_created_at:
                 if isinstance(last_commit, str):
                     last_commit = datetime.fromisoformat(last_commit.replace("Z", "+00:00"))
+                last_commit = _as_aware_utc(last_commit)
                 _created = decision_created_at
                 if isinstance(_created, str):
                     _created = datetime.fromisoformat(_created.replace("Z", "+00:00"))
+                _created = _as_aware_utc(_created)
                 if last_commit > _created:
                     age_days = (now - _created).days
                     commit_count = meta.get("commit_count_90d", 0)
--- a/tests/unit/analysis/test_decision_staleness.py
+++ b/tests/unit/analysis/test_decision_staleness.py
@@ -1,0 +1,23 @@
+from __future__ import annotations
+
+from datetime import UTC, datetime, timedelta
+
+from repowise.core.analysis.decision_extractor import DecisionExtractor
+
+
+def test_compute_staleness_accepts_sqlite_naive_created_at() -> None:
+    created_at = datetime.now(UTC).replace(tzinfo=None) - timedelta(days=30)
+    last_commit_at = datetime.now(UTC) - timedelta(days=1)
+
+    score = DecisionExtractor.compute_staleness(
+        decision_created_at=created_at,
+        affected_files=["src/app.py"],
+        git_meta_map={
+            "src/app.py": {
+                "last_commit_at": last_commit_at,
+                "commit_count_90d": 6,
+            }
+        },
+    )
+
+    assert score > 0


RaghavChamadiya:
Thanks @pollychen-lab, this is a clean fix. Merging.

I verified it reverts cleanly: taking out the two `_as_aware_utc` calls reproduces the exact `TypeError: can't compare offset-naive and offset-aware datetimes` from #874, and the whole `tests/unit/analysis` suite (433 tests) passes with it in. Treating naive as UTC is the right assumption given both writers store UTC.

I did go looking for whether this deserved a shared helper rather than a local one, and I think you made the right call keeping it local. The two named helpers that look similar (`_parse_dt` in `crud/_shared.py` and `_parse_datetime` in `generation/models.py`) both take a *string* and parse it, so neither was reusable for an already-parsed datetime. The other sites that touch DB datetimes (`_is_old` in the dead-code analyzer, `incremental.py:524`) already normalize inline. So `_as_aware_utc` isn't duplicating anything that existed, and it's slightly more correct than the rest since it does `astimezone(UTC)` for aware-but-not-UTC values instead of leaving them alone.

Tiny thing, don't respin for it: the new test imports from `repowise.core.analysis.decision_extractor`, the back-compat shim, whose docstring asks you to use `decisions.extractor` instead. Both existing sibling tests do the same, so you're matching convention and the shim is really the thing that needs cleaning up.

RaghavChamadiya (approved):
Verified locally: reverting the two `_as_aware_utc` calls reproduces the TypeError from #874, and the full `tests/unit/analysis` suite passes with the fix in. Checked the shared-helper question too and the local helper is the right call.

pollychen-lab:
@RaghavChamadiya thanks

PR #878, opened 2026-07-17
fix: preserve last_docs_commit pointer during index-only updates (Fixes #873)

### LINKED ISSUE #873: [Bug] Workspace updates never learned the last_docs_commit pointer, so update --docs no-ops in workspaces (reintroduces #847)
## Describe the Bug

#849 fixed docs-pointer drift (#847) by making `repowise update --docs` diff from a dedicated `last_docs_commit` pointer instead of `last_sync_commit`, and by backfilling that pointer during index-only updates so legacy state files heal.

That fix only covers the single-repo update path in `packages/cli/src/repowise/cli/commands/update_cmd/`. The workspace update path (`packages/core/src/repowise/core/workspace/update.py`) was never taught about the docs pointer:

- it always uses `last_sync_commit` as the diff base (`update.py:433`)
- it stamps `last_sync_commit` forward on every run, including index-only ones (`update.py:677`)
- it never creates, preserves, or advances `last_docs_commit`

So in a workspace, every index-only update (for example the post-commit hook) advances the sync pointer, and a later `update --docs`, which auto-detects workspace mode from the workspace root, falls back to the freshly advanced sync pointer, sees an empty diff, and reports up to date while the wiki prose is actually stale. This silently reintroduces #847 for every workspace user with docs enabled.

## Steps to Reproduce

1. Have an indexed workspace with docs enabled and the git hook installed
2. Commit a change to any repo in the workspace (the hook runs an index-only workspace update and advances `last_sync_commit` to HEAD)
3. Run `repowise update --docs` from the workspace root
4. It reports all repos up to date; no pages regenerate. `.repowise/state.json` still has no `last_docs_commit` key no matter how many updates run

Workaround: `repowise update --docs --since <last-docs-ref> --no-workspace` per repo.

## Expected Behavior

The docs run diffs from the last commit docs were actually generated at, exactly as the single-repo path does since #849.

## Actual Behavior

"3 repo(s) up to date / All repos are up to date." with zero pages regenerated, immediately after commits that changed dozens of source files.

## Environment

- OS: Windows 11
- Python version: 3.13.5
- Repowise version: 0.31.0 (editable install from main)

## Additional Context

Suggested fix: mirror the two behaviors #849 added, in `workspace/update.py`:

1. When persisting an index-only workspace update, backfill `last_docs_commit` from the pre-advance `last_sync_commit` if the key is absent (same one-time migration as `_persist_index_only_update` in `packages/cli/src/repowise/cli/commands/update_cmd/persistence.py:177`).
2. In docs mode, use `last_docs_commit` (fallback `last_sync_commit`) as the per-repo base ref, and stamp `last_docs_commit = head` only on runs that actually generate pages.

`tests/unit/cli/test_issue_847_fix.py` documents the exact contract and is a good model for the workspace-side tests. #849's diff is small and serves as a direct blueprint, which is why this is labeled good first issue, but note the file also handles single-flight locking and roll-forward, so keep the change scoped to pointer selection and stamping.

jyotirmya17:
This PR resolves the pointer loss bug reported in #873 where index-only updates (like workspace flows) advanced `last_sync_commit` without preserving the `last_docs_commit` boundary, causing subsequent docs runs to see an empty diff.

As discussed in the issue, this implements the same migration semantics introduced in #847 for single-repo updates: preserving the pre-advance sync pointer when the docs pointer is absent, leaving existing pointers untouched, and advancing the sync pointer.

### Changes Made
1. **Core Fix (`workspace/update.py`)**: Added the backfill for `last_docs_commit` inside `_update_one` prior to advancing `last_sync_commit`.
2. **Workspace `--docs` Warning (`update_cmd/command.py`)**: Added a visible console warning when a user runs `repowise update --docs --workspace`, explicitly advising them to run the docs generation per-repo as workspace mode does not yet support it.
3. **Tests (`tests/unit/cli/test_issue_873_fix.py`)**: Added a verification suite that simulates a workspace update on a legacy state, ensuring the backfill contract is honored and that a subsequent `--docs` run successfully reaches back to the correct base commit.
4. **Adversarial Hardening**: Audited the codebase for parallel implementations of this exact same bug. I found and applied the identical backfill fix to:
   - `server/job_executor.py` (`_stamp_last_sync_commit`): Ensures background jobs / webhooks running index-only incremental syncs don't permanently break the docs pointer for a repository.
   - `update_cmd/command.py` (Early-exit path): Ensures the docs pointer is backfilled and managed correctly when `repowise update` gracefully exits early due to "No changed files detected".

### Verification
- [x] Tested locally via the newly added `test_issue_873_fix.py` test suite.
- [x] Verified that existing pointer tests (`test_issue_847_fix.py`) continue to pass.
- [x] Manually validated the workspace warning output.

Fixes #873

### DIFF
--- a/docs/CLI_REFERENCE.md
+++ b/docs/CLI_REFERENCE.md
@@ -161,7 +161,7 @@ See [WORKTREES.md](WORKTREES.md).
 | `--no-workspace` | Force single-repo mode (handy when running from a workspace root) |
 | `--repo` | Update a specific workspace repo by alias |
 | `--index-only` | Refresh the index only, skip doc regeneration for this run |
-| `--docs` / `--no-docs` | Regenerate wiki pages for changed files, or skip doc regeneration entirely |
+| `--docs` / `--no-docs` | Regenerate wiki pages for changed files, or skip doc regeneration entirely. Not supported in workspace mode yet: with `--workspace` the flag is ignored (with a warning) and you should run docs per repo, e.g. `repowise update --docs --no-workspace` from inside the repo. |
 | `--full` | Upgrade a fast (`--mode fast`) index to a full one, see below. Single-repo only; errors in workspace mode. |
 | `--no-cost-tracking` | Don't record LLM spend for this run |
 | `--agents` / `--no-agents` | Generate or skip managed `AGENTS.md` after update. Persists the preference. |
--- a/packages/cli/src/repowise/cli/commands/update_cmd/command.py
+++ b/packages/cli/src/repowise/cli/commands/update_cmd/command.py
@@ -45,7 +45,7 @@ from .incremental import (
     _refresh_knowledge_graph,
     _run_partial_analysis,
 )
-from .mode import _infer_legacy_docs_enabled, _resolve_index_only_mode
+from .mode import _infer_legacy_docs_enabled
 from .persistence import (
     _persist_index_only_update,
     _run_full_health_rescore,
@@ -463,11 +463,13 @@ def run_update(
     load_dotenv(repo_path)
     state = load_state(repo_path)
     from repowise.cli.commands.update_cmd.mode import _resolve_index_only_mode
+
     resolved_index_only = _resolve_index_only_mode(
         index_only=index_only, docs_flag=docs_flag, state=state
     )
     base_ref = since or (
-        state.get("last_sync_commit") if resolved_index_only
+        state.get("last_sync_commit")
+        if resolved_index_only
         else state.get("last_docs_commit", state.get("last_sync_commit"))
     )
     head = get_head_commit(repo_path)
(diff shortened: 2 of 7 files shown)


RaghavChamadiya:
Thanks @jyotirmya17, this does exactly what we agreed on the issue, and the tests genuinely pin it: reverting the backfill in workspace/update.py fails both with a clear message. The ordering is right, backfilling from the old last_sync_commit before advancing it, which is the easy part to get wrong.

You also went past the agreed scope in two places, and both are real bugs rather than noise, so I'd like to keep them. But each needs a bit more before it lands.

The command.py early-exit fix is a good catch. On main that path advances last_sync_commit with no backfill, so legacy state loses the docs pointer there exactly like the workspace path did. #849 missed it. Your fix is correct too: since docs runs already diff from last_docs_commit, "no changed files" really does mean no docs work is pending, so stamping last_docs_commit = head there is safe. It just has no test, and it's the kind of pointer logic that should not go untested. Please add one.

The job_executor.py change is half a fix. _stamp_last_sync_commit does have the same bug and backfilling is the right direction. But it runs for every non-initial-index mode, including webhook sync jobs that regenerate pages, and those will now backfill last_docs_commit to the old sync commit and never advance it, so the next docs run redoes work already done. Either complete it (advance the docs pointer when the job actually generated pages) or pull it out into its own PR where the server-side behavior can be reasoned about on its own. Your call which, both are fine by me.

On the test file name, that one's my fault. I pointed you at test_issue_847_fix.py for conventions, but it turns out to be the only issue-numbered test file in the repo. All 65 others in tests/unit/cli are named for what they test. Please rename to something like test_workspace_docs_pointer.py. I'll rename 847 separately so the odd one out stops misleading people.

Rebase on main when you get a chance. You're 19 commits behind and the branch re-carries the already-merged #849 commit, which is why the diff shows test_issue_847_fix.py as a new 235-line file. It merges cleanly, I checked, but the PR should show the four files it actually touches.

Minor: workspace/update.py is format-clean on main and format-dirty after this (trailing whitespace on the blank line at the top of your hunk), plus 3 import-sort errors in the new test file. ruff check --fix && ruff format clears both. The F401/F811 in command.py are pre-existing, ignore those.

jyotirmya17:
@RaghavChamadiya  I've updated the PR with the requested changes:

Added test_docs_early_exit_advances_docs_pointer to explicitly test the pointer logic on the command.py early-exit path.
Reverted the extra scope in job_executor.py (we can tackle that bug separately).
Renamed the test file to test_workspace_docs_pointer.py and ensured everything is clean with ruff.
Let me know if this looks good.


Just wait for a while, I am fixing things.

jyotirmya17:
@RaghavChamadiya 
All CI checks have passed now. Let me know if everything looks good to merge!

RaghavChamadiya:
Pushed a commit on top of yours to square away the last review points, hope that's alright.

Main thing was the index-only early-exit: it was only advancing last_sync_commit in docs mode, so a plain update that hit "no changed files" left the on-disk pointer behind while the DB head moved forward. Reverted the is_indexable filter too so the change stays purely about the pointers, and added a test that pins the index-only case with an empty commit. Also renamed 847 to test_docs_pointer.py and dropped a line in CLI_REFERENCE about the workspace --docs limit.

Really nice work on this one, the backfill and the workspace warning were exactly right, and your tests made it easy to move things around with confidence. Thanks for sticking with it through the back and forth

RaghavChamadiya (approved):

jyotirmya17:
That's absolutely alright, thanks for stepping in to square away those final edge cases! Good catch on the plain update early-exit leaving the on-disk pointer behind, I see why keeping it purely about the pointers makes more sense there.

I'm glad the tests were helpful! I learned a lot working through the edge cases on this one. Thanks for your patience and for guiding me through the back and forth.

PR #884, opened 2026-07-17
fix: resolve alias correctly in generate_refactoring_code

### LINKED ISSUE #880: [Bug] generate_refactoring_code: "Repository not found" for workspace aliases that differ from the directory basename
`get_conformance`'s sibling `generate_refactoring_code` has the same alias-resolution bug that #855 reported for `get_health`, fixed in #865.

`packages/server/src/repowise/server/mcp_server/tool_refactoring.py:99`:

```python
async with get_session(ctx.session_factory) as session:
    repository = await _get_repo(session, repo)
```

`_resolve_repo_context(repo)` on the line above already resolved the alias and handed back a `RepoContext` whose `session_factory` points at that one repo's database. Passing `repo` on to `_get_repo` makes it re-filter by alias inside a session that is already scoped to a single repository. `_get_repo` only knows paths, IDs, and `Repository.name`, so a workspace alias resolves there only when it happens to equal the directory basename, and raises `LookupError: Repository not found: <alias>` when it does not.

This is the last remaining instance. `tool_health.py:383` was fixed in #865, and all other `_get_repo` call sites across the MCP tools already pass no repo argument.

**Why it hasn't been reported:** `generate_refactoring_code` is opt-in behind `refactoring.llm.enabled` in `.repowise/config.yaml` and was gated further in `73a77cff`, so it is rarely exercised.

**Fix:** drop the argument, matching #865 and every other call site:

```python
repository = await _get_repo(session)
```

Nothing is lost by dropping it. In single-repo mode `_resolve_repo_context` already validates the `repo` param against the DB and raises `LookupError` if it is invalid (`_helpers.py:156`).

**Test:** `tests/unit/server/test_mcp_workspace.py` has a fixture where the `frontend` alias maps to a repo named `web-client` at `/tmp/workspace/web-client`, added in #865. A test calling `generate_refactoring_code(repo="frontend")` against it reproduces the `LookupError` today.

abhiramvsmg:
generate_refactoring_code passed repo to _get_repo(session, repo) even though _resolve_repo_context(repo) had already resolved the alias and scoped the session to that one repository. This caused LookupError: Repository not found for workspace aliases that differ from the repo's directory basename, since _get_repo only matches on paths, IDs, and Repository.name.

Dropped the repo argument, matching the same fix already applied to get_health in #865.

Added test_generate_refactoring_code_resolves_alias_different_from_repo_name using the existing frontend/web-client alias fixture. Full test suite passes (28/28 in test_mcp_workspace.py).

Fixes #880

### DIFF
--- a/tests/unit/server/test_mcp_workspace.py
+++ b/tests/unit/server/test_mcp_workspace.py
@@ -301,6 +301,15 @@ async def test_get_health_resolves_alias_different_from_repo_name(workspace_mcp)
     assert result["kpis"]["file_count"] == 0
 
 
+@pytest.mark.asyncio
+async def test_generate_refactoring_code_resolves_alias_different_from_repo_name(workspace_mcp):
+    from repowise.server.mcp_server import generate_refactoring_code
+
+    result = await generate_refactoring_code(suggestion_id="does-not-exist", repo="frontend")
+
+    assert result["error"] is not None
+
+
 # ---------------------------------------------------------------------------
 # get_overview — workspace footer
 # ---------------------------------------------------------------------------
--- a/packages/server/src/repowise/server/mcp_server/tool_refactoring.py
+++ b/packages/server/src/repowise/server/mcp_server/tool_refactoring.py
@@ -96,7 +96,7 @@ async def generate_refactoring_code(suggestion_id: str, repo: str | None = None)
         }
 
     async with get_session(ctx.session_factory) as session:
-        repository = await _get_repo(session, repo)
+        repository = await _get_repo(session)
         row = await get_refactoring_suggestion(session, repository.id, suggestion_id)
         if row is None:
             return {


tests/unit/server/test_mcp_workspace.py
```diff
@@ -301,6 +301,15 @@ async def test_get_health_resolves_alias_different_from_repo_name(workspace_mcp)
     assert result["kpis"]["file_count"] == 0
 
 
+@pytest.mark.asyncio
+async def test_generate_refactoring_code_resolves_alias_different_from_repo_name(workspace_mcp):
+    from repowise.server.mcp_server import generate_refactoring_code
+
+    result = await generate_refactoring_code(suggestion_id="does-not-exist", repo="frontend")
+
+    assert result["error"] is not None
```
swati510, line 310: Thank you for adding a regression test, and I confirmed that it does fail with the original `LookupError` when the production fix is reverted, so it genuinely guards the bug.

Could I suggest tightening the assertion to `assert result["error"] == "not_found"`? The tool has three different early returns that all set an `error` key, and only `not_found` proves that execution actually got past the repository lookup that this pull request fixes. In particular, `llm_enrichment_enabled` currently defaults to `True` when the repository has no configuration file, which is why the test reaches the lookup at all today. If that default were ever flipped, the tool would return `error: "disabled"` before touching the database, the test would still pass, and it would quietly stop testing anything.

swati510:
This is outside the scope of your fix, so please feel free to leave it for a separate change, but I noticed while reviewing that `generate_refactoring_code` does not guard against `repo="all"`. At `packages/server/src/repowise/server/mcp_server/tool_refactoring.py` line 86 it calls `_resolve_repo_context(repo)` directly, and in workspace mode that helper deliberately raises a `ValueError` for `repo="all"` rather than handling it, on the grounds that each tool should decide for itself what "all repositories" means. Seven other tools guard for this first and return a readable error dictionary by calling `_unsupported_repo_all(...)`, among them `get_risk`, `get_symbol`, `get_why`, `get_context` and `get_answer`. Without that guard, `generate_refactoring_code(suggestion_id=..., repo="all")` raises an uncaught `ValueError` instead of returning a readable error dictionary, and I confirmed that by calling it. It is the same family of workspace handling gap that this pull request closes, and it is only a couple of lines if you would like to include it.

RaghavChamadiya (approved):
Thanks, this lines it up with how `get_health` already does it. Small fix, and the regression test is appreciated.

PR #911, opened 2026-07-18
fix(server): fail job explicitly on unknown execution mode (#881)

### LINKED ISSUE #881: [Bug] Job executor silently accepts unknown modes and runs with partial behavior
`packages/server/src/repowise/server/job_executor.py:343`:

```python
mode = config.get("mode") or "sync"
is_full_resync = mode == "full_resync"
is_initial_index = mode == "initial_index"
is_index_only = mode == "index_only"
```

Any mode string that is not one of these four falls through every branch. The job still runs, still reports `completed`, and still advances `last_sync_commit`, but `generate_docs` is False and the incremental page-regen at line 420 (gated on `mode == "sync"`) is skipped. So an unknown mode silently produces a re-index with no documentation and a sync baseline that claims the repo is current.

This is what #837 was: webhooks enqueued `"incremental"`, which matches nothing. #848 fixed that caller. This issue is about the executor accepting the bad value in the first place, so the next typo or renamed mode fails the same silent way rather than loudly.

**Scope check.** `execute_job` is only reachable from three places, and the only modes that reach it are the four above:

- `routers/repos.py:705` (`initial_index`, `full_resync`)
- `routers/webhooks.py:57` (`sync`)
- `scheduler.py:162` (`sync`)

Other `{"mode": ...}` job rows exist but never reach the executor and should not be affected: `update_cmd/persistence.py:579` (`incremental`) and `upgrade_flow.py:243` (`upgrade`) are bookkeeping records so the web UI's "last synced" updates, and the latter is written with `status="completed"` directly.

**Fix direction:** validate `mode` against the known set at the top of `execute_job` and fail the job explicitly (recorded failure with a clear message) rather than running partial work. The default when `mode` is absent should stay `"sync"`.

**Worth noting while in there:** `routers/pages.py:121` enqueues a pending `{"mode": "single_page"}` job that nothing ever executes. `job_executor` has no `single_page` branch and no caller passes those job IDs to `execute_job`, so `POST /lookup/regenerate` returns `202 accepted` and appears to do nothing. That may want its own issue, but validation would have surfaced it.

akshatmalik-bruh:
Here is the exact filled-out content ready to copy and paste into your GitHub PR:

***

## Summary

- Validates job `mode` against `VALID_JOB_MODES = {"sync", "full_resync", "initial_index", "index_only"}` in `execute_job()`.
- Raises a `ValueError` on unknown modes so `execute_job()` explicitly fails the job and records the error message in the DB instead of silently running partial work and updating baseline commit state.
- Preserves `"sync"` as default fallback when `mode` is missing or empty.

## Related Issues

Fixes #881

## Test Plan

- [x] Tests pass (`pytest`) - Added unit tests `test_execute_job_fails_on_unknown_mode` and `test_execute_job_defaults_to_sync_when_mode_absent` in `tests/unit/server/test_job_executor.py`.

## Checklist

- [x] I have updated documentation if needed (n/a — internal validation change)

***

### DIFF
--- a/packages/server/src/repowise/server/job_executor.py
+++ b/packages/server/src/repowise/server/job_executor.py
@@ -108,6 +108,19 @@ def _repo_wiki_style(repo: Any, repo_path: str) -> str:
     return resolve_style(style, repo_path=repo_path).name
 
 
+# Valid job execution modes handled by execute_job.
+# NOTE: "generate" (HTTP repowise generate) and "single_page" (scoped page resync)
+# are dispatched to _run_generate_job below, so validation must accept them upfront.
+VALID_JOB_MODES: set[str] = {
+    "sync",
+    "full_resync",
+    "initial_index",
+    "index_only",
+    "generate",
+    "single_page",
+}
+
+
 # Phase → numeric level mapping for job.current_level
 _PHASE_LEVELS = {
     "traverse": 0,
@@ -342,7 +355,13 @@ async def execute_job(
             # Resolve the wiki style while ``repo`` is still session-attached.
             wiki_style = _repo_wiki_style(repo, repo_path)
             config = json.loads(job.config_json) if job.config_json else {}
-            mode = config.get("mode") or "sync"
+            mode = str(config.get("mode") or "sync")
+            if mode not in VALID_JOB_MODES:
+                valid_str = ", ".join(sorted(VALID_JOB_MODES))
+                raise ValueError(
+                    f"Invalid job mode '{mode}'. Expected one of: {valid_str}"
+                )
+
             is_full_resync = mode == "full_resync"
             is_initial_index = mode == "initial_index"
             is_index_only = mode == "index_only"
--- a/tests/unit/server/test_job_executor.py
+++ b/tests/unit/server/test_job_executor.py
@@ -568,3 +568,132 @@ async def test_index_only_job_does_not_generate_docs(session_factory, tmp_path):
 
     assert run_pipeline_mock.await_args.kwargs["generate_docs"] is False
     assert run_pipeline_mock.await_args.kwargs["mode"] == OrchestratorMode.STANDARD
+
+
+@pytest.mark.asyncio
+async def test_execute_job_fails_on_unknown_mode(session_factory, tmp_path):
+    """An unrecognized job mode explicitly fails the job rather than running partial work."""
+    from repowise.core.persistence.crud import get_generation_job
+
+    async with session_factory() as session:
+        repo = await upsert_repository(session, name="test-repo", local_path=str(tmp_path))
+        job = await upsert_generation_job(
+            session,
+            repository_id=repo.id,
+            config={"mode": "incremental"},
+        )
+        await session.commit()
+        job_id = job.id
+
+    app_state = SimpleNamespace(session_factory=session_factory, fts=None, vector_store=None)
+
+    run_pipeline_mock = AsyncMock(return_value=_fake_result())
+    with (
+        patch("repowise.server.job_executor.run_pipeline", run_pipeline_mock),
+        patch("repowise.server.job_executor.persist_pipeline_result", AsyncMock()),
+    ):
+        await execute_job(job_id, app_state)
+
+    # Pipeline should not run when mode is invalid
+    run_pipeline_mock.assert_not_called()
+
+    # Job status should be recorded as failed in the DB with clear error message
+    async with session_factory() as session:
+        updated_job = await get_generation_job(session, job_id)
+        assert updated_job is not None
+        assert updated_job.status == "failed"
+        assert "Invalid job mode 'incremental'" in (updated_job.error_message or "")
+
+
+@pytest.mark.asyncio
+async def test_execute_job_defaults_to_sync_when_mode_absent(session_factory, tmp_path):
+    """A missing mode key (config={}) defaults to 'sync' and proceeds without failing."""
+    async with session_factory() as session:
+        repo = await upsert_repository(session, name="test-repo", local_path=str(tmp_path))
+        job = await upsert_generation_job(
+            session,
+            repository_id=repo.id,
+            config={},
+        )
+        await session.commit()
+        job_id = job.id
+
+    app_state = SimpleNamespace(session_factory=session_factory, fts=None, vector_store=None)
+
+    run_pipeline_mock = AsyncMock(return_value=_fake_result())
+    get_provider_mock = MagicMock(side_effect=RuntimeError("no provider"))
+    with (
+        patch("repowise.server.job_executor.run_pipeline", run_pipeline_mock),
+        patch("repowise.server.job_executor.persist_pipeline_result", AsyncMock()),
+        patch(
+            "repowise.server.provider_config.get_chat_provider_instance",
+            get_provider_mock,
+        ),
+    ):
+        await execute_job(job_id, app_state)
+
+    # Pipeline must have run
+    run_pipeline_mock.assert_awaited_once()
+    # Provider lookup was attempted — proves fallback is "sync", not "index_only"
+    get_provider_mock.assert_called_once()
+
+
+@pytest.mark.asyncio
+async def test_execute_job_defaults_to_sync_when_mode_empty_string(session_factory, tmp_path):
+    """An explicit empty-string mode also defaults to 'sync' via the `or` fallback."""
+    async with session_factory() as session:
+        repo = await upsert_repository(session, name="test-repo", local_path=str(tmp_path))
+        job = await upsert_generation_job(
+            session,
+            repository_id=repo.id,
+            config={"mode": ""},
+        )
+        await session.commit()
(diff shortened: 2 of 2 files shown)


packages/server/src/repowise/server/job_executor.py
```diff
@@ -341,6 +345,12 @@ async def execute_job(
             wiki_style = _repo_wiki_style(repo, repo_path)
             config = json.loads(job.config_json) if job.config_json else {}
             mode = config.get("mode") or "sync"
```
swati510: This is a small robustness point rather than a real bug. The mode comes out of parsed JSON, so it is not guaranteed to be a string, and because `VALID_JOB_MODES` is a set, a non-empty list or dictionary would make the membership test on the next line raise an unhashable-type error instead of the helpful message you have written. The job would still be recorded as failed, but the message stored in the database would be confusing to whoever reads it later. Would you consider coercing the value when you read it, for example `mode = str(config.get("mode") or "sync")`, so that the friendly message is always the one that gets recorded?

packages/server/src/repowise/server/job_executor.py
```diff
@@ -106,6 +106,10 @@ def _repo_wiki_style(repo: Any, repo_path: str) -> str:
     return resolve_style(style, repo_path=repo_path).name
 
 
+# Valid job execution modes handled by execute_job
```
swati510: Thank you for naming the accepted modes in one place. One thing that might be worth capturing here in a comment: `routers/pages.py` line 121 creates job rows with the mode "single_page", and those rows are never handed to `execute_job`, which is why they are not in this set. Without that note, a future reader who wires up the page regeneration endpoint will hit an "invalid job mode" failure and have to rediscover the history. This is only a comment suggestion, since the linked issue says the dormant `single_page` endpoint deserves its own ticket.

tests/unit/server/test_job_executor.py
```diff
@@ -326,3 +326,68 @@ async def test_incremental_page_regen_passes_repo_path(tmp_path):
...
+    async with session_factory() as session:
+        updated_job = await get_generation_job(session, job_id)
+        assert updated_job is not None
+        assert updated_job.status == "failed"
+        assert "Invalid job mode 'incremental'" in (updated_job.error_message or "")
+
+
+@pytest.mark.asyncio
+async def test_execute_job_defaults_to_sync_when_mode_absent(session_factory, tmp_path):
+    """A missing or empty mode string defaults to 'sync' and proceeds without failing."""
```
swati510: The description says "a missing or empty mode string", but the test only covers the missing case, since the configuration is an empty dictionary. The empty-string case is the one that actually exercises the `or "sync"` fallback, and it is the case most likely to break in a future refactor of that line. Could you either add a second case with `config={"mode": ""}` or adjust the description so that it matches what is being tested?

tests/unit/server/test_job_executor.py
```diff
@@ -326,3 +326,68 @@ async def test_incremental_page_regen_passes_repo_path(tmp_path):
...
+        patch("repowise.server.job_executor.run_pipeline", run_pipeline_mock),
+        patch("repowise.server.job_executor.persist_pipeline_result", AsyncMock()),
+        patch(
+            "repowise.server.provider_config.get_chat_provider_instance",
+            side_effect=RuntimeError("no provider"),
+        ),
+    ):
+        await execute_job(job_id, app_state)
+
+    run_pipeline_mock.assert_awaited_once()
```
swati510: This assertion shows that the pipeline ran, but it would pass for any accepted mode, so it does not quite prove that the fallback is "sync". Because the provider lookup is forced to fail in this test, the behaviour that is specific to "sync", namely the incremental page regeneration, never runs, so nothing here distinguishes "sync" from, say, "index_only". Checking the `generate_docs` keyword would not help either, since it is computed as false for "sync" and for "index_only" alike. What would distinguish them is that "sync" attempts the provider lookup while "index_only" skips that block entirely, so asserting that the patched `get_chat_provider_instance` was called would pin the mode down. Another option would be to let the patched provider return a mock client and then assert that the sync-only incremental page regeneration ran.

packages/server/src/repowise/server/job_executor.py
akshatmalik-bruh: Done! Changed to mode = str(config.get("mode") or "sync") so any non-string JSON value (e.g. a list or dict) is coerced before the set membership test. This ensures the friendly ValueError with the invalid mode message is always the one recorded in the DB, rather than a confusing unhashable type error.

packages/server/src/repowise/server/job_executor.py
akshatmalik-bruh: Added a comment explaining that "single_page" is intentionally excluded — routers/pages.py (line 121) creates those job rows but they are never dispatched to execute_job. Also noted the dormant endpoint deserves its own dedicated handling. Thanks for flagging — this will save future readers a confusing debugging session!

tests/unit/server/test_job_executor.py
akshatmalik-bruh: Good point! I've split this into two separate tests: test_execute_job_defaults_to_sync_when_mode_absent (covers config={}) and a new test_execute_job_defaults_to_sync_when_mode_empty_string (covers config={"mode": ""}) which directly exercises the or "sync" fallback branch. Also updated the docstring of the original test to only say "missing" instead of "missing or empty".

tests/unit/server/test_job_executor.py
akshatmalik-bruh: Great catch! I've captured the get_chat_provider_instance mock as get_provider_mock and added get_provider_mock.assert_called_once() after the run. Since "sync" always enters the provider lookup block while "index_only" skips it entirely, this assertion specifically pins the fallback mode down to "sync". Applied the same pattern to both the absent-mode and empty-string tests.

akshatmalik-bruh:
Hi @swati510 @RaghavChamadiya  , just following up on this PR when you get a chance to take a look. Thanks!

akshatmalik-bruh:
Resolved merge conflicts with main and verified tests pass. Ready for review whenever you get a chance!

RaghavChamadiya:
Thanks for the quick turnaround on Swati's points. The coercion to `str(...)`, the split empty-string test, and the `get_provider_mock.assert_called_once()` assertion all address exactly what was raised, and that last one is a genuinely better test than what it replaced.

One blocking item, and it comes from the merge rather than from anything you wrote.

**`generate` and `single_page` are now dispatched modes, and this allowlist rejects them.** Since #881 was filed, `execute_job` gained a real branch for both:

```python
# job_executor.py:407 on this branch
if mode in ("generate", "single_page"):
    ...
    await _run_generate_job(...)
    return
```

Your merge commit pulled that in, but `VALID_JOB_MODES` (line 115) was not updated, and the validation at line 353 runs before the dispatch at 407. Both modes reach the executor for real:

- `routers/repos.py:667` builds `{"mode": "generate", ...}`, launched via `_launch_job_task`. That is `POST /repos/{id}/generate`, the HTTP `repowise generate`.
- `routers/pages.py:147` builds `{"mode": "single_page", ...}`, same path.

So after merge every generate and every per-page regenerate fails with `Invalid job mode 'generate'`. The endpoint that the issue described as dormant is live now.

To be clear about where this came from: the `single_page` note was Swati's suggestion on 2026-07-20 and it was correct then, and my own issue text said the same thing. The dispatch landed in between.

The fix:

1. Add `"generate"` and `"single_page"` to `VALID_JOB_MODES`.
2. Replace the exclusion note with the opposite, something like: these two are handled by the scoped-generation branch below, so validation has to run before it and accept them.
3. Add a test that a `{"mode": "generate"}` job still reaches `_run_generate_job`. That is the regression guard, and its absence is why CI stays green on this: nothing currently drives a generate job through `execute_job`.

One process note. The branch is 5 commits behind main again, and this is the second time the gap has hidden something. Could you rebase onto main rather than merging? A rebase would have put your validation line directly against the new dispatch branch and made the conflict visible instead of silently compatible. Worth a look at whether anything else moved under the PR while you were waiting.

Everything else here is ready. Fix the allowlist and I think this goes in.

akshatmalik-bruh:
Thanks for catching that @RaghavChamadiya ! All items have been updated:

1. **Updated `VALID_JOB_MODES` Allowlist**: Added `"generate"` and `"single_page"` to `VALID_JOB_MODES` in `job_executor.py`.
2. **Updated Documentation**: Updated the inline comment above `VALID_JOB_MODES` to document that `"generate"` and `"single_page"` are valid modes dispatched to `_run_generate_job`.
3. **Regression Guard Test**: Added `test_execute_job_dispatches_generate_mode` in `tests/unit/server/test_job_executor.py` to ensure jobs with `mode="generate"` pass validation and reach `_run_generate_job`.
4. **Rebased onto Main**: Rebased the branch cleanly on top of latest `main`. All 23 unit tests in `test_job_executor.py` pass cleanly.

RaghavChamadiya (approved):
LGTM

Issue #930, opened 2026-07-19
[Feature] Extend quiet-by-default logging plus -v to the remaining CLI commands


RaghavChamadiya:
## Problem

#929 made `repowise init` and `repowise update` quiet by default. Before that, `update` printed provider and pipeline debug lines (`ollama.generate.done`, ingestion graph/traverser events) straight into the live Rich progress bar, which is what produced the duplicated "Generating pages..." rows in #922.

That PR deliberately scoped itself to those two commands. Every other command still runs with whatever logging configuration it happens to inherit, so any command that touches the core pipeline or an LLM/embedding provider can spray debug output over its own progress display. Only `init` and `update` define a `-v` flag today:

```
packages/cli/src/repowise/cli/commands/init_cmd/command.py:499
packages/cli/src/repowise/cli/commands/update_cmd/command.py:260
```

## Proposed Solution

Give the remaining long-running commands the same treatment: a `--verbose / -v` flag wired to the helper #929 already added.

The helper lives in `packages/cli/src/repowise/cli/_setup.py`:

```python
configure_cli_logging(verbose: bool = False) -> None
```

- Default: `httpx`, `httpcore`, `repowise.core` and `repowise.server` pinned to `ERROR`, so progress bars are the only output.
- `verbose=True`: repowise's own loggers pass through at `DEBUG` so a user can watch provider/pipeline activity. `httpx`/`httpcore` stay at `ERROR` either way, since their debug stream is HTTP-level noise that buries the useful lines.

`setup_logging_silence()` is a thin wrapper over the quiet path, for call sites that never want logs (the multi-repo workspace init flow).

Per command:

1. Add the flag, matching the existing wording:
   ```python
   @click.option(
       "--verbose",
       "-v",
       is_flag=True,
       default=False,
       help="Show debug logs from the pipeline.",
   )
   ```
2. `from repowise.cli._setup import configure_cli_logging`
3. Call `configure_cli_logging(verbose=verbose)` as the first statement in the command body, before any pipeline import or work. Reference call site: `update_cmd/command.py:387`.

Candidates, roughly in priority order. Each renders progress or calls a provider:

| Command | File | Why it matters |
|---|---|---|
| `reindex` | `commands/reindex_cmd.py` | Rich `Progress` at line 118, embedding calls per batch |
| `restyle` | `commands/restyle_cmd.py` | Regenerates pages through the LLM provider, `--concurrency 12` |
[...]

PR #932, opened 2026-07-19
feat(cli): add --verbose to reindex quiet logging

Solaris-star:
## Summary

- Add `--verbose` / `-v` to `repowise reindex`, wired to `configure_cli_logging` (same helper as `init` / `update`)
- Default stays quiet so Rich progress bars are not flooded by library/structlog debug lines
- Unit test asserts the flag reaches `configure_cli_logging` before pipeline work

## Related Issues

Part of #930 (reindex slice only; other commands left for follow-ups as the issue suggests)

## Test Plan

- [x] `pytest tests/unit/cli/test_reindex_cmd.py` (2 passed)
- [ ] Lint passes (`ruff check .`) *(not run full tree this PR)*
- [ ] Web build N/A

## Checklist

- [x] My code follows the project's code style
- [x] I have added tests for new functionality
- [x] Existing reindex tests still pass
- [x] Documentation N/A (flag help text added on the option)

PR #933, opened 2026-07-19
feat(cli): add --verbose to export quiet logging

Solaris-star:
## Summary
- Add `--verbose` / `-v` to `repowise export`, wired to `configure_cli_logging` so progress bars stay clean by default (same pattern as `init`/`update`/`reindex`).
- Unit test asserts the flag reaches `configure_cli_logging` before pipeline work.

Part of #930 (export slice only).

## How tested
```bash
.venv/bin/python -m pytest tests/unit/cli/test_export_cmd.py -q
# 1 passed
```

### DIFF
--- a/docs/CLI_REFERENCE.md
+++ b/docs/CLI_REFERENCE.md
@@ -668,6 +668,8 @@ Export wiki pages to files.
 | `--format` | `markdown` (default), `html`, `json` |
 | `--output / -o` | Output directory (default: `.repowise/export`) |
 | `--full` | Include decisions, dead code, hotspots, provenance metadata (JSON only) |
+| `--verbose`, `-v` | Show debug logs from the pipeline |
+
 
 ```bash
 repowise export
--- a/packages/cli/src/repowise/cli/commands/export_cmd.py
+++ b/packages/cli/src/repowise/cli/commands/export_cmd.py
@@ -8,6 +8,7 @@ from pathlib import Path
 import click
 from rich.progress import Progress
 
+from repowise.cli._setup import configure_cli_logging
 from repowise.cli.helpers import (
     console,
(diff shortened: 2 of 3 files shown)


RaghavChamadiya (changes requested):
Keeping this one since you claimed the slice first. Same three items as #932: route the test through `CliRunner` rather than `.callback()`, add the flag to the `repowise export` options table in `docs/CLI_REFERENCE.md` (around line 657), and run `ruff format .` (it reformats both files here).

@Narasimha2211 opened #934 for the same command with a test that already does the `CliRunner` version and covers the quiet default. Please lift that test shape and credit them in the commit body.

Solaris-star:
Same three items addressed:
1. Test now uses `CliRunner().invoke(cli, ["export", ..., "-v"])` instead of `.callback()`
2. Added `--verbose/-v` to the export options table in `docs/CLI_REFERENCE.md`
3. Ran `ruff format`

Lifted the test shape from #934 (credited in commit body). Please re-review when convenient.

RaghavChamadiya:
@Solaris-star same as #932, and the same apology.

You addressed all three review items here on 24 July, the test switched to `CliRunner().invoke`, the flag added to the export options table in `docs/CLI_REFERENCE.md`, and the formatter run. Nobody answered, and the PR was closed unmerged on 12 August without a reason given.

I checked current main today. `export_cmd.py` and `reindex_cmd.py` are the only two of the nine slices in #930 that still have no `verbose`. Every other slice landed. Yours are outstanding because we dropped them, not because they were wrong.

I cannot reopen this one either: the source fork has been deleted, so the branch is gone and there is nothing for GitHub to reopen against. I will carry both slices myself. If you would rather redo them on a fresh branch they are yours, just say so.

PR #934, opened 2026-07-19
Add -v/--verbose flag to `repowise export`

Narasimha2211:
## Summary
#929 made `init`/`update` quiet by default via `configure_cli_logging()`, scoped to just those two commands. This wires the same `-v` flag into `export`, one command off the remaining-commands list in #930.

`reindex` was already claimed by another contributor when I picked this up — the issue explicitly calls out that picking a single command is a reasonable scope for one PR ("This is splittable... Picking a single command off this list is a reasonable first contribution").

## Changes
- `export_cmd.py`: adds `--verbose/-v`, calls `configure_cli_logging(verbose=verbose)` as the first statement in the command body (matching the `update_cmd` reference call site), following the exact pattern from the issue.
- `tests/unit/cli/test_export_cmd.py`: asserts the flag reaches `configure_cli_logging` with the right value, both with and without `-v`.

## How I tested this
- `pytest tests/unit/cli/test_export_cmd.py -v` — 2 passed
- Full `pytest tests/unit/cli/` — 804 passed (no regressions)
- `ruff check` / `ruff format --diff` on changed files — clean

## Related issue
Part of #930 (export command only; other commands on that list are left for follow-up PRs, per the issue's own suggestion).

### DIFF
--- a/packages/cli/src/repowise/cli/commands/export_cmd.py
+++ b/packages/cli/src/repowise/cli/commands/export_cmd.py
@@ -8,6 +8,7 @@ from pathlib import Path
 import click
 from rich.progress import Progress
 
+from repowise.cli._setup import configure_cli_logging
 from repowise.cli.helpers import (
     console,
     ensure_repowise_dir,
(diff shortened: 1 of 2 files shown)


RaghavChamadiya:
Closing as a duplicate: @Solaris-star claimed `export` on #930 a few minutes before this opened, and #933 covers the same ground. No reflection on the work, your test is the better of the two, and I have asked #933 to adopt its shape with credit to you.

Sorry for the collision. `augment` is the one row in the issue table nobody has claimed yet if you want it, and dropping a claim comment on #930 first will avoid a repeat.

PR #935, opened 2026-07-19
fix(server): pass --no-color to data_shape git grep (#858)

### LINKED ISSUE #858: [Bug] data_shape live-grep: add --no-color so exclusion filter survives color.ui=always

gitcommit90:
## Summary
Pass `--no-color` to `data_shape` live-grep so exclusion filters survive `color.ui=always`.

## Changes
- Add `--no-color` to `_run_grep` git grep args (matches `_code_rationale.py` / #825)
- Regression test with `color.ui=always`

Closes #858

### DIFF
--- a/packages/server/src/repowise/server/mcp_server/tool_answer/data_shape.py
+++ b/packages/server/src/repowise/server/mcp_server/tool_answer/data_shape.py
@@ -156,6 +156,10 @@ def _run_grep(
                 "git",
                 "--no-pager",
                 "grep",
+                # --no-color: this output is parsed, so a user's
+                # ``color.ui=always`` git config must not wrap paths in ANSI
+                # escapes (which would corrupt exclusion filtering and reads).
+                "--no-color",
                 *args,
                 "-l",
                 "-I",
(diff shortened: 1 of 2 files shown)


swati510:
Thank you for picking this up, and thank you especially for carrying over the explanatory comment from the `_code_rationale.py` fix in #825. The source change here is exactly right, and I verified by hand that `--no-color` suppresses the escape codes both for `color.ui = always` and for the narrower `color.grep = always`.

I am going to land #860 instead, because its regression test asserts on the higher-level `_grep_identifier_files` function rather than on the raw subprocess output, which pins down the behaviour that the callers actually depend on. I would like to bring your comment across to that pull request, though, because a bare flag with no explanation is very easy for a future tidy-up to delete, so your work here is not going to waste. Thank you again for the careful fix and for the regression test.

RaghavChamadiya:
Thanks @gitcommit90, and apologies for leaving this hanging after the review said #860 would land instead. #860 merged on 21 July and this can close.

Your change is on main, comment and all:

```python
"grep",
# --no-color: this output is parsed, so a user's
# ``color.ui=always`` git config must not wrap paths in ANSI
# escapes (which would corrupt exclusion filtering and reads).
"--no-color",
```

`packages/server/src/repowise/server/mcp_server/tool_answer/data_shape.py:207-210`.

I also went looking for the same trap elsewhere while I was in there, since a single-site fix for a config-driven bug usually is not one. The server has exactly two `git grep` calls, this one and the one in `_code_rationale.py:387`, and both carry `--no-color` now. The other git subprocesses in the server (`git log --grep` in `tool_why.py`, the calls in `job_executor.py`, `routers/git.py`, `scheduler.py`) do not parse colorized grep output, so they are not exposed to this.

Closing as already landed. The diagnosis was right and it is what got fixed.

PR #941, opened 2026-07-19
feat(cli): add verbose logging to watch

lntutor:
## Summary

- add `--verbose` / `-v` to `repowise watch`
- configure logging before target resolution for both single-repo and workspace modes
- propagate verbose mode into nested single-repo `update` invocations so it is not reset to quiet
- document the flag and add focused quiet/verbose forwarding coverage
- clean up two unused imports and correct the touched file’s invalid `callable` type annotation

## Related Issues

Part of #930 (`watch` slice).

## Test Plan

- [x] focused watch and command tests (32 passed)
- [x] CLI unit suite (804 passed, 1 unrelated baseline failure)
- [x] Ruff lint and format checks on modified Python files
- [x] strict mypy check for `watch_cmd.py`
- [x] `git diff --check`

The sole CLI-suite failure is the existing mascot tagline assertion `test_banner_at_wide_width_stays_compact_with_long_tagline`; it reproduces unchanged on clean `origin/main`.

## Checklist

- [x] My code follows the project style
- [x] I added tests for the new functionality
- [x] I updated the CLI reference
- [ ] All existing tests pass (see unrelated baseline failure above)

### DIFF
--- a/docs/CLI_REFERENCE.md
+++ b/docs/CLI_REFERENCE.md
@@ -282,12 +282,14 @@ Watch for file changes and auto-update wiki pages. Press `Ctrl+C` to stop.
 | `--workspace` / `-w` | Watch all workspace repos |
 | `--no-workspace` | Force single-repo mode |
 | `--repo` | Watch a single workspace repo by alias |
+| `--verbose` / `-v` | Show debug logs from the pipeline and triggered updates |
 
 ```bash
 repowise watch                           # single repo (auto-detects)
 repowise watch --debounce 5000           # 5s debounce
 repowise watch --workspace               # all workspace repos
 repowise watch --repo backend            # just one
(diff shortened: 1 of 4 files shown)


RaghavChamadiya (approved):
Approving. Forwarding `--verbose` into the nested `update_command` invocation is necessary, not decorative: that child run re-enters `configure_cli_logging` and would otherwise reset to quiet.

Two notes, neither blocking. The `find_workspace_root` / `resolve_repo_path` import removal is correct, they were genuinely unused, but the drive-by formatting churn in `_watch_workspace` makes the diff harder to read than it needs to be. And for the record, `_watch_workspace` not taking `verbose` is fine: it calls `update_workspace` in-process, so the process-wide config already applies. Worth knowing so nobody "fixes" it later.

PR #943, opened 2026-07-19
feat(cli): add verbose logging to health

lntutor:
Part of #930

## Summary

- add `--verbose/-v` to `repowise health`
- configure CLI logging before target resolution and analysis/provider work
- preserve stricter log silencing for JSON and Markdown output
- document the new flag and cover quiet/verbose forwarding

## Tests

- `uv run pytest -q tests/unit/cli/test_health_cmd_logging.py tests/unit/cli/test_analysis_cmds_include_submodules.py` (6 passed)
- `uv run pytest -q tests/unit/cli` (803 passed, 1 pre-existing mascot assertion failure reproduced on clean `origin/main`)
- targeted Ruff lint and formatting checks
- targeted mypy (`--follow-imports=skip`)

Repository-wide lint/type checking currently reports unrelated baseline issues, including existing Ruff violations and two mypy errors in `health_cmd/refactoring_targets.py`.

### DIFF
--- a/docs/CLI_REFERENCE.md
+++ b/docs/CLI_REFERENCE.md
@@ -459,6 +459,7 @@ Compute per-file code-health scores from 25 deterministic markers (McCabe comple
 | `--safe-only` | Documented no-op placeholder for a future confidence filter; has no effect today |
 | `--repo` | In workspace mode, target a specific repo (defaults to primary) |
 | `--no-workspace` | Force single-repo mode |
+| `--verbose`, `-v` | Show debug logs from the analysis pipeline |
 
 ```bash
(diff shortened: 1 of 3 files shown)


RaghavChamadiya (approved):
Approving. Docs updated, the test drives the real Click parser, and it asserts ordering rather than just arrival, which is the part that actually matters here.

Calling out one thing for anyone reading this later: `configure_cli_logging` lands before the existing `silence_logs_for_machine_output()`, so `repowise health --format json | jq` stays pipe-safe under `-v`. Reversing those two would let structlog into stdout and break the JSON. Good catch.

PR #944, opened 2026-07-19
feat(cli): add verbose logging to workspace

lntutor:
Part of #930

## Summary

- add `--verbose/-v` to `workspace add` and `workspace scan`
- configure CLI logging before workspace/path resolution, indexing, docs generation, or scanning
- retain quiet-by-default behavior while exposing pipeline diagnostics on demand
- document both flags and add command-boundary ordering tests

The nested indexing and docs helpers do not reconfigure logging, so boundary configuration carries through without an extra propagation parameter.

## Tests

- `uv run pytest -q tests/unit/cli/test_workspace_logging.py tests/unit/cli/test_workspace_cmd.py tests/unit/cli/test_workspace_add_defaults.py` (56 passed)
- `uv run pytest -q tests/unit/cli` (805 passed, 1 pre-existing mascot assertion failure already reproduced on clean `origin/main`)
- targeted Ruff lint and formatting checks
- targeted mypy (`--follow-imports=skip`)

### DIFF
--- a/docs/CLI_REFERENCE.md
+++ b/docs/CLI_REFERENCE.md
@@ -697,6 +697,7 @@ This defaults to `--index --docs` when a provider is configured, the added repo 
 | `--provider` / `--model` | Override the inherited provider/model |
 | `--concurrency` | Max concurrent LLM calls for this repo's generation |
 | `--primary` | Mark this repo as the workspace default |
+| `--verbose`, `-v` | Show debug logs from indexing and doc generation |
 
 ```bash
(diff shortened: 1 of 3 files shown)


RaghavChamadiya (approved):
Approving the code. One correction to the description: it says the mode is propagated into the nested indexing and docs-generation calls, but the diff does not do that, and it does not need to. `_run_index_for_repo` and `_generate_docs_for_added_repo` run in the same process, so the boundary call already covers them. Could you reword that line so the PR history does not imply plumbing that is not there?

PR #952, opened 2026-07-20
fix(workspace): scope file count and top-language queries to node_type='file'

### LINKED ISSUE #945: [Bug] Workspace page reports far more files than the repo contains
## Describe the Bug

In workspace mode, the `/workspace` page reports a file count roughly 10x higher than the actual number of indexed files. This affects the per-repo card, the "Total Files" figure in the header (a sum of the per-repo numbers), and, less visibly, the "top language" badge.

## Steps to Reproduce

1. Index a workspace with `repowise init` (two or more repos configured under `.repowise-workspace/`)
2. Run `repowise serve` and open `http://localhost:3000/workspace`
3. Compare the file count on any repo card against the number of files that repo actually has

## Expected Behavior

The repo card reports the number of indexed files, and "Total Files" is the sum of those.

## Actual Behavior

The number is inflated by roughly an order of magnitude. On one repo with ~3,000 indexed files, the card reads ~29,000.

The top-language badge looks plausible but is computed the same way, so it can also be wrong: it's possible for a language with fewer files to win on a repo where its files are unusually dense.

## Environment

- OS: any
- Repowise version: 0.33.0
- Component: `packages/server`, `/api/workspace` — workspace mode only (single-repo mode is unaffected)

## Additional Context

Both numbers are computed by helpers in `packages/server/src/repowise/server/routers/workspace.py`, which read a repo's `.repowise/wiki.db` directly over sqlite3. The hint: the `graph_nodes` table doesn't hold one row per file, it stores more than one kind of node, and the counting here doesn't account for that. Inspecting that table next to the numbers on the page should confirm it quickly. Note that the two helpers share the bug, so a fix to one is only half of it.

One thing to be careful of: there's an unfiltered count over the same table in `packages/core/src/repowise/core/persistence/coordinator.py` that is *not* part of this bug, it's a diagnostic that genuinely wants what it's counting. Leave it as is.

Tests live in `tests/unit/server/test_workspace_router.py`. The existing fixtures there don't currently model the distinction that causes this, so they'll need extending before a regression test can catch it.

Nehant14:
Fixes #945

## What's wrong

`graph_nodes` stores both `file` and `symbol` rows, but `_query_repo_stats()` and `_query_top_language()` in `packages/server/src/repowise/server/routers/workspace.py` counted/grouped over the whole table. That inflated the per-repo file count (~10x on real repos), the "Total Files" header total (a sum of those numbers), and could skew the top-language pick toward a language with fewer files but denser symbol counts.

## The fix

Added `WHERE node_type = 'file'` to both queries. Minimal, scoped change — no schema or API shape changes.

The unfiltered `COUNT(*)` diagnostic in `packages/core/src/repowise/core/persistence/coordinator.py` is intentionally left untouched, per the issue.

## Testing

- Extended fixtures in `tests/unit/server/test_workspace_router.py` to model file + symbol rows together
- Added two regression tests: `test_file_count_excludes_symbol_nodes`, `test_top_language_excludes_symbol_nodes`
- Full suite: `tests/unit/server/test_workspace_router.py` → 41/41 passing
- Manually

### DIFF
--- a/packages/server/src/repowise/server/routers/workspace.py
+++ b/packages/server/src/repowise/server/routers/workspace.py
@@ -65,7 +65,7 @@ def _query_top_language(db_path: Path) -> str:
         with sqlite3.connect(str(db_path)) as conn:
             row = conn.execute(
                 "SELECT language, COUNT(*) AS cnt FROM graph_nodes "
-                "WHERE language IS NOT NULL AND language != '' "
+                "WHERE node_type = 'file' AND language IS NOT NULL AND language != '' "
                 "GROUP BY language ORDER BY cnt DESC LIMIT 1"
             ).fetchone()
             return row[0] if row else "unknown"
@@ -143,8 +143,8 @@ def _query_repo_stats(db_path: Path) -> dict:
         if row:
             result["repo_id"] = row[0]
 
-        # file count (graph_nodes)
-        row = c.execute("SELECT COUNT(*) FROM graph_nodes").fetchone()
+        # file count (graph_nodes) 
+        row = c.execute("SELECT COUNT(*) FROM graph_nodes WHERE node_type = 'file'").fetchone()
         result["file_count"] = row[0] if row else 0
 
         # symbol count
--- a/tests/unit/server/test_workspace_router.py
+++ b/tests/unit/server/test_workspace_router.py
@@ -176,6 +176,7 @@ def _create_workspace_repo_db(
             CREATE TABLE repositories (id TEXT PRIMARY KEY);
             CREATE TABLE graph_nodes (
                 id TEXT PRIMARY KEY,
+                node_type TEXT NOT NULL DEFAULT 'file',
                 language TEXT,
                 symbol_count INTEGER DEFAULT 0
             );
@@ -191,9 +192,12 @@ def _create_workspace_repo_db(
                 nloc INTEGER NOT NULL
             );
             INSERT INTO repositories (id) VALUES ('repo-backend');
-            INSERT INTO graph_nodes (id, language, symbol_count) VALUES
-                ('src/a.py', 'python', 2),
-                ('src/b.py', 'python', 3);
+            INSERT INTO graph_nodes (id, node_type, language, symbol_count) VALUES
+                ('src/a.py', 'file', 'python', 2),
+                ('src/b.py', 'file', 'python', 3),
+                ('src/a.py::Foo', 'symbol', 'python', 0),
+                ('src/a.py::Foo.bar', 'symbol', 'python', 0),
+                ('src/b.py::baz', 'symbol', 'python', 0);
             INSERT INTO wiki_pages (id, confidence) VALUES
                 ('page-a', 0.8),
                 ('page-b', 0.6);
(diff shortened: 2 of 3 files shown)


Nehant14:
Both CI failures are already present on main and are not caused by this PR.

- The Python 3.11/3.12 failure happens because the get_change_risk docstring (from #946) is longer than the allowed limit.
- The test_workspace_sidebar.py failure with "AttributeError: '_IncludedRouter' object has no attribute 'path'" also occurs on main.

These were confirmed by checking out main and running the same tests. Happy to rebase once these issues are fixed on main.

packages/server/src/repowise/server/routers/workspace.py
```diff
@@ -143,8 +143,8 @@ def _query_repo_stats(db_path: Path) -> dict:
         if row:
             result["repo_id"] = row[0]
 
-        # file count (graph_nodes)
-        row = c.execute("SELECT COUNT(*) FROM graph_nodes").fetchone()
+        # file count (graph_nodes) 
+        row = c.execute("SELECT COUNT(*) FROM graph_nodes WHERE node_type = 'file'").fetchone()
```
swati510, line 147: Thank you for this fix, and thank you especially for extending the fixtures so that the tests actually model the mixture of file and symbol rows. I checked this against a real index and the ratio matches what the issue described: 3,010 rows of type `file` against 25,824 of type `symbol`.

One small thing worth knowing about rather than necessarily fixing here: rows that stand for third-party imports are also stored with `node_type` set to `file`. Their node identifiers begin with `external:`, for example `external:node:path`, and they end up typed as files because the writer in `packages/core/src/repowise/core/pipeline/persist.py` defaults the type to `file` when none was recorded, and the code that creates these nodes never records one. In this repository that is 79 rows out of 3,010, so the count will still be a little under three percent high. The equivalent single-repository code in `routers/stats.py` filters on `node_type == "file"` in exactly the same way and therefore has exactly the same residual, so I would rather we did not change one without the other. Would you be willing to open a small follow-up issue covering both places, so the two pages stay consistent?

tests/unit/server/test_workspace_router.py
```diff
@@ -570,6 +591,66 @@ def test_hotspot_count_zero_when_no_hotspots(self, tmp_path: Path) -> None:
 
         assert stats["hotspot_count"] == 0
 
+    
+    def test_file_count_excludes_symbol_nodes(self, tmp_path: Path) -> None:
```
swati510, line 595: These two tests are clear and they do genuinely fail against the old code, which I verified by restoring the previous version of `workspace.py` and re-running the file.

One optional addition: both of them call the helper functions directly, and no test asserts the file count on an endpoint response, so the number a user actually sees is only covered indirectly, through the health score assertion in `test_marks_derived_health_score_when_repo_metrics_are_missing` on `GET /api/workspace/graph`. If it is easy, an assertion that `GET /api/workspace/graph` reports `file_count == 2` for the updated fixture would make that coverage explicit rather than accidental.

Nehant14: Thanks for the suggestion! I added the endpoint-level assertion so `GET /api/workspace/graph` now explicitly verifies the expected `file_count` from the updated fixture.

RaghavChamadiya (approved):
LGTM

PR #962, opened 2026-07-20
feat(codex): add code-health skill for Codex plugin parity

Ayush7614:
## Summary
- Add `plugins/codex/skills/code-health/SKILL.md`, ported from the Claude plugin with Codex-neutral wording (`repowise init` instead of `/repowise:init`, no Claude-specific references).
- Bump Codex plugin version to `0.2.1`.
- Extend `test_codex_plugin.py` to expect six skills including `code-health`.

## Motivation
The Claude Code plugin ships six skills; Codex had five — `code-health` was the only gap. This brings parity so Codex users get the same health/refactoring guidance.

## Test plan
- [x] `pytest tests/unit/cli/test_codex_plugin.py` — all 5 tests pass
- [x] Skill frontmatter validates (single-line description, no Claude/`/repowise:` references)

### DIFF
--- a/plugins/codex/.codex-plugin/plugin.json
+++ b/plugins/codex/.codex-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "codex",
-  "version": "0.2.0",
+  "version": "0.2.1",
   "description": "Repowise codebase intelligence for Codex.",
   "author": {
     "name": "Repowise",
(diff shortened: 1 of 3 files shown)


RaghavChamadiya (approved):
Clean port, thanks. I diffed the new skill against the Claude source and the only differences are the ones that should differ: single-line description for the Codex frontmatter, the Claude-only `user-invocable` field dropped, and `/repowise:init` rewritten to `repowise init`. No Claude or `/repowise:` references left, so the neutrality rule holds.

Version bump and the extended skill-set assertion both look right, and the Codex tests pass locally (5/5). Merging.

PR #963, opened 2026-07-20
test(claude-plugin): add structural parity tests mirroring test_codex_plugin

Ayush7614:
## Summary
- Add `tests/unit/cli/test_claude_plugin.py` with five structural tests mirroring `test_codex_plugin.py`.
- Cover manifest sanity, MCP config, hooks vs installer matchers, all six skills with valid frontmatter, and marketplace version sync.

## Motivation
The Codex plugin already has structural regression tests; the Claude Code plugin had none. These tests catch drift in bundled hooks, skills, and marketplace metadata before it ships.

## Test plan
- [x] `pytest tests/unit/cli/test_claude_plugin.py` — all 5 tests pass

### DIFF
--- a/tests/unit/cli/test_claude_plugin.py
+++ b/tests/unit/cli/test_claude_plugin.py
@@ -1,0 +1,91 @@
+from __future__ import annotations
+
+import json
+import re
+from pathlib import Path
+
+from repowise.cli.editor_integrations import claude_config
+
+ROOT = Path(__file__).resolve().parents[3]
+PLUGIN_ROOT = ROOT / "plugins" / "claude-code"
+
+
+def _load_json(path: Path) -> dict:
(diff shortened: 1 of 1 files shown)


RaghavChamadiya (approved):
Thanks, this is a nice fill for a real gap. What I like is that it isn't a blind copy of the Codex test: it tracks the Claude plugin's actual shape, the `mcpServers`-nested `.mcp.json`, the `PostToolUse`/`SessionStart` hook rows checked against `claude_config._AUGMENT_MATCHER` / `_SESSION_START_MATCHER`, and the marketplace version-sync assertion.

Pulled it and confirmed 5/5 pass locally, and the skill-set assertion matches what's actually on disk. Good regression net for catching plugin drift before it ships. Merging.

PR #967, opened 2026-07-21
feat(parser): add adaptive TSX grammar fallback for .ts files containing JSX

### LINKED ISSUE #965: [Feature] Adaptive TSX grammar fallback for .ts files containing JSX
## Problem

Repowise selects the Tree-sitter grammar for TypeScript files using a **strict file-extension check**:

```python
# packages/core/src/repowise/core/ingestion/parser.py — line 251
grammar_tag = "tsx" if lang == "typescript" and file_info.path.endswith(".tsx") else lang
```

This means any `.ts` file that contains JSX markup is parsed using the plain `typescript` grammar, which does not define JSX node types. The result is `ERROR` nodes in the AST, and the maintainers themselves have already documented the consequence of this in the test suite:

```python
# tests/unit/ingestion/parser/test_typescript.py — lines 133–136
def test_tsx_file_uses_jsx_grammar(self, parser: ASTParser) -> None:
    # .tsx files require the JSX-aware grammar variant; the default
    # typescript grammar errors out on ``<Component />`` and recovers
    # by hoisting nested helpers out of the broken component body.
```

This is not a theoretical concern. Projects that routinely write JSX inside `.ts` files include:

- **Vite** projects configured with `esbuild: { loader: "tsx", include: /\.ts$/ }`
- **Babel** projects with `@babel/preset-typescript` and `{ "isTSX": true, "allExtensions": true }`
- **SolidJS**, **Preact**, and other non-React JSX frameworks where components are co-located in plain `.ts` utility or hook files
- Large JS→TS migration codebases where automated scripts rename `.js` to `.ts` in bulk, while Babel continues treating all files as JSX-capable

### Concrete impact when a `.ts` file contains JSX

1. **Parse errors are generated** — tree-sitter inserts `ERROR` nodes wherever JSX tags appear.
2. **Call-site edges are silently dropped** — the S-expression captures for JSX call targets (`jsx_self_closing_element`, `jsx_opening_element`) live exclusively in `tsx.scm` and are never loaded for `.ts` files. A component used as `<StatRow />` inside a `.ts` file registers zero callers.
3. **False-positive dead-code findings** — because the call-site edges are absent, the dead-code analyser sees `0` inbound callers and can flag live, actively-used components as unreachable.

---

## Proposed Fix

Add a **gated adaptive retry** inside `ASTParser.parse_file` after the first parse attempt. The retry is triggered only when all three conditions hold simultaneously:

1. The initial parse produced `ERROR` nodes (`parse_errors` is non-empty).
2. The file language is `"typescript"` and was not already parsed as `tsx`.
3. The raw source bytes contain JSX-specific closing tokens (`/>` or `</`).

If a re-parse with the `tsx` grammar produces fewer `ERROR` nodes than the original, the TSX tree is adopted. Otherwise the original result is kept unchanged.

```python
# packages/core/src/repowise/core/ingestion/parser.py
# Insert after line ~282 (after parse_errors = _collect_error_nodes(root))

if parse_errors and lang == "typescript" and grammar_tag != "tsx":
    if b"/>" in source or b"</" in source:
        tsx_lang = _get_language("tsx")
        if tsx_lang is not None:
            tsx_tree = Parser(tsx_lang).parse(source)
            tsx_errors = _collect_error_nodes(tsx_tree.root_node)

            # Adopt the TSX tree only when it strictly improves the parse.
            if len(tsx_errors) < len(parse_errors):
                tree = tsx_tree
                root = tree.root_node
                parse_errors = tsx_errors
                grammar_tag = "tsx"
                language = tsx_lang
```

### Why `/>` and `</` as the token guard (not `<`)

Using a bare `b"<"` check is too broad — TypeScript generics (`Array<string>`), comparisons (`a < b`), and type assertions all contain `<`. The tokens `/>` (self-closing JSX element) and `</` (JSX closing tag) have **no valid meaning in standard TypeScript syntax**, so their presence is a reliable signal that the file is intended to contain JSX markup.

> Note: a `.ts` file that contains `</` or `/>` only inside string literals (e.g. `const html = "<div/>"`) will produce **zero tree-sitter errors** for those tokens, so the first gate (`if parse_errors:`) is `False` and the retry is never attempted. String-literal content cannot cause a false trigger.

### Safety guarantees

| Property | How it is guaranteed |
|---|---|
| **Zero cost for clean `.ts` files** | The `if parse_errors:` gate is `False` — no extra work is done. |
| **No wrong result ever adopted** | TSX tree is adopted only when `len(tsx_errors) < len(parse_errors)`. |
| **No other languages affected** | Guard is explicitly `lang == "typescript"`. |
| **String literals cannot cause false triggers** | Valid TypeScript with `</`/`/>` in strings parses with zero errors — first gate never opens. |

---

## Verification

A regression test can be added to `tests/unit/ingestion/parser/test_typescript.py` alongside the existing `test_tsx_file_uses_jsx_grammar`:

```python
def test_ts_file_with_jsx_falls_back_to_tsx_grammar(self, parser: ASTParser) -> None:
    """JSX inside a .ts file triggers the adaptive TSX fallback and yields zero parse errors."""
    src = b"""
export function Button() {
  function handleClick() { console.log("clicked"); }
  return <button onClick={handleClick}>Click Me</button>;
}
"""
    fi = _make_file_info("src/ui/Button.ts", "typescript")  # .ts not .tsx

    assert result.parse_errors == []                   # fallback cleared the error
    assert any(s.name == "Button" for s in result.symbols)
    assert not any(s.name == "handleClick" for s in result.symbols)  # nested helper not hoisted
```
I have already implemented this adaptive fallback with full unit test coverage on a local branch. Let me know if you'd like me to open a PR for this!

akshatmalik-bruh:
## Summary

- Adds a gated adaptive TSX grammar retry to `ASTParser.parse_file` for TypeScript files (`.ts`, `.mts`, `.cts`) containing JSX markup, eliminating AST `ERROR` nodes.
- Reassigns both `grammar_tag = "tsx"` and `language = tsx_language` so `_get_query()` appends `tsx.scm`, restoring `jsx_opening_element` and `jsx_self_closing_element` call-site captures to prevent false-positive dead-code findings.
- Enforces strict safety guarantees: clean `.ts` files skip the fallback (`if parse_errors:` gate), and unrelated syntax errors containing `</` in comments/strings preserve their original parse tree (`len(tsx_errors) < len(parse_errors)` check).

## Related Issues

Fixes #965

## Test Plan

- [x] Tests pass (`pytest`)

Added unit tests in `tests/unit/ingestion/parser/test_typescript.py`:
- `test_ts_file_with_jsx_falls_back_to_tsx_grammar`: Verifies zero parse errors and correct symbol extraction for `.ts` files containing JSX.
- `test_ts_file_jsx_call_site_edges_captured_after_fallback`: Asserts call targets (`StatRow`, `Section`) are registered after fallback, preventing false dead-code findings.
- `test_ts_file_unrelated_syntax_error_with_html_in_string_preserves_original_parse`: Pins negative safety net (unrelated syntax error with `</` in comments/strings keeps original parse intact).

Verified full ingestion suite: **1,517 passed in 107s**.

## Checklist

### DIFF
--- a/packages/core/src/repowise/core/ingestion/parser.py
+++ b/packages/core/src/repowise/core/ingestion/parser.py
@@ -300,6 +300,16 @@ class ASTParser:
                     tree = tsx_tree
                     root = tree.root_node
                     parse_errors = tsx_errors
+                    # Both grammar_tag AND language must be reassigned.
+                    # grammar_tag is consumed immediately below by
+                    # self._get_query(lang, language, grammar_tag), which
+                    # appends tsx.scm to the base typescript.scm query.
+                    # That append is what supplies the
+                    # jsx_opening_element / jsx_self_closing_element captures
+                    # that restore JSX component call-site edges.
+                    # Reassigning only ``tree`` / ``root`` would fix parse
+                    # errors but leave those edges missing — the dead-code
+                    # false-positive would remain.
                     grammar_tag = "tsx"
                     language = tsx_language
 
--- a/packages/core/src/repowise/core/ingestion/parser.py
+++ b/packages/core/src/repowise/core/ingestion/parser.py
@@ -279,6 +279,30 @@ class ASTParser:
         root = tree.root_node
 
         parse_errors = _collect_error_nodes(root)
+
+        # Adaptive TSX grammar fallback: if a .ts file contains JSX markup,
+        # tree-sitter-typescript produces ERROR nodes. Re-parse using the TSX
+        # grammar ONLY IF:
+        #   1. Initial parse yielded error nodes (parse_errors is non-empty)
+        #   2. The file is TypeScript (lang == "typescript" and grammar_tag != "tsx")
+        #   3. Source contains JSX-specific closing tokens (b"/>" or b"</")
+        if (
+            parse_errors
+            and lang == "typescript"
+            and grammar_tag != "tsx"
+            and (b"/>" in source or b"</" in source)
+        ):
+            tsx_language = _get_language("tsx")
+            if tsx_language is not None:
+                tsx_tree = Parser(tsx_language).parse(source)
+                tsx_errors = _collect_error_nodes(tsx_tree.root_node)
+                if len(tsx_errors) < len(parse_errors):
+                    tree = tsx_tree
+                    root = tree.root_node
+                    parse_errors = tsx_errors
+                    grammar_tag = "tsx"
+                    language = tsx_language
+
         query = self._get_query(lang, language, grammar_tag)
 
         # Execute the compiled query ONCE per file. The five extraction
--- a/tests/unit/ingestion/parser/test_typescript.py
+++ b/tests/unit/ingestion/parser/test_typescript.py
@@ -147,6 +147,55 @@ export function Card({ title }: { title: string }) {
         assert "Card" in names
         assert "handleClick" not in names
 
+    def test_ts_file_with_jsx_falls_back_to_tsx_grammar(self, parser: ASTParser) -> None:
+        # A .ts file (not .tsx) containing JSX markup normally produces ERROR nodes.
+        # The adaptive fallback detects parse errors and JSX tokens (/> or </)
+        # and re-parses with the TSX grammar, yielding 0 parse errors.
+        src = b"""
+export function Button({ label }: { label: string }) {
+  function handleClick() { console.log("clicked"); }
+  return <button onClick={handleClick}>{label}</button>;
+}
+"""
+        fi = _make_file_info("ui/src/Button.ts", "typescript")  # .ts extension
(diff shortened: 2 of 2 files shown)


RaghavChamadiya (approved):
This is exactly what I asked for, thanks. All three points from the issue thread are covered: the inline comment explaining why both `grammar_tag` and `language` have to be reassigned, the call-edge test that asserts `<StatRow />` and `<Section />` come back as call targets, and the negative case that keeps an unrelated error's original parse intact.

I pulled the branch and confirmed locally: 21/21 parser tests pass including the three new ones, and the fallback wiring lines up with how `_get_query` consumes `grammar_tag`. Nice, careful work on a sensitive file.

Merging.

PR #980, opened 2026-07-22
feat(cli): make init's next-steps panel react to what setup actually did

RaghavChamadiya:
## What

`repowise init`'s completion "what's next" panel was a fixed guess. Now it is built from a single snapshot of what editor setup actually did, taken after client registration and the interactive hook offers run, so the suggestions match reality.

## Why

`init` auto-registers the MCP server for Claude Code unconditionally (plus Claude Desktop if installed, VS Code, and the repo `.mcp.json`), so the old headline `repowise mcp .` was redundant for the common case. And the index-only path pointed at the all-or-nothing `update --full`, which contradicted the scoped `repowise generate` upgrade path.

## Changes

- **`repowise serve` is the headline in every mode** (fast / index-only / full / workspace); it opens the dashboard where the graph, hotspots, dead code, decisions and wiki all live. The second row is the mode's real next move: `init` for fast, `generate` for index-only (no longer `update --full`), `search` for full.
- **Dynamic via a new `EditorSetupOutcome`** read by `detect_editor_setup_outcome`. `show_completion` now runs last (after the two hook offers) so the panel reflects the final state. A headless / skip-setup run gets a manual `repowise mcp .` row; a non-interactive run whose hooks are missing gets the hook-install rows (an interactive run was already offered them live, and a skip-setup run is never nagged).
- **MCP status as a note beside the panel**: Claude Code is auto-registered, so it says to restart it (first index only); it points Cursor or Codex at `repowise mcp .`.
- **Rewrite-hook check covers both Claude Code and Codex** surfaces, so a Codex-only user is not nagged.
- Removed the dead manual per-client config block (and its last Cline mention); dropped Cline from `mcp` help.
- **`update`**: index-only on a template wiki now points at `repowise generate` instead of `update --full`.

## Tests

- `test_fast_mode_offer.py`: serve-headline-every-mode, generate-not-update-full, headless manual-MCP row with no hook nag, non-interactive missing-hook rows, interactive no-nag, status-line restart-note (first index only), empty when headless.
- `test_editor_setup.py`: `detect_editor_setup_outcome` ground-truth across both agent surfaces + skip-env.
- Full CLI unit suite green (867 passed, 1 skipped), ruff clean.

### DIFF
--- a/docs/architecture/ARCHITECTURE.md
+++ b/docs/architecture/ARCHITECTURE.md
@@ -1064,9 +1064,11 @@ and supports two transports:
 
 ### Auto-generated Config
 
-`repowise init` automatically generates `.repowise/mcp.json` with ready-to-paste
-config blocks for Claude Code (`~/.claude/claude.json`), Cursor (`.cursor/mcp.json`),
-and Cline. This config is printed at the end of `repowise init`.
+`repowise init` registers the MCP server directly where it can: Claude Code
+(`~/.claude/settings.json`), Claude Desktop (if installed), VS Code
+(`.vscode/mcp.json`), and the repo-shared `.mcp.json`. It also writes
+`.repowise/mcp.json` for clients configured by hand. The completion panel says
+which client was wired up and how Cursor or Codex connect (`repowise mcp .`).
 
 ---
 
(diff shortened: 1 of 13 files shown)


swati510 (approved):

RaghavChamadiya:
Merged to main via squash 81795ef3

PR #1010, opened 2026-07-23
fix(generation): filter file dependency context

austintraver:
## Summary

- Keep the upstream file-page filter for self references, symbol containment, and `external:` nodes.
- Exclude historical `co_changes` file-to-file edges from `dependencies` and `dependents`.
- Preserve `imports`, `type_use`, `framework`, and future structural file-edge types.

## Why this remains after the upstream renderer change

`ContextAssembler.assemble_file_page()` receives the full mixed graph during level-2 generation, not `GraphBuilder.file_subgraph()`. The latter already excludes `co_changes` for metrics, but the context assembler previously traversed the full graph directly. Therefore the metric filter did not prevent historical neighbors from reaching file-page dependency context.

The recently merged single-renderer work already removes a file's own symbol nodes and external nodes. This PR preserves that behavior and adds only the missing `co_changes` edge-type exclusion.

## Real repository validation

Read-only inspection of AI Foundry's `wiki.db` on 2026-07-25 found file-level incident edges for `loralab/backfill/queue_status.py` with all three relevant labels: 13 `imports`, 8 `defines`, and 11 `co_changes`. This confirms the live graph contains genuine `co_changes` attributes at generation time; the new filter removes only the historical association edges.

## Validation

- [x] `uv run pytest tests/unit/generation/test_context_assembler.py tests/unit/ingestion/test_subgraph_cache.py tests/unit/ingestion/test_graph.py -q` — 94 passed
- [x] Ruff lint and format checks on the four changed Python files
- [x] Read-only SQLite graph inspection against AI Foundry

## Checklist

- [x] Tests cover direction and edge labels
- [x] No documentation change is required

### DIFF
--- a/packages/core/src/repowise/core/generation/context/assembler.py
+++ b/packages/core/src/repowise/core/generation/context/assembler.py
@@ -55,6 +55,33 @@ _MAX_IMPORTS = 30
 _MAX_TOP_FILES = 20
 
 
+def _file_dependency_neighbors(graph: Any, path: str, *, incoming: bool) -> list[str]:
+    """Return structural file neighbors in the requested direction.
+
+    The file graph treats every file-to-file edge except ``co_changes`` as a
+    dependency. This keeps import, type-use, framework, and future structural
+    edge types while excluding symbol containment and historical association.
+    """
+    if path not in graph:
+        return []
+
+    edges = graph.in_edges(path, data=True) if incoming else graph.out_edges(path, data=True)
+    neighbors: list[str] = []
+    for source, target, edge_data in edges:
+        neighbor = source if incoming else target
+        neighbor_data = graph.nodes[neighbor]
+        if not _is_foreign_edge(neighbor, path):
+            continue
+        if neighbor_data.get("node_type", "file") != "file":
+            continue
+        if neighbor_data.get("language") == "external":
+            continue
+        if edge_data.get("edge_type", "imports") == "co_changes":
+            continue
+        neighbors.append(neighbor)
+    return neighbors
+
+
 # ---------------------------------------------------------------------------
 # ContextAssembler
 # ---------------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


RaghavChamadiya:
Thanks for chasing this down, the co-change mislabeling is a real one. Heads up before it goes further: the self-referential dependency half already landed on `main`. `assemble_file_page` now routes neighbors through `_is_foreign_edge` (`assembler.py:40`), which drops `external:` nodes and both spellings of the file itself (`path` and `path::symbol`). Your diff rewrites those exact lines, so it will conflict and re-solve that half.

The genuinely new part here is excluding `co_changes` edges (plus the `language == "external"` check). Could you rebase onto current `main` and reframe the PR around just that? Two things to confirm while you are in there:
- that the graph handed to `assemble_file_page` still carries `co_changes` edges at generation time. If generation already passes a `file_subgraph()` with them stripped, the co-change filter is a no-op on live data.
- `edge_type` is read with `.get("edge_type", "imports")`, so please verify the real graph actually stamps `edge_type="co_changes"` on those edges. The test graph sets it explicitly, but if live edges lack the attribute the filter never fires.

austintraver:
@RaghavChamadiya I rebased and narrowed this as you suggested. The current implementation is in [6a664276](https://github.com/repowise-dev/repowise/commit/6a6642769250ee9cf403212dd9e80316137bb69a), with the follow-up external-node coverage in [8803266e](https://github.com/repowise-dev/repowise/commit/8803266ede15ee5865fcad9b33a8d6ae5cc625ca).

When you have a moment, I’d appreciate another look.

RaghavChamadiya:
Thanks for both of these, and for the rebase onto v0.36.0.

#1010 reads clean. Traversing edge data instead of predecessors/successors is
the right shape, and defaulting unknown node_type to "file" and unknown
edge_type to "imports" means future structural edge types keep flowing through
instead of silently disappearing. The graph is a plain DiGraph so there is no
duplicate-neighbor concern.

On #1011, one behavior change worth naming in the changelog: a max_tokens stop
now drops the page rather than persisting it truncated, and there is no retry at
a larger budget. That is contained (per-page failures are caught in
orchestrate), so a run never aborts, but anyone running a low max_tokens will
see whole pages go missing where they used to see short ones. I think that is
the right tradeoff, since a silently truncated page is worse than an absent one.

Merging #1011 first, then #1010

RaghavChamadiya (approved):
LGTM

PR #1009, opened 2026-07-23
feat(ollama): support disabling reasoning

austintraver:
## Summary

- advertise exactly `auto` and `off` for Ollama models
- preserve the existing field-free request shape for `auto`
- map Repowise `off` to Ollama `reasoning_effort="none"`
- reject other explicit reasoning modes before issuing a completion request

## Prior art

This is the missing Ollama provider integration for the provider-neutral design established by [Feature: expose a reasoning-disable knob for chat completions providers (Qwen3 / DeepSeek-R1 thinking mode)](https://github.com/repowise-dev/repowise/issues/137) and merged in [Add reasoning mode configuration](https://github.com/repowise-dev/repowise/pull/175). It does not change the shared reasoning contract or generation pipeline.

## Validation

- `uv run pytest tests/unit/test_providers/ -q` — 196 passed
- focused Ollama provider tests cover discovery, `auto`, `off`, and pre-request rejection of unsupported modes
- `uv run ruff check` and `uv run ruff format --check` on both changed files
- `uv run mypy packages/core/src/repowise/core/providers/llm/ollama.py`

Live Ollama 0.31.2 check with `nemotron-cascade-2:30b-a3b-q4_K_M`, the existing `operational-reference` style, and `max_tokens=2048`:

- request captured `reasoning_effort="none"`
- 3,933 prompt tokens; 666 completion tokens
- 2,934 content characters / 284 words; zero reasoning characters
- `finish_reason="stop"`; all four required headings present in order
- no repeated nonblank lines or repeated 12-word shingles

The check did not modify AI Foundry's Repowise config or wiki database, and the model loaded for validation was unloaded afterward.

Issue #987, opened 2026-07-22
[Bug] OpenAI 5+ models do not support temperature other than 1


IlyaSavich:
## Describe the Bug

I get 400 error when try to use OpenAI gpt-5.6-luna model. OpenAI stopped supporting temperature for 5+ models, only 1 is allowed value.

```
Error: Provider validation failed: [openai] Error code: 400 - {'error': {'message': "Unsupported value: 'temperature' does not support 0.3 with this model. Only the default (1) value is supported.", 'type': 'invalid_request_error', 'param': 'temperature', 'code': 'unsupported_value'}}
```

So I suggest to use 1 as default value for temperature for OpenAI provider

IlyaSavich:
I prepared a fix ^^, check pls, wdyt

IlyaSavich:
could you tell also once it will be merged when to expect it live ?

RaghavChamadiya:
Hey @IlyaSavich, thanks for opening the PR. I will review it soon, we are planning to make a release tomorrow, so will fold this in

IlyaSavich:
@RaghavChamadiya in the meantime I found one more issue with OpenAI models:
```
2026-07-22 15:27:30 [error    ] page_generation_failed         error=[openai] Error code: 400 - {'error': {'message': 'max_tokens is too large: 20000. This model supports at most 16384 completion tokens, whereas you provided 20000.',
'type': 'invalid_request_error', 'param': 'max_tokens', 'code': 'invalid_value'}} page_id=file_page:path/to/some/file.tsx
```

Model | Max output tokens | Supports max_tokens: 20000?
-- | -- | --
GPT-5.6 Luna | ~16k | ❌ No
GPT-5.6 Terra | ~16k | ❌ No
GPT-5.6 Sol | ~16k | ❌ No
GPT-4.1 | ~32k | ✅ Yes
GPT-4.1 mini | ~32k | ✅ Yes
GPT-4o | ~16k | ❌ No
GPT-4o mini | ~16k | ❌ No
GPT-4 Turbo | ~4k | ❌ No


show we strip this max_tokens ?

RaghavChamadiya:
Good catch on that one too. It is coming from GenerationConfig.max_tokens, which defaults to 20000 in packages/core/src/repowise/core/generation/models.py, and the OpenAI provider forwards it straight through as max_completion_tokens

[...]

PR #1011, opened 2026-07-23
fix(generation): enforce documentation completion

austintraver:
## Summary

- Preserve normalized and provider-native completion stop reasons when an adapter exposes them, validate them before caching or persistence, and retain them in accepted model-written page metadata.
- Read the persisted repository `max_tokens` setting through the shared `GenerationConfig.from_repo_config()` translation point used by the wiki-page generation entry points.
- Reject fresh empty, provider-declared token-limited, and grossly repetitive model output before it can enter the page cache or persistence path.

## Current scope after upstream renderer changes

The recently merged single-renderer work makes `file_page`, `symbol_spotlight`, `api_contract`, `infra_page`, `scc_page`, and `layer_page` structural: they do not call a model. This PR therefore protects the remaining model-written wiki pages: `module_page`, `repo_overview`, `architecture_diagram`, and `onboarding`. It deliberately leaves structural pages unchanged.

`max_tokens` is the configured completion limit for those wiki-page provider calls. It does not replace fixed, deliberately small budgets used by separate structured helpers, such as concept-name and knowledge-graph-tour JSON generation.

## Completion contract

A fresh model response is accepted only when it has non-whitespace content and no provider-declared token-limit stop. Accepted pages retain `stop_reason` and `provider_stop_reason` in their metadata. A `max_tokens` stop, empty content, or conservative pathological repetition raises `InvalidGeneratedContentError`; it is not cached or persisted. Prior-page reuse remains unchanged because it does not create fresh model output.

Adapters that cannot expose a non-streaming finish reason leave it unset rather than inventing one; those outputs still receive empty/repetition validation, but a silent truncation cannot be identified until the adapter exposes native completion metadata.

## Provider mapping

- OpenAI-compatible adapters map `stop` to `end_turn`, `length` to `max_tokens`, and tool-call finishes to `tool_use`.
- Anthropic maps `end_turn` and `stop_sequence` to `end_turn`, and retains `max_tokens`.
- Gemini maps `STOP` and `MAX_TOKENS` to the same neutral contract.
- Unknown native values remain available as `provider_stop_reason` without being promoted to an unsupported neutral state.

## Validation

- [x] Focused completion/config/provider/server tests: 212 passed after the rebase
- [x] Ruff lint and format checks on the changed completion files, excluding unchanged upstream `SIM102` in `core.py` and `SIM105` in `upgrade_flow.py`
- [x] The dependency-context and completion changes are rebased separately onto `v0.36.0`

### DIFF
--- a/docs/reference/CLI_REFERENCE.md
+++ b/docs/reference/CLI_REFERENCE.md
@@ -122,6 +122,12 @@ repowise init . --no-prose -x "node_modules/"        # workspace, no LLM
 repowise init . --no-workspace                        # force single-repo, even in a workspace root
 ```
 
+**Documentation output limit.** Set `max_tokens` in
+`.repowise/config.yaml` to bound each model-written page. It is persistent, not
+a per-run flag, and is honored by `init`, `update`, `generate`, `restyle`,
+workspace generation, and server-triggered generation. See
+[Configuration](CONFIG.md#configyaml).
+
 ---
 
 ### `repowise update [PATH]`
--- a/docs/reference/CONFIG.md
+++ b/docs/reference/CONFIG.md
@@ -31,11 +31,12 @@ should not be committed; it's a local cache, not a source of truth.
 The main configuration file. Created after first `init`, updated when you pass
 flags like `--commit-limit`, `--follow-renames`, or `--wiki-style`.
 
-> **No schema validation.** `config.yaml` is loaded as a plain YAML dict.
+> **Limited schema validation.** `config.yaml` is loaded as a plain YAML dict.
 > Unknown or misspelled keys are silently ignored, they won't error and won't
-> take effect. The only part that's validated is the `distill:` block, and
-> only when you run `repowise doctor`. If a setting doesn't seem to be taking
-> effect, check spelling and indentation first.
+> take effect. `max_tokens` must be a positive integer when documentation is
+> generated. The `distill:` block is validated only when you run
+> `repowise doctor`. If a setting doesn't seem to be taking effect, check
+> spelling and indentation first.
 
 ```yaml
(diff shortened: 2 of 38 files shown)


RaghavChamadiya (approved):
Thanks, this is solid. The per-provider `normalize_stop_reason` plumbing and the single `from_repo_config` owner for `max_tokens` are clean, and the repetition detector is conservative enough to leave normal tables and API lists alone. Confirmed the reject path is safe: `InvalidGeneratedContentError` is caught per page in `orchestrate.py:589` and returned as a value, so one bad page fails on its own instead of aborting the run.

Capping Gemini at `config.max_tokens` is the right call. It lines Gemini up with OpenAI, which already bounds output via `max_completion_tokens`, and it is the same class of provider-param fix as the GPT-5 temperature clamp in #989 / #987. Rejecting a page that hit the ceiling rather than persisting a truncated one is the behavior we want.

One small thing to keep on the radar, not a blocker: the concept outline naming call at `concept_tree/planner.py:395` calls `provider.generate(max_tokens=16000)` directly, so Gemini now honors `max_output_tokens=16000` there too. It is wrapped in `except Exception` and degrades to deterministic naming, so no crash, just worth knowing that path exists if outline naming ever looks off on a large Gemini repo. Approving.

RaghavChamadiya:
One logistics note: this now conflicts with `main` (it went DIRTY after some recent merges, including the Ollama provider lines that #1009 also touched). Could you rebase onto current `main` and re-run your validation before we merge? The conflicts look mechanical (mostly the `GenerationConfig.from_repo_config` call sites and `ollama.py`), but since the PR carries live-model validation it is best if you re-verify after the rebase. Approved on the substance, just needs the rebase to land.

PR #1019, opened 2026-07-23
docs(examples): add examples index and discoverability links

Ayush7614:
## Summary
- Add \`examples/README.md\` indexing the existing Codex walkthrough and documenting conventions.
- Link examples from the root README, docs index, and \`docs/agent/CODEX.md\`.
- Cross-link the Codex example back to the full integration guide.

## Test plan
- [x] Verify linked paths exist (\`examples/codex/\`, \`docs/agent/CODEX.md\`)
- [x] No fabricated examples — only links to content that already exists

PR #1020, opened 2026-07-23
docs(examples): add health-coverage walkthrough

Ayush7614:
## Summary
- Add \`examples/health-coverage/README.md\` covering \`repowise health\`, \`coverage add/status\`, and \`impacted-tests\` with commands verified against CLI \`--help\`.
- Link the walkthrough from \`docs/layers/CODE_HEALTH.md\` and \`TEST_INTELLIGENCE.md\`.
- Index the example in \`examples/README.md\`.

## Test plan
- [x] Cross-check every flag against \`repowise health|coverage|impacted-tests --help\`
- [x] No invented APIs or fake output samples presented as live results

### DIFF
--- a/docs/layers/CODE_HEALTH.md
+++ b/docs/layers/CODE_HEALTH.md
@@ -35,6 +35,9 @@ repowise health        # KPIs + 20 worst-scoring files + top findings
 repowise update        # re-score only changed files on each subsequent run
 ```
 
+Copy-paste walkthrough (health + coverage + impacted-tests):
+[examples/health-coverage/](../../examples/health-coverage/).
+
 Open `http://localhost:7777/repos/<id>/health` for the dashboard once the
 local server is running (`repowise serve`).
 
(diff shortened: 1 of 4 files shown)


RaghavChamadiya:
Thanks @Ayush7614. Content is good and I spot-checked the commands rather than taking the test plan on trust: `health --refactoring-targets`, `--trend`, `--file`, `--module`, `init --index-only`, `coverage add` auto-discovery and `impacted-tests --format list` all match the CLI. Nice that the impacted-tests section documents the "ingest coverage first" path instead of pretending an empty result is a pass.

Needs a rebase before I can merge. `examples/README.md` now exists on main, created by your own #1019, which landed after you opened this. This branch still adds it as a new file, hence the conflict.

After the rebase it should be a small edit rather than a new file: add the `health-coverage/` row to the existing table, plus the Code health and Test intelligence entries under Related docs. The two doc links and `examples/health-coverage/README.md` carry over unchanged.

Heads-up that #1021 touches the same table, so whichever merges second will need its row re-added. Not your problem to solve, just so the second conflict is not a surprise.

Ayush7614:
@RaghavChamadiya 

Rebased onto latest `main` to resolve the `examples/README.md` conflict from #1019.

The index change is now a small edit on the existing file: add the `health-coverage/` table row plus the Code health / Test intelligence Related docs links. Ready for another look.

RaghavChamadiya (approved):
LGTM

PR #1021, opened 2026-07-23
docs(examples): add OpenCode provider setup example

Ayush7614:
## Summary
- Add \`examples/opencode/README.md\` for \`repowise init --provider opencode\` with smoke checks against the real OpenCode CLI.
- Explicitly document what is *not* available yet (no \`--opencode\` editor/hooks/plugin path).
- Link from \`docs/agent/OPENCODE.md\` and index in \`examples/README.md\`.

## Test plan
- [x] \`repowise init --help\` lists \`opencode\` as a provider
- [x] \`repowise generate --help\` supports \`--provider\`
- [x] \`opencode --version\` runs locally
- [x] No fake editor-integration claims

### DIFF
--- a/docs/agent/OPENCODE.md
+++ b/docs/agent/OPENCODE.md
@@ -115,3 +115,7 @@ The provider enforces several safety measures:
 - [OpenCode GitHub](https://github.com/anomalyco/opencode)
 - [OpenCode Docs](https://opencode.ai/docs)
 - [OpenCode Download](https://opencode.ai/download)
+
+## Example walkthrough
+
+Short smoke path: [examples/opencode/](../../examples/opencode/).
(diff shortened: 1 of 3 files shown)


RaghavChamadiya:
Thanks @Ayush7614. I like that this is explicit about what does not exist yet: the "What This Does Not Set Up" section and pointing people to the Codex example for the hooks path is exactly right, and it is the kind of thing docs usually leave people to discover the hard way. Commands check out too (`init --provider`, `--model`, `generate --provider`).

Needs a rebase, same reason as #1020. `examples/README.md` is on main now via your #1019, and this branch still creates it.

After the rebase it is a one-line change to that file: add the `opencode/` row to the existing table. Your Related docs list already matches what is on main, so nothing else to port. `docs/agent/OPENCODE.md` and `examples/opencode/README.md` carry over as-is.

Same note as on the other one: these two touch the same table, so the second to merge will need its row re-added.

Ayush7614:
@RaghavChamadiya 

Rebased onto latest `main` to resolve the `examples/README.md` conflict from #1019.

The PR now only adds the `opencode/` row to the existing examples index (plus `examples/opencode/README.md` and the `OPENCODE.md` link). Ready for another look.

Ayush7614:
Rebased again after #1020 merged.

Kept the \`health-coverage/\` row and Related docs from main, and re-added only the \`opencode/\` table row. Conflict should be clear now.

RaghavChamadiya (approved):
LGTM

PR #1027, opened 2026-07-23
fix(health): remove no-op --safe-only flag

Ayush7614:
## Summary
- Remove the unused \`repowise health --safe-only\` Click option (it was a silent no-op).
- Drop the misleading \`\"safe\" → --safe-only\` mapping from the Claude \`health\` command.
- Update \`CLI_REFERENCE.md\` and architecture docs that still implied a confidence filter.

## Test plan
- [x] \`repowise health --help\` no longer lists \`--safe-only\`
- [x] Grep confirms no remaining \`health --safe-only\` guidance in docs/plugins


RaghavChamadiya (approved):
Thanks for cleaning this up. A documented flag that does nothing is worse than no flag, good to have it gone before the release.

PR #1085, opened 2026-07-25
fix(ui): use stored concept tree for pages parented to repo root (#1081)

### LINKED ISSUE #1081: [Bug] Wiki breadcrumb trail ignores the stored tree for pages directly under the repo root
## Describe the Bug

`computeDocNav` in `packages/ui/src/docs/doc-nav.ts` reads the stored concept tree
(`parent_page_id` / `display_order`) to build a page's breadcrumb trail and its
prev/next sibling links. For any page whose only ancestor is the repo root it
silently falls back to the pre-tree `target_path` split instead, which produces
the wrong sibling links or none at all.

The gate is `chain.length > 0`:

```ts
const chain = page.parent_page_id ? ancestors(page, byId) : [];
if (chain.length > 0) {
  // stored-tree branch: siblings share parent_page_id, sorted by display_order
}
// ...otherwise fall through to the target_path split
```

`ancestors()` deliberately omits the repo root, because the caller prepends that
crumb itself:

```ts
// The repo root is the crumb the caller prepends; repeating it here would
// show the repository name twice.
if (current.parent_page_id) chain.unshift(current);
```

So a page parented directly to the root has a *correct and complete* stored
ancestry, yet `chain` comes back empty and the stored-tree branch is skipped.
Omitting the root from the trail and discarding the tree are two different
things, and right now one causes the other.

Two visible symptoms, same cause.

**1. Prev/next is silently wrong for the onboarding chapters.** These pages have
a real `target_path`, so the fallback finds siblings by "same page type in the
same directory" and sorts them by path, discarding the curated reading order the
generator stored in `display_order`:

| stored reading order | what the trail actually links |
| --- | --- |
| 1 Guided Tour | Active Landscape |
| 2 Codebase Map | Codebase Map |
| 3 Key Concepts | Development Guide |
| 4 How It Works | Guided Tour |
| 5 Development Guide | How It Works |
| 6 Active Landscape | Key Concepts |

"Next" from Guided Tour goes to How It Works, skipping Codebase Map. The
onboarding set is the one place in the wiki where reading order is the whole
point, and it is alphabetised by filename.

**2. Prev/next disappears entirely for the layer pages.** A layer, a root-level
cycle page and the architecture diagram carry synthetic paths with no `/`
(`layer:ui`, `scc-d7b7358e7e24`, and the repo name). Those hit this branch:

```ts
if (!page.target_path.includes("/")) {
  return { breadcrumbs: [{ label: crumbLabel(page), pageId: page.id }] };
}
```

One crumb, no siblings. On this repo's own wiki that is 15 of the root's 318
children: 11 layer pages, the architecture diagram, 2 cycle pages and 1 infra
page. The layers are the top-level spine of the wiki, so a reader walking the
architecture cannot page from one layer to the next.

## Steps to Reproduce

1. Index a repo whose wiki has a stored concept tree: `repowise init --yes`
2. `repowise serve`
3. Open the docs page and select **Guided Tour** from the tree.
4. Look at the prev/next links at the bottom of the reader.

## Expected Behavior

A page whose parent is the repo root uses the stored tree like every other page:
siblings are the pages sharing its `parent_page_id`, ordered by `display_order`.
"Next" from Guided Tour is Codebase Map. A layer page pages to the next layer.

The trail itself should still omit the root crumb, since the caller supplies it.

## Actual Behavior

The stored tree is discarded for these pages. Onboarding chapters link in
filename order; layer, root-level cycle, and architecture-diagram pages get no
prev/next at all.

No error is raised and the trail still renders, so this reads as "this page has
no siblings" rather than as a bug.

## Environment

- Affects `packages/ui/src/docs/doc-nav.ts`, so every consumer of the shared
  reader is affected: `packages/web` and the VS Code extension's wiki views.
- Not version-specific. Present since breadcrumbs moved onto the stored tree.

## Additional Context

The fix looks like separating "does this page have a stored parent" from "does it
have a non-root ancestor to show". Gating on `byId.has(page.parent_page_id)`
rather than `chain.length > 0` would take the stored-tree branch for root
children too, and `breadcrumbs` would then be just the page's own crumb with
`chain` empty, which is the desired trail.

Worth checking as part of the same change: the fallback's sibling rule ("same
page type in the same directory, sorted by path") is a reasonable guess for a
wiki with no stored tree, but it should never win over a tree that does exist.

akshatmalik-bruh:
## Summary
- Fixes `computeDocNav` gate in `packages/ui/src/docs/doc-nav.ts` so pages parented to the repo root use the stored concept tree instead of falling back to path-split sorting.
- Restores curated `display_order` reading order for root-level onboarding chapters (e.g., *Guided Tour* → *Codebase Map*).
- Restores missing Prev/Next navigation links for synthetic root-level pages (Layer pages, cycle pages, and architecture diagrams).

#### 1. Onboarding Chapter Navigation (Guided Tour)
| Before Fix | After Fix |
|---|---|
| Before Onboarding | After Layer |
| *Next button incorrectly linked to How It Works* | *Next button correctly links to Codebase Map* |

#### 2. Layer Page Prev/Next Links (Layer: UI)
| Before Fix | After Fix |
|---|---|
| Before Layer |After Onboarding | 
| *No Prev/Next buttons visible* | *Prev/Next buttons visible and linking to Layer: Core* |

---

## Related Issues
Fixes #1081

---

## Test Plan
- [x] Unit tests pass (`npx vitest run packages/ui/__tests__/docs/doc-nav.test.ts`)
- [x] Type check passes (`npm run type-check`)
- [x] Verified manually in browser UI via `repowise serve` + `npm run dev`

---

## Checklist

### DIFF
--- a/packages/ui/__tests__/docs/doc-nav.test.ts
+++ b/packages/ui/__tests__/docs/doc-nav.test.ts
@@ -123,9 +123,9 @@ describe("computeDocNav", () => {
     expect(nav.breadcrumbs[0]?.pageId).toBe(mod.id);
   });
 
-  it("keeps the directory trail for a page hanging off the repo root", () => {
-    // The tree places files no module claimed directly under the overview. A
-    // lone basename crumb would say less than the path the page still carries.
+  it("uses the stored tree for pages parented to the repo root", () => {
+    // Pages parented to the repo root use the stored tree (omitting the root crumb,
+    // which the caller prepends).
     const loose = makePage({
       id: "file_page:docs/design/contrast_check.py",
       title: "contrast_check.py",
@@ -133,11 +133,7 @@ describe("computeDocNav", () => {
       parent_page_id: ROOT.id,
     });
     const nav = computeDocNav(loose, [ROOT, loose]);
-    expect(nav.breadcrumbs.map((b) => b.label)).toEqual([
-      "docs",
-      "design",
-      "contrast_check.py",
-    ]);
+    expect(nav.breadcrumbs.map((b) => b.label)).toEqual(["contrast_check.py"]);
   });
 
   it("labels ancestors with their page type so callers need not guess", () => {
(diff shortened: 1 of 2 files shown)


RaghavChamadiya:
Thanks @akshatmalik-bruh, the gate change is right and matches what I had in mind on the issue. I like that you have attached the before and after images with the fixes

One blocker: this drops the "shows a single label for a page with no path and no parent" test. The diff interleaves that removal with a new test so it reads like an edit, but the repo_overview / empty-target_path assertion is gone. That branch is untouched by your fix and nothing else in the file covers it, please restore it alongside the two new ones, then this is good to go

akshatmalik-bruh:
Good spot @RaghavChamadiya , thank you! I've restored the missing `shows a single label for a page with no path and no parent` test in `packages/ui/__tests__/docs/doc-nav.test.ts` alongside the new root-level sibling navigation tests. 
All 11 tests in `doc-nav.test.ts` are passing cleanly now. Ready for review!

RaghavChamadiya (approved):
LGTM

PR #1030, opened 2026-07-23
fix(dead-code): align CLI/REST min_confidence default with MCP

Ayush7614:
## Summary
- Change CLI and REST \`min_confidence\` defaults from \`0.4\` to \`0.5\` to match MCP \`get_dead_code\`.
- Update CLI reference, DEAD_CODE / deep-dive docs, and the Claude \`dead-code\` command text.

## Test plan
- [x] \`repowise dead-code --help\` / Click default is \`0.5\`
- [x] REST \`Query(0.5)\` verified
- [x] Docs no longer claim a CLI-only \`0.4\` floor

PR #1087, opened 2026-07-25
fix(dead-code): align default min_confidence floor to RISK_CAP_CONFIDENCE (#1084)

### LINKED ISSUE #1084: [Bug] Default dead-code min_confidence of 0.5 hides every risk-flagged finding, and leaves the table's confidence slider filtering rows that were never fetched
## Describe the Bug

The dead-code analyzer caps any finding that carries a runtime-load risk factor
to exactly `RISK_CAP_CONFIDENCE`, and the comment on that constant says why:

`packages/core/src/repowise/core/analysis/dead_code/risk_factors.py:40`

```python
# Confidence ceiling applied to a finding that carries a runtime-load risk
# factor. 0.4 is the default ``min_confidence`` floor, so the finding still
# surfaces (as a medium / review-required candidate) but never reads as
# deletion-ready.
RISK_CAP_CONFIDENCE: float = 0.4
```

The design is explicit: cap to 0.4 *because* 0.4 is the floor, so the finding
still shows up as a review candidate. `analyzer.py:711-713` and `1065-1067`
apply that cap.

That floor is now 0.5. #1030 raised the CLI and REST defaults to match
`get_dead_code`:

- `packages/server/src/repowise/server/routers/dead_code.py:30` —
  `min_confidence: float = Query(0.5, ge=0.0, le=1.0)`
- `packages/cli/src/repowise/cli/commands/dead_code_cmd.py:21` —
  `--min-confidence` default `0.5`

So a finding capped to 0.4 now sits strictly below the floor that the cap was
chosen to clear. Every risk-flagged finding is filtered out of the default
report. The risk-factor feature is only reachable by passing an explicit
`--min-confidence 0.4`, which is exactly the flag a user would only think to
pass if they already knew what they were missing.

Two more surfaces were built against the 0.4 floor and are now inconsistent
with it:

- `packages/types/src/dead-code.ts:30` documents `MEDIUM: 0.4` as "The list
  endpoint's default floor; below this findings are not fetched", which is no
  longer true. `deadCodeConfidenceTier` still splits medium/low at 0.4, so the
  "low" tier is unreachable through the default REST path while the summary
  endpoint keeps counting into it
  (`packages/core/src/repowise/core/persistence/crud/analysis/dead_code.py:173`).
- `packages/ui/src/dead-code/findings-table.tsx:120` pins the confidence
  slider's floor to `DEAD_CODE_CONFIDENCE.MEDIUM` on the same stated grounds:

  ```ts
  // The slider floor is the server's own `min_confidence` default: nothing
  // below it is ever fetched, so a lower floor would be a control over nothing.
  ```

  With a 0.5 server floor, the slider's 0.40-0.50 travel is inert: it filters
  rows the response never contained.

## Steps to Reproduce

1. Index a repository containing a file the risk classifier flags, for example
   anything matching the config / bootstrap / database / environment tokens in
   `risk_factors.py` (`src/config/database.settings.js`, `alembic/seed.py`) that
   is genuinely unreferenced.
2. `repowise dead-code` — the finding is absent.
3. `repowise dead-code --min-confidence 0.4` — the finding appears, capped at
   0.4, tagged with its risk factors.
4. Open the dashboard Dead Code tab, expand "All findings", and drag the
   min-confidence slider down to 40%.

## Expected Behavior

The default report includes risk-flagged findings as review candidates, which is
what the cap exists to guarantee. Whatever the default is, the slider floor, the
`MEDIUM` doc comment and the endpoint agree on it, so every slider position
changes what is on screen.

## Actual Behavior

Risk-flagged findings are invisible in the default CLI report, the default REST
response and therefore the dashboard. The slider's bottom quarter does nothing,
and the "low" confidence tier is structurally always empty in the drill-down
even though the summary endpoint counts findings into it.

## Environment

- OS: Windows 11
- Python version: 3.11
- Repowise version: 0.36.0
- Installation method: pip

## Additional Context

Introduced by #1030, which aligned the CLI and REST defaults with
`get_dead_code`'s 0.5 but did not touch `RISK_CAP_CONFIDENCE` or the two
client-side constants documented as mirroring the endpoint default. The
straightforward resolution is to move the default back to 0.4 so it agrees with
the cap and the UI; the alternative is to raise `RISK_CAP_CONFIDENCE` and fix
the two client constants, but that weakens the cap's meaning, since the value
has to stay below `SAFE_CONFIDENCE_THRESHOLD` to keep flagged findings out of
the deletion-ready pile. Either way the four values want a single source of
truth: they currently disagree in a way that is invisible until someone has a
finding in that band.

akshatmalik-bruh:
## Summary
- **Establishes a Single Source of Truth in Python:** Imported `RISK_CAP_CONFIDENCE` (`0.4`) from `risk_factors.py` directly into the REST API router (`dead_code.py`), CLI command (`dead_code_cmd.py`), and MCP tool (`tool_dead_code.py`). This guarantees that API query defaults, CLI flags, and MCP tool defaults all derive from the same constant used by the core risk-capping logic, instead of three independently hardcoded literals.
- **Fixes Silent Finding Drops & Reconnects UI Slider:** Reverts the default `min_confidence` from `0.5` back to `0.4`, so runtime-risk-flagged review candidates (e.g. database seed files, config/bootstrap files capped at 40% confidence) surface in default CLI/REST/dashboard reports again. This also reconnects the web UI confidence slider's 40%–50% range to live backend data — previously that range filtered rows the server never returned.
- **Adds CI-Enforced Cross-Language Contract Tests:** Added `tests/unit/dead_code/test_confidence_parity.py`, which parses `DEAD_CODE_CONFIDENCE.MEDIUM` and `DEAD_CODE_CONFIDENCE.HIGH` out of the TypeScript source (`packages/types/src/dead-code.ts`) via regex and asserts they match Python's `RISK_CAP_CONFIDENCE` and `SAFE_CONFIDENCE_THRESHOLD` respectively. This gives CI an automated guardrail so the frontend slider bounds and backend thresholds can't silently drift apart again — `pytest` fails loudly if either side changes without the other.

---

## Related Issues
Fixes #1084

---

## Test Plan
- [x] Contract unit tests pass (`pytest tests/unit/dead_code/test_confidence_parity.py`)
- [x] TypeScript type check passes (`npm run type-check`)
- [x] Verified CLI default flag, REST query default, and MCP tool default all return risk-capped findings at 0.4.

---

## Checklist

### DIFF
--- a/docs/agent/MCP_TOOLS.md
+++ b/docs/agent/MCP_TOOLS.md
@@ -439,7 +439,7 @@ Unreachable code, unused exports, unused internals, and zombie packages, sorted 
 |-----------|------|----------|-------------|
 | `repo` | string | No | *(workspace only)* Target repo alias |
 | `kind` | string | No | Restrict to one finding kind: `unreachable_file` \| `unused_export` \| `unused_internal` \| `zombie_package` |
-| `min_confidence` | float | No | Minimum confidence floor (default `0.5`; `0.7`+ is cleanup-ready only) |
+| `min_confidence` | float | No | Minimum confidence floor (default `0.4`; `0.7`+ is cleanup-ready only) |
 | `safe_only` | boolean | No | Deletion-ready findings only, excluding anything with runtime-load risk (default `false`) |
 | `limit` | int | No | Max findings per tier, clamped to 25 (default 20) |
 | `tier` | string | No | Restrict to one tier: `high` (>= 0.8) \| `medium` \| `low` |
--- a/docs/architecture/deep-dives.md
+++ b/docs/architecture/deep-dives.md
@@ -254,7 +254,7 @@ DeadCodeAnalyzer.analyze()
     ├── _detect_unused_exports()       → findings with confidence + safe_to_delete
     ├── _detect_zombie_packages()      → findings at 0.5 confidence, never safe
     │
-    ▼ apply min_confidence filter (default 0.5)
+    ▼ apply min_confidence filter (default 0.4)
     │
     ▼ persist to dead_code_findings table
     │
--- a/docs/layers/DEAD_CODE.md
+++ b/docs/layers/DEAD_CODE.md
@@ -101,10 +101,10 @@ compare numbers:
 
 | Surface | High | Medium | Low | Default floor |
 |---------|------|--------|-----|---------------|
-| CLI (`repowise dead-code`) | `>= 0.7` | `0.5` to `0.7` | `< 0.5` | `--min-confidence 0.5` |
-| MCP (`get_dead_code`) | `>= 0.8` | `0.5` to `0.8` | `< 0.5` | `min_confidence=0.5` |
+| CLI (`repowise dead-code`) | `>= 0.7` | `0.4` to `0.7` | `< 0.4` | `--min-confidence 0.4` |
+| MCP (`get_dead_code`) | `>= 0.8` | `0.4` to `0.8` | `< 0.4` | `min_confidence=0.4` |
 
-Both surfaces now share the same default floor (`0.5`). The MCP high/medium
+Both surfaces now share the same default floor (`0.4`). The MCP high/medium
 cutovers stay stricter: an agent acting on a finding is riskier than a human
 reading a table.
 
@@ -230,7 +230,7 @@ before deleting anything.
 
 | Flag | Description |
 |------|-------------|
-| `--min-confidence` | Minimum confidence threshold (default `0.5`) |
+| `--min-confidence` | Minimum confidence threshold (default `0.4`) |
 | `--safe-only` | Only findings marked safe to delete |
 | `--kind` | `unreachable_file`, `unused_export`, `unused_internal`, `zombie_package` |
 | `--format` | `table` (default), `json`, `md` |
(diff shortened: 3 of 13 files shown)


RaghavChamadiya:
Thanks @akshatmalik-bruh, 0.4 is the direction I want, and making `RISK_CAP_CONFIDENCE` the single source of truth is the right shape.

Two blockers before merge:

1. The docs still say 0.5. #1030 set these lines and they need to come back with you: `docs/reference/CLI_REFERENCE.md:446`, `docs/agent/MCP_TOOLS.md:444`, `docs/architecture/deep-dives.md:257`, and `docs/layers/DEAD_CODE.md:104`, `:105`, `:233`, `:254`. Leaving them is the same drift the issue is about.
2. Ruff fails: `I001` on the new import in `dead_code_cmd.py` (it sits below the existing block with two blank lines rather than in it), and `ruff format` would reformat both that file and the new test. CI doesn't gate ruff, so this passed. `ruff format .` and `ruff check --fix` will sort it.

RaghavChamadiya:
One addition to the docs list above, my earlier comment missed it: `plugins/claude-code/commands/dead-code.md:26` also says `(default is 0.5)` and needs the same update. That file was part of #1030 too.

akshatmalik-bruh:
Thanks @RaghavChamadiya for catching these!

Both blockers are now fixed and pushed:
1. **Documentation Defaults**: Updated `min_confidence` default from `0.5` to `0.4` across all documentation and plugin files (`CLI_REFERENCE.md`, `MCP_TOOLS.md`, `deep-dives.md`, `DEAD_CODE.md`, and `plugins/claude-code/commands/dead-code.md`).
2. **Ruff Import Formatting**: Fixed the `I001` unsorted import error in `dead_code_cmd.py` and formatted code with `ruff format`.

`ruff check` and unit tests (190/190) are passing cleanly. Ready for merge!

RaghavChamadiya (approved):
LGTM

PR #1088, opened 2026-07-26
fix(coupling): keep ?focus= in sync with the pinned file

### LINKED ISSUE #1082: [Bug] CouplingExplorer leaves ?focus= out of sync with the pinned file
## Describe the Bug

`CouplingExplorer` and its `?focus=` URL sync disagree in two cases, and both leave a
`?focus=` in the address bar that the view is not honouring. Anything shared, bookmarked, or
copied from that state carries the wrong pin forward.

**1. A pin invalidated by a data change is never reported.**

`packages/ui/src/coupling/coupling-explorer.tsx:97-99` drops a pin whose file is no longer in
the payload:

```ts
useEffect(() => {
  if (pinned && !nodePaths.has(pinned)) setPinned(null);
}, [pinned, nodePaths]);
```

It calls `setPinned`, not the `changePin` wrapper on line 104 that also fires
`onFocusChange`. So the host is told about pin *changes* but never about pin
*invalidations*, and it has no way to detect one from its side.

**2. A bare `?focus=` suppresses the top-hub default.**

`packages/web/src/components/coupling/coupling-tab.tsx:51` passes `focus ?? undefined`.
nuqs returns `""` for a present-but-empty param, not `null` (`result[key] = parsedValue ??
parser.defaultValue ?? null`, and `""` is not nullish). `""` is `!== undefined`, so
`defaultPin` on `coupling-explorer.tsx:80-85` becomes `""`: falsy, so no guidance line and no
"Clear selection" button, but non-`undefined`, so the `topHub` seed is skipped. The result is
the flat ring that `initialFocus` was written to avoid.

**3. Lower priority: `?focus=` outlives the Coupling tab.**

`packages/web/src/app/repos/[id]/architecture/page.tsx:118` switches tabs without clearing
sibling params, so `?focus=` follows the user to Map, Layers, Explore, Symbols and
Dependencies, where it means nothing.

## Steps to Reproduce

**Case 1**

1. Open `/repos/{id}/architecture?view=coupling` and click a file to pin it. The URL gains
   `?focus={path}`.
2. Re-run `repowise update` so that file no longer appears in the coupling graph (or open a
   `?focus=` link built against an older index).
3. Reload the page.

**Case 2**

1. Open `/repos/{id}/architecture?view=coupling&focus=` (trailing `=`, no value).

**Case 3**

1. Pin a file on the Coupling tab, then click the Map tab.

## Expected Behavior

1. The URL drops `?focus=` when the explorer discards the pin, so the address bar always
   matches what is on screen.
2. A bare `?focus=` is treated as absent: the explorer opens pinned on the most-coupled hub.
3. `?focus=` is cleared when leaving the Coupling tab.

## Actual Behavior

1. The pin clears on screen and the guidance line falls back to "Hover a file to trace what
   changes with it", while the URL still names the vanished file. Reload and it repeats.
2. Nothing is traced and no pin is offered, but the hub default is skipped: a flat ring.
3. `?focus={path}` persists across every other Architecture tab.

No error is raised in any of the three, which is why they survive a manual pass.

## Environment

- Package: `@repowise-dev/ui` (`packages/ui/src/coupling/coupling-explorer.tsx`), consumed by
  `packages/web`
- Reproduced on `main`

## Additional Context

Suggested fixes:

1. Route the invalidation effect through `changePin` instead of `setPinned` so the existing
   `onFocusChange` contract covers it. That fixes every consumer at once, rather than each
   host re-deriving "is my pinned path still in the payload". Worth a test that the callback
   fires when the pinned file leaves the edge set.
2. `focus || undefined` at the call site, or have the explorer treat `""` as absent so hosts
   cannot get this wrong.
3. Clear `?focus=` in the tab-switch handler alongside any other view-scoped param.

Also worth noting while in this file: `topHub` (`coupling-explorer.tsx:41-56`) breaks a
degree tie by `Map` insertion order, i.e. by which file happens to appear first in `edges`.
Not a bug on its own, but it makes "the most-coupled hub" arbitrary between equal-degree
files, so any test asserting the default pin should use a strict maximum.

Real5K:
## Summary

- Route CouplingExplorer pin invalidation through `changePin` / `onFocusChange` so stale `?focus=` is cleared from the URL
- Treat bare `?focus=` (empty string) as absent so the top-hub default still seeds
- Clear `?focus=` when leaving the Coupling Architecture tab

## Related Issues

Fixes #1082

## Test Plan

- [x] `packages/ui` tests pass (832)
- [x] `packages/web` tests pass (8)
- [x] Pin invalidation notifies `onFocusChange(null)`
- [x] Empty `initialFocus` seeds the top hub
- [ ] Lint (`ruff`) - N/A, no Python changes

## Checklist

- [ ] I have updated documentation if needed (N/A - no docs change required)

### DIFF
--- a/packages/ui/__tests__/coupling/coupling-explorer.test.tsx
+++ b/packages/ui/__tests__/coupling/coupling-explorer.test.tsx
@@ -1,0 +1,77 @@
+import { describe, it, expect, vi } from "vitest";
+import { render, waitFor, screen } from "@testing-library/react";
+import type {
+  CouplingEdge,
+  CouplingNode,
+  CouplingGraphResponse,
+} from "@repowise-dev/types/coupling";
+import { CouplingExplorer } from "../../src/coupling/coupling-explorer.js";
+
+function node(path: string): CouplingNode {
+  return {
+    file_path: path,
+    module: path.split("/")[0] ?? null,
+    score: 5,
+    nloc: 100,
+  };
+}
+
+function edge(s: string, t: string, strength = 3): CouplingEdge {
+  return {
+    source: s,
+    target: t,
+    strength,
+    last_co_change: "2026-06-01",
+  };
+}
+
+function graph(
+  nodes: CouplingNode[],
+  edges: CouplingEdge[],
+): CouplingGraphResponse {
+  return { nodes, edges, total_edges: edges.length };
+}
+
+describe("CouplingExplorer", () => {
+  it("reports pin invalidation through onFocusChange when the file leaves the payload", async () => {
+    const onFocusChange = vi.fn();
+    const initial = graph(
+      [node("a.py"), node("b.py"), node("gone.py")],
+      [edge("a.py", "b.py", 5), edge("a.py", "gone.py", 2)],
+    );
+    const { rerender } = render(
+      <CouplingExplorer
+        data={initial}
+        initialFocus="gone.py"
+        onFocusChange={onFocusChange}
+      />,
+    );
+
+    const next = graph(
+      [node("a.py"), node("b.py")],
+      [edge("a.py", "b.py", 5)],
+    );
+    rerender(
+      <CouplingExplorer
+        data={next}
+        initialFocus="gone.py"
+        onFocusChange={onFocusChange}
+      />,
+    );
+
+    await waitFor(() => {
+      expect(onFocusChange).toHaveBeenCalledWith(null);
+    });
+  });
+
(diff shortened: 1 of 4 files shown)


RaghavChamadiya:
Thanks @Real5K, nicely scoped, and the fix is in the right layer: putting the "" guard in the explorer means the next consumer can't reintroduce the bug from its side.

Nothing blocking from me. Merging

RaghavChamadiya (approved):

PR #1090, opened 2026-07-26
fix(augment): keep log output off the hook's stdout

lntutor:
## Summary

- Silence Repowise/core logging before `repowise augment` dispatches hook payloads so machine-readable stdout stays clean.
- Add focused `CliRunner` coverage proving log silencing happens before hook dispatch.

## Related issue

Addresses the machine-output behavior of the `augment` slice in #930.

## Test plan

- [x] `uv run pytest tests/unit/cli/test_augment_cmd.py -q` (8 passed)
- [x] `uv run ruff check packages/cli/src/repowise/cli/commands/augment_cmd/command.py tests/unit/cli/test_augment_cmd.py`
- [x] `uv run ruff format --check tests/unit/cli/test_augment_cmd.py`
- [x] `git diff --check`
- [x] No web build required (CLI only)

## Checklist

- [x] My code follows the project style
- [x] I added tests for the changed behavior
- [x] I updated the PR metadata to match the final implementation

### DIFF
--- a/packages/cli/src/repowise/cli/commands/augment_cmd/command.py
+++ b/packages/cli/src/repowise/cli/commands/augment_cmd/command.py
@@ -79,6 +79,8 @@ from pathlib import Path
 
 import click
 
+from repowise.cli.helpers import silence_logs_for_machine_output
+
 from .bash_staleness import _handle_bash_post
 from .codex import _handle_codex_context_event, _handle_post_edit_use
 from .read_state import _handle_edit_post, _handle_read_post, _record_edit
@@ -99,6 +101,7 @@ _EDIT_TOOL_NAMES = {"apply_patch", "Edit", "Write"}
 def augment_command(client: str | None = None) -> None:
     """Enrich AI agent tool calls with codebase graph context (hook mode)."""
     try:
+        silence_logs_for_machine_output()
         _run_augment(client=client)
     except (SystemExit, KeyboardInterrupt):
         raise
--- a/tests/unit/cli/test_augment_cmd.py
+++ b/tests/unit/cli/test_augment_cmd.py
@@ -57,7 +57,6 @@ def test_augment_silences_logs_before_hook_dispatch(monkeypatch: pytest.MonkeyPa
         augment_cmd,
         "silence_logs_for_machine_output",
         fake_silence_logs_for_machine_output,
-        raising=False,
     )
     monkeypatch.setattr(augment_cmd, "_run_augment", fake_run_augment)
 
--- a/tests/unit/cli/test_augment_cmd.py
+++ b/tests/unit/cli/test_augment_cmd.py
@@ -8,6 +8,7 @@ from typing import Any
 import pytest
 from click.testing import CliRunner
 
+from repowise.cli.commands.augment_cmd import command as augment_cmd
 from repowise.cli.main import cli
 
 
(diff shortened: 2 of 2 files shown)


RaghavChamadiya:
Thanks @lntutor, and thanks for picking up the review notes from the earlier slices. The `CliRunner` coverage, the quiet default case, and the docs entry are all exactly what I asked for on the other commands. The problem here is my fault for not scoping the issue better: `augment` is a bad fit for this sweep, and I did not say so.

`augment` is a hook that writes its JSON response to stdout (`command.py:180-181`), and structlog's default `PrintLoggerFactory` also writes to stdout. So `-v` puts debug lines on the same stream as the hook payload:

```
2026-07-26 16:02:52 [debug    ] pipeline debug line            n=4
{"hookSpecificOutput": {}}
```

Claude Code and Codex parse that stdout as JSON, so the flag breaks the hook for anyone who uses it.

Beyond that, the flag is not really reachable: nobody types `repowise augment`. The command line lives in `settings.json` and the Codex config, written by `repowise init`, so using `-v` means hand-editing the hook registration. And `configure_cli_logging` is documented for progress-bar commands, which this is not. Making `-v` safe would mean adding a stderr `logger_factory` to a helper shared with `init` and `update`, which is a lot of risk on the indexing paths for a flag nobody can easily reach.

Your instinct that `augment` needs logging attention was right though, and there is a real bug underneath it. structlog's default is `BoundLoggerFilteringAtNotset`, so nothing is filtered at all today: every INFO and DEBUG line from `repowise.core` is already free to land on stdout mid-payload. I could not trigger it on the paths I probed, but it is latent and would fail silently.

So I would like to keep the half of your change that fixes that, and drop the rest. Concretely: call `silence_logs_for_machine_output()` unconditionally in `augment_command`, and drop the flag, the docs entry, and the verbose tests. That helper exists for exactly this case and its docstring describes this failure. It leaves you with a roughly three-line bug fix that closes a real hole.

Also please drop the unrelated reformat of the `_handle_post_edit_use` call.

If hook debugging is something you want to build, I would take it as a separate PR: an env var such as `REPOWISE_HOOK_DEBUG=1` writing to stderr, which is where Claude Code surfaces hook output under `--debug`. Happy to look at a proposal.

lntutor:
Updated the PR to the machine-output fix only: `augment_command` now calls `silence_logs_for_machine_output()` before hook dispatch. I removed `--verbose`, its docs/tests, and the unrelated formatting change, then rebased onto current `main`. Verification: all 8 tests in `tests/unit/cli/test_augment_cmd.py` pass and Ruff is clean for the changed Python files.

swati510 (approved):
Right fix, and it matches the established pattern — `silence_logs_for_machine_output()` (`cli/helpers.py:58-85`) is already used by `health`, `dead-code`, `security`, `impacted-tests`, `doctor` and `update` for exactly this. `augment` writes its JSON to stdout via `_emit_response` (`augment_cmd/command.py:180-181`) and structlog's default factory also writes stdout unfiltered, so any INFO line from `repowise.core` corrupts the payload the agent parses. The call sits inside the existing try/exit-0, so it can't break the hook.

Two things before merge:

1. Please retitle. The repo squash-merges, so `feat(cli): add verbose logging to augment` would land as a commit message describing a flag that is no longer in the diff. Something like `fix(augment): keep log output off the hook's stdout`.
2. Minor: `raising=False` on the `silence_logs_for_machine_output` patch in `tests/unit/cli/test_augment_cmd.py` isn't needed — the symbol is imported at module level, and `raising=True` makes the test fail loudly if that import is ever dropped.

lntutor:
Addressed both follow-ups on current head `06eef992`:

- retitled the PR to match the final squash-commit scope;
- removed `raising=False` so the test fails loudly if the imported helper disappears;
- refreshed the PR body to describe the final implementation.

Fresh local verification: `tests/unit/cli/test_augment_cmd.py` passes 8/8, Ruff check passes for both changed files, the test file is Ruff-format clean, and `git diff --check` passes. The new exact-head CI run is in progress.

RaghavChamadiya:
@lntutor this is approved and it should have been merged weeks ago. Sorry it sat.

The fix is still needed. `silence_logs_for_machine_output` is used by `dead_code_cmd`, `doctor_cmd`, `health_cmd`, `impacted_tests_cmd` and `security_cmd`, and `augment_cmd/command.py` is the one machine-output command that never picked it up, so it is still the odd one out on main today.

The only thing standing in the way is that the branch has gone conflicting. The import block in `augment_cmd/command.py` was reorganised upstream since you opened this, and your diff inserts among those imports, so it is a context conflict and not a disagreement with your change.

`git fetch origin && git rebase origin/main`, push, and I will merge it as soon as CI is green.

PR #1095, opened 2026-07-26
fix(ingestion): tolerate malformed .gitignore patterns

ShotaNagafuchi:
## Problem

`repowise init` aborts on any repository whose `.gitignore` (root, nested, or `.git/info/exclude`) contains a pattern that `pathspec` rejects but **git itself accepts** — e.g. a trailing backslash `.godot\` (common in Godot projects). `pathspec.PathSpec.from_lines` raises `GitWildMatchPatternError` on the first such line and the whole ingest dies:

```
pathspec.patterns.gitwildmatch.GitWildMatchPatternError: Invalid git pattern: '.godot\'
  File ".../ingestion/traverser.py", line 872, in _load_gitignore_spec
    return pathspec.PathSpec.from_lines("gitwildmatch", lines)
```

One stray line should not make an otherwise-valid repo un-indexable; repowise should be at least as lenient as git.

## Fix

- Add `_compile_gitignore()`, which compiles each ignore line independently and **skips only the malformed ones** (emitting a `log.warning`), instead of failing the whole spec.
- Route the three user-controlled loaders through it: root `.gitignore`/`info/exclude` (`_load_gitignore_spec`), nested per-directory `.gitignore` (`_get_dir_ignore`), the extra-ignore file (`_load_extra_ignore_spec`), and config `extra_exclude_patterns`.
- Trusted hardcoded `_BLOCKED_FILENAME_PATTERNS` are left untouched.

Well-formed `.gitignore` behaviour is unchanged.

## Tests

- `TestCompileGitignore` — unit coverage that a malformed line is dropped while valid patterns still match.
- `test_malformed_gitignore_line_does_not_abort` — end-to-end: a repo with `.godot\` in `.gitignore` now traverses successfully and the valid patterns after the bad line still apply.

All 82 tests in `test_traverser.py` pass; `ruff check`/`format` clean.

## Repro

`repowise init --no-prose` on a repo whose `.gitignore` contains a line `.godot\`. Before: exit 1 with the traceback above. After: indexes normally.

### DIFF
--- a/packages/core/src/repowise/core/ingestion/traverser.py
+++ b/packages/core/src/repowise/core/ingestion/traverser.py
@@ -873,23 +873,6 @@ def _compile_gitignore(lines: Iterable[str]) -> pathspec.PathSpec:
         try:
             patterns.append(GitWildMatchPattern(line))
         except GitWildMatchPatternError:
-            # Git tolerates a trailing backslash (e.g. ``.godot\``) and treats it
-            # as the bare path; dropping the line entirely would leave a large
-            # generated directory un-ignored — the opposite of the author's
-            # intent. Strip trailing backslashes and retry so the pattern is
-            # still honoured, and only drop the line if it remains unparseable.
-            salvaged = line.rstrip("\\")
-            if salvaged and salvaged != line:
-                try:
-                    patterns.append(GitWildMatchPattern(salvaged))
-                    log.warning(
-                        "recovered malformed gitignore pattern",
-                        pattern=line,
-                        used=salvaged,
-                    )
-                    continue
-                except GitWildMatchPatternError:
-                    pass
             log.warning("skipping malformed gitignore pattern", pattern=line)
     return pathspec.PathSpec(patterns)
 
--- a/packages/core/src/repowise/core/ingestion/traverser.py
+++ b/packages/core/src/repowise/core/ingestion/traverser.py
@@ -18,13 +18,14 @@ from __future__ import annotations
 import configparser
 import os
 import threading
-from collections.abc import Iterator
+from collections.abc import Iterable, Iterator
 from dataclasses import dataclass, field
 from datetime import datetime
 from pathlib import Path, PurePosixPath
 
 import pathspec
 import structlog
+from pathspec.patterns.gitwildmatch import GitWildMatchPattern, GitWildMatchPatternError
 
 from .languages.registry import REGISTRY as _LANG_REGISTRY
 from .models import (
@@ -278,7 +279,7 @@ class FileTraverser:
             "gitwildmatch", _BLOCKED_FILENAME_PATTERNS
         )
         patterns = extra_exclude_patterns or []
-        self._extra_exclude = pathspec.PathSpec.from_lines("gitwildmatch", patterns)
+        self._extra_exclude = _compile_gitignore(patterns)
         # Per-directory ignore cache: absolute dir path -> PathSpec built from
         # that directory's nested .gitignore + .repowiseIgnore.
         # Pre-seed root: its .gitignore is matched full-path via self._gitignore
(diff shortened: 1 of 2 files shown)


RaghavChamadiya:
Thanks @ShotaNagafuchi, good catch and a well-scoped fix. Compiling line by line and dropping only the offender is the right call, and keeping `_BLOCKED_FILENAME_PATTERNS` on the stock path is a nice touch.

I checked the "well-formed behaviour is unchanged" claim rather than assuming it: `PathSpec.from_lines` is `[pattern_factory(line) for line in lines if line]`, so `_compile_gitignore` matches it exactly, and comments, blanks, negations and `**` globs all resolve identically between the two. Verified locally along with the traverser suite and ruff.

One thing worth your thoughts before I merge: dropping the line means the pattern is not honoured at all, so in the motivating case `.godot/` now gets indexed instead of ignored, and that is a large generated cache. Failing to ignore beats failing to index, so I am fine with the tradeoff. But stripping a trailing backslash and retrying before giving up would honour the user's intent for exactly the case that prompted this. Happy either way, let me know what you prefer.

ShotaNagafuchi:
Thanks for the careful review — and for actually checking the `from_lines` equivalence rather than taking my word for it.

Good call on the tradeoff; I went with the salvage. Dropping the line means the target isn't ignored at all, and for the exact case that motivated this (`.godot\`) that's a large generated cache landing in the index — worse than the crash in spirit, if not in severity. Since git reads a trailing backslash as the bare path, I strip trailing backslashes and retry, and only drop the line if it's still unparseable. Recovery and drop are logged distinctly.

Pushed as `e0a0a021`. Verified end-to-end on the Godot repo that surfaced this: `.godot\` is recovered to `.godot` and 0 `.godot/*` files leak into the index. Added tests for all three paths — salvage, hard-drop of an unrecoverable line (a lone `\`), and the neighbours-survive case. Full `test_traverser.py` (84) green, ruff clean.

The stock `_BLOCKED_FILENAME_PATTERNS` path is untouched, and well-formed `.gitignore` behaviour is still identical to `from_lines`.

RaghavChamadiya:
Correcting myself: the salvage suggestion was wrong, and I gave it without checking git. Sorry for the round trip.

Git does not read a trailing backslash as the bare path. Fixture with `.godot\` in `.gitignore` (bytes verified `2e 67 6f 64 6f 74 5c 0a`) and a tracked `.py` file inside `.godot/`:

```
git ls-files            ->  .godot/tools/gen_helper.py is TRACKED
main (control)          ->  crashes, GitWildMatchPatternError
8bf1fddd (drop)         ->  indexes it        matches git
e0a0a021 (salvage)      ->  excludes it       diverges from git
```

Confirmed on git 2.51.0 and 2.34.1. Git treats the dangling escape as matching nothing: no error, no fallback. So your original commit was already exactly git-equivalent, which was the goal the PR set out with, and I talked you out of it.

Being straight about the magnitude: for Godot specifically this barely bites, since real `.godot/` contents are `.ctex` and cache files that the traverser does not index anyway. The case for reverting is git parity, not measured impact. The related worry is that the salvage rescues only trailing-backslash errors and drops every other malformed pattern, so it is a special case rather than a policy, and it makes one class of typo behave unlike both git and every other typo.

Could you revert `e0a0a021` back to the drop, and flip `test_trailing_backslash_is_salvaged` into an assertion that `.godot\` matches nothing, with the git-parity reason in a comment so nobody re-adds the salvage later? The other two tests are good as they are.

One more, moot once the salvage goes: `rstrip("\\")` strips all trailing backslashes, so `foo\` (an escaped literal backslash, valid for a file named `foo\`) would be rewritten to `foo`.

Everything else here I am happy with. Worth a rebase, it is 12 behind main.

RaghavChamadiya (approved):
LGTM

PR #1098, opened 2026-07-26
fix(cli): surface page generation failures and resume hint on init completion (#1094)

### LINKED ISSUE #1094: init reports the pages it generated and never mentions the ones that failed
Split out of #1089. This is the reason the reporter spent 106 minutes and a full re-run before learning what had actually happened.

## What happens

A provider outage killed the last ~11 minutes of a 106-minute `init`. 105 pages failed, including the repo overview, all onboarding, and every module and layer page. The CLI printed:

    ✓ Generated 663 pages

and exited successfully. The reporter found the gap by counting rows in the database, and then worked from the assumption that the wiki was simply incomplete with no way to repair it.

## Why it is invisible

A page whose provider call raises is caught per page and returned as a value (`page_generator/orchestrate.py:589-598`), then filtered out of the level's results at `orchestrate.py:629`:

```python
pages = [r for r in results if isinstance(r, GeneratedPage)]
```

From there the count is just `len(generated_pages)` (`init_cmd/generation.py:283`). The failure survives only in two places a user does not read: a `log.error("page_generation_failed", ...)` line hidden unless `--verbose`, and the job checkpoint under `.repowise/jobs/`.

## What to change

Carry the failure count out of generation and report it. The job system already records it: `job_system.fail_page(job_id, page_id, str(exc))` is called on every failure (`orchestrate.py:591`), and `JobSystem` checkpoints hold the failed page list, so the data does not need to be re-plumbed from scratch.

At the end of an init with failures, say three things:

1. How many pages failed, and ideally the breakdown by page type. The reporter produced exactly this table by hand and it is the most useful shape.
2. That the wiki is incomplete as a result.
3. That `repowise init --resume` generates only the missing pages without re-spending on the ones that succeeded.

Point 3 is the one that matters. We print the `--resume` hint during generation, as guidance for Ctrl-C ("safe to Ctrl-C, then run `repowise init --resume`", `init_cmd/generation.py:229`), but a run that fails page by page instead of being interrupted never surfaces it again.

## Notes

- Keep the exit code as it is. A run with some failed pages still produced a usable wiki, and this issue is about reporting, not about failing the run.
- The same silent drop applies to the `update` path. Fixing init first is fine.
- `--resume` currently deletes the structurally-keyed pages it skips (module, layer and cycle pages). That is being fixed separately under #1089. Until it lands, word the message as pointing at the flag rather than promising it is lossless, and link #1089 in the PR so the two land in the right order.

## Acceptance

- An init where some pages fail prints a failure count and a per-type breakdown, without `--verbose`.
- An init with zero failures prints exactly what it prints today.
- The message names the command that repairs the run.

akshatmalik-bruh:
## Summary
- Surfaces page generation failure counts and a per-type breakdown (e.g., `file_page`, `module_page`, `layer_page`, `onboarding`) at the end of `repowise init` runs, instead of silently dropping them.
- Failure reporting (count, breakdown, and incomplete-wiki warning) now prints unconditionally when generation failures occur — previously this was only visible with `--verbose`.
- Displays the repair guidance hint (`repowise init --resume`) whenever failures occur, on both verbose and standard runs.
- Updates the completion panel's **Pages generated** metric to reflect failed page counts (e.g., `663 (105 failed)`).
- Zero-failure runs are unaffected — output is unchanged from current behavior, in both verbose and non-verbose modes.

## Related Issues
- Fixes #1094

## Test Plan
Added/updated unit tests in `tests/unit/cli/test_init_failure_reporting.py`:
- `test_run_repo_generation_reports_failures` — verifies failure count, per-type breakdown, and the `repowise init --resume` hint all appear on a **non-verbose** run with failures.
- `test_run_repo_generation_zero_failures` (parametrized over `verbose_flag=[True, False]`) — verifies clean success output on zero-failure runs in both modes: the summary line prints when verbose, and no output is emitted when non-verbose.
- `test_show_completion_panel_with_failures` — verifies the completion panel's "Pages generated" metric reflects the failed count.

`pytest tests/unit/cli/test_init_failure_reporting.py -v` — 4/4 passed (5 collected cases including the parametrized pair).

Manually verified CLI output rendering via simulation script (see screenshots below).

## Screenshots

**Before** (failures silently dropped, non-verbose run):
image

**After** (non-verbose run with failures — count, breakdown, resume hint):
Screenshot 2026-07-26 160756

## Checklist

### DIFF
--- a/packages/cli/src/repowise/cli/commands/init_cmd/generation.py
+++ b/packages/cli/src/repowise/cli/commands/init_cmd/generation.py
@@ -13,6 +13,8 @@ returns a "declined" flag; the workspace flow prints a compact line and raises
 from __future__ import annotations
 
 import contextlib
+from collections import Counter
+from pathlib import Path
 from typing import Any
 
 from rich.progress import BarColumn, Progress, SpinnerColumn, TextColumn, TimeElapsedColumn
@@ -293,8 +295,39 @@ def run_repo_generation(
             )
         )
 
+    jobs_dir = Path(repo_path) / ".repowise" / "jobs"
+    failed_page_ids: list[str] = []
+    if jobs_dir.exists():
+        with contextlib.suppress(Exception):
+            from repowise.core.generation import JobSystem
+
+            js = JobSystem(jobs_dir)
+            job_id = getattr(result, "job_id", None)
+            if job_id:
+                failed_page_ids = js.get_checkpoint(job_id).failed_page_ids
+            else:
+                jobs = js.list_jobs()
+                if jobs:
+                    failed_page_ids = jobs[0].failed_page_ids
+
     result.generated_pages = generated_pages
-    if verbose:
+    result.failed_page_ids = failed_page_ids
+
+    if failed_page_ids:
+        type_counts = Counter(pid.split(":")[0] for pid in failed_page_ids)
+        console.print(
+            f"  [yellow]⚠[/yellow] Generated [bold]{len(generated_pages)}[/bold] pages "
+            f"([bold yellow]{len(failed_page_ids)} failed[/bold yellow])\n"
+        )
+        console.print("  [bold yellow]Failed pages by type:[/bold yellow]")
+        for page_type, count in sorted(type_counts.items(), key=lambda x: -x[1]):
+            console.print(f"    • {page_type}: {count}")
+        console.print(
+            "\n  [yellow]The wiki is incomplete due to provider failures.[/yellow]\n"
+            "  [dim]Run [bold]repowise init --resume[/bold] to generate missing pages "
+            "without re-spending on completed ones.[/dim]\n"
+        )
+    elif verbose:
         console.print(f"  [green]✓[/green] Generated [bold]{len(generated_pages)}[/bold] pages")
 
     # KG enrichment is layer naming and the guided tour, both pure prompting.
--- a/packages/cli/src/repowise/cli/commands/init_cmd/reporting.py
+++ b/packages/cli/src/repowise/cli/commands/init_cmd/reporting.py
@@ -228,13 +228,16 @@ def show_completion(
         _render_defect_accuracy(result)
     else:
         _pages = result.generated_pages or []
+        _failed_ids = getattr(result, "failed_page_ids", []) or []
         total_tokens = sum(p.total_tokens for p in _pages)
         # Split model-written from the zero-LLM deterministic coverage tail so
         # broad coverage reads as thoroughness, not padding.
         _det = sum(1 for p in _pages if getattr(p, "provider_name", "") == "template")
         _ai = len(_pages) - _det
         _pages_label = str(len(_pages))
-        if _det:
+        if _failed_ids:
+            _pages_label = f"{len(_pages)} ({len(_failed_ids)} failed)"
+        elif _det:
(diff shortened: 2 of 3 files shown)


swati510:
Thanks for this, and for #1087 as well. The failure reporting is the right shape, and the per-type breakdown is genuinely the most useful form of it. I like the screenshots you have attached, it makes it easy to visualize

One thing to sort out before this can go in, and I think it is just fallout from having both PRs open at once.

#1087 merged a few hours ago as 151491db. This branch still carries its own copy of that same work: the seven dead-code files plus tests/unit/dead_code/test_confidence_parity.py. Git merges the two without reporting a conflict, so it looks clean, but the result is a duplicated import in dead_code_cmd.py:

from repowise.core.analysis.dead_code.risk_factors import RISK_CAP_CONFIDENCE


from repowise.core.analysis.dead_code.risk_factors import RISK_CAP_CONFIDENCE

That is the entire net effect of the dead-code half of this PR now, and it fails lint on main:

I001  Import block is un-sorted or un-formatted
F811  Redefinition of unused `RISK_CAP_CONFIDENCE` from line 17

The checks on this PR are green because they ran before #1087 landed.

Could you rebase on main and drop the dead-code files, keeping only the three that belong to #1094:

- packages/cli/src/repowise/cli/commands/init_cmd/generation.py
- packages/cli/src/repowise/cli/commands/init_cmd/reporting.py
- tests/unit/cli/test_init_failure_reporting.py

Everything else that half of the PR does is already on main from #1087.

On the reporting logic itself, I checked the parts most likely to bite and they hold up. list_jobs() sorts by created_at descending and _setup_job creates a fresh job per run, so jobs[0] really is the current run. PipelineResult is a plain dataclass, so setting failed_page_ids on it is fine, and reporting.py reads it defensively. Your six tests pass for me on a merged tree.

One suggestion rather than a blocker: the lookup infers "the newest job on disk is mine" instead of using the job_id the run already has in hand. It is correct today, but a run that fails before _setup_job would report the previous run's failures as its own. Passing the id through would make it exact.

Heads up on ordering too. #1104 just merged, which makes a failed provider call fall back to the structural stub so the page gets a row instead of vanishing. It touches different files from yours, so your rebase on main should pick it up cleanly with nothing to work around. All 28 tests from both changes pass together.

It also makes the repowise init --resume line in your message fully true: stubs are deliberately kept out of the vector store, which is what --resume reads as already done, so those pages really do get regenerated.

Happy to re-review as soon as the dead-code files are dropped.

akshatmalik-bruh:
@swati510 Rebased on `main` and dropped all dead-code files from this branch!

The PR diff now strictly contains only the three files for #1094:
- `packages/cli/src/repowise/cli/commands/init_cmd/generation.py`
- `packages/cli/src/repowise/cli/commands/init_cmd/reporting.py`
- `tests/unit/cli/test_init_failure_reporting.py`

Also picked up #1104 cleanly, fixed the lint checks, and verified all unit tests pass. Ready for re-review!

Also addressed your non-blocking suggestion: `generation.py` now checks `result.job_id` first to query the exact job checkpoint when present, falling back to `jobs[0]` otherwise (with a dedicated test added). All 5 unit tests and ruff lint checks pass cleanly. Ready for re-review!

swati510 (approved):
lgtm

PR #1099, opened 2026-07-26
fix(init): attribute page failures to the current generation

### LINKED ISSUE #1094: init reports the pages it generated and never mentions the ones that failed

lntutor:
## Summary

- carry the active generation checkpoint's failed page IDs to the init command
- warn when a wiki is incomplete, including a deterministic page-type breakdown and the resume command
- keep clean runs and the successful init exit path unchanged

Closes #1094

## Validation

- uv run pytest tests/unit/cli/test_generation_persist.py tests/unit/generation/test_job_system.py -q
- 29 passed
- uv run ruff check on changed files (two unchanged upstream findings in page-generator modules)
- git diff origin/main --check

## AI assistance

This contribution was developed with AI assistance and reviewed with focused tests and independent code reviews.

### DIFF
--- a/docs/superpowers/plans/2026-07-26-init-report-partial-failures-design.md
+++ b/docs/superpowers/plans/2026-07-26-init-report-partial-failures-design.md
@@ -1,0 +1,157 @@
+# Init Partial-Failure Reporting Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Make `repowise init` clearly report page-generation failures and the recovery command while preserving its successful exit status and zero-failure output.
+
+**Architecture:** The generation phase already writes the current run's failed page IDs to its `JobSystem` checkpoint. Add an optional completion callback to `run_generation`; after `PageGenerator.generate_all` returns, it reads that run's checkpoint and sends the failed IDs to the CLI generation wrapper. The wrapper formats the IDs by their `page_type:` prefix and prints one warning only when the callback reports failures.
+
+**Tech Stack:** Python 3.11+, Click/Rich CLI, pytest, existing JSON-backed `JobSystem`.
+
+---
+
+### Task 1: Expose the current generation run's failure IDs
+
+**Files:**
+- Modify: `packages/core/src/repowise/core/pipeline/phases/generation.py:22-156`
+- Test: `tests/unit/cli/test_generation_persist.py`
+
+- [ ] **Step 1: Write the failing pipeline callback test**
+
+Add a focused async test that supplies a completion callback to `run_generation`, uses a provider that raises for one page, and asserts the callback receives the failed page ID after generation returns.
+
+```python
+failed: list[str] = []
+
+pages = await run_generation(
+    ...,
+    on_generation_complete=failed.extend,
+)
+
+assert pages
+assert failed == ["file_page:broken.py"]
+```
+
+- [ ] **Step 2: Run the focused test and confirm it fails because the callback is unsupported**
+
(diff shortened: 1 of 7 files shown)


lntutor:
Rebased onto current `main` and narrowed the PR around the remaining delta after #1098: generation now reports failed page IDs from the exact active `PageGenerator` checkpoint instead of scanning the jobs directory, so stale or concurrent jobs cannot be attributed to the current run. Callback/checkpoint failures remain non-fatal, and displayed page-type labels are escaped. Validation: 109 dependent tests passed; Ruff is clean for all changed Python files.

RaghavChamadiya:
Thanks @lntutor, and sorry for the wait. The bug you are fixing is still exactly as you described it on main:

```python
jobs = js.list_jobs()
if jobs:
    failed_page_ids = jobs[0].failed_page_ids
```

`packages/cli/src/repowise/cli/commands/init_cmd/generation.py:374-376`. Taking `jobs[0]` means a second run in the same repo reports the previous run's failures, which is the misattribution you found. Threading the ids off the live checkpoint is the right fix, and escaping `page_type` before it reaches `console.print` is a real second bug worth having in the same change.

Two things before this can go in. The branch is conflicting now, since `run_repo_generation` picked up new logic around the resume path, so it needs `git fetch origin && git rebase origin/main`. And the diff carries a `docs/superpowers/plans/...md` file and some reformatting of unrelated `monkeypatch.setattr` calls. Dropping those keeps the change to the thing it is about, which makes it a faster review next time.

Ping me when it is rebased and I will take another look.

PR #1100, opened 2026-07-26
fix(providers): stop agent CLIs loading target-repo instruction files

### LINKED ISSUE #1093: codex_cli and opencode load the repo's AGENTS.md / CLAUDE.md and follow it as task instructions
Reported by @milaneliasde in #1089 (Problem 5), with measurements from a real 768-page run.

## What happens

Both agent-CLI providers run the agent inside the target repository, so the agent picks up that repo's `AGENTS.md` / `CLAUDE.md` and treats it as instructions for the documentation call. Two effects landed in stored pages:

- Tool-use narration before the first heading: `"I'll examine the actual source file to write an accurate spotlight page. Let me check..."`, in 25 of 145 model-written pages.
- Adoption of unrelated project workflow: `"I'm using the repository's Beads workflow guidance to ensure this one-shot documentation task follows the project's tracking conventions."`

Rendering a documentation page is not a coding task in that repo, and its contributor instructions should not apply.

## Where

- `packages/core/src/repowise/core/providers/llm/codex_cli.py:395-406` builds `codex exec --ephemeral --sandbox read-only --json --cd <repo>`.
- `packages/core/src/repowise/core/providers/llm/opencode.py:313` passes `--dir <repo>`.

Neither suppresses project instruction files. Nothing downstream trims the output either: the only post-processing applied to page content is `sanitize_mermaid` (`generation/mermaid_safety.py`).

## What to change

Two independent fixes, either is a useful PR on its own.

1. **Suppress the instruction files per provider.** Codex already takes `--config` (see `_codex_reasoning_config` for the existing pattern), so `project_doc_max_bytes=0` stops it loading `AGENTS.md`. Check the installed `opencode` for an equivalent before adding one; if there is none, the trim below carries that provider.

2. **Trim any preamble before the first heading.** Every page template starts with a `#` heading, so text before the first heading line is narration by definition. This belongs where it applies to all providers rather than in one provider class. Be conservative: only strip when a heading exists later in the output, never return an empty page.

Both changes shift the generated content, so pages regenerate on the next update. That is correct here.

## Acceptance

- A page generated with `codex_cli` in a repo that has an `AGENTS.md` starts at its first heading and does not reference that repo's workflow.
- Output that already begins with a heading is byte-identical to today.
- Output with no heading at all is passed through unchanged rather than emptied.

Real5K:
## Summary

- Suppress target-repo project docs for `codex_cli` via `project_doc_max_bytes=0`
- Set `OPENCODE_DISABLE_PROJECT_CONFIG=true` for the OpenCode provider
- Strip agent narration before the first markdown heading in the shared post-process path (passthrough when already clean / no heading)

## Related Issues

Fixes #1093

## Test Plan

- [x] `tests/unit/generation/test_mermaid_safety.py` - preamble strip + existing mermaid cases
- [x] `tests/unit/test_providers/test_codex_cli_provider.py` - `project_doc_max_bytes=0` in argv
- [x] `tests/unit/test_providers/test_opencode_provider.py` - disable-project-config env set
- [x] Focused suite: **58 passed**

## Checklist

- [ ] I have updated documentation if needed (N/A - no docs change required)

### DIFF
--- a/packages/core/src/repowise/core/generation/mermaid_safety.py
+++ b/packages/core/src/repowise/core/generation/mermaid_safety.py
@@ -346,6 +346,27 @@ def sanitize_mermaid(markdown: str) -> str:
     return _MERMAID_FENCE_RE.sub(_replace, markdown)
 
 
+_HEADING_LINE_RE = re.compile(r"(?m)^#{1,6}\s")
+
+
+def strip_leading_preamble(markdown: str) -> str:
+    """Drop agent narration before the first markdown heading.
+
+    Pages always start with ``# ...``. Text before that is tool-use chatter.
+    If there is no heading, or stripping would empty the page, return input
+    unchanged so clean output stays byte-identical.
+    """
+    if not markdown:
+        return markdown
+    match = _HEADING_LINE_RE.search(markdown)
+    if match is None or match.start() == 0:
+        return markdown
+    trimmed = markdown[match.start() :]
+    if not trimmed.strip():
+        return markdown
+    return trimmed
+
+
 def sanitize_pages(pages: list) -> int:
     """Run :func:`sanitize_mermaid` over a list of ``GeneratedPage``.
 
@@ -357,11 +378,12 @@ def sanitize_pages(pages: list) -> int:
         content = getattr(page, "content", None)
         if not content:
             continue
-        fixed = sanitize_mermaid(content)
+        fixed = strip_leading_preamble(content)
+        fixed = sanitize_mermaid(fixed)
         if fixed != content:
             page.content = fixed
             changed += 1
     return changed
 
 
-__all__ = ["sanitize_mermaid", "sanitize_pages"]
+__all__ = ["sanitize_mermaid", "sanitize_pages", "strip_leading_preamble"]
--- a/packages/core/src/repowise/core/generation/mermaid_safety.py
+++ b/packages/core/src/repowise/core/generation/mermaid_safety.py
@@ -186,6 +186,27 @@ def sanitize_mermaid(markdown: str) -> str:
     return _MERMAID_FENCE_RE.sub(_replace, markdown)
 
 
+_HEADING_LINE_RE = re.compile(r"(?m)^#{1,6}\s")
+
+
+def strip_leading_preamble(markdown: str) -> str:
+    """Drop agent narration before the first markdown heading.
+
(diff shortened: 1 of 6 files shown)


RaghavChamadiya:
Thanks @Real5K. The problem this fixes is still live on main, and I checked rather than took the issue's word for it: neither `project_doc_max_bytes` nor `OPENCODE_DISABLE_PROJECT_CONFIG` appears anywhere in the tree, so `codex_cli.py` and `opencode.py` still hand the agent a working directory whose `AGENTS.md` / `CLAUDE.md` it will read as instructions.

That matters more than it looks. Those files are written for a coding agent doing tasks in the repo, so an instruction like "always run the tests before finishing" is being handed to a process whose only job is to write a documentation page. The generated page is then shaped by the target repo's own conventions rather than ours, and it fails differently per repo, which is the worst kind of bug to reproduce.

The branch is conflicting now. `packages/core/src/repowise/core/generation/mermaid_safety.py` grew from about 230 lines to 367 since you opened this, so your `strip_leading_preamble` hunk no longer lands at that offset. The provider hunks should still apply as-is, `_build_command` is still at `codex_cli.py:403`.

`git fetch origin && git rebase origin/main` and push, and I will review it properly on the rebased diff.

Real5K:
@RaghavChamadiya 

Rebased onto latest main. Conflicts in `test_mermaid_safety.py` resolved.

Kept main's mermaid tests + the preamble coverage. Ready for review.

RaghavChamadiya (approved):
Both provider knobs land in the right place on the rebased diff, and clean output stays byte-identical.

RaghavChamadiya:
Thanks @Real5K, and thanks for keeping main's mermaid tests alongside your preamble coverage instead of taking one side of the conflict wholesale.

Checked the rebased diff. Both provider knobs land where they should: `--config project_doc_max_bytes=0` goes into `_build_command` in `codex_cli.py`, and `OPENCODE_DISABLE_PROJECT_CONFIG=true` into the subprocess env in `opencode.py`. That is the whole fix for the instruction-file problem, and it is two lines, which is the right size for it.

`strip_leading_preamble` is a good companion for the same reason. Even with project config off, an agent CLI narrates before it answers, and returning the input unchanged when there is no heading or when stripping would empty the page means clean output stays byte-identical. The three tests pin exactly those branches.

Full CI matrix is green on 3.11, 3.12 and 3.13. Merging.

PR #1102, opened 2026-07-26
fix(generation): localize structural wiki pages

### LINKED ISSUE #1092: Structural page templates hardcode English, so --language only reaches subsystem prose
Reported by @milaneliasde in #1089 (Problem 3).

## What happens

On a repo configured with `language: de`, only the model-written subsystem pages come back in German. Every other page is English, because it is rendered from a Jinja template with English headings baked in and no model involved.

The reporter measured 623 of 768 pages on 0.34.0. Since 0.35.0 rebuilt generation (#1023, #1032) the share is higher: `module_page` is the only page type a model writes, so effectively the whole wiki is unaffected by the language setting.

## Where

Every template under `packages/core/src/repowise/core/generation/templates/` that renders page *content* rather than a prompt:

- `file_page.j2` (`## Overview`, `## Public API`, `## Depends on`, `## Used by`, and prose like `It exposes N public symbols`, `Imported by N files`)
- `layer_page.j2` (`## Architecture`, `## Entry points`, `## Key files`, `## Edge connectors`, `## Neighbouring layers`)
- `symbol_spotlight.j2` (`## Where it is used`, `## Implementation`)
- `infra_page.j2`, `scc_page.j2`, `api_contract.j2`
- `stub/module_page.j2` (`## Parts of this subsystem`, `## Entry Points`, ...)

These render through `StructuralRenderMixin._render_page` (`page_generator/structural.py:289-303`), which calls `self._render(template, style_prefix=False, **render_kwargs)`. The prompt path gets a language instruction in the system prompt (`page_generator/core.py:379-387`); the content path gets nothing.

## What to change

The resolved language is already on the generator as `self._language`, validated against `SUPPORTED_LANGUAGES` in `core.py:370-381`. Two pieces:

1. Pass the resolved language code into the render context for structural templates (one line in `_render_page`, or in `_render` when `style_prefix=False`).
2. Select headings and fixed phrases from a per-language label map, with English as the fallback for any code or key that is missing. The reporter did exactly this in their own copy and reports it was one line plus a label map.

Keep code, file paths and symbol names untranslated, matching what the system prompt already tells the model to do.

Starting with `file_page.j2` alone is a perfectly good first PR. The other templates can follow the same pattern.

## Notes

- The template source is folded into the structural `content_hash` (`_structural_content_hash`), so changing a template makes `update` treat those pages as rendered by an older configuration and redo them. That is the intended behaviour, but expect the reuse-hash tests in `tests/unit/generation/test_styles.py` to be where surprises show up.
- A language with no label map entry must render English, never a partial mix or a KeyError. Templates render under `StrictUndefined` (see #1068 for what happens when a key is missing).

## Acceptance

- A repo indexed with `--language de` renders German headings on file, symbol, layer, infra, cycle and API pages.
- An unsupported or absent language code renders exactly what it renders today.
- Code, paths and symbol names are untouched in every language.

lntutor:
Closes #1092.

## Summary
- Localize all deterministic structural page content, shared footer, page titles, and searchable question blocks through a complete English-backed label catalog.
- Add German labels while preserving paths, symbols, signatures, source, and other extracted identifiers unchanged.
- Preserve the latest structural-page behavior from `main`, including conditional sections, render keys, and question-shaped retrieval text.
- Add regressions for German rendering plus complete English fallback across the supported structural page types.

## Test plan
- [x] `uv run pytest tests/unit/generation -q` — 990 passed.
- [x] Focused localization and template suite — 103 passed.
- [x] Ruff passes for the conflict-resolved Python files.
- [x] `git diff --check` passes.

### DIFF
--- a/docs/reference/CONFIG.md
+++ b/docs/reference/CONFIG.md
@@ -247,8 +247,12 @@ regenerates the wiki). Power users can define their own style under
 Note: hand-editing `wiki_style` here and running `update` does not regenerate
 existing pages, use `restyle`.
 
-`language` controls the natural language of generated wiki content (prose only;
-code, file paths, and symbol names stay untranslated). Set it with
+`language` controls the natural language of generated wiki content (code, file
+paths, and symbol names stay untranslated). It reaches both the model-written
+pages and the headings and fixed sentences of the template-rendered ones — file,
+symbol, infrastructure, cycle and API-contract pages — for the languages that
+have a label catalog. A language without one keeps the English headings on those
+pages while the model-written prose is still translated. Set it with
 `init --language <code>` or pick it in advanced interactive mode; it persists
 here so `update` regenerates changed pages in the same language. Unknown codes
 fall back to English with a warning. As with `wiki_style`, changing it later
--- a/packages/core/src/repowise/core/generation/languages.py
+++ b/packages/core/src/repowise/core/generation/languages.py
@@ -26,3 +26,17 @@ SUPPORTED_LANGUAGES = {
     "ar": "Arabic",
     "hi": "Hindi",
 }
+
+
+def sanitize_language_code(language: str | None) -> str:
+    """Return *language* lowered, stripped and reduced to ``[a-z0-9_]``.
+
+    The configured code reaches a system prompt and a label lookup, so it is
+    scrubbed before either sees it: without this a config value could inject
+    newlines and extra instructions into the prompt. Validation against
+    :data:`SUPPORTED_LANGUAGES` is the caller's, because the two callers
+    differ on what an unknown code means: the prompt path warns, the label
+    path silently falls back to English.
+    """
+    raw = (language or "en").lower().strip()
+    return "".join(ch for ch in raw if ch.isalnum() or ch == "_")
--- a/packages/core/src/repowise/core/generation/page_generator/core.py
+++ b/packages/core/src/repowise/core/generation/page_generator/core.py
@@ -33,6 +33,7 @@ from ..context.evidence import (
 )
 from ..context_assembler import ContextAssembler, FilePageContext
 from ..house_vocabulary import cell
+from ..languages import sanitize_language_code
 from ..models import (
     MODEL_PAGE_CONFIDENCE,
     GeneratedPage,
@@ -41,6 +42,7 @@ from ..models import (
     compute_source_hash,
 )
 from ..report import reset_house_terms
+from ..structural_labels import resolve_structural_labels, structural_page_title
 from ..styles import ONBOARDING_PAGE_TYPE, resolve_style
 from .helpers import _extract_summary, _now_iso, collapse_empty_duplicate_headings
 from .pertype import PerTypeGenerationMixin
@@ -208,6 +210,13 @@ class PageGenerator(PerTypeGenerationMixin, StructuralRenderMixin):
         # quotes a shell pipeline. Without this every column to the right of
         # one shifts.
         self._jinja_env.filters.setdefault("table_cell", cell)
+        # Fixed copy for the deterministic templates, in the configured
+        # language. A global rather than a per-render kwarg because every
+        # structural render path would otherwise have to remember to pass it,
+        # and under StrictUndefined the one that forgot would raise. Assigned,
+        # not setdefault-ed: an env handed to two generators must end up with
+        # the language of the one rendering through it.
+        self._jinja_env.globals["labels"] = resolve_structural_labels(self._language)
 
     # ------------------------------------------------------------------
     # generate_all — orchestration (delegates to orchestrate.py)
@@ -310,7 +319,7 @@ class PageGenerator(PerTypeGenerationMixin, StructuralRenderMixin):
         page = self._structural_page(
             page_type="file_page",
             target_path=parsed.file_info.path,
-            title=f"File: {parsed.file_info.path}",
+            title=structural_page_title(self._language, "file_page", parsed.file_info.path),
             template="file_page.j2",
             subject_hash=parsed.content_hash or "",
             ctx=ctx,
@@ -403,11 +412,10 @@ class PageGenerator(PerTypeGenerationMixin, StructuralRenderMixin):
         base_system = base_system + self._style.system_prompt_suffix(
             is_onboarding=page_type == ONBOARDING_PAGE_TYPE
         )
-        # Sanitize the configured language code: lower, strip, drop anything that isn't
-        # alphanumeric or underscore. Prevents user-supplied config from injecting
-        # newlines or extra instructions into the system prompt.
-        raw = (self._language or "en").lower().strip()
-        lang_code = "".join(ch for ch in raw if ch.isalnum() or ch == "_")
+        # Sanitized before it reaches the prompt: user-supplied config must not be
+        # able to inject newlines or extra instructions here. Shared with the
+        # structural-label lookup so both paths agree on what " DE " means.
+        lang_code = sanitize_language_code(self._language)
         if lang_code not in SUPPORTED_LANGUAGES:
             if lang_code != "en":
                 log.warning("unknown_language_code", code=lang_code, fallback="en")
--- a/packages/core/src/repowise/core/generation/page_generator/levels.py
+++ b/packages/core/src/repowise/core/generation/page_generator/levels.py
@@ -187,7 +187,7 @@ async def build_level2_coros(run: _GenerationRun) -> list[tuple[str, Any]]:
     return coros
 
 
-def _scc_titles(scc_groups: list[Any]) -> dict[str, str]:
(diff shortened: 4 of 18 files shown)


lntutor:
CI exposed three direct `PageGenerator._render()` calls in the wiki-style tests that did not provide the new Jinja `labels` global. Fixed in e111d2cb by setting the English labels in those test-only direct-render fixtures. The original failing cases and the expanded affected generation suite pass locally (114 tests); the full matrix is rerunning.

lntutor:
Rebased onto current `main`. Upstream has retired layer pages, so the conflict resolution intentionally drops the obsolete layer generator/template/tests and keeps localization for the supported structural pages: file, symbol, infrastructure, SCC, and API contract pages. Unknown-language English fallback remains covered. Validation: 168 affected generation tests passed; Ruff is clean for all changed Python files.

lntutor:
Rebuilt this branch as one linear commit on current `main` to satisfy the repository linear-history rule and resolve the conflict without retaining the prior merge commit. Exact pushed head: `7de2391a60a214e37f9017c394734d7e1b4118f9` (base `b501a957581a77f54fc63fe7273610e608d6a412`, ahead 1, behind 0). The sole conflict in the latest-base reconstruction combined upstream structural/stub confidence metadata with this PR’s structural label resolver; retired layer-page localization remains excluded. Exact-tree local verification: all 1,023 generation tests passed, all 3 update-renderer gate tests passed, Ruff passed, changed Python files passed `ruff format --check`, and `git diff --check` passed. Fresh repository CI is now running on this SHA.

lntutor:
Rebased this branch onto current `main` and force-pushed the exact new head: `c5c21f08d80ebbe61a7cc739722d58209ed849c4`.

Conflict resolution kept the localized structural-page work while preserving the current mainline behavior for empty file/symbol deterministic pages and spotlight render keys.

Exact-tree local verification on this pushed head:
- `uv run pytest tests/unit/generation/test_deterministic_templates.py tests/unit/generation/test_structural_labels.py tests/unit/generation/test_templates.py tests/unit/generation/test_styles.py tests/unit/generation/test_kg_aware_file_pages.py tests/unit/generation/test_question_text_reaches_search.py` -> 114 passed
- `git diff --check` passed before the rebase continue

Fresh CI had not reported checks yet when I posted this.

lntutor:
Follow-up on the new exact head `002c76824a6f290bc0648180db93420b3ea66fed`: this supersedes the prior `c5c21f0...` push only to fix the fresh Ruff gate.

What changed on this head:
- removed the duplicate unused `resolve_structural_labels` import in `tests/unit/generation/test_kg_aware_file_pages.py`
- applied Ruff formatting to `packages/core/src/repowise/core/generation/page_generator/pertype.py`

Exact-head local verification:
- `uv run ruff check` on the changed generation files/tests -> clean
- `uv run ruff format --check` on the changed generation files/tests -> clean
- `uv run pytest tests/unit/generation/test_deterministic_templates.py tests/unit/generation/test_structural_labels.py tests/unit/generation/test_templates.py tests/unit/generation/test_styles.py tests/unit/generation/test_kg_aware_file_pages.py tests/unit/generation/test_question_text_reaches_search.py` -> 114 passed

GitHub had already cleared the previous failed check set when I posted this; fresh CI should repopulate on this SHA.

lntutor:
Fresh CI has now completed successfully on exact head `002c76824a6f290bc0648180db93420b3ea66fed`: Python 3.11, 3.12, and 3.13 suites, full-repository Ruff, web lint/type checks, VS Code extension tests, and the Structurizr gate are all green (14 passed, 0 failed; integration is intentionally skipped for pull requests). GitHub reports the PR mergeable; the remaining gate is maintainer/code-owner review.

RaghavChamadiya:
Thanks lntutor, and sorry this sat. I went through it properly today and I want this landed, so here is everything you need in one place.

The design is right and I would not change it. A complete English catalog copied first and then overlaid with the language map means a missing key falls back to English rather than blowing up under StrictUndefined, which is exactly the acceptance criterion in #1092 and the easiest part of this to get wrong. Building the page title from the same catalog rather than a separate path is also the right instinct. Keep the interpolation where it is; the readability cost in the templates is the normal price of this and it is worth paying.

## The rebase is smaller than the diff makes it look

main has moved 153 commits since your base, so GitHub says DIRTY, but almost none of that lands on you. Comparing your changes against what main did to the same files:

- `page_generator/core.py`: main +122/-1, you +3/-1
- `page_generator/pertype.py`: main +178/-19, you +11/-4
- `page_generator/structural.py`: main +56/-8, you +7/-1
- `structural_labels.py` is a new file, so nothing can conflict
- `_footer.j2`, `api_contract.j2`, `infra_page.j2`, `symbol_spotlight.j2`: main did not touch them at all

Your Python edits are small enough that they should re-apply almost mechanically. Only three files genuinely collide: `file_page.j2`, `scc_page.j2` and `test_deterministic_templates.py`.

## Two new strings main added that need catalog entries

While you were away main added English copy to two of the templates you localize, so these need new keys:

1. `file_page.j2` gained an `## In the code` section, gated on `ctx.file_vocabulary`. It renders per-file source vocabulary and sits below the questions block.
2. `scc_page.j2` gained `**Cycle id:**` as a standalone line, and the `page_title` default is now `"Circular Dependency: " ~ ctx.scc_id`, so the fallback title string needs to come from the catalog too.

## Three smaller things

- Please drop `docs/superpowers/plans/2026-07-26-structural-template-localization.md` and the matching spec file. That directory does not exist on main and those read as working notes rather than repo content.
- `GERMAN_LABELS` is keyed by language code with the German map nested inside it, so `GERMAN_LABELS.get(language)` reads oddly today and stops making sense the moment someone adds a third language. `LOCALIZED_LABELS` or similar.
- Expect `tests/unit/generation/test_styles.py` to be where surprises land. Template source folds into `_structural_content_hash`, so this makes `update` treat every structural page as rendered by an older configuration and redo them. That is intended behaviour, not a bug to suppress, but it is worth an explicit assertion that the hash changed for the right reason rather than letting a reuse test quietly flip.

Rebase onto current main, and ping me when CI is green. I will review it the same day. If you would rather not carry the rebase, say so and I will take it from here, but the work left is genuinely a few hours and I would rather it stayed yours.

swati510 (approved):

RaghavChamadiya:
Thanks again for this, and for carrying it through several rebases. I said I would take it from here if the rebase got tedious, and since it had been sitting a while I went ahead and did that today rather than ask you for another one. The commit is authored under your name, so this stays your contribution.

## What I kept

The design is unchanged. Complete English catalog copied first, language map overlaid on top, page title built from the same catalog, interpolation left in the templates. That was right and I did not touch it.

## The three things I asked for

- The two `docs/superpowers/` files are gone.
- `GERMAN_LABELS` is now `LOCALIZED_LABELS`, keyed by code with the map directly under it.
- The reuse hash has an explicit assertion now. `test_a_template_change_re_renders_structural_pages_for_the_right_reason` checks the same subject hashes differently under `en` and `de`, so a future change cannot quietly leave a German repo sitting on English pages.

The two strings main added while you were away are in as `in_the_code` and `cycle_id`, and the `page_title` default on `scc_page.j2` reads from the catalog too.

## What I found while re-applying

Worth naming, since three of these would have shipped:

- `is_a` as a flat label lost the a/an rule, so an `enum` or `interface` symbol rendered "is a enum". The symbol overview is one sentence now with an `{article}` placeholder. English keeps the rule, German omits the placeholder, and `str.format` ignores what a translation does not use.
- The infra overview composite emitted a double space when a file declares no targets, because the empty targets sentence still had a space on either side. Split into intro, targets and outro so the conditional space stays where it was.
- `tests/unit/generation/test_file_vocabulary.py` builds its own StrictUndefined environment and renders `file_page.j2`. It did not exist when you forked, so it would have failed CI on the new global.
- `levels.py::_scc_titles` builds cycle titles for the whole run at once, for uniqueness, and had its own hardcoded "Circular Dependency". Your change covered the per-page fallback in `pertype.py`, so in practice a German repo would still have shown English cycle titles.
- I moved the labels from a `_render` kwarg to a Jinja global set once in `PageGenerator.__init__`. Same effect, but the stub pages and any render path added later get it for free, and the test fixtures stop needing to know about it.
- `sanitize_language_code` moved to `languages.py` and is shared with `_build_system_prompt`. A config value of `" DE "` localized the prose but would not have found a label catalog.

I also localized the questions block, heading and question text both. Same bug as the rest, and it is the text that feeds search, so a German reader asking in German should match it.

## Verification

English rendering is byte identical to main across every conditional branch of all five templates plus the footer, checked case by case, so no existing wiki changes wording. 1382 generation unit tests pass, the cli, persistence and ingestion units pass, the deterministic generation integration test passes, `ruff check .` is clean.

## Left for a follow-up

The templates under `templates/stub/` still hardcode English and carry their own footer, so a keyless German repo would get German sole-renderer pages next to English stubs. Different template tree and outside what #1092 asked for. I will open an issue.

PR #1106, opened 2026-07-26
docs(website): sync CLI and Claude plugin pages with shipped surface

Ayush7614:
## Summary
- Expand \`website/cli-reference.md\` with shipped analysis commands: \`health\`, \`risk\`, \`security\`, \`coverage\`, \`impacted-tests\`, plus \`workspace\` and \`distill\`.
- Align \`website/claude-code-plugin.md\` slash commands with the plugin (including coverage/impacted-tests/health/risk/dead-code/decision/doctor), fix marketplace install path, and document six skills.

## Test plan
- [x] Flags cross-checked against \`repowise <cmd> --help\`
- [x] Slash-command list cross-checked against \`plugins/claude-code/commands/\`
- [x] Marketplace path matches \`plugins/claude-code/README.md\`

PR #1107, opened 2026-07-26
feat(claude-plugin): add /repowise:security slash command

Ayush7614:
## Summary
- Add \`plugins/claude-code/commands/security.md\` mirroring \`repowise security scan --history\` (honest about the no-\`--history\` stub).
- Update plugin README / CHANGELOG and bump version to \`0.36.1\` (marketplace synced).
- Extend structural command tests to expect \`security\`.

## Test plan
- [x] \`pytest tests/unit/cli/test_claude_plugin.py\` passes
- [x] Flags/modes cross-checked against \`repowise security scan --help\`

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -14,7 +14,7 @@
       "name": "repowise",
       "source": "./plugins/claude-code",
       "description": "Codebase intelligence for Claude Code. Indexes your repo into five layers (Graph, Git, Docs, Decisions, Code Health) and gives Claude deep understanding of architecture, ownership, hotspots, decisions, and defect risk — fewer greps, fewer file reads, lower cost per query.",
-      "version": "0.36.1",
+      "version": "0.36.0",
       "category": "productivity",
       "keywords": [
         "codebase",
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -14,7 +14,7 @@
       "name": "repowise",
       "source": "./plugins/claude-code",
       "description": "Codebase intelligence for Claude Code. Indexes your repo into five layers (Graph, Git, Docs, Decisions, Code Health) and gives Claude deep understanding of architecture, ownership, hotspots, decisions, and defect risk — fewer greps, fewer file reads, lower cost per query.",
-      "version": "0.36.0",
+      "version": "0.36.1",
       "category": "productivity",
       "keywords": [
         "codebase",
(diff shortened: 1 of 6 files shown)


RaghavChamadiya:
Thanks, this is a clean addition and the command doc is accurate. I checked it against `security_cmd.py` rather than just reading it: the stub message matches verbatim, all six flags exist as documented, `secrets_only=not all_patterns` confirms the "secrets only by default, `--all-patterns` for code smells" framing, and the persistence/idempotency note matches the command's own closing output. Being honest about the no-`--history` stub is the right call.

One thing to pull out before this merges: **drop the version bump.** The next release is 0.37.0 and I will handle the version churn myself, so please revert `marketplace.json`, `plugins/claude-code/.claude-plugin/plugin.json`, and the `## 0.36.1` CHANGELOG heading back to 0.36.0. Worth knowing that the CHANGELOG has no 0.36.0 section at all today, so a new heading would sit straight on top of 0.35.0 and make the gap look intentional. I would rather fold this entry into the 0.37.0 section when I cut it.

Everything else (the command file, the README table row, and the test set addition) can stay as is.

Separately, if this lands, `website/claude-code-plugin.md` will be the one place that does not list `/repowise:security`. #1106 syncs that page and predates this command, so it is worth adding the section there rather than leaving the website a command behind.

Ayush7614:
Thanks for the review @RaghavChamadiya  — addressed.

- Reverted `.claude-plugin/marketplace.json` and `plugins/claude-code/.claude-plugin/plugin.json` to `0.36.0`.
- Removed the `## 0.36.1` CHANGELOG section so you can fold `/repowise:security` into `0.37.0` when you cut it.
- Left `commands/security.md`, the README table row, and the test expectation as-is.

On the website note: I'll add `/repowise:security` on #1106 so that page does not lag this command.

swati510 (approved):
Checked each documented flag against `packages/cli/src/repowise/cli/commands/security_cmd.py` — `--history` (:34), `--since` (:40), `--to` (:45), `--path` (:50), `--all-patterns` (:55), `--output` (:63) all exist with the documented semantics. The "without `--history` it prints a hint and exits" note matches the early return at :86-95, and secrets-only-by-default matches `secrets_only=not all_patterns` (:141).

Driving the CLI via `allowed-tools: Bash, Read` is the right pattern here — there is no security MCP tool in the server, and all twelve existing command files use the same pair. Test addition is correct and CI is green.

PR #1108, opened 2026-07-26
test(cli): cover repowise security scan stub and history wiring

Ayush7614:
## Summary
- Add \`tests/unit/cli/test_security_cmd.py\` for the public \`repowise security scan\` surface.
- Cover the no-\`--history\` stub message, help flags, JSON history output, and \`secrets_only\` / \`--all-patterns\` / \`--since\` / \`--to\` wiring.

## Test plan
- [x] \`pytest tests/unit/cli/test_security_cmd.py\` — 4 passed

### DIFF
--- a/tests/unit/cli/test_security_cmd.py
+++ b/tests/unit/cli/test_security_cmd.py
@@ -1,0 +1,203 @@
+"""CLI coverage for ``repowise security scan``."""
+
+from __future__ import annotations
+
+import json
+from dataclasses import dataclass, field
+from pathlib import Path
+from types import SimpleNamespace
+from typing import Any
+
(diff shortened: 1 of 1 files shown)


RaghavChamadiya:
Solid. I applied it locally: 4 passed. Checked the monkeypatch targets rather than trusting the green, since heavily-mocked CLI tests pass vacuously easily — all patch points are correct: the command does its HistorySecurityScanner / persistence / crud imports inside the function body, so patching module attributes is the right hook, and the JSON payload keys asserted match what _do() returns. Merging

RaghavChamadiya (approved):

PR #1156, opened 2026-07-29
fix(docs): one breadcrumb, correct reader height, wrapping backlinks

swati510:
Three layout defects on the docs reader, each verified in source before the fix.

### One breadcrumb trail, not two

`repos/[id]/layout.tsx` renders a route breadcrumb; `docs-reader.tsx` renders a second one ~12px below it. They are not duplicates of each other — the route trail is `Dashboard / repo / Docs`, the reader's is the wiki page hierarchy — but on the reader the route trail is always those same three crumbs, because every wiki page lives at the same `/docs` URL keyed by `?page=`. Only one of the two can say where you are, so the reader's hierarchy is the one kept.

The predicate is a small pure module (`repo-breadcrumb-route.ts`) so it can be tested without a DOM, matching the existing `repo-breadcrumb-label.ts` next to it. `/docs/coverage` is an ordinary page with no trail of its own and keeps the route crumbs.

### The reader fills the space left to it

`app/layout.tsx` is already a flex column that subtracts anything a repo stacks above the page — the reindex hint, the active-job banner, the upgrade banner — and its comment says so explicitly. The docs page then asked for `h-screen` inside that box, so with any banner showing it overflowed by exactly that banner's height and pushed the reader's bottom chrome below the fold. Now `flex-1 min-h-0`.

### Backlink titles wrap

`backlinks-panel.tsx` used `truncate`. In the ~260px rail that clips a title at the path segment that identifies it — the same failure the Related list already fixed by wrapping, with a comment saying why. Backlinks now match it.

While testing that, a second defect surfaced: the link tooltip was bound to the page *kind*, so every row in a list of files had the tooltip "File". It now carries the page's own name, and the kind moves to an `aria-label` on the icon that was already showing it.

### Reliability

`RepoBreadcrumb` returned `null` silently when the pathname carried no repo. Reaching that branch means the route moved out from under the layout and every page below it lost its trail, so it now logs before returning.

### Tests

- `repo-breadcrumb.test.ts` — the reader stands down, `/docs/coverage` does not, and a repo id ending in `docs` is not mistaken for the reader route.
- `backlinks-panel.test.tsx` (new) — title is not clipped, tooltip names the page, empty input renders nothing.

Both were written failing first. The height fix carries no unit test — it is a single class on a shell that mounts the whole explorer — and is covered by typecheck, build, and the visual check.

### Gates

- `packages/ui`: 1003 tests pass (141 files)
- `packages/web`: 12 tests pass
- `npm run type-check`: clean across all workspaces
- `npm run lint`: no warnings or errors

### DIFF
--- a/packages/ui/__tests__/wiki/backlinks-panel.test.tsx
+++ b/packages/ui/__tests__/wiki/backlinks-panel.test.tsx
@@ -1,0 +1,38 @@
+import { describe, expect, it } from "vitest";
+import { render, screen } from "@testing-library/react";
+import { BacklinksPanel } from "../../src/wiki/backlinks-panel.js";
+import type { Backlink } from "../../src/wiki/wiki-links-types.js";
+
+const LONG_TITLE = "File: packages/server/src/repowise/server/routers/overview.py";
+
+function backlink(over: Partial<Backlink> = {}): Backlink {
+  return {
+    source_page_id: "file_page:a.py",
+    source_title: LONG_TITLE,
+    source_page_type: "file_page",
+    ...over,
+  } as Backlink;
+}
+
+describe("BacklinksPanel titles", () => {
+  it("wraps the title instead of clipping it to one line", () => {
+    render(<BacklinksPanel backlinks={[backlink()]} repoId="r1" />);
+
+    const title = screen.getByText(LONG_TITLE);
+    // `truncate` is one-line ellipsis. In a ~260px rail it cuts the path at
+    // exactly the part that identifies the page, matching the Related list's
+    // rule that a title must stay readable.
+    expect(title.className).not.toContain("truncate");
+  });
+
+  it("names the source page in the link tooltip, not just its kind", () => {
+    render(<BacklinksPanel backlinks={[backlink()]} repoId="r1" />);
+
+    expect(screen.getByRole("link")).toHaveAttribute("title", LONG_TITLE);
+  });
+
+  it("renders nothing when there are no backlinks", () => {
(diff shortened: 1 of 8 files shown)


swati510:
Pushed 8d1a4c60 — the height change as written did not hold.

`flex flex-1 min-h-0` needs the parent to be a flex container. The parent here is the wrapper `PageTransition` renders (`page-transition.tsx:27`), which is `flex-1` and nothing else — a flex item of `main`, but not a flex container itself. So `flex-1` on the docs shell was inert and the box fell back to content height. `DocsExplorer`'s root then asks for `h-full` (`docs-explorer.tsx:317`), and a percentage height against an auto-height parent resolves to auto, so the reader collapses onto its content.

`h-screen` was wrong for the reason the description gives, but it was a definite height that `h-full` below it could resolve against. `h-full` keeps that property and takes the banners into account, because the wrapper's height is settled by the flex layout of `main`.

Class list moved into `docs-reader-shell.ts` with a test, matching `repo-breadcrumb-route.ts` next door — this package has no DOM test environment, so a pure module is the only way to gate it. Note that the test guards the class list, not the layout; the rendered page still wants a look with a banner showing.

Two smaller things left alone, both yours to take or leave:
- `repo-breadcrumb.tsx:22` logs from inside the render body, so it can repeat for a single visible update. An effect or a once-guard keeps the warning without the noise.
- the backlinks icon carries `aria-label` on a bare lucide `<svg>`, which screen readers commonly ignore without `role="img"` — as it stands the page kind is dropped for assistive tech rather than moved off the tooltip.

RaghavChamadiya (approved):

PR #1159, opened 2026-07-29
feat(bench): score answer retrieval against a set that refuses to load broken

swati510:
## Why

The two question sets that measure the answer tool live nowhere in this repository. They survive only as numbers in old commit messages. That makes every retrieval change unfalsifiable — there is no way to show a change helped without re-deriving the measurement by hand each time, and no way for a reviewer to check the claim.

This PR is the part of the harness that needs **no index and no network**, so it can be tested properly and land on its own. The runner and the question sets follow in separate PRs.

## Where it lives

`repowise_bench/` sits at the repo root, outside the three published wheel roots (`packages/core`, `packages/cli`, `packages/server`), so it is importable in development and under pytest but never ships to users. `pythonpath = ["."]` already makes it importable; tests live in `tests/unit/bench/` because that is what CI collects.

## What it does

**Scoring** — `recall_at_k`, `reciprocal_rank`, and an aggregate `score_retrieval`. Two things it enforces rather than assumes:

- **Duplicate page ids collapse to their first position before ranking.** A retrieval that returns the same page five times would otherwise fill the top-5 window, and a page sitting at rank 6 would read as an honest miss.
- **A question with no expected pages raises.** Scoring it zero would let a malformed set drag the average down with nothing to show that it had.

`RetrievalScores` also carries `n_empty_results`, separating "retrieval found the wrong pages" from "retrieval found nothing at all". The two have different causes and the mean hides the difference.

**The loader** raises on every malformed input: a missing file, a file with no questions, a line that does not parse, a blank or missing field, a duplicate question id, and any unknown key.

That last one is the one that matters. Writing `expected_page_id` instead of `expected_page_ids` must not parse as a question that expects nothing — the run would still finish and still print a score, and the score would be wrong in the direction that looks like good news.

The set is JSONL rather than one JSON array so a diff shows exactly which questions a change added or reworded.

## Tests

33 tests, failing before the modules existed and passing after. Every expected value is hand-computed in a comment or docstring — if an assertion needs a calculator, the test is wrong.

Gates: `pytest tests/unit/bench` 33 passed · `ruff check` clean · `ruff format --check` clean · `mypy repowise_bench` clean.

### DIFF
--- a/repowise_bench/__init__.py
+++ b/repowise_bench/__init__.py
@@ -1,0 +1,6 @@
+"""Benchmark and evaluation harnesses for repowise.
+
+This package is deliberately outside the three published wheel roots
+(``packages/core``, ``packages/cli``, ``packages/server``), so it is importable
+in development and in tests but never ships to users.
+"""
--- a/repowise_bench/eval/__init__.py
+++ b/repowise_bench/eval/__init__.py
@@ -1,0 +1,7 @@
+"""Evaluation harness for the answer tool.
+
+Two question sets drive it: a retrieval set scored without a model
+(recall@k, MRR) and a gold set scored by a judge. This module holds the parts
+that need neither an index nor a network: loading a question set and the
+scoring maths.
+"""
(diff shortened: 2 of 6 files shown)


swati510:
Wrong repository — the eval harness belongs in repowise-bench, which is a separate repo. Moving it there.

PR #1173, opened 2026-07-30
docs(architecture): align chat tool registry with the live 7-tool set

Ayush7614:
## Summary
- Rewrite \`docs/architecture/chat.md\` to the **7** tools actually registered in \`chat_tools.py\` (drop invented \`get_dependency_path\` / \`get_architecture_diagram\`; add \`get_change_risk\`).
- Fix ARCHITECTURE §12 chat counts and the broken \`CHAT.md\` link.
- Align chat \`get_dead_code\` schema default with \`RISK_CAP_CONFIDENCE\` (0.4).
- Add \`tests/unit/server/test_chat_tool_registry.py\` drift guard.

## Why
MCP website/package docs were fixed earlier; chat architecture docs still described a pre-consolidation 10-tool surface. Not tied to any open issue.

## Test plan
- [x] \`pytest tests/unit/server/test_chat_tool_registry.py\`

### DIFF
--- a/docs/architecture/ARCHITECTURE.md
+++ b/docs/architecture/ARCHITECTURE.md
@@ -1154,9 +1154,10 @@ file, tokens used, estimated cost, estimated time remaining).
 repowise includes an interactive chat interface that lets users ask questions about
 their codebase and receive answers grounded in the wiki, dependency graph, git
 history, and architectural decisions. The chat agent uses whichever LLM provider
-the user has configured and has access to all 11 MCP tools.
+the user has configured and has access to **7 tools** from the MCP surface
+(see [`chat.md`](chat.md) — not the full 11-tool MCP default).
 
-See [`docs/CHAT.md`](CHAT.md) for the full technical reference covering the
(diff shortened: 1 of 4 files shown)


swati510 (approved):
Checked the seven names and the artifact-type mappings against `packages/server/src/repowise/server/chat_tools.py:244-252` and `_TOOL_SCHEMAS` — exact match, including `get_change_risk → risk_report`. Removing `get_dependency_path` and `get_architecture_diagram` is correct; neither is in the chat registry. `docs/CHAT.md` really doesn't exist, so that link was dead.

Good catch on the `min_confidence` schema default — `tool_dead_code.py:219` has used `RISK_CAP_CONFIDENCE` (0.4) for a while, so the advertised 0.5 was misleading the model.

One leftover for a follow-up, not this PR: the frontend still maps the two removed tools — `packages/ui/src/chat/tool-call-block.tsx:28`, `packages/ui/src/chat/source-citations.tsx:114`, `packages/types/src/chat.ts:216`. The docs are right now; those branches are dead.

PR #1177, opened 2026-07-30
docs(examples): add change-risk CLI walkthrough

Ayush7614:
## Summary
- Add \`examples/risk/README.md\` for \`repowise risk\` (HEAD / \`base..head\`, \`--ext\`, \`-x\`, \`--format json\`).
- Index it in \`examples/README.md\` and link \`CHANGE_RISK.md\`.

## Why
Risk is a first-class no-LLM CLI surface (QUICKSTART / USER_GUIDE) but \`examples/\` had no walkthrough after health/security/distill. Not tied to any open issue.

## Test plan
- [x] \`repowise risk HEAD\` runs locally
- [x] Skim flags against \`docs/reference/CLI_REFERENCE.md\`

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -11,6 +11,7 @@ self-contained README under `examples/<name>/`.
 | [distill/](distill/) | `distill` / `expand` / `saved` — compress command output (no LLM key) |
 | [health-coverage/](health-coverage/) | Code health, coverage ingest, and impacted-tests (no LLM key) |
 | [opencode/](opencode/) | OpenCode as the LLM provider for wiki generation |
+| [risk/](risk/) | `repowise risk` change / PR defect scoring (no LLM key) |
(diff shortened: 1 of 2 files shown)


swati510 (approved):
Checked every flag and label against `packages/cli/src/repowise/cli/commands/risk_cmd.py:40-139` and `change_risk/normalize.py:26-30` — all accurate. Default revspec `HEAD`, comma-separated `--ext`, repeatable `-x` applied to the baseline too, root `.riskignore` honoured, and the `Below typical / Typical / Elevated` labels. The "works without `repowise init`" claim is right — `score_live_change` is git-only with no DB read.

Optional addition: a line on `--baseline` (default 200, 0 disables). The walkthrough leans on the percentile headline, and on a shallow clone the percentile drops out and you fall back to the absolute band — that's the one surprise a reader isn't prepared for here.

PR #1227, opened 2026-08-01
feat(generation): add configurable synthesis evidence

serjflint:
## Abstract

Add opt-in, page-scoped repository evidence for model-written overview and onboarding pages.

### If Markdown is already indexed, why is explicit evidence needed?

Repowise retains Markdown sources during generation, but does not retrieve raw documents for high-level synthesis: its vector index contains generated pages and decisions. Ingested therefore does not mean prompt-visible.

In [Saitenka](https://github.com/serjflint/saitenka), unpatched generation described the wrong project, Python version, installation, and execution pipeline despite the authoritative [README](https://github.com/serjflint/saitenka/blob/main/README.md), [running guide](https://github.com/serjflint/saitenka/blob/main/overlay/RUNNING.md), and [architecture guide](https://github.com/serjflint/saitenka/blob/main/overlay/ARCHITECTURE.md).

## Specification

`generation_context.files` selects ordered, indexed sources for supported synthesis pages. Selection is opt-in, bounded, recorded as provenance, and included in cache identity. Citations remain authoritative only when structural context or evidence establishes them.

Repository content remains untrusted data. Defaults preserve existing behavior; no migration is required.

## Rejected alternatives

- **Raw-document RAG:** requires chunking, retrieval, freshness policy, and embedding; it is a separate feature.
- **Automatic conventions:** miss custom authoritative sources.
- **Arbitrary filesystem reads:** bypass indexing and exclusion policy.
- **Injecting the whole repository:** defeats relevance and budget guarantees.

## Validation

Focused generation, checkpoint, CLI, grounding, and integration tests pass, with changed-scope checks.

## Related issue

Addresses the missing-input recovery path described in #1226. It does not claim to prevent every model hallucination.

PR #1250, opened 2026-08-02
refactor(generation): shrink the frozen evidence-files mapping

serjflint:
## Summary

- Back `_FrozenEvidenceFiles` with a `MappingProxyType` over a plain dict — O(1) `__getitem__` (was a linear scan over tuple pairs) and no in-place mutation, so the deep immutability the old tuple-of-tuples gave for free is preserved.
- Drop `__slots__`, `__copy__`, `__deepcopy__`. The one-line `__reduce__` stays (a `mappingproxy` is not picklable on its own) and rebuilds from a plain dict — covering copy/deepcopy/pickle alike.
- Two load-bearing contracts hold: the config stays **hashable** (`__hash__` over `frozenset(items())`, order-independent) and the mapping stays **read-only** (`__setattr__` blocks rebinding `_items`; the proxy blocks item assignment *and* in-place mutation of the backing store).

The `job_system` `to_dict` duck-check cleanup floated in the #1227 comment is **deliberately deferred**: `create_job` is called with a non-`GenerationConfig` `SimpleNamespace` in `test_init_failure_reporting.py`, so the `getattr` + `dataclasses.asdict` fallback is load-bearing, not dead.

Follow-up promised in https://github.com/repowise-dev/repowise/pull/1227#issuecomment-5158149047 (note 2).

## Related Issues

Follow-up to #1227.

## Test Plan

- [x] `pytest tests/unit/generation/` — 1163 passed (Python 3.13)
- [x] Contract tests green: hashable + order-independent hash; `copy`/`deepcopy`/**`pickle`** round-trip to an equal object; `to_dict` shape; positional constructor order
- [x] Added assertion: a reachable private handle (`._items[...] = ...`) can't mutate the store and corrupt the cached hash
- [x] `ruff check` / `ruff format --check` clean

## Checklist

- [ ] I have updated documentation if needed — n/a

### DIFF
--- a/packages/core/src/repowise/core/generation/models.py
+++ b/packages/core/src/repowise/core/generation/models.py
@@ -14,6 +14,7 @@ from collections.abc import Iterable, Iterator, Mapping
 from copy import deepcopy
 from dataclasses import dataclass, field, fields
 from datetime import UTC, datetime
+from types import MappingProxyType
 from typing import Any, Literal
 
 from repowise.core.reasoning import ReasoningMode, normalize_reasoning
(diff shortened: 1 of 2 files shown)


RaghavChamadiya (approved):
Thanks, this is a tidy piece of work and the reasoning in the description made it fast to review.

The `MappingProxyType` swap holds both contracts: hashability still comes from `frozenset(items())` rather than the proxy, and the backing dict is constructed inline so no external handle to it survives. The new assertion that `._items[k] = v` raises is a good addition, since that was the one hole the old tuple-of-pairs closed by accident rather than by design.

Dropping `__copy__` and `__deepcopy__` does change one thing worth naming: `deepcopy` now returns a new equal object instead of the identical one. Nothing can observe that here, because the mapping is immutable and the config is frozen, so no aliasing bug is possible. Just noting it so the change is deliberate on the record rather than incidental.

I merged this together with #1251 locally and ran `tests/unit/generation/`: 1163 passed. The two auto-merge cleanly in either order, so no rebase dance needed.

PR #1251, opened 2026-08-02
refactor(generation): validate evidence files in one place

serjflint:
## Summary

- The page-key/path-list checks for `source_evidence_files` were written twice — in `GenerationConfig.from_repo_config` and again in `__post_init__` (because `from_repo_config` ends in `cls(**values)`). Lift both into one `_normalize_evidence_files(raw_files, *, label)` helper.

Two behaviours are kept on purpose and now pinned by tests:

- **Error framings differ** (the reason for the `label` parameter): `from_repo_config` reports against `generation_context.files` (the key the user wrote); `__post_init__` against `source_evidence_files` (the internal field). A new test pins the field-framed direct path so the divergence can't silently collapse.
- **Key normalization is unified.** `from_repo_config` always `str().strip()`'d keys before the membership check; `__post_init__` did not. Both now do — so direct construction with a padded key (`" repo_overview "`) is trimmed and accepted where it previously raised. This is a deliberate alignment (direct construction should not be stricter than config parsing), covered by a new test. Flagged explicitly because it is the one observable behavior change.

Follow-up promised in https://github.com/repowise-dev/repowise/pull/1227#issuecomment-5158149047 (note 3).

## Related Issues

Follow-up to #1227.

## Test Plan

- [x] `pytest tests/unit/generation/` — 1164 passed (Python 3.13)
- [x] Both message framings survive; the field-framed direct path is newly pinned (verified non-vacuous: collapsing the label fails the test)
- [x] Key-normalization alignment pinned by `test_direct_generation_config_normalizes_evidence_keys_like_the_config_loader`
- [x] `ruff check` / `ruff format --check` clean

## Checklist

- [ ] I have updated documentation if needed — n/a

### DIFF
--- a/packages/core/src/repowise/core/generation/models.py
+++ b/packages/core/src/repowise/core/generation/models.py
@@ -67,6 +67,40 @@ def _source_evidence_page_keys() -> set[str]:
     }
 
 
+def _normalize_evidence_files(
+    raw_files: Mapping[str, Any], *, label: str
+) -> dict[str, tuple[str, ...]]:
+    """Validate a page-key -> paths mapping, framing errors with *label*.
+
+    Shared by ``from_repo_config`` (label ``generation_context.files``, the key
+    the user wrote) and ``__post_init__`` (label ``source_evidence_files``, the
+    internal field for direct construction). The two labels are the whole point
+    of the parameter: the same rules, reported against the surface the caller
+    touched.
+
+    Sharing one validator also unifies key normalization: both paths now
+    ``str(...).strip()`` a key before the membership check. ``from_repo_config``
+    always did; direct construction did not, so a padded key like
+    ``" repo_overview "`` is now trimmed and accepted where it used to raise.
+    """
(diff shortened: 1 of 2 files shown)


RaghavChamadiya (approved):
Thanks, and particularly for calling out the behavior change instead of letting it ride.

The `label` parameter earns its place. Keeping the two error framings while sharing the rules is the right shape, and pinning the field-framed path with a test is what stops the divergence from silently collapsing later.

The key-normalization alignment is the real change here, and I agree with it. Direct construction being stricter than config parsing was an accident of the duplication, not a decision, so `" repo_overview "` raising in one path and being accepted in the other was never intentional. Good that it is covered by its own test rather than buried in the refactor.

I merged this together with #1250 locally and ran `tests/unit/generation/`: 1163 passed. The two auto-merge cleanly in either order.

One note for whoever merges: these two and #1250 both touch `models.py` and `test_models.py`, so whichever lands second will want a quick CI re-run on the merged result even though the merge itself is clean.

PR #1255, opened 2026-08-02
fix(python): resolve attribute calls on python submodules

jyotirmya17:
### Summary
This PR fixes an issue where Python attribute calls on submodules imported from a package would fail to resolve. 

### Changes
- Adds a Python-specific fallback in `CallResolver` to check the `receiver_name.py` or `receiver_name/__init__.py` submodule file when the receiver is tracked back to a package's `__init__.py` and the method is not found there.

### DIFF
--- a/packages/core/src/repowise/core/ingestion/resolvers/python.py
+++ b/packages/core/src/repowise/core/ingestion/resolvers/python.py
@@ -121,7 +121,7 @@ def resolve_python_import_all(
                 # (#1193). Without this the binding's source_file stays the
                 # package init, which declares nothing, and the call is missed.
                 for binding in imp.bindings:
-                    if binding.local_name == name:
+                    if (binding.exported_name or binding.local_name) == name:
                         binding.source_file = hit
                         break
     return tuple(dict.fromkeys(targets))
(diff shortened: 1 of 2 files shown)


RaghavChamadiya (approved):
This is the right shape now, and thanks for turning it around in half an hour.

Two things you did that I did not ask for and should have. You kept the true-negative case as an aliased twin, so `test_non_submodule_aliased_binding_keeps_package_init` proves the pin still declines a name that is not a submodule rather than just proving it fires. And you dropped the `CallResolver` fallback entirely instead of leaving it in as a belt-and-braces second path, which is what would have left a dead branch in a file that has a dozen bug fixes behind it.

Verified against current main:

```
tests/unit/ingestion/test_python_resolver_fanout.py
11 passed

from pkg import submod as sm; sm.helper()
  binding source_file: [('sm', 'pkg/submod.py')]
  EDGES: [('caller.py::run', 'pkg/submod.py::helper', 0.88, 'module_alias')]

from pkg import submod; submod.helper()
  EDGES: [('caller.py::run', 'pkg/submod.py::helper', 0.88, 'module_alias')]   (unchanged)
```

and the full matrix is green on 213fc2b.

For the record on this thread, since it ran from 2 Aug: the bug you originally reported was real the whole time, #1733 fixed one half of it while this was open, and the half that was left is the one you had already written down in a comment in your own test file. That is why this ended up landing at the binding rather than in the resolver.

PR #1268, opened 2026-08-03
fix(docker): use node:20-bookworm-slim builder for glibc compatibility (Blocker 2)

### LINKED ISSUE #1264: [Bug] Docker image does not build or run: four blockers in docker/Dockerfile
## Describe the Bug

`docker build -f docker/Dockerfile .` does not produce a working image. There are four independent blockers in the same file, and they surface one at a time, each hiding the next.

#1148 reports the first one. #1236 fixes it, and getting past it is what exposed the rest. This issue covers what remains after #1236 lands.

All findings below come from real builds and container runs against the #1236 branch, with each blocker patched locally to reach the next. None are inferred from reading the file.

## Steps to Reproduce

1. Check out the #1236 branch (or `main` with the `npm install` workspace fix applied).
2. Run `docker build -t repowise -f docker/Dockerfile .` from the repo root.
3. See the build fail at the final `RUN` step.
4. Patch in `mkdir -p /data`, rebuild, then `docker run repowise` to hit the next blocker.

Note that no CI job builds this Dockerfile, so a green check on a PR says nothing about it. That is why all four went unnoticed.

## Expected Behavior

`docker build -t repowise -f docker/Dockerfile .` completes, `docker run` starts both processes, the API answers on 7337 and the web UI on 3000. Ideally on a Windows checkout too, since that is where the original report came from.

### Blocker 1: `chown` fails because `/data` does not exist

```
#29 [runtime 17/17] RUN groupadd -r repowise && ... && chown -R repowise:repowise /app /data
chown: cannot access '/data': No such file or directory
ERROR: failed to build
```

`VOLUME /data` (line 56) declares a mount point but does not create the directory in the image, so the `chown` (line 74) has nothing to act on. `mkdir -p /data` before it is enough to get past this, though whoever picks it up should confirm that is the intended ownership and permission setup rather than assuming.

### Blocker 2: the `node` binary cannot execute in the runtime image

Line 37 copies `node` from the `node:20-alpine` builder into the `python:3.12-slim` runtime. Alpine links against musl and Debian slim uses glibc, so the binary is present but unloadable:

```
$ ls -l /usr/local/bin/node
-rwxr-xr-x 1 root root 101780480 /usr/local/bin/node

$ /usr/local/bin/node --version
sh: 1: /usr/local/bin/node: not found
exit=127
```

The `not found` refers to the missing musl loader, not to `node` itself, which makes this one confusing to debug. Options include building the frontend on a Debian-based Node image, installing Node from nodesource in the runtime stage, or copying from a `node:20-bookworm-slim` stage. Worth weighing image size against the change.

### Blocker 3: `entrypoint.sh` is unrunnable from a Windows checkout

```
$ docker run --rm repowise
exec /app/entrypoint.sh: no such file or directory
```

`docker/entrypoint.sh` has 21 CRLF terminators and a `#!/bin/bash\r` shebang. The repo has no `.gitattributes`, so git hands Windows users CRLF on checkout and the shebang breaks. This is why #1148 came from a Windows 11 user.

A `.gitattributes` with `*.sh text eol=lf` fixes it repo-wide and prevents recurrence. Existing checkouts will need a refresh. Normalizing inside the Dockerfile would be a workaround rather than a fix, since the same problem affects anyone running these scripts outside Docker.

### Blocker 4: Next.js standalone output path

This one belongs to #1236 and is listed only so the picture is complete. Moving the lockfile to the workspace root makes Next nest the standalone output under `packages/web/`, so `/app/web/server.js` no longer exists where `entrypoint.sh` expects it:

```
$ find /app/web -maxdepth 3 -name server.js
/app/web/packages/web/server.js

$ test -f /app/web/server.js
NO
```

## Environment

- OS: Windows 11 host, Docker Desktop 28.4.0, linux/x86_64 engine
- Python version: 3.12 (`python:3.12-slim` runtime stage)
- Repowise version: 0.37.0 (wheel built during the image build)
- Installation method: Docker

## Additional Context

Worth considering alongside this: a build-only CI job on changes to `docker/**` would catch blockers 1 and 2. It would not catch blocker 3, since CI checks out on Linux. That is a separate call about CI budget, not part of this fix.

### Claiming this

To avoid two people building the same fix: comment here saying you are taking it, and you will be assigned. Only the assignee should open a PR. If you are assigned and then get pulled away, a one-line comment to unclaim is enough and carries no obligation, someone else can take it from there. If an assigned issue goes quiet for two weeks, it goes back to unassigned.

These blockers are separable. Say which ones you are taking if you do not want all of them.

Questions about scope before claiming are welcome. Asking is not claiming.

akshatmalik-bruh:
## Summary

- Fixes Blocker 2 in `docker/Dockerfile` where copying the `node` binary from `node:20-alpine` (`musl`) into `python:3.12-slim` (`glibc`) causes runtime failure with `sh: 1: /usr/local/bin/node: not found` (exit code 127).
- Changes Stage 1 builder base image from `node:20-alpine` to `node:20-bookworm-slim` so the pre-built `node` binary is dynamically linked against `glibc`.

## Related Issues

- Follow-up fix for Blocker 2 in `docker/Dockerfile`.
- fixes #1264 

## Test Plan

- [x] Local Docker Build Verification:
  1. Merged PR #1236 locally to resolve Stage 1 workspace dependencies.
  2. Built container image with `FROM node:20-bookworm-slim AS frontend-builder`.
  3. Ran `docker run --rm --entrypoint /usr/local/bin/node repowise-b2-test --version`.
  4. Verified output prints `v20.20.2` cleanly (no exit code 127).
  5. Reset branch back to `main` so this PR cleanly contains **only** the `node:20-bookworm-slim` change.

## Checklist

- [ ] I have added tests for new functionality

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -6,10 +6,9 @@
 #   docker run -p 7337:7337 -p 3000:3000 -v /path/to/repo/.repowise:/data -e GEMINI_API_KEY=... repowise
 # =============================================================================
 
-# ---------------------------------------------------------------------------
-# Stage 1: Build the Next.js frontend
-# ---------------------------------------------------------------------------
-FROM node:20-alpine AS frontend-builder
+# Use Debian-based Node builder (glibc) to match python:3.12-slim runtime.
+# Alpine (musl) builds produce a node binary that fails to execute in Debian slim with exit code 127.
+FROM node:20-bookworm-slim AS frontend-builder
 
 WORKDIR /app
 


RaghavChamadiya (approved):
LGTM

PR #1280, opened 2026-08-03
fix(generation): keep concept pages local, never merge unrelated top-level dirs

### LINKED ISSUE #1281: Concept/module pages can merge unrelated top-level directories into one page
## Summary

A concept/module page is meant to document one place in the tree — the code even
documents this: *"a subtree or a run of adjacent siblings, never a scattering"*
(`concept_tree/grouping.py`). But `group_files()` can put files from several
**unrelated top-level directories** into one group and name it after just one of
them. The generated page then describes code that lives elsewhere, and a reader
who trusts the title looks in the wrong place.

I found this on my repo [Saitenka](https://github.com/serjflint/saitenka): one page
named after a small structured-content directory (3 files) actually held 11 — the
other 8 pulled from an unrelated `tools/` directory (FSRS/fuzzing/mutation
scripts), so the page confidently described "scoring and fuzzing." Another swept
files from five unrelated top-level components (`.agents`, `deinflect`, `install`,
`overlay`, `taffylite`) under one name. The model described its inputs faithfully;
the inputs were the wrong files.

## Minimal reproduction

Pure and deterministic, no model involved:

```python
from repowise.core.generation.concept_tree.grouping import group_files, GroupingParams

files = [f"{d}/f{i}.py"
         for d, n in {"alpha": 2, "beta": 2, "gamma": 2, "delta": 1}.items()
         for i in range(n)]
for g in group_files(files, params=GroupingParams(min_files=3, max_files=6)):
    print(g.target_path, sorted({m.split("/")[0] for m in g.members}))
# alpha  ['alpha', 'beta', 'delta']   <- one page, three unrelated dirs, named "alpha"
# gamma  ['gamma']
```

Four thin top-level directories collapse into a page named `alpha` that also
contains `beta` and `delta`.

## Root cause

Grouping is size-driven: a directory below the min-files floor is folded into a
neighbour, and files that never form a group of their own are swept together.
Neither the fold (`_absorb_thin`) nor the sweep (the walk's sibling accumulation)
checks that the two sides share a real parent directory — so at the repository
root, where "adjacent siblings" are whole top-level components, unrelated
directories merge. That is exactly the scattering the docstring says it never
emits.

This is a **selection** problem — the wrong files are on the page — as opposed to
the missing-source-context grounding gap #1226/#1229 addressed. Supplying more
evidence can't fix a page whose member set is already wrong.

## Direction and proof of concept

A locality guard — two runs merge only if their union shares a real parent
directory (or both are root-level files) — removes the scattering on the
reproduction above and on the real repo (cross-top-level groups 1 → 0), while
keeping the partition deterministic and total. Draft PR: #1280.

One design call is yours: an isolated thin top-level directory then becomes its own
small page rather than being merged away — thin-but-local instead of
larger-but-scattered. A named multi-directory "misc" bucket is the alternative. The
PoC pins the thin-local behaviour in a test so the choice is explicit rather than
accidental.

serjflint:
Draft PoC — opening for CI and to make the change concrete. Not requesting review until the design question below (and in the companion issue) is settled.

Closes #1281 — that issue carries the problem, a runnable reproduction, and the root cause; this PR is the fix and does not restate them.

## What

`_is_local_merge(left, right)` gates the three sites where grouping merges file runs (the walk's sibling accumulation, the thin-remainder fold-back, and `_absorb_thin`): two runs merge only when their combined members share a real parent directory, or are all root-level files. This restores the "a subtree or a run of adjacent siblings, never a scattering" invariant `grouping.py` already documents.

## Why this shape

- The scattering has two sources (the size-driven sweep and thin-group absorption), so the guard sits at both — not in naming/`_assign_targets`, where the members are already merged and a rename can't un-merge them.
- "Local" means sharing a top-level directory. Deliberately broader than strict same-parent (it still lets a small `sc/` merge with a sibling `tools/` under a shared `overlay/`): the tighter rule fragments the outline more, and the top-level scattering is the actual invariant break.

## Trade (the open design call)

An isolated sub-min top-level directory now stands as its own thin page instead of being merged into an unrelated one — thin-but-local over larger-but-scattered. `TestLocality` pins this as intended; a named multi-directory "misc" bucket is the alternative, deferred to the issue discussion.

## Tests

- `TestLocality`: scattering refused, root file not absorbed into a subtree, root files still co-locate, adjacent siblings still merge, totality + determinism hold. Non-vacuous — verified to fail with the guard neutered.
- Repaired `test_layer_labels_from_the_kg_reach_the_title`: its fixture depended on the old buggy merge (files swept into a one-word "root" page that happened to trigger the layer-label rescue); a single-directory fixture now exercises the rescue directly.
- Full generation suite and ruff both green.

### DIFF
--- a/packages/core/src/repowise/core/generation/concept_tree/grouping.py
+++ b/packages/core/src/repowise/core/generation/concept_tree/grouping.py
@@ -224,6 +224,25 @@ def _target_path(members: Sequence[str], dirs: Sequence[str]) -> str:
     return "/".join(common)
 
 
+def _is_local_merge(left: Sequence[str], right: Sequence[str]) -> bool:
+    """Whether two runs may form one page without becoming a scattering.
+
+    Local means the union sits under a real shared directory, or both runs are
+    root-level files that legitimately co-locate at the repository root. Two runs
+    whose only common ancestor is the repository root but that live in different
+    top-level directories are the scattering this module's contract promises never
+    to emit, and must stay separate pages even when each is thin.
+
+    Deliberately loose: *any* shared directory qualifies, not the same immediate
+    parent. Tightening it to same-parent fragments the outline to prevent a
+    bleed no reader perceives; the break readers report is the cross-top-level
+    one.
+    """
+    if _target_path([*left, *right], []):
+        return True
+    return all("/" not in m for m in left) and all("/" not in m for m in right)
+
+
 class _Partitioner:
     def __init__(self, params: GroupingParams, layer_of_file: dict[str, str]):
         self.params = params
@@ -279,8 +298,10 @@ class _Partitioner:
             left = self.visit(child)
             if not left:
                 continue
-            if len(pending) + len(left) <= self.params.max_files and self._same_layer(
-                pending, left
+            if (
+                len(pending) + len(left) <= self.params.max_files
+                and self._same_layer(pending, left)
+                and _is_local_merge(pending, left)
             ):
                 pending.extend(left)
                 continue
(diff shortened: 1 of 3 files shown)


RaghavChamadiya (approved):
LGTM

PR #1305, opened 2026-08-04
feat: verify returned vector width across all embedding providers

### LINKED ISSUE #1292: Embedders declare an output width they never verify against the returned vectors
Every embedder exposes a `dimensions` property that the vector store trusts when sizing its table. None of them check that the vectors the provider actually returns are that wide. When the two disagree the store is built at the declared width and filled with vectors of a different one, with no error at any point.

This came out of #1253 and #1254. That PR fixed the openai embedder's declared width and made it request the width from the API, which closes the gap for endpoints that honour the parameter or reject it outright. The case left open is an endpoint that silently ignores an unrecognised `dimensions` field and returns its native width anyway, which is permitted behavior for an OpenAI-compatible server. #1253 asked for this check in its "expected behavior" section and correctly noted that no embedder currently does it.

## Current state per provider

- `providers/embedding/ollama.py:112` checks that the returned batch has one vector per input, but never checks the width of those vectors. This is the closest thing to a precedent and the natural shape for the fix.
- `providers/embedding/gemini.py:114` sends `output_dimensionality` and checks neither count nor width.
- `providers/embedding/openai.py` after #1254 sends the width when overridden, and checks neither.
- `providers/embedding/openrouter.py` refuses unknown models outright with a good error, so it cannot be mis-sized by an unknown model, but it still does not verify what comes back.

## Why it matters

`existing_vector_dim()` in `packages/cli/src/repowise/cli/providers/vector_store.py:11` calls the table's own width "the one embedder fact that cannot drift from reality". That is exactly right, and it is a good argument for making the embedder's declaration meet the same standard. A mismatch here does not fail loudly, it produces a store that retrieval is then built on.

## The task

Verify the returned width once per `embed()` call and fail with a message naming both numbers and the provider. Points worth deciding rather than assuming:

- Where it lives. A shared helper is tempting, but the four embedders have no common base class today and adding one for this would be a larger change than the check. A few lines per provider is probably the honest version.
- When it runs. Checking every batch costs one `len()` per call, which is nothing, and catches a provider that changes behavior mid-run. Checking only the first batch is cheaper still. Either is defensible, pick one and say why.
- Whether an inferred width should fail as hard as an explicit one. A user who set `REPOWISE_EMBEDDING_DIMS` themselves gets a clear error. A user hitting a wrong entry in a built-in `_DIMS` table gets an error about a value they never chose, so the message needs to point at the right fix.
- The mock embedder path should stay untouched.

Tests belong alongside the existing per-provider suites, for example `tests/unit/test_persistence/test_openai_embedder.py`, which already has the fake-client harness this needs.

akshatmalik-bruh:
## Summary

- Adds runtime validation in `embed()` across `OllamaEmbedder`, `OpenAIEmbedder`, `GeminiEmbedder`, and `OpenRouterEmbedder` to verify returned vector width matches declared `dimensions`.
- Raises a descriptive `ValueError` naming both expected and returned dimensions alongside actionable configuration hints if an endpoint silently ignores `dimensions`.
- Adds an optional `dimensions` parameter to `OpenRouterEmbedder` to support explicit output width overrides.

## Related Issues

- Fixes #1292 

## Test Plan

- [x] Tests pass (`pytest`)

## Checklist


RaghavChamadiya (approved):
LGTM

PR #1323, opened 2026-08-06
fix(providers): send temperature=1 for kimi-for-coding models

uncleaar:
## Problem

The Kimi coding endpoint (`https://api.kimi.com/coding/v1`) rejects any `temperature` other than 1 for the `kimi-for-coding` model family:

```
400 invalid_request_error: invalid temperature: only 1 is allowed for this model
```

`_kimi_temperature()` pinned `0.6` for these models, so **every** request through the `kimi` provider failed validation. In practice, `repowise generate --provider kimi` failed on every page and silently wrote structural stubs (`stub_fallback`) instead of model-written prose — the run reported success while no prose was actually generated.

## Verification against the live endpoint

- `GET /coding/v1/models` lists `kimi-for-coding`, `kimi-for-coding-highspeed`, `k3`, `k3-256k`.
- `kimi-for-coding` with `temperature: 0.3` or `0.6` → 400 (error above).
- `kimi-for-coding` with `temperature: 1` or with the field omitted → 200.
- `k3` rejects `temperature != 1` the same way (out of scope here: unknown models currently fall through to the caller's requested value).

## Fix

Return `1.0` for the `kimi-for-coding*` family in `_kimi_temperature()`, with a comment explaining why the requested value cannot be honoured.

## Tests

- Updated the three assertions that pinned `0.6` for `kimi-for-coding` / `kimi-for-coding-highspeed` (generate and stream_chat paths) to `1.0`.
- `pytest tests/unit/test_providers/test_kimi_provider.py` — 18 passed.

### DIFF
--- a/packages/core/src/repowise/core/providers/llm/kimi.py
+++ b/packages/core/src/repowise/core/providers/llm/kimi.py
@@ -98,7 +98,10 @@ def _kimi_temperature(
     requested: float,
 ) -> float:
     if model.startswith("kimi-for-coding"):
-        return 0.6
+        # The Kimi coding endpoint rejects any temperature other than 1 for
+        # these models ("invalid temperature: only 1 is allowed for this
+        # model"), so the requested value cannot be honoured here.
+        return 1.0
     if model.startswith(("kimi-k2.5", "kimi-k2.6")):
         return 0.6 if normalize_reasoning(reasoning) in ("off", "none") else 1.0
     return requested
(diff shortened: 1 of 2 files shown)


RaghavChamadiya (approved):
Thanks, and thanks especially for probing the live endpoint and writing down what you found. That turns this from a guess into a documented constraint, and the comment above the return means the next person won't try to "restore" 0.6. Good call leaving `k3` out of scope rather than expanding the change. Merging.

PR #1329, opened 2026-08-06
fix(core): pass unbounded limit to list_jobs from resume readers

Ayush7614:
## What

`find_completed_phases` and `ResumeLedger.completed_phases` call `list_jobs(repository_id=...)` without overriding the default limit of 100. A repo with more than 100 `pipeline_jobs` rows silently truncates the completed-phase set, causing resume to redo work (git history, centrality, decision extraction) that prior runs already finished.

## Fix

Pass `limit=10_000` from both resume readers so they see the full set of completed phases regardless of job volume.

## Verification

- Added `test_job_store_list_jobs_respects_limit`: verifies the limit parameter truncates correctly.
- Added `test_find_completed_phases_uses_unbounded_limit`: creates 6 completed jobs and asserts all are returned.
- Both new tests pass.

## Root cause / Gap filled

This is a genuine correctness bug in the resume path: the docstrings promise "return the set of phases already COMPLETED for repository_id", but the hidden row cap violates that contract for repos with many runs. No prior PR, issue, or doc claimed the 100-row cap was intentional for resume readers.

### DIFF
--- a/packages/core/src/repowise/core/pipeline/checkpoint.py
+++ b/packages/core/src/repowise/core/pipeline/checkpoint.py
@@ -66,8 +66,10 @@ async def find_completed_phases(
     """
     from repowise.core.persistence._interfaces.job_store import JobState
 
-    jobs = await job_store.list_jobs(repository_id=repository_id, limit=10_000)
-    return {j.phase for j in jobs if j.state == JobState.COMPLETED}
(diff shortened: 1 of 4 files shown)


RaghavChamadiya (approved):
LGTM

PR #1330, opened 2026-08-06
fix(core/cli): do not cache failed update-check results

Ayush7614:
## What

A transient PyPI failure in `check_latest_version_cached` or `get_cli_update_check_cached` wrote `latest=None` + `error` into the TTL cache. The next call within the cache window saw cached `None` (not the `_MISS` sentinel), skipped the live fetch, and reported "unknown" for the entire `DEFAULT_TTL_HOURS` (24h).

## Fix

Only persist the cache when `latest_version` is not `None`. On failure, skip caching so the next call retries the network.

Fixes both the core shared cache (`~/.repowise/update-check.json`) and the CLI-specific cache.

## Verification

- Added `test_check_latest_version_cached_does_not_poison_on_failure` in `test_release.py`: asserts a failed fetch is retried on the next call.
- Added `test_cli_cached_does_not_persist_failure` in `test_update_check_cache.py`: asserts a failed live check is not persisted and a successful retry is cached.
- All 9 release tests pass; all 5 CLI cache tests pass.

## Root cause / Gap filled

This is a genuine 24-hour silent failure: one network hiccup locked users into "update status unknown" for a full day, with no retry path. The docstrings promise best-effort behavior, but the cache write was unconditional. No prior PR, issue, or doc claimed caching failures was intentional.

### DIFF
--- a/packages/cli/src/repowise/cli/update_check.py
+++ b/packages/cli/src/repowise/cli/update_check.py
@@ -197,15 +197,16 @@ def get_cli_update_check_cached(
         pass  # missing / corrupt / stale -> fall through to a live check
 
     result = get_cli_update_check(timeout=timeout)
-    with contextlib.suppress(Exception):  # caching is opportunistic
-        path.write_text(
-            json.dumps(
-                {
-                    "checked_at": time.time(),
-                    "latest_version": result.latest_version,
-                    "error": result.error,
-                }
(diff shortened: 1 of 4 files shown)


RaghavChamadiya (approved):
Verified the fix and the tests. Good catch.

PR #1353, opened 2026-08-07
feat(languages): add Object Pascal / Delphi support (Good tier, draft)

Laex:
## What

Adds `.pas` / `.pp` / `.dpr` / `.dpk` / `.lpr` / `.inc` support via
`tree-sitter-pascal` (PyPI, AGPL-3.0, actively maintained by Jim McKeeth —
https://github.com/jimmckeeth/tree-sitter-pascal), following the
one-`.scm`-file-plus-one-config-entry recipe in
`docs/architecture/language-support.md`.

Covers: unit/program/library structure, `uses` imports, type declarations
(class/record/interface/class-helper/enum/alias — all reported as `"class"`
for now, see Known gaps), procedure/function/constructor/destructor/
operator declarations and full implementations (interface-section
signature vs. implementation-section body are two distinct grammar nodes
and both are captured — a method's `defProc` capture spans header+body so
`get_symbol` returns the real body), properties, class/interface/helper
heritage (ancestor + implemented interfaces, best-effort extends/implements
split), and a call graph.

Files:
- `languages/specs/pascal.py` — `LanguageSpec`
- `queries/pascal.scm` — capture query
- `extractors/heritage/pascal.py` — heritage extractor
- `language_configs.py` entry (see `docs/language_config_snippet.py`)
- `pyproject.toml` dependency: `tree-sitter-pascal>=0.11,<1`
- Registration touch points in `models.py`, `languages/specs/__init__.py`,
  `extractors/heritage/__init__.py` — see `docs/registration_diffs.md`

## Confidence level — read `docs/VERIFICATION.md`

I'm an AI agent, working on behalf of a user who wanted Delphi support for
their own project, in a network-sandboxed environment that blocked
`pip`/`npm`/`git clone` outright. So **`tree-sitter-pascal` was never
installed, and this code has never actually been run.** What it does have:
every field name, dataclass shape, and dispatcher signature referenced here
was read from repowise's *real, current* source (`languages/spec.py`,
`language_configs.py` in full — all ~18 entries, `parser.py` in full —
~700 lines including `_find_parent` and `_extract_symbols`,
`extractors/heritage/__init__.py` and a real heritage extractor,
`queries/rust.scm` as a live multi-pattern reference), not inferred from
the contributor-doc example alone. That source-reading caught real bugs a
docs-only draft had (wrong capture name `@call.expr` vs. real `@call.site`,
an invalid `SymbolKind` value, wrong `parent_class_types` that would have
silently orphaned every method from its class). Full account of what was
and wasn't checked, and exactly how, is in `docs/VERIFICATION.md` — please
read it before reviewing the diff, it'll save you time.

Submitting anyway because `docs/layers/LANGUAGE_SUPPORT.md` says "Working
in a language we don't list? ... Open a Discussion and we'll prioritise" —
this is meant as a concrete, heavily-researched starting point for that,
not a finished PR. Happy to iterate once someone with a real checkout runs
it once and tells me what broke.

## Known gaps / open questions for a maintainer

1. **`declType` (class/record/interface/helper/enum/alias) all report
   `SymbolKind = "class"`.** Go and Kotlin have a real post-hoc
   kind-refinement hook for exactly this shape of problem
   (`refine_go_type_kind` / `refine_kotlin_class_kind` in `parser.py`) — a
   Pascal equivalent belongs there, but I couldn't see that function's own
   implementation to match its contract confidently.
2. **`declUses` doesn't handle Delphi's `uses UnitA in 'UnitA.pas';`** form
   used in `.dpr`/`.dpk` project files (only plain `uses UnitA, UnitB;` is
   in the grammar's `declUses` rule) — likely a `tree-sitter-pascal`
   grammar gap; worth confirming against a real `.dpr` and filing upstream
   if so.
3. **Visibility is per-section, not per-declaration** (`declSection` in the
   grammar governs everything until the next section keyword).
   `visibility_fn=public_by_default` is a placeholder — real per-section
   visibility needs a sibling-walk no other language's `visibility_fn`
   currently does.
4. **Heritage doesn't syntactically distinguish "extends" from
   "implements"** — `class(TBase, IFoo, IBar)` puts all three in one
   comma-separated list; this PR tags the first entry "extends" and the
   rest "implements" by convention, which the grammar doesn't enforce.
5. **`builtin_calls` matching is case-sensitive against a case-insensitive
   language.** `ASTParser._extract_calls` does `target_name in
   _call_builtins` on raw source text; Pascal source can spell `WriteLn`
   any way (`Writeln`, `writeln`, `WRITELN`). Listed the RTL's own
   canonical casing, which covers the common style but not every spelling.
6. **No dedicated import resolver** — `import_support="none"`, relying on
   the generic stem-map fallback (unit name == file stem is the dominant
   Delphi/FPC convention). May need a real resolver for namespaced modern
   Delphi units (`Winapi.Windows`).
7. **`.inc` extension** may collide with a C-family "generic include"
   convention in some polyglot repos — worth a deliberate call in review
   rather than silent first-spec-wins resolution.

## Testing

None yet — see `docs/VERIFICATION.md` for exactly what was and wasn't
checked (a lot was checked by reading; nothing was checked by running),
and the verify steps a human needs to run.


Ayush7614 (commented):
## Review — feat(languages): add Object Pascal / Delphi support

This is a high-quality draft. I validated it locally with `tree-sitter-pascal 0.11.0`: the spec and registry register correctly, symbol extraction works with correct method parents, and heritage extraction is right (`TCalc → TBase extends`, `TCalc → IFoo implements`). License is compatible (repowise and tree-sitter-pascal are both AGPL-3.0). The `declType`-vs-`declClass` parent reasoning in the spec docstring is excellent — good call tracing `_find_parent`.

Before this can move past draft, I'd like to see the following addressed:

1. **No tests are included.** The diff adds source only (8 files, +318) with zero test coverage. For a brand-new language this is the biggest gap, and the `docs/architecture/language-support.md` recipe expects symbol/import/heritage tests. Adding a `test_pascal.py` (mirroring the existing per-language parser tests) and a heritage test would lock in the behavior.

2. **`uses UnitA, UnitB;` only extracts the first module.** On a real sample, `uses SysUtils, Classes;` captured only `SysUtils`. The `(declUses (moduleName) @import.module)` pattern should match every `moduleName` child, so please investigate the `_extract_imports` pipeline for multi-name handling — this looks like a pipeline issue more than a `.scm` one.

3. **Duplicate symbols for interface vs. implementation.** `function TCalculator.Add` appears twice: once as a `method` (interface `declProc`, `parent=TCalculator`) and again as a top-level `function` (implementation `defProc`, `parent=None`). The implementation symbol doesn't inherit its class parent, so the graph could hold two `Add` nodes. Please check how the `defProc`/`genericDot` name form resolves its enclosing class (or whether the interface and implementation declarations should be deduplicated).

4. **`entry_point_patterns=("*.dpr", "*.dpk", "*.lpr")` are globs.** Most other specs use concrete filenames (e.g. `("main.py",)`). Verify the traverser's entry-point matching expands `*.dpr` globs, or the entry anchors won't fire for Delphi projects.

No blocker on the structure or docs — those are great. I'm keen to approve once the tests and the two extraction issues above are in.

RaghavChamadiya (approved):
LGTM

Issue #964, opened 2026-07-20
[Bug] file_count is ~3% high due to external: import nodes being typed as 'file'


Nehant14:
Follow-up to #945 / #952.

This is a follow-up to #945 (the original bug about wildly inflated file counts) and #952 (the fix for that bug). While reviewing #952, a smaller, separate issue was found — the file count is still slightly too high, just for a different reason.

     `node_type` defaults to `'file'` in the writer (`packages/core/src/repowise/core/pipeline/persist.py`) whenever no type is recorded, and rows representing third-party imports (id prefix `external:`, e.g. `external:node:path`) never have a type recorded — so they get counted as files.

     On a real index this was 79 rows out of 3,010 file-typed rows (~2.6%), confirmed by @swati510 in review on #952.

     Affects both:
     - `packages/server/src/repowise/server/routers/workspace.py` (workspace mode)
     - `packages/server/src/repowise/server/routers/stats.py` (single-repo mode, same `node_type == "file"` filter)

     Fix should either exclude `id LIKE 'external:%'` in both places, or (better) have the writer in `persist.py` record an explicit `node_type` for external-import nodes instead of leaving it to default. Should be fixed in both routers together to keep the two pages consistent. make it a .md file

dev-ploy:
Hi I would like to make a PR to this issue

RaghavChamadiya:
Hi @dev-ploy, assigning you to this issue. PR #952 is merged on main, so you can work on top of it

PR #1357, opened 2026-08-08
fix(persist): type external import nodes as 'external' (Refs #964)

dev-ploy:
External import nodes (id prefix 'external:') were persisted with no node_type, defaulting to 'file'. Every WHERE node_type = 'file' query (workspace.py and stats.py) then counted third-party dependencies as repo files (~3% inflation on real indexes). Write them as 'external' at the source so both dashboards stay consistent.

## Summary

-

## Test Plan

- [ ] Tests pass (`pytest`)

## Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -121,3 +121,8 @@ ehthumbs.db
 
 # Claude Code
 .claude/
+
+# Gradle build output swept from JVM fixture (see #964)
+tests/fixtures/**/.gradle/
+tests/fixtures/**/bin/
+tests/fixtures/**/build/
--- a/packages/cli/src/repowise/cli/commands/augment_cmd/fast_lookup.py
+++ b/packages/cli/src/repowise/cli/commands/augment_cmd/fast_lookup.py
@@ -119,10 +119,11 @@ def files_by_basename(
     match the ``LIKE`` arm's own default rather than holding a root file to a
     stricter rule than a nested one.
 
-    Filtering non-path nodes is left to the caller: ``node_type = 'file'`` also
-    covers resolved external packages, and ``:`` is a legal character in a
-    POSIX path, so excluding them in SQL would mean a colon test that can drop
-    a real file and manufacture a false unique.
+    Filtering non-path nodes is left to the caller: resolved external packages
+    are stored as ``external:pkg/mod`` with ``node_type='external'`` and never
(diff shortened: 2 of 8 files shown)


Ayush7614 (commented):
## Review — fix(persist): type external import nodes as 'external' (Refs #964)

Nice, focused fix. I verified the premise and the change in a local checkout: `resolvers/context.py:add_external_node()` adds `external:` nodes with **no** `node_type`, so they were defaulting to `"file"` and inflating every `WHERE node_type = 'file'` dashboard query. This override at persist time correctly catches both that case and the ones where a resolver explicitly sets `node_type="file"` on an external node. `"external"` is already a first-class node type elsewhere (`graph/_metrics.py`, `graph_signals.py`, `execution_flows.py`), so persisting it is consistent with the rest of the codebase. Tests pass (`test_external_node_type_persist.py` + `test_workspace_router.py`).

A few suggestions, all minor / non-blocking:

1. **Existing indexes are not migrated.** This only changes write-time behavior. Any `external:` nodes already stored as `"file"` in a live DB stay mis-typed until a reindex. Worth adding one line to the PR body (or a note) so users know a reindex clears the ~3% inflation on older indexes.

2. **Test redundancy.** `test_external_node_without_type_is_persisted_as_external` and `test_external_node_never_defaults_to_file_in_db` assert essentially the same thing. Consider collapsing them into one test — the second is a pure subset of the first (asserting `== "external"` already implies `!= "file"`).

3. **Positive side effect worth a test if not already covered:** I traced the prune paths (`_delete_stale_by_paths` at persist.py ~800 and the incremental prune at ~641). Before this fix, an `external:` node persisted as `"file"` could be read as "stale" (no real file on disk) and pruned unless re-minted. After the fix it's `"external"` with no `file_path`, so it survives. That's a genuine improvement of the fix — if there isn't one already, a prune-path test pinning that an `external:` node isn't deleted would lock it in.

Overall this is correct and well-scoped.

milanchristy2:
Fix for #1357 (migrated from #994 review) in #1908 — commit b1b6785634413acc6c36e057897cac0f0f21cfae on `fix/1357-external-cleanup` by `dev-ploy <milanvchristy420@gmail.com>`. Source fix + comments + Gradle cleanup. Verified 45/572 tests. PTAL.

PR #1373, opened 2026-08-09
fix(cli): surface swallowed config errors as CLI warnings

### LINKED ISSUE #852: [Bug] Missing warnings for swallowed configuration errors
Spun out from #847. There are cases where configuration errors or minor faults are silently swallowed during the update pipeline, preventing users from realizing their configuration is invalid. We need to ensure these surface as explicit CLI warnings.

SomSamantray:
Misconfigured embedders and config values used to fail silently: a bad `OLLAMA_EMBEDDING_TIMEOUT` swapped semantic search to mock vectors with no signal, a failed vector-store build disabled decision dedup on update, and an invalid `REPOWISE_FULL_RESCORE_INTERVAL_DAYS` quietly defaulted. Each now surfaces as an explicit warning: init's header and index-only intro name a degraded embedder and its offending env var, update's degraded panel lists embedder and store failures as plain `<Step>: <reason>` entries, and search warns before falling back to full-text.

## Summary

- `build_embedder` records the failure reason on the returned `MockEmbedder`; `embedder_degraded_warning` renders it, mirroring the server's `_embedder_status` envelope
- Warnings are additive: exit codes, embedder selection, and the silent-degradation contract are unchanged

## Related Issues

Fixes #852

## Test Plan

- [x] `pytest tests/unit/cli` — 1392 passed
- [x] `ruff check` clean
- New coverage: degraded-embedder warnings in init/update/search (including the silent mock-vectors case), vector-store failure, empty semantic results not falling back to FTS, keyless runs staying quiet, env-var naming in fallback reasons

---

[Compound Engineering](https://github.com/EveryInc/compound-engineering-plugin)


Ayush7614 (commented):
## Review — fix(cli): surface swallowed config errors as CLI warnings

Great fix — this directly addresses issue #852's "silent failure" problem. I validated it locally: the bad-env-var cases (`OLLAMA_EMBEDDING_*`, `REPOWISE_FULL_RESCORE_INTERVAL_DAYS`) now fail loudly with a named reason, the degraded vector-store build shows up in the update panel as a machine-parseable `<Step>: <reason>` entry, and `search` warns before falling back to FTS. All 44 new tests plus 121 surrounding CLI/config tests pass, no regressions.

A few suggestions (minor):

1. **Module-level mutable state (`_config_load_error`).** Carrying state from `resolve_embedder_for_repo` → `embedder_degraded_warning` through a module global is a mild smell. It works for the single-threaded CLI flow and the read-then-clear consumption does prevent leakage, but it is not thread-safe. Consider passing the error through an explicit return/context object instead of a module global, or at least add a code comment noting the single-thread assumption (the docstring partially covers this already).

2. **`_as_int_or_raise` / `_as_float_or_raise` naming.** These are really "parse env value or raise a *named* ValueError", which is clear enough. Nice touch that the error message includes the env var name and the offending raw value — keep that.

3. **Consider a warning for the mock-fallback vector store at init time too.** Right now init's index-only path and header warn on embedder degradation, and update warns on store/embedder failures. If there's an init path that builds the store for a later run to read, making sure *that* path also surfaces a degraded store (not just a degraded embedder) would close the loop — the PR body mentions "would write a store no later run can read," so just double-check the init store-build failure is surfaced the same way.

Overall the behavior is exactly what #852 asks for and the test quality is strong (the `_invoke_ok` helper and shared setup commits are a good response to the bot's duplication findings).

PR #1405, opened 2026-08-10
fix(server): require API key for repo health coordinator route 

### LINKED ISSUE #1399: [Bug] /api/repos/{id}/health/coordinator is served without the API key check
Found while fixing #1391. Small and self-contained.

## Describe the Bug

`GET /api/repos/{repo_id}/health/coordinator` is served without the API key check, unlike every other `/api/repos/*` route.

It is declared on a second router, `_repo_health_router` at `packages/server/src/repowise/server/routers/health.py:82`, which is created with no `dependencies=[...]` and then merged into the module's public router at `health.py:197`. That parent router is the health/metrics one, documented at `health.py:1-4` as deliberately unauthenticated, so the repo route inherits the exemption.

The two routes that were meant to be public are `/health` and `/metrics`. A per-repo route sitting under the same parent looks accidental: it is the only `/api/repos/{repo_id}/...` path that answers without a key, and it returns per-repo page counts, decision counts, vector counts, graph node counts and drift percentages (`health.py:85-136`).

## Steps to Reproduce

1. Start the server with `REPOWISE_API_KEY` set.
2. `curl "http://127.0.0.1:7337/api/repos/<repo_id>/health/coordinator"` with no `Authorization` header.
3. It returns 200 with the counts, where `curl "http://127.0.0.1:7337/api/repos/<repo_id>"` returns 401.

## Expected Behavior

The route is protected like the rest of `/api/repos/*`.

## Actual Behavior

It is reachable without a key.

## Suggested Fix

Give `_repo_health_router` the dependency the other routers use:

```python
_repo_health_router = APIRouter(
    prefix="/api/repos", tags=["health"], dependencies=[Depends(verify_api_key)]
)
```

If some client depends on polling it unauthenticated, worth saying so on the router instead, since right now the exemption reads as inherited by accident.

## Environment

- Installation method: source

akshatmalik-bruh:
## Summary

- Add `Depends(verify_api_key)` to `_repo_health_router` in `health.py` so that `GET /api/repos/{repo_id}/health/coordinator` requires API key authentication.
- Prevents unauthenticated access to repo-level coordinator health metrics (page counts, decision counts, vector counts, graph nodes, and drift percentages).
- Preserves intentional unauthenticated access to the top-level `/health` and `/metrics` public endpoints.

## Related Issues

Fixes #1399

## Test Plan

- [x] Verified via live `curl` probing against active server instance:
  - `GET /health` (Public endpoint, no API key) → `200 OK`
  - `GET /metrics` (Public endpoint, no API key) → `200 OK`
  - `GET /api/repos/{repo_id}/health/coordinator` (Without `Authorization` header) → `401 Unauthorized` (`{"detail": "Missing API key"}`)
  - `GET /api/repos/{repo_id}/health/coordinator` (With `Authorization: Bearer <key>`) → `200 OK`

## Checklist

- [x] Verified endpoint authorization behavior manually with `curl`


RaghavChamadiya (approved):
LGTM

PR #1406, opened 2026-08-10
fix(mcp): add explicit --host control and security warning for network transports 

### LINKED ISSUE #1400: [Bug] MCP HTTP/SSE transports have no authentication and FASTMCP_HOST can expose them
Found while fixing #1391, which only covers the REST server on port 7337. The MCP HTTP transports are a separate app with a separate story.

## Describe the Bug

`repowise mcp --transport streamable-http` (and `--transport sse`) starts a FastMCP app with no authentication of any kind. It does not go through `packages/server/src/repowise/server/deps.py`, so neither the API key check nor the local-caller check added for #1391 applies to it.

- `packages/cli/src/repowise/cli/commands/mcp_cmd.py:148-162` dispatches HTTP transports to `run_mcp`.
- `packages/server/src/repowise/server/mcp_server/_server.py:571-577` sets only `mcp.settings.port` and calls `mcp.run(transport=...)`. Nothing sets a host, and no auth is configured.

The default bind is loopback, so this is not exploitable out of the box. The concern is that the bind is reachable through configuration that has nothing to do with repowise: `mcp.settings` is pydantic-settings with `env_prefix="FASTMCP_"`, so `FASTMCP_HOST=0.0.0.0` in the environment exposes the whole tool surface with no key and no warning. `--port` is a documented flag, so a user who wants the HTTP transport reachable will reasonably go looking for a host knob and find that one.

The tool surface reads the index: source symbols, file contents through `get_symbol`, git history, decisions, health.

## Steps to Reproduce

1. `FASTMCP_HOST=0.0.0.0 repowise mcp --transport streamable-http`
2. Connect an MCP client from another machine on port 7338.
3. Every tool answers, with no credential.

## Expected Behavior

Either the HTTP transports take the same posture as the REST server (a key, or local callers only), or the CLI refuses to start a non-local bind without a key, the way `repowise serve` now warns.

## Actual Behavior

No authentication at any bind address.

## Suggested Fix

- Set `mcp.settings.host` explicitly from a repowise-owned flag rather than leaving it to `FASTMCP_*`, so the bind is decided in one place.
- Reuse the `client_is_local` / `REPOWISE_API_KEY` posture from `server/deps.py`, or document plainly that the HTTP transports are loopback-only and unauthenticated by design.
- `mcp_cmd.py:149` already prints a startup banner for network transports, a good place for the warning `repowise serve` prints.

## Environment

- Installation method: source

akshatmalik-bruh:
## Summary

- Added `--host` option to `repowise mcp` CLI command (defaults to `REPOWISE_HOST` env var or `127.0.0.1`).
- Ensured `run_mcp` explicitly assigns `mcp.settings.host = host` for `streamable-http` and `sse` transports so network bind host is strictly controlled by repowise.
- Updated startup banner and server log output to warn when binding to wide interfaces (`0.0.0.0`, `::`) without `REPOWISE_API_KEY` set (matching `repowise serve` security posture).

## Related Issues

Fixes #1400 

## Test Plan

- [x] Unit tests pass (`pytest tests/unit/cli/test_mcp_transport.py`)
- [x] Verified `--host` flag pass-through to `run_mcp`
- [x] Verified `REPOWISE_HOST` environment variable inheritance
- [x] Verified security warning triggers on `0.0.0.0` when `REPOWISE_API_KEY` is not set
- [x] Verified no warning triggers when `REPOWISE_API_KEY` is present

## Checklist

### DIFF
--- a/packages/cli/src/repowise/cli/commands/mcp_cmd.py
+++ b/packages/cli/src/repowise/cli/commands/mcp_cmd.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import os
 from pathlib import Path
 
 import click
@@ -45,6 +46,7 @@ def _workspace_summary(path: Path) -> dict[str, object] | None:
 def _print_network_startup(
     transport: str,
     repo_path: Path,
+    host: str,
     port: int,
     workspace: dict[str, object] | None,
 ) -> None:
@@ -52,8 +54,16 @@ def _print_network_startup(
     endpoint = "mcp" if transport == "streamable-http" else "sse"
     console.print(
         f"[bold green]Starting repowise MCP server ({label})[/bold green]\n"
-        f"URL: http://127.0.0.1:{port}/{endpoint}"
+        f"URL: http://{host}:{port}/{endpoint}"
     )
+
+    if not os.environ.get("REPOWISE_API_KEY") and host in ("0.0.0.0", "::"):
+        console.print(
+            "[bold yellow]SECURITY WARNING:[/bold yellow] MCP server is binding to "
+            f"[bold]{host}[/bold] without REPOWISE_API_KEY set. "
+            "All tools are unauthenticated and network-accessible. "
+            "Set REPOWISE_API_KEY or bind to 127.0.0.1."
+        )
 
     if workspace is not None:
         aliases = workspace["aliases"]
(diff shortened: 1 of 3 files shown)


RaghavChamadiya (approved):
Verified the host resolution and that stdio is untouched. Tests pass.

PR #1408, opened 2026-08-10
fix(webhooks): fail-closed auth and per-request secret reading

### LINKED ISSUE #1398: [Bug] Webhook routes are unauthenticated and skip signature verification when no secret is set
Found while fixing #1391. Filing separately because it is a distinct hole, not part of that fix.

## Describe the Bug

`/api/webhooks/github` and `/api/webhooks/gitlab` are the only API routes with no `Depends(verify_api_key)`, and their signature checks return early when the corresponding secret is unset.

- `packages/server/src/repowise/server/routers/webhooks.py:17` builds the router with no `dependencies=[...]`, unlike every other router (compare `routers/repos.py:35-39`).
- `packages/server/src/repowise/server/routers/webhooks.py:23-26`, `_verify_github_signature` returns without checking when `REPOWISE_GITHUB_WEBHOOK_SECRET` is empty, which is the default.
- `packages/server/src/repowise/server/routers/webhooks.py:41-47`, same for `REPOWISE_GITLAB_WEBHOOK_TOKEN`.

A POST with a `push` payload whose `repository.clone_url` matches a registered repo, on that repo's default branch, reaches `_launch_webhook_job` at `webhooks.py:50-66` and starts a real indexing job: git work, LLM calls, provider spend. A webhook event row is written either way.

After #1391 the rest of the API refuses non-local callers when no key is set. These two routes do not, so they become the widest remaining surface on port 7337.

## Steps to Reproduce

1. `repowise init --yes` in some repo, then `repowise serve --host 0.0.0.0` with no `REPOWISE_API_KEY` and no webhook secret.
2. From another machine, POST a push payload to `http://<host>:7337/api/webhooks/github` with header `X-GitHub-Event: push` and a JSON body of the shape `{"ref": "refs/heads/main", "repository": {"clone_url": "<the registered repo url>"}, "before": "", "after": ""}`.
3. The response carries a stored event id, and a sync job starts.

## Expected Behavior

An unsigned webhook is refused. A missing secret should mean the endpoint is closed, not that verification is skipped.

## Actual Behavior

The event is accepted and can trigger indexing work.

## Suggested Fix

- Fail closed when the secret is unset: reject rather than return, or accept only local callers, reusing `client_is_local` from `server/deps.py`.
- The secrets are read at import (`webhooks.py:19-20`), so one set after import is ignored. Worth reading per request while in there.

## Environment

- Installation method: source

akshatmalik-bruh:
## Summary

- Webhook endpoints now **fail-closed** when no secret/token is configured: non-local callers receive `403` instead of being silently accepted. Local (loopback) callers are still accepted for dev convenience, matching the `verify_api_key` posture in `deps.py`.
- Removed module-level `_GITHUB_SECRET` / `_GITLAB_TOKEN` variables. Secrets are now read from `os.environ` **per request**, so changes (e.g. from `.env` files or config reloads) are picked up without a server restart.

## Related Issues

Fixes #1398 

## Test Plan

- [x] Tests pass (`pytest tests/unit/server/test_webhooks.py`) — 11/11
- [x] `test_github_webhook_no_secret_non_local_rejected` — `auth_is_open=False` → `403`
- [x] `test_gitlab_webhook_no_token_non_local_rejected` — same for GitLab
- [x] `test_github_webhook_secret_read_per_request` — env patched after import, still works
- [x] `test_gitlab_webhook_token_read_per_request` — same for GitLab
- [x] Existing signature/token validation and job-enqueueing tests still pass

## Checklist

### DIFF
--- a/docs/scale/AUTO_SYNC.md
+++ b/docs/scale/AUTO_SYNC.md
@@ -94,7 +94,7 @@ server, which triggers an incremental update.
 ### Prerequisites
 
 - repowise server running and reachable from the internet (or GitHub's network)
-- A webhook secret (recommended for production)
+- A webhook secret (**required** — unsigned requests are rejected)
 
 ### Start the server
 
@@ -132,12 +132,12 @@ Recent Deliveries). You should see a `200` response with:
 
 ### Security
 
-When `REPOWISE_GITHUB_WEBHOOK_SECRET` is set, every incoming request is
-verified using HMAC-SHA256 against the `X-Hub-Signature-256` header. Requests
-with invalid or missing signatures are rejected with `401 Unauthorized`.
-
-If the secret is **not** set, signature verification is skipped (convenient for
-local development, but not recommended for production).
+`REPOWISE_GITHUB_WEBHOOK_SECRET` is **required**. Every incoming request must
+carry a valid HMAC-SHA256 signature in the `X-Hub-Signature-256` header.
+Requests with an invalid signature are rejected with `401 Unauthorized`.
+Requests with **no** signature — including those sent without a secret
+configured on your server — are rejected with `403 Forbidden`. There is no
+unsigned mode.
 
 ---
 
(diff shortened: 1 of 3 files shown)


RaghavChamadiya (approved):
Auth logic verified: HMAC construction, header names, constant-time compares, per-request secret reads, and fail-closed via auth_is_open.

PR #1416, opened 2026-08-11
fix(cli): use DeadCodeFinding.kind in export --full JSON

Ayush7614:
## Summary
- `repowise export --format json --full` crashed with `AttributeError: DeadCodeFinding has no attribute finding_type`
- Serialize `kind` (the real ORM column) instead
- Add regression tests locking the model contract

## Why
Verified locally: full JSON export failed as soon as dead-code rows existed. No open/merged PR covers this.

## Test plan
- [x] `pytest tests/unit/cli/test_export_full_dead_code.py`
- [x] Confirm `DeadCodeFinding` has `kind`, not `finding_type`


RaghavChamadiya (approved):
Verified against the ORM model and the export path. Good catch.

PR #1418, opened 2026-08-11
fix(ui): replace dead architecture-diagram chat suggestion

Ayush7614:
## Summary
- Empty-state suggestion still asked for an architecture diagram from a removed chat tool
- Replace with a `get_change_risk`-shaped prompt

## Why
Users clicking the suggestion would hit a tool the chat registry no longer exposes.

## Test plan
- [x] `DEFAULT_SUGGESTIONS` no longer mentions architecture diagram


RaghavChamadiya (approved):
Verified against the live chat tool registry.

PR #1419, opened 2026-08-11
fix(server): drop dead get_architecture_diagram chat summary

Ayush7614:
## Summary
- Remove unreachable `_build_tool_summary` branch for `get_architecture_diagram`

## Why
Tool was removed from the chat registry; the summary handler could never fire.

## Test plan
- [x] Diff removes only the dead branch


RaghavChamadiya (approved):
Confirmed unreachable.

PR #1420, opened 2026-08-11
fix(server): summarize get_change_risk in chat tool results

Ayush7614:
## Summary
- Chat registers `get_change_risk` but summaries fell through to generic `Completed`
- Emit review priority + percentile (or score) instead
- Add unit tests

## Why
Without a real summary, the UI tool step is opaque for the main change-risk path.

## Test plan
- [x] `pytest tests/unit/server/test_chat_change_risk_summary.py`


RaghavChamadiya (approved):
Payload keys verified against change_risk_payload.

PR #1421, opened 2026-08-11
fix(server): align chat system prompt with the 7-tool registry

Ayush7614:
## Summary
- System prompt claimed **8** specialized tools; live registry has **7**
- Mention `get_change_risk` in guidelines
- Drift test against `get_tool_registry()`

## Why
Wrong count mis-trains the chat model about available tools.

## Test plan
- [x] `pytest tests/unit/server/test_chat_system_prompt_tools.py`
- [x] `len(get_tool_registry()) == 7`


RaghavChamadiya (approved):
Count verified against the registry.

PR #1422, opened 2026-08-11
fix(mcp): align FastMCP instructions with the live tool surface

Ayush7614:
## Summary
- MCP server blurb still advertised architecture diagrams and default dependency paths
- Update instructions to match default tools (change-risk, health) + workspace extras

## Why
Agents read FastMCP `instructions` at connect time; stale copy steers them toward removed/opt-in surfaces.

## Test plan
- [x] Diff against `docs/agent/MCP_TOOLS.md` default set


RaghavChamadiya (approved):
Verified against the registered tool surface.

PR #1423, opened 2026-08-11
fix(types): document live vs legacy chat artifact tool bindings

Ayush7614:
## Summary
- Document that `graph` / `diagram` artifacts are legacy wire shapes from removed chat tools
- Note `risk_report` covers both `get_risk` and `get_change_risk`

## Why
Types package comments still named removed tools as if they were current.

## Test plan
- [x] Comment-only; keeps wire types for historical SSE payloads


RaghavChamadiya (approved):
Comment-only, claims verified.

PR #1424, opened 2026-08-11
fix(types): update ChatArtifact contract comments for live tools

Ayush7614:
## Summary
- Refresh `contracts.test.ts` comments to the live 7-tool chat registry
- Add `risk_report` narrowing coverage in the type test

## Why
Contract test comments still pointed at removed `get_architecture_diagram` / wrong backend paths.

## Test plan
- [x] Type imports resolve (`RiskReportArtifact`)


RaghavChamadiya (approved):
Fills a real gap in the union test.

PR #1425, opened 2026-08-11
fix(server): drop unused get_dependency_path chat summary

Ayush7614:
## Summary
- Remove unreachable `_build_tool_summary` branch for `get_dependency_path`

## Why
Chat never registers that tool (MCP opt-in only); the summary branch was dead code.

## Test plan
- [x] Diff removes only the dead branch


RaghavChamadiya (approved):
Confirmed dead on the chat path.

PR #1439, opened 2026-08-11
fix(cache): HMAC-seal on-disk pickle caches before unpickling

### LINKED ISSUE #1390: Critical: RCE via unsafe pickle deserialization of .repowise parse/centrality/duplication caches
## Summary

`repowise` unpickles cache files that live **inside the repo being indexed**, with no integrity verification. A repo that ships a crafted `.repowise/parse_cache.pkl` (or `centrality_cache.pkl` / `duplication_cache.pkl`) executes arbitrary code on the machine that runs `repowise init` / `repowise update` against it. CWE-502 — deserialization of untrusted data → RCE.

## Affected code

- `packages/core/src/repowise/core/ingestion/parse_cache.py:131` → `payload = pickle.load(fh)` and `:183` → `pickle.loads(blob)`
- `packages/core/src/repowise/core/ingestion/graph/_centrality_cache.py:94` → `pickle.load(fh)`
- `packages/core/src/repowise/core/analysis/health/duplication/token_cache.py:55` → `pickle.load(fh)`

Cache locations (inside the indexed repo):

- `packages/core/src/repowise/core/pipeline/phases/ingestion.py:204` → `ParseCache(repo_path / ".repowise")`
- `packages/core/src/repowise/core/pipeline/incremental.py:115,123` → `.repowise/` caches

## Why the existing guard does not protect

The version / fingerprint checks in all three `load()` methods run **after** `pickle.load()` has already executed. Unpickling a crafted top-level payload is the attack — the checks never see it. There is no MAC, signature, or allowlist over the cached bytes.

## Reproduction (verified)

1. Craft a repo with a malicious cache:

   ```python
   import pickle, os
   class Evil:
       def __reduce__(self):
           return (os.system, ("touch /tmp/PWNED",))
   payload = {"version": 1, "fingerprint": "anything", "files": {("a/x.py", "HASH"): Evil()}}
   with open("repo/.repowise/parse_cache.pkl", "wb") as fh:
       pickle.dump(payload, fh)
   ```

2. `git add -f .repowise/parse_cache.pkl` and commit (`.repowise/` is gitignored, but `-f` overrides that — an attacker publishing a repo can ship it).
3. Victim runs `repowise init` / `repowise update` on the cloned repo.
4. Observed: the payload ran (a marker file was created during `ParseCache.load()`).

## Impact

Arbitrary code execution as the indexing user when indexing any untrusted repo — a supply-chain RCE. Anyone who indexes a repo from an untrusted source (or has a colleague commit a `.repowise/` dir) is a target.

## Suggested fix

- Use a trusted serializer for on-disk caches (e.g. `json` / `msgpack`), OR
- Verify a keyed HMAC over the pickle bytes before deserializing, using a key not stored alongside the cache, OR
- Never read the cache file from inside the repo tree.

Ayush7614:
## Summary
- Seal `.repowise/` pickle caches (parse, centrality, duplication token, duplication pairs) with a machine-local HMAC **before** `pickle.loads`.
- Unsigned legacy files and forged MACs are refused and degrade to a full recompute — the version/fingerprint checks no longer run after an attacker-controlled unpickle.
- Key lives outside the repo (`$XDG_CONFIG_HOME/repowise/cache_hmac_key`, or `REPOWISE_CACHE_HMAC_KEY` / `REPOWISE_CACHE_KEY_PATH` overrides).

## Related Issues
Fixes #1390

## Test plan
- [x] `pytest tests/unit/test_cache_seal.py tests/unit/ingestion/test_parse_cache.py tests/unit/ingestion/test_centrality_cache.py tests/unit/health/test_duplication_cache.py tests/unit/health/test_duplication_incremental.py` (50 passed)
- [x] Regression: unsigned pickle with `__reduce__` gadget does not execute / does not load
- [x] Forged HMAC under a different key is a miss

### DIFF
--- a/packages/core/src/repowise/core/analysis/health/duplication/pair_index.py
+++ b/packages/core/src/repowise/core/analysis/health/duplication/pair_index.py
@@ -38,6 +38,8 @@ from pathlib import Path
 
 import structlog
 
+from repowise.core.cache_seal import seal, unseal
+
 from .limits import DuplicationLimits
 
 log = structlog.get_logger(__name__)
@@ -95,8 +97,7 @@ def load_pair_index(
     """Load and validate the artifact; ``None`` on any mismatch/error."""
     path = Path(cache_dir) / _INDEX_FILENAME
     try:
-        with path.open("rb") as fh:
-            payload = pickle.load(fh)
+        payload = pickle.loads(unseal(path.read_bytes()))
         if (
             payload.get("version") != _INDEX_VERSION
             or payload.get("window_tokens") != window_tokens
@@ -116,7 +117,7 @@ def load_pair_index(
         )
     except FileNotFoundError:
         return None
-    except Exception as exc:  # corrupt / unreadable -> full re-detect
+    except Exception as exc:  # corrupt / unsigned / unreadable -> full re-detect
         log.debug("duplication_pair_index_load_failed", error=str(exc))
         return None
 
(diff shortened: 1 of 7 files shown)


RaghavChamadiya (approved):
LGTM

PR #1440, opened 2026-08-11
fix(webhooks): match repos by normalized URL, not contains prefix

### LINKED ISSUE #1392: High: webhook endpoints unauthenticated and signature verification skipped when no secret configured
## Summary

`POST /api/webhooks/github` and `POST /api/webhooks/gitlab` carry **no bearer/auth dependency**, and both HMAC/token checks `return` early ("dev mode") when the secret env var is unset — which is the shipped default. Anyone who can reach the port can POST a forged `push` payload and enqueue sync/re-generation jobs, even when `REPOWISE_API_KEY` **is** set.

## Affected code

- `packages/server/src/repowise/server/routers/webhooks.py:17` — `router = APIRouter(prefix="/api/webhooks", tags=["webhooks"])` (no `Depends(verify_api_key)`)
- `packages/server/src/repowise/server/routers/webhooks.py:19-20` — `_GITHUB_SECRET` / `_GITLAB_TOKEN` default to `""`
- `packages/server/src/repowise/server/routers/webhooks.py:23-27` — `_verify_github_signature`: `if not _GITHUB_SECRET: return`
- `packages/server/src/repowise/server/routers/webhooks.py:41-44` — `_verify_gitlab_token`: `if not _GITLAB_TOKEN: return`
- `packages/server/src/repowise/server/routers/webhooks.py:114-117` / `:172-175` — repo matching via `Repository.url.contains(repo_url[:50])`
- `packages/server/src/repowise/server/routers/webhooks.py:119-132` / `:177-190` — job upsert + `_launch_webhook_job`

## Attack scenario

```bash
curl -X POST http://<host>:7337/api/webhooks/github \
  -H "X-GitHub-Event: push" \
  -H "Content-Type: application/json" \
  -d '{"ref":"refs/heads/main","before":"0000","after":"deadbeef",
       "repository":{"clone_url":"https://github.com/repowise-dev/repowise.git"}}'
```

Because the signature header is never checked (secret unset) and the endpoint requires no API key, this reliably enqueues a `sync` generation job (LLM spend, CPU, DB writes) for the matching repo. An attacker can also choose the target by matching another registered repo's URL, and can point the `before` ref to force an expensive `git diff`-driven page regen (`job_executor.py:1171-1229`).

## Impact

- **Resource exhaustion / unbounded LLM cost**: any network peer (or any webpage that can induce a cross-site POST) can trigger re-indexing and page regeneration at will.
- Circumvents the `REPOWISE_API_KEY` protection that protects every other route: these two routes sit in the same `app.include_router` list (`app.py:480`) but outside the key gate.

## Suggested fix

- Add `Depends(verify_api_key)` (or a dedicated webhook-token check) to both handlers, and
- Treat a missing signature as **reject** rather than pass: fail closed when `REPOWISE_GITHUB_WEBHOOK_SECRET` / `REPOWISE_GITLAB_WEBHOOK_TOKEN` is unset, or document + require them.

Ayush7614:
## Summary
- Replace `Repository.url.contains(repo_url[:50])` with exact normalized host/path matching for GitHub and GitLab push webhooks.
- A short forged URL (e.g. `https://github.com/org/re`) can no longer enqueue sync jobs for a longer registered repo (`…/repowise`).
- Still accepts common URL variants (`.git` suffix, trailing slash, `git@host:path`).

Auth fail-closed when the webhook secret/token is unset was already fixed in #1408 (against #1398). This PR closes the remaining matching gap called out in #1392.

## Related Issues
Fixes #1392

## Test plan
- [x] `pytest tests/unit/server/test_webhooks.py` (14 passed)
- [x] Prefix-substring attack does not enqueue a job
- [x] `.git` / bare URL variants still match the registered repo
- [x] Existing signature / fail-closed / enqueue tests still pass

### DIFF
--- a/packages/server/src/repowise/server/routers/webhooks.py
+++ b/packages/server/src/repowise/server/routers/webhooks.py
@@ -29,9 +29,10 @@ def _normalize_scm_url(url: str) -> str:
     """Collapse clone / web URL variants to a comparable host/path key.
 
     Strips scheme, userinfo, ``.git`` suffix, trailing slashes, and lowercases
-    the host. Used so ``https://github.com/org/repo.git`` matches a stored
-    ``https://github.com/org/repo`` without falling back to a substring
-    ``contains`` match (which let a short forged URL hit the wrong repo).
+    the **whole** key (host and path). GitHub and GitLab resolve repo paths
+    case-insensitively; the previous ``contains`` match was also case-insensitive
+    under SQLite, so preserving path case would silently stop syncing when a
+    hand-registered URL differed only in casing from the webhook's ``clone_url``.
     """
     raw = (url or "").strip()
     if not raw:
@@ -47,7 +48,7 @@ def _normalize_scm_url(url: str) -> str:
     path = path.strip("/")
     if path.endswith(".git"):
         path = path[: -len(".git")]
-    return f"{host.lower()}/{path}".rstrip("/")
+    return f"{host}/{path}".rstrip("/").lower()
 
 
 async def _find_repo_by_scm_url(session: AsyncSession, url: str) -> Repository | None:
--- a/packages/server/src/repowise/server/routers/webhooks.py
+++ b/packages/server/src/repowise/server/routers/webhooks.py
@@ -6,11 +6,15 @@ import hashlib
 import hmac
 import json
 import os
+import re
+from urllib.parse import urlparse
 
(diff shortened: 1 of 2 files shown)


RaghavChamadiya (approved):
Approved. Casing fix verified on the head commit.

PR #1441, opened 2026-08-11
fix: suppress banner in dead-code --format json output

### LINKED ISSUE #1436: [Bug] dead-code --format json prints a banner to stdout, so the output is not valid JSON
## Describe the Bug

`repowise dead-code --format json` prints a human-readable banner to **stdout** before the JSON payload, so the output is not parseable JSON. Every other agent-facing command that grew `--format json` in this cycle keeps stdout clean; `dead-code` is the one that doesn't.

The README documents a usage that produces a broken file:

```
repowise dead-code --format json > dead-code.json    # packages/cli/README.md:286
```

`dead-code.json` starts with `repowise dead-code — /path/to/repo`, so `json.load()` on it raises.

## Steps to Reproduce

```bash
cd <any indexed repo>
repowise dead-code --format json | python -c "import json,sys; json.load(sys.stdin)"
```

## Expected Behavior

Exit 0, JSON parses. In `--format json` mode stdout carries the payload and nothing else — the banner either goes to stderr or is suppressed.

## Actual Behavior

```
  ...
json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)
```

First lines of stdout:

```
repowise dead-code — C:\...\scratchpad\md
[
  {
    "kind": "unreachable_file",
```

## Where

`packages/cli/src/repowise/cli/commands/dead_code_cmd.py:131`

```python
console.print(f"[bold]repowise dead-code[/bold] — {repo_path}")
```

Unconditional — no `format` check. There is a second stdout leak on the same path at line 126, which fires only in workspace mode:

```python
console.print("[dim]  (Tip: pass --repo <alias> to analyze a different repo.)[/dim]")
```

Both should be gated on the output format (or routed to stderr).

## Verification done

Ran all agent-facing commands with `--format json` against a freshly indexed repo and parsed **stdout only**. 11 of 12 are clean; `dead-code` is the sole failure:

| command | stdout parses as JSON |
|---|---|
| `ask`, `context`, `context` (symbol), `symbol`, `why`, `search`, `search --mode symbol` | yes |
| `risk <revspec>`, `risk -t`, `risk -t --changed-file`, `health`, `status` | yes |
| `dead-code` | **no** |

## Environment

- OS: Windows 11
- Python: 3.11
- Repowise version: 0.40.0 (editable checkout at `f82ebb0d`, i.e. `main` after the 0.41.0 cycle merged)
- Installation method: editable / uv

mercael91:
## Summary

`repowise dead-code --format json` printed a human-readable banner to stdout before the JSON payload, making the output invalid JSON. Every other agent-facing command that grew `--format json` keeps stdout clean; `dead-code` is the one that doesn't.

## Change

Wrap the banner in a `fmt` check so it only prints for `table` and `md` formats:

```python
# Before
console.print(f"[bold]repowise dead-code[/bold] — {repo_path}")

# After
if fmt != "json":
    console.print(f"[bold]repowise dead-code[/bold] — {repo_path}")
```

Fixes #1436

### DIFF
--- a/packages/cli/src/repowise/cli/commands/dead_code_cmd.py
+++ b/packages/cli/src/repowise/cli/commands/dead_code_cmd.py
@@ -14,6 +14,7 @@ from repowise.cli.helpers import (
     run_async,
     silence_logs_for_machine_output,
 )
+from repowise.cli.output import notice_console
 from repowise.core.analysis.dead_code.risk_factors import RISK_CAP_CONFIDENCE
 
 
@@ -110,7 +111,8 @@ def dead_code_command(
         no_workspace_flag=no_workspace,
         repo_alias=repo_alias,
     )
-    target.notice(console, command="dead-code")
+    notices = notice_console(fmt)
+    target.notice(notices, command="dead-code")
 
     if target.is_workspace:
         if target.repo_filter is not None:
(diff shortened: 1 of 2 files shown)


RaghavChamadiya (approved):
Thanks for this. Routing the banner and the tip through `notice_console(fmt)` is exactly the house pattern, so notices land on stderr instead of vanishing, and a human running with `--format json` still sees why the payload looks the way it does. Adding the command to the rollout test list is the part that keeps it from regressing. Merging.

PR #1445, opened 2026-08-12
fix(get_answer): completeness scope gate for truncated symbol bodies 

### LINKED ISSUE #1444: [Bug] get_answer can claim a cause is the only cause, including over a symbol body it truncated
## Describe the Bug

`get_answer` synthesises prose over the top-k retrieval hits. Nothing in that
pipeline can establish that it has seen *every* relevant site, but the prose is
free to claim it has — and does.

On a defect whose cause was split across two functions, the answer named one of
them and stated that the outcome "depends entirely on `<function_a>`". The
second site was in the payload: its signature, docstring and call site were all
included. Only its **body** was omitted, and the payload correctly marked it
`"truncated": true` with a `continuation` pointer and a
`call get_symbol, do NOT Read` hint.

So the structured fields were right and the prose was wrong, in the same
response. A caller that reads the prose — the part written to be read — gets a
completeness claim the tool has no basis for, pointing away from the half of the
cause it was actually shown.

### Why this is wrong by construction, not just unlucky

This is not a ranking miss or a retrieval gap. Retrieval returned the right
material. The issue is that **exhaustiveness is not a property `get_answer` can
observe.** It sees a truncated, budgeted, top-k slice of the repo. Words like
"entirely", "only", "the sole place" assert a global property from a local
sample, so they are never justified from this pipeline — not occasionally wrong,
but unfounded in every case where they appear.

The failure mode is worst exactly where the tool is most useful: "why does X
happen" questions on multi-site causes, where a confident single-site answer
reads as a complete diagnosis and terminates the search early.

### Where it comes from

`packages/server/src/repowise/server/mcp_server/tool_answer/config.py`

`_SYSTEM_PROMPT` currently says:

> "Only hedge (say 'inspect the source' / 'the excerpts do not contain…') when
> there is genuinely no relevant signature, docstring, or source body in the
> excerpts."

and `_USER_TEMPLATE` repeats it:

> "Only hedge if no signature, docstring, or source body in the excerpts is
> relevant."

That instruction conflates two different hedges:

1. **Epistemic** — "I was not given enough material to answer." Suppressing this
   is correct and is clearly what the instruction is aimed at; hedging here
   makes answers useless.
2. **Completeness** — "this may not be the only place this happens." The
   instruction suppresses this one too, as collateral damage.

Because relevant excerpts *were* present, the model followed the instruction
exactly and did not qualify — converting "here is what I was shown" into "this
is the whole story". The prompt is working as written; what it is written to do
is wrong for (2).

## Steps to Reproduce

Any repo with a defect whose cause spans two functions, where retrieval returns
both but budgets the second one's body:

1. Ask `get_answer` a "why does X happen" question about the behaviour.
2. Check whether the prose contains an unqualified exclusivity token
   ("entirely", "only", "the sole").
3. Check whether any symbol cited in that claim carries `"truncated": true`.

(2) and (3) being simultaneously true is the bug. Both are assertable from the
response object, so this can be pinned by a test rather than reviewed by eye.

## Expected Behavior

The answer states what the retrieved material shows, without asserting it is the
complete set of sites. For example:

> "the excerpts show `<function_a>` doing X; other sites may also participate"

rather than:

> "the final result depends entirely on `<function_a>`"

Roughly one clause longer, and it keeps the caller looking.

## Actual Behavior

The answer asserts a single site is the whole cause, while the same response
carries `"truncated": true` on a second symbol it cited as part of the
mechanism. A consumer acting on the prose edits the one named site, sees a green
test suite (existing tests did not cover the second site — that is why the
defect existed), and stops. The result is a plausible, partial fix that looks
verified.

This is more damaging than a low-confidence answer, because a hedged answer
would have kept the search going. It also silently undercuts the `truncated` /
`continuation` / `get_symbol` affordances: they are correct, present, and simply
outranked by the prose.

## Environment

- OS: Windows 11
- Python version: 3.13.5
- Repowise version: 0.41.0 (`repowise --version`)
- Installation method: editable install from source

### Suggested fix

Smallest change that addresses the cause, in `config.py`:

1. **Split the hedging instruction.** Keep "do not hedge about having enough
   material". Add that completeness is never assumed: an answer may state what
   the retrieved material shows, and may not state that it is the only or
   complete set of sites.
2. **Ban unqualified exclusivity language** from synthesis output — "entirely",
   "only", "the sole" — unless attributed to something in the retrieved material
   that actually says so (an assertion, a type constraint, an explicit comment).
   Words like "always" / "never" are legitimate when quoting such a constraint,
   so the rule should be about *unattributed* exclusivity rather than a
   vocabulary blocklist.
3. **Make truncation and causal claims mutually exclusive.** If a symbol's body
   arrived `truncated: true`, the answer must not make an unqualified causal
   claim about that symbol's behaviour. This one is detectable in-process — the
   flag is already on the entry — so it can be a check rather than a hope.

### Trade-off, stated rather than buried

Completeness hedging makes answers slightly longer, and `get_answer` already
returns the largest payload of any tool in the surface. This trades a small
amount of output for a correctness property. That seems clearly worth it for
"why does X happen" questions; it may be unnecessary for pure "where is X
defined" lookups, so scoping the requirement to causal/mechanism questions is a
reasonable middle.

### On verifying the fix

Worth proving in both directions: the check should fail on `main` today and pass
after the change. A check whose failure path has never fired is not a check —
and here the natural wrong version ("does the answer mention the second
function?") would already pass today, because the answer *did* mention it. The
defect is the exclusivity claim, not the omission, so that is what the assertion
has to target.

akshatmalik-bruh:
## Summary

- Adds a 7th post-synthesis confidence gate (`_has_unqualified_exclusivity_over_truncated`) that downgrades confidence (`high` → `medium`) and surfaces an explicit completeness note when prose asserts an unqualified exclusivity claim (`"entirely"`, `"the sole"`, `"depends only on"`) over a symbol whose body arrived with `truncated: true`.
- Splits prompt instructions in `_SYSTEM_PROMPT` and `_USER_TEMPLATE` (`config.py`) to separate epistemic hedge suppression ("don't say you lack material") from completeness scoping ("never assert exhaustiveness on top-k slices").
- Adds a dedicated test suite (`test_answer_completeness_scope.py`) proving the gate fires on claims over truncated bodies while preserving `high` confidence on non-truncated or properly-scoped negative controls.

## Related Issues

Fixes #1444

## Proof of Fix (Before vs After)

image

image

image

---

## Test Plan

- [x] New unit and e2e tests pass (`uv run pytest tests/unit/server/mcp/test_answer_completeness_scope.py -v`)
- [x] Existing calibration test suite passes (`uv run pytest tests/unit/server/mcp/test_answer_calibration.py -v`)

## Checklist

- [x] I have updated documentation if needed```

### DIFF
--- a/packages/server/src/repowise/server/mcp_server/tool_answer/answer.py
+++ b/packages/server/src/repowise/server/mcp_server/tool_answer/answer.py
@@ -117,6 +117,7 @@ from repowise.server.mcp_server._neighbor_rerank import (
 from repowise.server.mcp_server.tool_answer.confidence import (
     _answer_is_hedged,
     _frame_term_grounding,
+    _has_unqualified_exclusivity_over_truncated,
     _is_value_question,
     _ungrounded_numbers,
 )
@@ -1729,6 +1730,20 @@ async def get_answer(
         else:
             frame_unsupported = []
 
+    # Seventh gate — completeness scope over truncated bodies: when prose asserts
+    # an unqualified exclusivity claim ("entirely", "the sole", "the only") while
+    # any cited symbol body arrived truncated: true, the answer asserts a global
+    # property from a sample it knows is incomplete (#1444).
+    # Guard: only fires at high (like gates 3 / 5 / 6) so it cannot stack with
+    # other downgrades and push a single response two levels for one problem.
+    exclusivity_over_truncated = False
+    if confidence == "high" and not hedged:
+        exclusivity_over_truncated = _has_unqualified_exclusivity_over_truncated(
+            answer_text, symbol_bodies
+        )
+        if exclusivity_over_truncated:
+            confidence = "medium"
+
     # Non-dominant ceiling: ambiguous retrieval is the calibration cost of
     # always synthesizing — the answer may be right, but with no single dominant
     # page it must never read "high" (cite without verifying). Cap at medium even
@@ -1873,6 +1888,22 @@ async def get_answer(
                 f"Verify the mechanism before citing: the asserted term(s) "
                 f"{frame_unsupported} are not in the retrieved material."
             )
+        elif exclusivity_over_truncated:
+            # Note names the axis of doubt (what to be uncertain about), not the
+            # check that triggered it — so a reader can tell which kind of doubt
+            # this is without consulting the source code.
+            payload["note"] = (
+                "Answer may not cover every relevant site: a cited symbol's body "
+                "was truncated and the answer makes an unqualified causal claim. "
+                "Other functions may also participate; call get_symbol for the "
+                "full body or verify against "
+                f"{fallback_targets[0] if fallback_targets else 'the cited source'}."
+            )
+            if fallback_targets:
+                payload["next_action_hint"] = (
+                    f"Read {fallback_targets[0]} to verify whether other functions "
+                    "participate beyond what the truncated symbol body shows."
+                )
         elif confidence == "high":
             payload["note"] = (
                 "High confidence: top retrieval result clearly dominates "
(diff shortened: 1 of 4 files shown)


swati510 (approved):
lgtm

PR #1452, opened 2026-08-12
fix(health): clamp share_of_repo_gap_pct to 100%

### LINKED ISSUE #1437: [Bug] get_health share_of_repo_gap_pct exceeds 100% (up to 1467%) — gross deficit over net gap
## Describe the Bug

`get_health`'s `share_of_repo_gap_pct` is a percentage that routinely exceeds 100% — up to **1467.5%** on a mostly-healthy repo. It appears in the `directive` block (the first thing an agent reads) and on every `high_leverage_files` row.

The cause is that the numerator and denominator measure two different quantities:

- **Numerator** — the file's *gross* deficit, `max(8.0 - score, 0) * nloc`, summed only over files below 8.0.
- **Denominator** — `gap_analysis.weighted_gap_points`, the *net* gap, `8.0 * total_nloc - Σ(score * nloc)`. Files scoring **above** 8.0 contribute negative terms and cushion this total.

On a repo whose weighted average is already near 8.0, the net gap is small while individual gross deficits are not, so the ratio explodes. The healthier the repo, the more absurd the number.

This is why dogfooding never caught it: on this repo (avg 7.46) the net gap is large and the shares look plausible.

## Steps to Reproduce

Index a repo whose weighted average health is just under 8.0 (microdot at avg 7.96 works), then:

```python
r = await get_health()
r["directive"]["share_of_repo_gap_pct"]   # 1467.5
r["gap_analysis"]["weighted_gap_points"]  # 507
r["directive"]["recovers_points"]         # 7440
```

## Expected Behavior

A field named `share_of_repo_gap_pct` is a share of a whole and should land in 0–100, with the rows summing to at most 100%.

## Actual Behavior

Measured on two repos with the same build (`f82ebb0d`):

**microdot — avg 7.96, 174 files, net gap 507**

```
     1467.5%  weighted_deficit=  7440  src/microdot/microdot.py
      317.5%  weighted_deficit=  1610  libs/micropython/datetime.py
      149.4%  weighted_deficit=   758  src/microdot/asgi.py
      135.1%  weighted_deficit=   685  libs/micropython/unittest.py
      105.6%  weighted_deficit=   536  src/microdot/multipart.py
  sum of all 8 rows = 2545.4%
  gross deficit of those rows = 12906  vs  NET repo gap = 507
```

**repowise — avg 7.46, 3334 files, net gap 265005** (looks fine, masking the bug)

```
        3.6%  weighted_deficit=  9576  packages/core/src/repowise/core/pipeline/persist.py
        3.6%  weighted_deficit=  9562  packages/cli/src/repowise/cli/commands/update_cmd/command.py
  sum of all 20 rows = 52.1%
```

## Where

Two call sites divide a gross deficit by the net gap:

- `packages/server/src/repowise/server/mcp_server/tool_health.py:607` — the `directive` field
  ```python
  "share_of_repo_gap_pct": (round(100.0 * recovers / gap_points, 1) if gap_points else None),
  ```
- `packages/server/src/repowise/server/mcp_server/tool_health.py:1458-1467` — each `high_leverage_files` row (same formula inline)

The net-gap denominator comes from `_gap_analysis` at `tool_health.py:681`:

```python
net_gap = HEALTHY_MIN * total_nloc - weighted_sum
```

## Suggested fix

Pick one and use it in both places:

1. **Divide by the gross deficit** — `Σ max(8.0 - score, 0) * nloc` over all files below target. Then the value is a true share of the repo's total below-target deficit, rows sum to ≤100%, and the "few files, not the long tail" framing the comment describes still holds.
2. **Keep the net denominator but rename and clamp** — the quantity "this file alone covers 14.7x the net gap to 8.0" is meaningful, but it is not a share and must not be named `_pct`.

Option 1 preserves the field name and the documented meaning, so it is probably the smaller change. Whichever is chosen, `gap_analysis.weighted_gap_points` should stay net — `files_to_reach_target` depends on it and is correct as-is.

## History

Not a fresh regression, but newly prominent:

- The formula has been on `high_leverage_files` since `3d4042a2` (2026-07-28, #1142); the net denominator dates to `79fa4076` (2026-07-08, #719).
- `23cfd68b` (2026-08-11, #1414) promoted the same expression into `directive`, which is the block the tool docstring tells agents to read first.

Found while dogfooding the 0.41.0 release candidate; not a release blocker, since the number is misleading rather than load-bearing.

## Environment

- OS: Windows 11
- Python: 3.11
- Repowise version: 0.40.0 (editable checkout at `f82ebb0d`, i.e. `main` after the 0.41.0 cycle merged)
- Installation method: editable / uv

Ameysr:
## Summary

Fixes #1437 — `get_health`'s `share_of_repo_gap_pct` is now capped at 100%.
It previously reported percentages up to 1467.5% because the numerator and
denominator measured different quantities: a single file's **gross** deficit
vs the repo's **net** gap.

## Root cause

In `_directive` (`tool_health.py`), the percentage was computed as
`100 * recovers / gap_points` where:

- **Numerator** — the top file's gross deficit: `max(8.0 - score, 0) * nloc`,
  summed only over files below 8.0.
- **Denominator** — `weighted_gap_points`, the repo's **net** gap:
  `8.0 * total_nloc - Σ(score * nloc)`. Files scoring **above** 8.0 contribute
  negative terms that shrink this total.

On a mostly-healthy repo, the healthy files' surplus cushions the net gap, so
one large low-scoring file's gross deficit can exceed the whole net gap — the
share reads as >100% (the test even asserted the buggy `103.7`).

## Change

Clamp the percentage to ≤100% — a single file cannot close more than the
entire remaining gap:

```python
"share_of_repo_gap_pct": (
    round(min(100.0 * recovers / gap_points, 100.0), 1) if gap_points else None
),
```

- `recovers_points` (a raw point count) is unchanged.
- `weighted_gap_points` keeps its net-gap value — the documented intent
  ("points the weighted average needs to reach 8.0") is preserved.
- When there is no net gap (weighted average already ≥ 8.0), the directive is
  absent entirely — there is nothing to recommend.

## Files changed

- `packages/server/src/repowise/server/mcp_server/tool_health.py`
- `tests/unit/server/mcp/test_health.py`
- `docs/agent/MCP_TOOLS.md`

## Tests

- Updated `test_get_health_dashboard_leads_with_a_directive`: the fixture's
  700-point gross deficit vs 675-point net gap now asserts `100.0` (was the
  buggy `103.7`).
- New `test_directive_share_capped_when_gross_exceeds_net_gap`: a 2000-point
  gross deficit vs 1800-point net gap (111.1% unclamped) clamps to `100.0`.
- New `test_directive_share_is_none_when_repo_already_healthy`: no directive
  when nothing is below the Healthy floor.

## Verification

- `tests/unit/server/mcp/test_health.py` — 56 passed.
- Full MCP server suite — no failures (ran through 81% before the 10-min
  suite timeout; all health tests passed).
- `ruff check` — clean.

### DIFF
--- a/docs/agent/MCP_TOOLS.md
+++ b/docs/agent/MCP_TOOLS.md
@@ -546,7 +546,10 @@ per-dimension scores, and the score breakdown. Each finding carries a `dimension
 
 **Lead with `directive`.** Dashboard mode opens with the single file to fix
 first, its dominant finding, `recovers_points` / `share_of_repo_gap_pct` (what
-fixing it buys the headline), and `then`, the next two by leverage. Every other
+fixing it buys the headline; the share is bounded by 100% and sums to 100%
+across `high_leverage_files` — the gross deficit of below-target files is the
+denominator, not the net gap, so a single file cannot read as closing more than
+the whole remaining gap), and `then`, the next two by leverage. Every other
 block ranks and describes; this one recommends. Same role as `get_risk`'s
 `directive`. Rank by `weighted_deficit`, not `score` — the score floors at 1.0.
 
@@ -611,8 +614,10 @@ make that actionable rather than a mystery:
   `gap_analysis.weighted_gap_points` share one unit — *score-points x NLOC* —
   which compares against itself and nothing else. Every `high_leverage_files`
   row and the `directive` also carry `share_of_repo_gap_pct`, the same quantity
-  with a denominator; that plus `gap_analysis.files_to_reach_target` is what
-  answers "is this worth doing".
+  over the gross deficit of all below-target files
+  (`gap_analysis.weighted_gross_gap_points`), so the shares are bounded by 100%
+  and sum to 100% by construction; that plus
+  `gap_analysis.files_to_reach_target` is what answers "is this worth doing".
 - `kpis.non_code_files` and `kpis.average_health_code_only` say how much of the
   headline is markdown/JSON/YAML. No biomarker walks those files, so they score
   a mechanical 10.0 meaning "nothing looked at this" — on this repo, 233 of
--- a/packages/server/src/repowise/server/mcp_server/tool_health.py
+++ b/packages/server/src/repowise/server/mcp_server/tool_health.py
@@ -602,9 +602,14 @@ def _directive(
         "reason": lead.get("primary_reason") or f"scores {round(top.score, 2)}",
         # Points the repo headline recovers if this one file reaches Healthy,
         # and what share of the total gap that is — the "few files, not the
-        # long tail" argument made concrete for a single file.
+        # long tail" argument made concrete for a single file. The denominator
+        # is the *gross* deficit of all below-target files (not the net gap,
+        # which healthy files cushion): per-file shares are then bounded by
+        # 100% and sum to 100% by construction (issue #1437).
         "recovers_points": recovers,
-        "share_of_repo_gap_pct": (round(100.0 * recovers / gap_points, 1) if gap_points else None),
+        "share_of_repo_gap_pct": (
+            round(100.0 * recovers / gap_points, 1) if gap_points else None
+        ),
         "then": [m.file_path for m in by_leverage[1:3]],
         # Projected, not bare. ``include`` adds a block without subtracting the
         # dashboard, and five ranked lists at the default ``limit`` compose: the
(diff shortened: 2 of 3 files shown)


RaghavChamadiya (approved):
Denominator moved to the gross gap, clamp dropped, and the net number kept where it belongs. Verified the 100% bound holds by construction.

PR #1454, opened 2026-08-12
fix(ingestion): resolve TS workspace members from pnpm-workspace.yaml

### LINKED ISSUE #1453: [Bug] pnpm monorepos resolve no workspace members: pnpm-workspace.yaml is never read
### What happens

`ts_workspace.py` resolves workspace members only from the root `package.json`'s
`workspaces` field. pnpm does not use that field — a pnpm monorepo declares its
members in `pnpm-workspace.yaml` — so on a pnpm repo `build_workspace_info()`
returns `{}` and **every cross-package import resolves to an `external:` node
instead of an intra-repo edge**.

The module docstring already advertises the behaviour that is missing:

```
"""npm/yarn/pnpm workspace package resolution for TypeScript imports.

Reads the root ``package.json``'s ``workspaces`` field ...
```

`resolve_via_workspaces()`'s docstring says "npm/yarn/pnpm" too, and
`grep -r pnpm-workspace` across the repo returns nothing, so this reads as an
unfinished intent rather than a deliberate scope choice.

Because everything funnels through `build_workspace_info()`, the blind spot is not
limited to import resolution — `build_ts_workspace_index()` and the dead-code
analyzer's `exports_entry_paths` are built from the same empty map, so a pnpm
monorepo also loses its "public surface" protection and can take false-positive
unreachable findings.

### Why it matters

pnpm is the default package manager for a large share of current TS monorepos
(Nx, Turborepo and Vue/Vite templates all scaffold `pnpm-workspace.yaml`). On
those repos the dependency-graph layer is blind at precisely the package
boundaries that carry the architecture.

### Reproduction

A pnpm monorepo — root `package.json` with **no** `workspaces` key, plus:

```yaml
# pnpm-workspace.yaml
packages:
  - "apps/*"
  - "packages/*"
```

```python
from pathlib import Path
from repowise.core.ingestion.resolvers.ts_workspace import build_workspace_map

build_workspace_map(Path("/path/to/pnpm/monorepo"))
# {}   — expected: every apps/* and packages/* member, plus the root
```

Measured on a real pnpm monorepo (`pnpm@11`, TS, 18 packages across `apps/*` +
`packages/*` + one plain `scripts` entry, plus a named private root), read-only:

| | members discovered | `@scope/domain` | `@scope/db` |
|---|---|---|---|
| today | 0 | `None` → `external:` | `None` → `external:` |
| reading `pnpm-workspace.yaml` | 19 | `packages/domain/src/index.ts` | `packages/db/src/index.ts` |

### A second, pre-existing bug this uncovers

A workspace member **at the repo root** carries `dir == "."`, and the eight join
sites build candidates as `f"{pkg_dir}/{sub}"` → `"./src/index.ts"`, which never
equals the repo-relative `"src/index.ts"` in `path_set`. This is not pnpm-specific
— it reproduces today on npm/yarn:

```python
# root package.json: {"name": "@org/root", "workspaces": ["."],
#                     "exports": {".": "./src/index.ts"}}
build_workspace_map(d)                      # {'@org/root': '.'}   ← mapped
resolve_via_workspaces("@org/root", ctx)    # None                 ← never resolves
```

It matters here because pnpm's `packages` setting says "the root package is always
included, even when custom location wildcards are used", so reading the pnpm
manifest makes root members common rather than rare.

### Suggested fix

Read `pnpm-workspace.yaml`'s `packages` list and feed it into the existing glob
expansion in `build_workspace_info()`. Four details matter, and three of them are
places where the obvious implementation is wrong:

1. **Precedence, not union.** pnpm ignores `workspaces` entirely, so when both
   manifests exist, merging them registers members pnpm never installs — turning a
   registry dependency into a fake intra-repo edge. The pnpm manifest has to win.
2. **The root is always a member**, and is not subject to the negated patterns.
   Omitting `packages` is legal and means root-only.
3. **`!` negation needs the include globber, not a gitignore matcher.** `pathspec`
   is the tempting choice and is wrong: under git-ignore semantics a slashless
   `foo` matches at *any* depth and a directory match swallows its descendants, so
   `!fixtures` and `!packages/*` silently drop members pnpm keeps. (Measured:
   `pathspec.match_file` returns True for `foo` vs `packages/foo/`, and for
   `packages/*` vs `packages/a/b/`.) Expanding negations with `Path.glob` — the
   same globber the includes already use — anchors at the root and honours
   single-segment `*`, matching fast-glob for these forms.
4. **Only `pnpm-workspace.yaml`.** `.yml` is an open request (pnpm/pnpm#1380), not
   a supported spelling.

No new dependency is needed — `pyyaml` is already declared and imported in five
`core` modules, including the sibling `ingestion/resolvers/sql.py`.

I have this implemented with unit tests, including the root-`dir` normalization,
and would like to take it if the scope looks right to you. Happy to split the
root-`dir` fix into its own PR if you would rather keep them separate — it is
about 3 lines plus tests, and the pnpm work is what makes it reachable.

alexander-mitelman:
## Summary

- Read `pnpm-workspace.yaml`'s `packages` list when building the TS/JS workspace
  map, so pnpm monorepos resolve cross-package imports to intra-repo files instead
  of `external:` nodes. The pnpm manifest is **authoritative** when present, not
  merged with `package.json`'s `workspaces`.
- Include the workspace root as a member on pnpm repos, per pnpm's `packages`
  setting, and fix the pre-existing `dir == "."` join bug that stopped any
  root-level member from resolving (it reproduces on npm/yarn today).
- Expand `!` negation entries with the same `Path.glob` the includes use, rather
  than a gitignore matcher, so pnpm's anchoring and single-segment semantics hold.
- No new dependency: `pyyaml` is already declared and imported in five `core`
  modules, including the sibling `ingestion/resolvers/sql.py`.

## Related Issues

Fixes #1453

## Why

`build_workspace_info()` was the only source of workspace members and read only the
`workspaces` field. pnpm does not use that field, so on a pnpm repo the map came
back empty. Since `build_ts_workspace_index()` and the dead-code analyzer's
`exports_entry_paths` are built from the same map, the effect went beyond import
resolution — a pnpm monorepo also lost its `exports` "public surface" protection.
The fix goes in that one function, so all three consumers are fixed together.

## Three places the obvious implementation is wrong

I wrote each of these the naive way first; all three are covered by tests that fail
on the naive version.

1. **Union would be wrong.** pnpm ignores `package.json#workspaces`, so merging the
   two manifests registers members pnpm never installs — a registry dependency
   becomes a fake intra-repo edge. Precedence is also simpler.
2. **`pathspec` would be wrong.** It implements git-ignore semantics, under which a
   slashless `foo` matches at any depth and a directory match swallows its
   descendants. Measured: `pathspec.match_file` returns True for `foo` vs
   `packages/foo/`, and for `packages/*` vs `packages/a/b/`. Both silently DROP
   real members. `Path.glob` anchors at the repo root and honours single-segment
   `*`, matching fast-glob (pnpm's matcher) for these forms — and it is already
   what the include path uses, so excludes are now consistent with includes by
   construction rather than by a second pattern language.
3. **`.yml` is not a pnpm manifest.** An open request (pnpm/pnpm#1380), not a
   supported spelling. Reading it would map members from a file pnpm ignores.

## The root-`dir` fix

pnpm's `packages` setting says "the root package is always included, even when
custom location wildcards are used", which makes root members common instead of
rare. They did not work: the eight join sites build `f"{pkg_dir}/{sub}"`, so a root
member's `"."` produced `"./src/index.ts"` and never matched the repo-relative
`"src/index.ts"` in `path_set`. Normalizing inside `_probe_path` covers seven sites;
the `exports` wildcard `base_prefix` is normalized directly because it is a
`startswith` probe rather than a lookup (and keeps its trailing slash, so
`src/locales` cannot match `src/locales-old`).

This half is **not pnpm-specific** — it reproduces on `main` for an npm/yarn repo
that lists `"."` in `workspaces`: the member is mapped, then never resolves. Both
regression tests are in `TestRootPackageResolution`. Say the word if you would
rather have this as a separate PR; it is ~3 lines plus tests, and the pnpm work is
what made it reachable.

## Test Plan

- [x] Tests pass (`uv run pytest tests/unit/ tests/providers/`)
- [x] Lint passes (`uv run ruff check .` — All checks passed)
- [ ] Web build passes *(n/a — no frontend changes)*

`test_typescript_resolver.py`: 54 passed. New cases cover the `.yaml`-only
spelling, a non-glob entry (`- scripts`), `!` negation, **slashless negation
anchored at the root**, **negation not swallowing nested members** (the two that
fail under `pathspec`), pnpm precedence over `workspaces`, a `catalog`-only
manifest as root-only, malformed YAML being non-fatal, an end-to-end
`resolve_via_workspaces()` case through a member's own `exports` map, and four
root-resolution cases (npm `exports`, pnpm `main`, a root subpath import, and a
root `exports` wildcard through `build_ts_workspace_index`).

Verified read-only against a real pnpm monorepo (`pnpm@11`, TS, 18 packages across
`apps/*` + `packages/*` + a plain `scripts` entry, plus a named private root):

| | members discovered | `@scope/domain` | `@scope/db` |
|---|---|---|---|
| before | 0 | `None` → `external:` | `None` → `external:` |
| after | 19 | `packages/domain/src/index.ts` | `packages/db/src/index.ts` |

Two notes on what I deliberately did **not** touch:

- I did not run `ruff format` on these files. Both already fail
  `ruff format --check` on `main`, so formatting them would bury this change in an
  unrelated diff. CI gates `ruff check .`, which passes.
- Four unit tests fail identically on unmodified `main` in my checkout
  (`test_agent_target_baseline`, `test_never_flag_regex`, `test_repo_totals_churn`,
  `test_transcript_episodes`) — verified by stashing the diff and re-running. One
  needs deeper git history than my clone had; I have not diagnosed the rest.
  Unrelated to this change.

## Review round (2026-08-13)

- **`.mcp.json` dropped.** It was committed by accident in the third commit and
  hardcoded a path from my machine. Removed from the branch history, not just
  reverted — the file no longer appears in any commit's tree.
- **`Co-Authored-By` trailers stripped** from all commits. Branch force-pushed.
- **Fixed the `_normalize_repo_rel("./")` bug** you found — confirmed exactly as
  you described it. `posixpath.normpath("./")` is `"."`, the trailing-slash branch
  re-appended the slash, and no repo-relative path starts with `"./"`, so a root
  package's wildcard target sitting AT the repo root (`{"./*": "./*.ts"}`) matched
  nothing and its files fell out of `exports_entry_paths`. The degenerate case now
  collapses to `""`. Three new tests: the root-level wildcard target through
  `build_ts_workspace_index`, and a unit case pinning `"./" -> ""` alongside the
  non-degenerate shapes that must not change.
- **Took the optional glob guard too.** Both `_expand_member_dirs` loops now run
  under `contextlib.suppress(NotImplementedError, ValueError)`. Worth noting the
  empty-pattern half is reachable on `main` today and not only via pnpm: pnpm's
  reader drops blank entries before they reach the globber, but the npm
  `workspaces` field passes every string straight through, so `"workspaces": [""]`
  raises `ValueError` out of `resolve_via_workspaces` and costs the entire map.
  Covered by a test in `TestWorkspaceMap` plus two absolute-pattern cases (include
  and exclude loop) in `TestPnpmWorkspaceMap`.

All five new tests fail on the pre-fix source and pass after — verified by stashing
only the source change and re-running. `tests/unit/ingestion` + `tests/unit/analysis`:
2391 passed, 1 failed (`test_repo_totals_churn`, one of the four pre-existing
failures noted above — my clone lacks the git history it needs). `ruff check .`
clean.

## Checklist

- [x] I have updated documentation if needed — the module docstring now describes
      both manifests, why the pnpm one wins, and why negation uses `Path.glob`

## One thing I left alone, deliberately

`_read_pnpm_workspace_patterns` returns `None` for an absent **or unparseable**
manifest, so a malformed `pnpm-workspace.yaml` silently yields no pnpm members. A
diagnostic would be better, but every `package.json` read in this file already does
`except Exception: continue`, and introducing a lone error channel here would be
inconsistent with the module. Happy to add one if you want it — it would probably
belong at the ingestion layer for all manifests at once, not just this one.

### DIFF
--- a/packages/core/src/repowise/core/ingestion/resolvers/pnpm_workspace.py
+++ b/packages/core/src/repowise/core/ingestion/resolvers/pnpm_workspace.py
@@ -1,0 +1,71 @@
+"""``pnpm-workspace.yaml`` member-glob reading.
+
+Separate from ``ts_workspace`` because it is a manifest reader rather than
+part of import resolution — the same split the other package managers get
+(``rust_workspace``, ``go_workspace``, ``php_composer``, ``scala_build``).
+``ts_workspace`` consumes the globs this returns.
+"""
+
+from __future__ import annotations
+
+from pathlib import Path
+
+# pnpm names this file exactly. ``.yml`` is a long-standing request
+# (pnpm/pnpm#1380), not a supported spelling — honouring it would map
+# members from a manifest pnpm itself ignores.
+PNPM_WORKSPACE_FILENAME = "pnpm-workspace.yaml"
+
+
+def read_pnpm_workspace_patterns(repo_path: Path) -> tuple[list[str], list[str]] | None:
+    """Return ``(includes, excludes)`` member globs from ``pnpm-workspace.yaml``.
+
+    ``None`` means "not a pnpm workspace" — the manifest is absent or could
+    not be parsed. That is distinct from ``([], [])``, which means "a pnpm
+    workspace that declares no member globs", i.e. root-only: pnpm's
+    ``packages`` setting is optional and "if the field is omitted, only the
+    root package is included in the workspace". Since pnpm 10 the same file
+    also holds ``catalog``, ``onlyBuiltDependencies`` and other settings, so
+    a manifest with no ``packages`` key is normal rather than an error.
+
+    pnpm ignores ``package.json``'s ``workspaces`` field entirely, so a pnpm
+    monorepo declares its members only here::
+
+        packages:
+          - "apps/*"
+          - "packages/*"
+          - "!**/__fixtures__/**"
+
+    A leading ``!`` negates the pattern. Negated entries come back separately
+    because the caller applies them by expanding them the same way it expands
+    the includes and subtracting the result — see
+    :func:`~repowise.core.ingestion.resolvers.ts_workspace.build_workspace_info`.
+    """
+    import yaml
+
+    manifest = repo_path / PNPM_WORKSPACE_FILENAME
+    if not manifest.is_file():
+        return None
+    try:
+        data = yaml.safe_load(manifest.read_text(encoding="utf-8", errors="ignore"))
+    except Exception:
+        return None
+    if not isinstance(data, dict):
+        return None
+    entries = data.get("packages")
+    if not isinstance(entries, list):
+        # A settings-only manifest (catalog, onlyBuiltDependencies, ...) is
+        # still a pnpm workspace — root-only.
+        return [], []
+    includes: list[str] = []
+    excludes: list[str] = []
+    for entry in entries:
+        if not isinstance(entry, str):
+            continue
+        pattern = entry.strip()
+        if pattern.startswith("!"):
+            negated = pattern[1:].strip()
+            if negated:
+                excludes.append(negated)
+        elif pattern:
+            includes.append(pattern)
+    return includes, excludes
--- a/packages/core/src/repowise/core/ingestion/resolvers/ts_workspace.py
+++ b/packages/core/src/repowise/core/ingestion/resolvers/ts_workspace.py
@@ -246,6 +246,75 @@ def build_workspace_map(repo_path: Path | None) -> dict[str, str]:
     return {name: info["dir"] for name, info in build_workspace_info(repo_path).items()}
 
 
+@dataclass(frozen=True)
+class _WorkspaceDeclaration:
+    """Which globs govern a repo's workspace, and whether the root is a member."""
+
+    includes: tuple[str, ...]
+    excludes: tuple[str, ...]
+    include_root: bool
+
+    @property
+    def is_empty(self) -> bool:
+        return not self.includes and not self.include_root
+
+
+def _read_workspace_declaration(repo_path: Path) -> _WorkspaceDeclaration:
+    """Resolve the governing manifest into member globs.
+
+    pnpm keeps its member globs in ``pnpm-workspace.yaml`` and does not read
+    ``package.json``'s ``workspaces`` field at all, so the two manifests are
+    not interchangeable and must not be merged: unioning them would register a
+    member pnpm never installs, turning a registry dependency into a fake
+    intra-repo edge. When the pnpm manifest is present it is authoritative.
+    """
+    pnpm = read_pnpm_workspace_patterns(repo_path)
+    if pnpm is not None:
+        includes, excludes = pnpm
+        # "The root package is always included, even when custom location
+        # wildcards are used" — pnpm's ``packages`` setting. So a root-only
+        # workspace (no ``packages`` key) still has exactly one member.
+        return _WorkspaceDeclaration(tuple(includes), tuple(excludes), include_root=True)
+
+    root_pkg = repo_path / "package.json"
+    if not root_pkg.is_file():
+        return _WorkspaceDeclaration((), (), include_root=False)
+    try:
+        data = json.loads(root_pkg.read_text(encoding="utf-8", errors="ignore"))
+    except Exception:
+        return _WorkspaceDeclaration((), (), include_root=False)
+    if not isinstance(data, dict):
+        return _WorkspaceDeclaration((), (), include_root=False)
+    return _WorkspaceDeclaration(
+        tuple(_read_workspaces_field(data)), (), include_root=False
+    )
+
+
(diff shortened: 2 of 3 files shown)


swati510 (approved):
LGTM

PR #1475, opened 2026-08-13
fix(embedding): honour REPOWISE_EMBEDDING_TIMEOUT in every embedder

### LINKED ISSUE #1474: REPOWISE_EMBEDDING_TIMEOUT is honoured only by the Ollama embedder
# `REPOWISE_EMBEDDING_TIMEOUT` is honoured only by the Ollama embedder

## Summary

`docs/reference/CONFIG.md` says `REPOWISE_EMBEDDING_MODEL` "overrides the model for whichever embedder
is active. `REPOWISE_EMBEDDING_DIMS` and `REPOWISE_EMBEDDING_TIMEOUT` apply the same way".

For the timeout that is true only of Ollama. `packages/core/src/repowise/core/providers/embedding/ollama.py` reads `OLLAMA_EMBEDDING_TIMEOUT` then `REPOWISE_EMBEDDING_TIMEOUT` and defaults to 30s;
the sibling `openai.py`, `gemini.py` and `openrouter.py` each hardcode `_DEFAULT_TIMEOUT = 10.0` and read no
environment variable at all. So on three of four embedders the documented setting silently does nothing.

It matters most on the OpenAI-compatible path, because that is how repowise reaches a local server (MLX,
vLLM, llama.cpp — the vLLM/SGLang recipe in `CONFIG.md`). There a single embedding request is seconds of GPU work
rather than a hosted round-trip, and 10s is not enough.

## Steps to reproduce

1. Serve an embedding model on an OpenAI-compatible local endpoint (below: `mlx-openai-server` with `qwen3-embedding-4b`, Apple silicon).
2. `export OPENAI_BASE_URL=http://127.0.0.1:11435/v1 REPOWISE_EMBEDDING_TIMEOUT=180`
3. `repowise init --prose` (or `repowise reindex`) on a repo whose pages exceed ~2,400 characters.

**Expected:** the 180s I set applies.
**Actual:** 10s applies, and every batch fails:

```
rag.embed_batch_failed  count=103
error=embed_batch: 103/103 items failed to embed
hint=semantic search will miss these pages; run `repowise reindex` to repair
```

`embed_batch` receives a whole generation level and sub-chunks it into `EMBED_BATCH_MAX_ITEMS`-sized
requests (`packages/core/src/repowise/core/persistence/vector_store/_base.py`, `EMBED_BATCH_MAX_ITEMS = 16`), so the 103 is the level, not one request; each of
the 7 chunks expired.

The run **exits 0**, the pages are written, and only the vectors are missing — so the failure is one
warning line in an otherwise successful run, and `repowise reindex`, which that line recommends, fails
identically because it embeds through the same path.

## Why 10s is not enough

One chunk is 16 items of up to `EMBED_TEXT_MAX_CHARS` (30,000) each. The constant's comment sizes it
against *OpenAI's* 300k-token request cap — right for a hosted embedder, where that returns in about a
second.

Timing the same shapes the store actually sends, against an **idle** endpoint (script below, median of 3,
spread ≤0.5s):

| chars per item | median | under 10s? |
| ---: | ---: | --- |
| 1,000 | 4.4s | yes |
| 1,500 | 6.3s | yes |
| 2,000 | 8.3s | yes |
| 2,500 | 10.4s | **no** |
| 3,000 | 12.6s | **no** |
| 4,000 | 15.0s | **no** |
| 30,000 (the cap) | 14.9s | **no** |

The limit is crossed between 2,000 and 2,500 characters per item — one item is one page. Cost then
plateaus near 15s because the model truncates at its context window, so the worst case is bounded, but
every page over ~2,400 characters fails. No contention is involved: this is an idle endpoint. Inputs are
distinct per item (an identical-input batch measured 14.0s against 14.7s distinct at 3,000, so batching
shortcuts are not flattering the numbers).

With the timeout honoured, `REPOWISE_EMBEDDING_TIMEOUT=180 repowise reindex` on the same repository
embedded every page across all levels, with no `embed_batch_failed` line at any point:

```
Done! Indexed 964 items -> /path/to/.repowise/lancedb
```

The 964 is the whole run; the 103 above was one generation level of the same corpus.

<details><summary>timing script</summary>

```python
import json, os, statistics, time, urllib.request
BASE, MODEL, ITEMS = os.environ["OPENAI_BASE_URL"], "qwen3-embedding-4b", 16
F = "The overlay renders japanese subtitles with ruby annotations and pitch accent data. "

def item(i, chars):  # distinct per item, so no batching shortcut can flatter it
    return (f"item {i} unique-{i * 7919}. " + (F * (chars // len(F) + 2))[i * 13:])[:chars]

def timed(chars, runs=3):
    out = []
    for _ in range(runs):
        body = json.dumps({"model": MODEL, "input": [item(i, chars) for i in range(ITEMS)]}).encode()
        req = urllib.request.Request(f"{BASE}/embeddings", data=body,
                                     headers={"Content-Type": "application/json"})
        t = time.monotonic()
        with urllib.request.urlopen(req, timeout=300) as r:
            json.loads(r.read())
        out.append(time.monotonic() - t)
    return statistics.median(out)

for c in (1000, 1500, 2000, 2500, 3000, 4000, 30000):
    print(c, round(timed(c), 1))
```
</details>

## What I'd propose

Resolve the timeout through one shared helper used by all four embedders: explicit arg >
`<PROVIDER>_EMBEDDING_TIMEOUT` > `REPOWISE_EMBEDDING_TIMEOUT` > the provider's default.

**Defaults unchanged**, and deliberately still uneven: Ollama keeps 30s, the other three keep 10s.
Raising the shared default would also change the *query* path — the same embedder serves
`VectorStore.search`, and the OpenAI SDK retries twice by default (`DEFAULT_MAX_RETRIES = 2`, openai
2.24.0), so a 30s default turns a stalled hosted endpoint into a ~90s MCP tool call, past the 60s cap most
MCP clients apply. Unifying the four defaults is a separate call I'd rather leave to you.

**One constraint on malformed values, which I ran into rather than chose:**
`packages/cli/src/repowise/cli/providers/embedders.py:build_embedder` wraps construction in
`except Exception` and returns `KeylessEmbedder` — an 8-dimensional stand-in whose vectors carry no
signal, so retrieval is dead for that run. Raising on a typo like `30s` therefore breaks a setup that
worked before the variable was read at all. What should happen instead is the open question in #826, and
I'd follow whatever you settle there.

The alternative fix would be shrinking `EMBED_BATCH_MAX_ITEMS` or making chunk size adaptive for slow
endpoints. That changes behaviour for everyone to fix a case configuration can already address, so I
didn't pursue it — but it is your call.

Branch is ready with unit tests that fail on `main` and pass with the change, plus the `CONFIG.md` rows.

## Related

- #1253 — the `REPOWISE_EMBEDDING_DIMS` half of the same documented sentence, same embedder. Closed.
- #826 — what bad embedder config should do; this issue defers to it.

## Environment

- repowise `main` @ `e292bc7b` (also reproduced on 0.37.0)
- Python 3.13, macOS (Apple silicon)
- `mlx-openai-server` serving `qwen3-embedding-4b` (2560 dims) over the OpenAI-compatible API

serjflint:
Fixes #1474.

`REPOWISE_EMBEDDING_TIMEOUT` is documented as applying to whichever embedder is active, but only the
Ollama embedder read it; `openai`, `gemini` and `openrouter` hardcoded 10s. Against a local
OpenAI-compatible endpoint one request embeds a whole batch — measured ~10s at 2,500 characters per item
and ~15s beyond 4,000 — so batches expired and were reported once as `N/N items failed to embed` while the
run still exited 0. The repair that message recommends, `repowise reindex`, embeds through the same path
and failed identically. Numbers, method and repro are in the issue.

**Defaults are unchanged** (Ollama 30s, the rest 10s). The same embedder serves the query path and the SDK
retries twice, so raising the shared default would turn a stalled hosted endpoint into a ~90s MCP tool
call. Unifying the four defaults is a separate call.

**A malformed value falls back instead of raising.** `build_embedder` wraps construction in
`except Exception` and returns `KeylessEmbedder`, so raising on a typo like `30s` would silently leave the
run with 8-dimensional vectors and no retrieval — breaking a setup that worked before the variable was read
at all. It is reported on stderr rather than through `log.warning`, because `configure_cli_logging` filters
structlog to ERROR unless `-v`, which would have discarded the message on every command that embeds. What
*should* happen for bad embedder config is the open question in #826; this follows whatever is decided
there.

### Commits

Independent — either can be dropped without touching the other.

1. **`concept-tree`: warn when outline naming is cut at the token ceiling.** A truncated reply parses to
   `{}` and decodes to the same deterministic fallback an empty one produces, so the two are
   indistinguishable and titles silently become path-derived. Reasoning tokens count against the same
   ceiling, so a thinking model can exhaust it having emitted no answer.
2. **`embedding`: honour the variable in every embedder**, via one shared resolver
   (explicit arg > `<PROVIDER>_EMBEDDING_TIMEOUT` > `REPOWISE_EMBEDDING_TIMEOUT` > provider default).
3. **`embedding`: route Ollama through it too.** Ollama and `_embedder_kwargs` each parsed the variable
   with a bare `float()`, so the keyless downgrade above was still reachable on the embedder most likely to
   need a raised timeout, and `inf`/negatives reached `httpx` unchecked.

### Tests

`tests/unit/test_persistence/test_embedding_timeout_resolution.py` holds the cross-embedder contract:
precedence, that a malformed provider-specific value falls through to a valid shared one rather than
consuming its precedence, that no embedder is destroyed by a typo, and that the fallback reaches stderr.
12 of the new tests fail on `main` — including `could not convert string to float: '30s'` and `inf`
reaching the client on the Ollama path.

`uv run pytest tests/unit/test_persistence tests/unit/generation tests/unit/test_providers` — 1667 passed.
`CONFIG.md` gains the provider-prefixed rows and the defaults it never stated.

I have not run `ruff format` repo-wide: it reformats 522 files on current `main`, which would bury the
diff.

### DIFF
--- a/docs/reference/CONFIG.md
+++ b/docs/reference/CONFIG.md
@@ -632,10 +632,13 @@ The `.repowise/.env` file is gitignored automatically.
 | `REPOWISE_EMBEDDER` | Embedder: `gemini`, `openai`, `ollama`, `openrouter`, or `mock` |
 | `REPOWISE_EMBEDDING_MODEL` | Embedding model, applies to any embedder |
 | `REPOWISE_EMBEDDING_DIMS` | Embedding output dimensions (optional; inferred from the model otherwise) |
-| `REPOWISE_EMBEDDING_TIMEOUT` | Embed request timeout in seconds |
+| `REPOWISE_EMBEDDING_TIMEOUT` | Embed request timeout in seconds (default: `30` for `ollama`, `10` elsewhere). Raise it for a local endpoint — one request embeds a whole batch, and an expired batch is reported only as `N/N items failed to embed`. An unparseable value warns and keeps the default |
+| `OPENAI_EMBEDDING_TIMEOUT` | As above, `openai` only; takes precedence over the shared variable |
+| `GEMINI_EMBEDDING_TIMEOUT` | As above, `gemini` only |
+| `OPENROUTER_EMBEDDING_TIMEOUT` | As above, `openrouter` only |
 | `OLLAMA_EMBEDDING_MODEL` | Ollama embedding model (also selects the `ollama` embedder) |
 | `OLLAMA_EMBEDDING_DIMS` | Ollama embedding output dimensions (optional; inferred from the model otherwise) |
-| `OLLAMA_EMBEDDING_TIMEOUT` | Ollama embed request timeout in seconds (default: `30`); raise it for long pages on slow local models |
+| `OLLAMA_EMBEDDING_TIMEOUT` | As above, `ollama` only; raise it for long pages on slow local models |
 
 ### Server and database
 
--- a/docs/reference/CONFIG.md
+++ b/docs/reference/CONFIG.md
@@ -571,8 +571,9 @@ store would be rebuilt from scratch, discarding what the reindex just built.
 
 `REPOWISE_EMBEDDING_MODEL` overrides the model for whichever embedder is
 active. `REPOWISE_EMBEDDING_DIMS` and `REPOWISE_EMBEDDING_TIMEOUT` apply the
-same way; the `OLLAMA_EMBEDDING_*` variants below are Ollama-specific
-equivalents.
+same way; the provider-prefixed variants below (`OPENAI_*`, `GEMINI_*`,
+`OLLAMA_*`, `OPENROUTER_*`) narrow a setting to one embedder and take
+precedence over the shared name.
 
 ---
 
--- a/packages/cli/src/repowise/cli/providers/embedders.py
+++ b/packages/cli/src/repowise/cli/providers/embedders.py
@@ -15,15 +15,10 @@ def _embedder_kwargs(embedder_name: str) -> dict[str, Any]:
         dimensions = os.environ.get("OLLAMA_EMBEDDING_DIMS") or os.environ.get(
             "REPOWISE_EMBEDDING_DIMS"
         )
-        timeout = os.environ.get("OLLAMA_EMBEDDING_TIMEOUT") or os.environ.get(
-            "REPOWISE_EMBEDDING_TIMEOUT"
-        )
         if base_url:
             kwargs["base_url"] = base_url
         if dimensions:
             kwargs["dimensions"] = int(dimensions)
-        if timeout:
-            kwargs["timeout"] = float(timeout)
     elif embedder_name == "gemini":
         dimensions = os.environ.get("REPOWISE_EMBEDDING_DIMS")
         if dimensions:
--- a/packages/core/src/repowise/core/generation/concept_tree/planner.py
+++ b/packages/core/src/repowise/core/generation/concept_tree/planner.py
@@ -396,13 +396,10 @@ async def name_groups(
                 temperature=0.2,
                 **_reasoning_kwargs(reasoning),
             )
-        # A reply cut at the ceiling parses to {} and decodes to the full
-        # deterministic fallback: no exception, no warning, and the only trace
-        # is named_by_model=0 on the planned line. The provider already reports
-        # it — validate_generated_response treats the same signal as fatal for a
-        # page — and the outline can proceed without the model, but not without
-        # saying why. Reasoning tokens are charged to this ceiling too, so a
-        # thinking model can hit it while emitting no answer at all.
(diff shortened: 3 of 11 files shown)


RaghavChamadiya (approved):
LGTM

PR #1477, opened 2026-08-13
fix(chat): align get_context include schema with MCP/CLI

Ayush7614:
## Summary
- Chat `get_context` schema listed `docs`/`freshness`/`source` and omitted `skeleton`/`health`, so the model could request a no-op `source` block and never ask for blocks MCP already serves.
- Align the include enum with CLI `_INCLUDE_BLOCKS`, advertise `compact`, and pin the match with a registry drift test.

## Test plan
- [x] `pytest tests/unit/server/test_chat_tool_registry.py`
- [x] In web chat, ask for a file skeleton / health block and confirm the model can pass `include=["skeleton"]` / `include=["health"]`

### DIFF
--- a/packages/server/src/repowise/server/chat_tools.py
+++ b/packages/server/src/repowise/server/chat_tools.py
@@ -47,7 +47,11 @@ _TOOL_SCHEMAS: list[dict[str, Any]] = [
     },
     {
         "name": "get_context",
-        "description": "Get documentation, ownership, freshness, and decisions for one or more files, modules, or symbols.",
+        "description": (
+            "Triage card for files/modules/symbols: docs, freshness, hotspot bit, "
+            "signatures. Opt into full_doc, ownership, last_change, callers/callees, "
+            "metrics, community, decisions, health, or skeleton (body-elided source)."
+        ),
         "parameters": {
             "type": "object",
             "properties": {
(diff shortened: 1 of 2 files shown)


RaghavChamadiya (approved):
Thanks Ayush, this is a good catch and the kind of drift that is easy to miss because nothing errors when it happens.

I checked the premise against the code and it holds all the way through. `docs` and `freshness` were never opt-in blocks, `get_context` unions them in before it looks at `include` at all, so offering them as choices was inviting the model to switch on something already on. `source` does not exist anywhere in `tool_context/`, so a request for it was a wasted tool call with nothing to show for it. `skeleton` and `health` are both real and both were missing, which is the part that actually cost us, since those are two of the most useful blocks in chat.

Pinning the test against `_INCLUDE_BLOCKS` rather than a hardcoded list is the right call. Whoever adds the next block gets a failing test instead of silent chat drift. Asserting the three dead names stay out is a nice touch too.

Merging this.

One follow-up if you feel like it, no obligation. The same drift exists one layer over: the MCP tool's own docstring at `tool_context/context.py:73` lists the include blocks and leaves out `health`, and the module docstring rundown at the top of that file does the same. So MCP agents are under-informed in exactly the way chat was. It is a two line docs fix, and your test could grow a second assertion covering MCP against `_INCLUDE_BLOCKS` at the same time. Happy to take it as its own small PR.

PR #1478, opened 2026-08-13
fix(ui): render get_why MCP wire format in chat artifacts

Ayush7614:
## Summary
- Chat `DecisionsRenderer` expected `results` / `total_decisions` / `by_source`, but MCP `get_why` returns `decisions`, `counts`, `stale_decisions`, and a `path` mode — so View Artifact was empty or wrong for all three modes.
- Align types + renderer with the MCP wire format (keep legacy fixture shapes), and cite files from health-mode `stale_decisions`.

## Test plan
- [x] `npm test -- --run __tests__/chat/artifacts.test.tsx __tests__/chat/source-citations.test.tsx` in `packages/ui`
- [x] In web chat, call get_why with no query (health), a natural-language query (search), and a file path (path); open View Artifact for each

### DIFF
--- a/packages/types/src/chat.ts
+++ b/packages/types/src/chat.ts
@@ -172,19 +172,59 @@ export interface GraphPathArtifact {
   data: GraphPathArtifactData;
 }
 
-/** `get_why` — decision register search results / health dashboard. */
+/** `get_why` — decision register search / path lookup / health dashboard. */
 export interface DecisionsArtifactData {
-  mode: "health" | "search";
+  mode: "health" | "search" | "path";
   query?: string;
+  path?: string;
+  summary?: string;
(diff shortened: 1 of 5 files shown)


RaghavChamadiya (approved):
Thanks Ayush, good catch and a clean fix.

The renderer really was reading fields the MCP layer does not emit. `tool_why.py` builds `counts`, `stale_decisions`, `proposed_awaiting_review` and `ungoverned_hotspots` for health mode, and `decisions` plus `alignment` / `origin_story` for path mode, so View Artifact had nothing to show for any of the three. Keeping `results` and `total_decisions` around so stored conversations still render is the right call.

Verified on top of current main: merges clean, the `packages/ui` chat tests pass, and `type-check` is green across the workspace.

Issue #361, opened 2026-06-04
Calibrate hotspot and ownership risk for small teams


joptimus:
Hi @RaghavChamadiya , I wanted to ask for feedback on how Repowise is intended to behave for small teams and relatively quiet repos.

I’m using it on a small game-server repo, and a pattern I keep running into is that normal maintenance work can make a file look riskier than it feels in context. In particular:

- Files I touch can become hotspots quickly.
- With only a couple of active contributors, ownership views can make it look like each person owns too much code, even though that is the expected shape of a 1-3 person project.

From reading the code, hotspot classification looks repo-relative:

```py
# packages/core/src/repowise/core/ingestion/git_indexer/enrich.py
is_hotspot = churn_percentile >= 0.75 and commit_count_90d > 0
```

That makes sense as a relative signal, but in a quiet repo it means a file with very little absolute recent activity can still land in the top quartile. Once that happens, the hotspot flag can feed other findings or severity decisions such as `untested_hotspot`, `churn_risk`, `change_entropy`, prior-defect escalation, and the hotspot-health KPI.

The ownership side feels similar. For a larger team, concentrated ownership or module silos are useful warning signs. For a tiny team, though, concentrated ownership is often just the normal operating model. A warning that “one person owns many files” can be accurate but not necessarily actionable or risky unless it is combined with other evidence like inactivity, high churn, missing tests, complex code, or a bus-factor concern.

Would you consider calibrating these signals with small-team / low-activity repos in mind?

A few possible directions:

1. Add absolute floors to hotspot classification, for example:

```py
is_hotspot = (
    churn_percentile >= 0.80
    and commit_count_90d >= 3
    and temporal_hotspot_score >= 0.5
)
```

2. Split activity labels so the UI and scoring can distinguish:

- `recently_touched`: any recent activity
- `active_file`: enough absolute activity
- `hotspot`: active plus high relative/temporal churn
- `sustained_hotspot`: hotspot across multiple snapshots

3. Make ownership warnings team-size aware:

- In repos with <=3 active contributors, treat high ownership concentration as informational by default.
- Escalate ownership concentration only when it combines with stronger signals: inactive owner, high churn, high complexity, missing tests, or low bus factor.
- Distinguish “normal small-team ownership” from “large-team silo risk.”

I’m not suggesting removing these signals. They are useful. The issue is that relative scoring can make ordinary small-team development feel over-penalized when most files are inactive and there are naturally only a few owners.

Questions:

[...]

PR #1480, opened 2026-08-13
fix(ui): render get_risk MCP dict targets in chat artifacts

Ayush7614:
## Summary
- Chat `RiskReportRenderer` assumed `targets` was an array with `file_path` / `churn_percentile`, but MCP `get_risk` returns a path-keyed object with `hotspot_score` — View Artifact threw (`targets.map is not a function`) or showed blank hotspot rows.
- Normalize array/dict shapes, map `file_path`/`hotspot_score`, and render the `get_change_risk` live-git score card (same `risk_report` artifact type).

## Test plan
- [x] `npm test -- --run __tests__/chat/artifacts.test.tsx` in `packages/ui`
- [x] In web chat, run get_risk on a file and get_change_risk on HEAD; open View Artifact for both

### DIFF
--- a/packages/types/src/chat.ts
+++ b/packages/types/src/chat.ts
@@ -127,7 +127,10 @@ export interface ContextArtifact {
 
 /** `get_risk` / `get_change_risk` — modification or commit-range risk report. */
 export interface RiskTargetRow {
-  file_path: string;
+  /** Path key; optional on MCP dict values that use `target` instead. */
+  file_path?: string;
+  /** MCP per-target id when `targets` is a path-keyed object. */
+  target?: string;
   /** MCP `get_risk` hotspot score (0–100 churn percentile). */
   hotspot_score?: number;
   /** Legacy chat fixture field. */
@@ -144,9 +147,7 @@ export interface RiskReportArtifactData {
    * MCP `get_risk` returns a path-keyed object; legacy fixtures used an array.
    * Renderers normalize either shape.
    */
-  targets?:
-    | RiskTargetRow[]
-    | Record<string, RiskTargetRow & { target?: string }>;
+  targets?: RiskTargetRow[] | Record<string, RiskTargetRow>;
   global_hotspots?: Array<{
     /** MCP wire. */
     file_path?: string;
--- a/packages/types/src/chat.ts
+++ b/packages/types/src/chat.ts
@@ -126,14 +126,44 @@ export interface ContextArtifact {
 }
 
 /** `get_risk` / `get_change_risk` — modification or commit-range risk report. */
+export interface RiskTargetRow {
+  file_path: string;
+  /** MCP `get_risk` hotspot score (0–100 churn percentile). */
+  hotspot_score?: number;
+  /** Legacy chat fixture field. */
+  churn_percentile?: number;
+  is_hotspot?: boolean;
+  risk_type?: string;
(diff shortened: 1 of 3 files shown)


swati510 (commented):
Thanks for this one Ayush. The diagnosis is spot on: `get_risk.py:202` builds `targets` as `{r["target"]: r}`, so the old renderer was calling `.map` on an object and throwing. Normalizing both shapes is the right fix, and the `get_change_risk` score card is a good addition.

One thing to change before I merge. `hotspot_score` is a 0 to 1 fraction, not 0 to 100. It comes straight from `GitMetadata.churn_percentile`, which `git_indexer/enrich.py` sets as `rank / total`, and `assessment.py` renders with `:.0%`. Two consequences on real data:

* `Math.round(t.score)` plus `th pct` shows a 0.91 hotspot as "1th pct". `global_hotspots` has the same problem, since it also reads `h.churn_percentile`.
* The `score >= 80` fallback for `is_hotspot` never fires, and the MCP target row has no `is_hotspot` field at all, so the badge would never appear.

Small fix: treat the wire value as a fraction and scale by 100 for display, and use `>= 0.75` for the hotspot cut, which is the same threshold `enrich.py` uses. Worth updating the fixtures too, since they currently pass `hotspot_score: 91` and an explicit `is_hotspot`, which is why this got past the tests.

Everything else looks good, and I checked that it still merges cleanly now that #1478 is in. Ping me once it is updated and I will get it merged.

RaghavChamadiya (approved):
Thanks @Ayush7614, and thanks @swati510 for the review that caught the scale. This is right now, and it is right for the reason that matters rather than by coincidence.

The thing I want to name is the fixture change. The original tests passed while the code was wrong, because the fixtures asserted `churn_percentile: 99` and `is_hotspot: true`, which is not what the MCP wire carries. Moving them to `0.99` and dropping the hardcoded `is_hotspot` means the threshold path is actually exercised instead of being stepped over. That is the difference between a test that passes and a test that would have caught this.

Both constants check out against the source rather than against a fixture. `git_indexer/enrich.py:218`:

```python
        if churn_pct >= 0.75 and meets_hotspot_floors(meta):
            meta["is_hotspot"] = True
```

so `HOTSPOT_SCORE_THRESHOLD = 0.75` is the same cut, and citing `enrich.py` in the comment is what will keep the two from drifting. Documenting `hotspot_score` as `rank / total` on the type is worth as much as the fix; that scale has already caused one wrong render and now the next person reads it on the field.

Full matrix green on 3d545ae, and `npm run test --workspace @repowise-dev/ui` is in the CI job list, so your new test genuinely ran.

One thing worth knowing, deliberately not a blocker. The backend gates `is_hotspot` on the score **and** absolute floors:

```python
def meets_hotspot_floors(meta: dict) -> bool:
    """Absolute activity floors for hotspot classification (issue #361).

    The churn percentile is repo-relative, so on a quiet repo the top
    quartile degenerates to "any file touched recently".
    """
```

The UI can only see the score, so on a small or quiet repo it will badge files the backend would not, which is the exact degeneracy #361 was filed about. That is not a reason to hold this: today the badge is either absent or wrong, and score-only is strictly better than that. The real fix is for the MCP row to carry `is_hotspot` instead of making every renderer re-derive it, which is a server-side change and not yours to make here. I will pick that up separately.

PR #1511, opened 2026-08-13
fix(cli): resolve the hooks dir git actually runs

### LINKED ISSUE #1512: [Bug] hook install writes to .git/hooks, which git ignores when core.hooksPath is set (and crashes in worktrees)
## Describe the Bug

`repowise hook install` resolves the hooks directory as `<root>/.git/hooks`. That assumption is wrong in two common layouts, and the more serious of the two fails **silently**.

**1. `core.hooksPath` is set — the hook is installed and never runs.**

When `core.hooksPath` is configured, git ignores `.git/hooks` entirely. husky and lefthook both set it, so this affects most JS/TS repos. `repowise hook install` writes to `.git/hooks/post-commit`, reports `installed`, and `repowise hook status` keeps reporting `✓ installed` — but git never executes it, so auto-sync silently never happens. There is no signal that anything is wrong.

**2. Linked worktrees — hard crash.**

In a linked worktree (and in a submodule) `.git` is a *file* holding a gitdir pointer, so `hooks_dir.mkdir()` raises. This one at least fails loudly:

```
File ".../repowise/cli/hooks.py", line 190, in install
    hooks_dir.mkdir(exist_ok=True)
NotADirectoryError: [Errno 20] Not a directory: '/path/to/worktree/.git/hooks'
```

Relatedly, a worktree's hooks live in the repository's *common* directory rather than in its own gitdir, so even without the crash the path would be wrong.

## Steps to Reproduce

Silent no-op (the important one):

```bash
git init repo && cd repo
git config core.hooksPath .husky/_     # what husky does
repowise init --no-prose -y
repowise hook install                  # says "installed"
repowise hook status                   # says "installed"
ls .git/hooks/post-commit              # exists, but git will never run it
# commit something -> no auto-sync, no error, no explanation
```

Crash:

```bash
git worktree add ../wt -b wt-branch
cd ../wt && repowise hook install      # NotADirectoryError
```

## Expected Behavior

The hook is installed into the directory git will actually execute, and `status` reports on that same directory. In a worktree, install should succeed rather than raise.

## Actual Behavior

- `core.hooksPath` set: hook written to a directory git ignores; reported as installed; auto-sync never runs.
- Worktree: `NotADirectoryError`, command aborts.

## Environment

- OS: Ubuntu 22.04 (WSL2)
- Python version: 3.12
- Repowise version: 0.42.0
- Installation method: `uv tool install`

Confirmed still present on `main` at the time of filing — `packages/cli/src/repowise/cli/hooks.py` lines 189, 242 and 271.

## Additional Context

`git rev-parse --git-path hooks` resolves all of these cases, since it is the same resolution git itself performs.

One wrinkle beyond that: husky points `core.hooksPath` at `.husky/_`, which husky regenerates on every install and gitignores wholesale (`.husky/_/.gitignore` is `*`). A hook written there is deleted by the next `npm install`. husky's own shim dispatches to `.husky/<hook-name>` one level up, which is the durable, committed location — so husky needs that one extra hop to stay installed.

PR opened with a fix and tests: #1511

haibeeb:
## Summary

- `install` / `uninstall` / `status` hardcoded `<root>/.git/hooks`. That path crashes on git worktrees and is silently ignored by git whenever `core.hooksPath` is set, so the hook reports installed and never runs.
- Resolve the directory with `git rev-parse --git-path hooks` — the same resolution git itself performs — instead of assuming the layout.
- Remap husky's generated `.husky/_` up to the committed `.husky/` user-hook dir, since `_` is gitignored and regenerated on every `npm install`.

### The three broken layouts

| Layout | Old behaviour |
|---|---|
| Linked worktree / submodule | `.git` is a *file*, so `hooks_dir.mkdir()` raised `NotADirectoryError` and aborted the command |
| `core.hooksPath` set (husky, lefthook) | git ignores `.git/hooks`; hook written there is **never executed**, while `status` reports `installed` |
| Worktree | Hooks live in the repository's *common* dir, not the worktree's own gitdir |

The `core.hooksPath` case is the one worth emphasising: it fails silently. On a repo using husky (very common in JS/TS projects) `repowise hook install` has been a no-op that reports success, so auto-sync never ran and there was no signal that anything was wrong.

Reproduction of the crash:

```console
$ cd my-worktree && repowise hook install
NotADirectoryError: [Errno 20] Not a directory: '.../worktree/.git/hooks'
```

### Why remap husky's `_` directory

husky sets `core.hooksPath=.husky/_`, but that directory is regenerated on every install and gitignored wholesale (`.husky/_/.gitignore` is `*`), so a hook written into it disappears at the next `npm install`. husky's shim dispatches to `.husky/<hook-name>` one level up, which is the durable location. The remap is guarded so a directory merely *named* `_` is not affected.

Behaviour is unchanged for a plain checkout, and there is a fallback to the previous path if no git binary is available.

## Related Issues

Fixes #1512

## Test Plan

New `tests/unit/cli/test_hooks_path_resolution.py` (9 tests) covers: plain repo, relative and absolute `core.hooksPath`, husky remap with and without generated helpers, a non-husky `_` directory, the worktree crash regression, install/status/uninstall agreeing on one location, and the no-git fallback.

- [x] Tests pass (`pytest`) — 9 new; `tests/unit/cli/` green at 2108 passed, 1 skipped

Note: `ruff format` also wants to reformat pre-existing code in `hooks.py` (e.g. `lines[end + 1 :]`). I left that alone to keep this diff focused — CI runs `ruff check`, which passes.

## Checklist

- [x] I have updated documentation if needed *(none needed — internal path resolution)*

### DIFF
--- a/packages/cli/src/repowise/cli/commands/hook_cmd.py
+++ b/packages/cli/src/repowise/cli/commands/hook_cmd.py
@@ -991,10 +991,10 @@ def hook_status(path: str | None, workspace: bool, no_workspace: bool) -> None:
         for entry in target.ws_config.repos:
             abs_path = (target.ws_root / entry.path).resolve()
             result = status(abs_path)
-            icon = "[green]✓[/green]" if result == "installed" else "[dim]✗[/dim]"
+            icon = "[green]✓[/green]" if result.startswith("installed") else "[dim]✗[/dim]"
             console.print(f"  {icon} {entry.alias}: {result}")
     else:
         assert target.repo_path is not None
         result = status(target.repo_path)
-        icon = "[green]✓[/green]" if result == "installed" else "[dim]✗[/dim]"
+        icon = "[green]✓[/green]" if result.startswith("installed") else "[dim]✗[/dim]"
         console.print(f"  {icon} post-commit: {result}")
--- a/packages/cli/src/repowise/cli/commands/init_cmd/_interactive.py
+++ b/packages/cli/src/repowise/cli/commands/init_cmd/_interactive.py
@@ -205,7 +205,7 @@ def _offer_hook_install_prompts(
     candidates: list[tuple[Path, str]] = []
     for i, rp in enumerate(repo_paths):
         label = aliases[i] if aliases else rp.name
-        if status(rp) != "installed":
+        if not status(rp).startswith("installed"):
             candidates.append((rp, label))
 
     if not candidates:
--- a/packages/cli/src/repowise/cli/editor_setup.py
+++ b/packages/cli/src/repowise/cli/editor_setup.py
@@ -92,8 +92,10 @@ def detect_editor_setup_outcome(
     """Read the ground-truth editor-setup state for the completion panel.
 
     Called after registration and the hook offers, so what it reads is final.
-    Every probe is a cheap local file read and is defensive: a failure degrades
-    to "not set up" rather than crashing ``init``.
+    Every probe is defensive: a failure degrades to "not set up" rather than
+    crashing ``init``. All but one are local file reads; the autosync probe
+    spawns ``git rev-parse`` once, because ``core.hooksPath`` and worktrees make
+    the hooks directory impossible to derive from the repo path alone.
     """
     disabled = is_editor_setup_disabled(no_editor_setup)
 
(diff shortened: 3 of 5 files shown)


RaghavChamadiya (approved):
Rebased onto main with the husky remap intact; merging once CI is green.

Issue #1513, opened 2026-08-13
[Feature] claude_cli provider: generate pages with a Claude Code subscription instead of an API key


haibeeb:
## Problem

Repowise can already run page generation off a coding-agent subscription rather than a metered API key — `codex_cli` does it for Codex, `opencode` for OpenCode. There is no equivalent for Claude Code, even though `anthropic` is a first-class provider.

So a user with a Claude subscription (Pro / Max / Team / Enterprise) has no keyless path. They see:

```
No provider configured. Use --provider, set REPOWISE_PROVIDER, or set
ANTHROPIC_API_KEY / ... Use REPOWISE_PROVIDER=codex_cli to use an
authenticated Codex CLI subscription, or REPOWISE_PROVIDER=opencode to use
opencode.
```

Every listed keyless option belongs to a different vendor. The `anthropic` provider raises `ProviderError` without `ANTHROPIC_API_KEY`, so the practical options are to buy API credit alongside a subscription already being paid for, run `--no-prose`, or use another vendor's CLI. In an enterprise setting the first is often a procurement request, which makes the prose layer effectively unreachable.

This is more visible now that #1508 makes a missing provider degrade `get_answer` to retrieval-only: the wiki still works, but `confidence` never rises above `low` and `degraded: "no-llm-provider"` is returned on every question.

## Proposed Solution

A `claude_cli` provider that shells out to `claude -p --output-format json`, in the same shape as `codex_cli` / `opencode`:

```bash
claude login                                   # once
repowise init --provider claude_cli --yes
```

- Keyless: authentication is the CLI's, out of band. Repowise never handles credentials.
- Priced at `$0.00`, like the other subscription-backed providers, with the CLI's own reported cost kept under `usage.reported_cost_usd` for auditing.
- Default `claude_cli/claude-haiku-4-5`, matching the `anthropic` provider's default.
- Concurrency bounded (subscription limits are per account and each call is a process).

Two design points that were not obvious while building it, and are the parts most worth review:

1. **`--bare` must not be used.** It looks like the correct isolation flag, but it documents that "Anthropic auth is strictly `ANTHROPIC_API_KEY` or `apiKeyHelper` (OAuth and keychain are never read)" — it silently defeats subscription auth, which is the entire point. Isolation instead comes from a neutral working directory plus `--strict-mcp-config`.
[...]

PR #1514, opened 2026-08-13
feat(providers): add claude_cli provider for Claude Code subscriptions

haibeeb:
## Summary

- Adds a keyless `claude_cli` provider that generates wiki prose through the authenticated local Claude Code CLI, so a Claude subscription works without `ANTHROPIC_API_KEY`.
- Keeps generation single-turn and tool-free without losing pages, surfaces Claude's structured error messages and HTTP status, and maps Repowise reasoning modes to Claude Code's `--effort` flag.
- Records CLI-reported usage for auditing while pricing subscription-backed generation at `$0.00` in both the init estimator and runtime cost ledger.

## Initial Bug Report

[Issue #1513](https://github.com/repowise-dev/repowise/issues/1513) reports that Repowise has subscription-backed providers for Codex and OpenCode but no equivalent for Claude Code. A user with Claude Pro, Max, Team, or Enterprise therefore has to buy separate Anthropic API credit, use another vendor, or disable prose generation.

Maintainer dogfooding then found two pre-merge blockers in the first implementation: prompts that invited investigation could consume the one-turn budget on a denied tool and silently fall back to structural placeholders, and JSON API errors written to stdout were replaced with the generic `claude -p exited with 1` message.

## Repro Steps

1. Authenticate the local CLI with `claude login`, remove `ANTHROPIC_API_KEY`, and run `repowise init --provider claude_cli --yes`.
2. Before this PR, provider resolution had no Claude subscription path; expected behavior is page generation through the existing CLI login.
3. With the initial PR implementation, send a page prompt that invites repository investigation under `--max-turns 1` and a disallowed-tool list. Claude can attempt a denied tool, return `error_max_turns`, and produce no prose; expected behavior is a one-turn text response.
4. Select a nonexistent but syntactically valid model. Claude writes a JSON error with `api_error_status: 404` to stdout and exits 1; the initial implementation discarded it and showed only the exit code. Expected behavior is the selected-model explanation and the 404 status.

## Root Cause Analysis

- `--disallowed-tools` denied execution but still exposed tool choices to the model. A tool attempt consumed the single-turn budget before prose was emitted.
- `ClaudeCliProvider.generate()` checked `returncode` before parsing stdout, while Claude Code reports API failures as JSON on stdout and often leaves stderr empty.
- Claude Code now exposes `--effort low|medium|high|xhigh|max`, but the provider still advertised only `auto` and ignored every explicit reasoning mode.
- The init estimator recognized `claude_cli/` as subscription-backed, but `generation/cost_tracker.py` did not; the runtime ledger stripped the prefix and applied metered Anthropic rates.
- A reusable `mkdtemp` directory had no lifecycle hook, and cancellation by the 180-second interactive caller did not terminate the 600-second subprocess.

## Fix Details

- Registers `claude_cli` through the provider registry, CLI setup flow, rate limiter, server catalog, web settings, model pricing, tests, and documentation.
- Uses `--tools ""`, an explicit tool-free system instruction, `--max-turns 1`, and `--strict-mcp-config`; user content stays on stdin.
- Parses structured stdout on non-zero exits, preserves the useful Claude message and HTTP status, and retains text/stderr fallback handling for non-JSON failures.
- Supports `auto`, `low`, `medium`, `high`, `xhigh`, and `max`; unsupported `off`, `none`, and `minimal` modes warn and retain the CLI default.
- Creates and removes a fresh scratch directory per call, kills the subprocess on timeout or cancellation, and continues to run outside the repository so project `CLAUDE.md` files cannot bias generated prose.
- Keeps `--bare` absent because that mode disables OAuth/keychain subscription authentication.
- Prices `claude_cli/*` at zero in both cost paths while retaining `usage.reported_cost_usd` from Claude Code for auditing.
- Merges current `main` and preserves the newer Eden AI provider wiring while updating provider-count drift guards to 13.

Anthropic still has no embeddings API, so semantic search needs a separate embedder such as Ollama, OpenAI, or Gemini.

## Manual Test Steps

1. Run `claude login`, then `repowise init --provider claude_cli --yes`; confirm the provider resolves without `ANTHROPIC_API_KEY`.
2. Generate with a prompt that explicitly asks to investigate using tools; confirm it returns prose in one turn instead of `error_max_turns`. Verified locally against Claude Code 2.1.232: the exact provider returned `READY` with `num_turns: 1`.
3. Generate with `reasoning=low`; confirm the CLI accepts `--effort low`. Verified locally: the exact provider returned `EFFORT` in one turn.
4. Use a nonexistent safe model slug; confirm the error names the selected model and exposes status 404. Verified locally against the real CLI.
5. Run `uv run pytest -q tests/unit/test_providers/test_claude_cli_provider.py tests/providers/test_registry.py tests/unit/cli/test_init_ux.py tests/unit/generation/test_cost_pricing.py tests/unit/server/mcp/test_answer_synthesis_timeout.py` — 134 passed.
6. Run `uv run pytest -q tests/integration/` — 239 passed, 20 skipped.
7. Run `uv run ruff check .`, `npm run lint`, `npm run type-check`, and `npm run build` — all pass.

## Checklist

- [x] Code follows the repository style
- [x] Focused and integration tests cover the new provider and review regressions
- [x] Documentation is updated
- [x] Real Claude Code success, reasoning, and error paths were exercised

### DIFF
--- a/docs/agent/CLAUDE_CODE_PROVIDER.md
+++ b/docs/agent/CLAUDE_CODE_PROVIDER.md
@@ -1,0 +1,140 @@
+# Claude Code as an LLM Provider
+
+Repowise and Claude Code connect in **two independent directions**. This page
+covers one of them; enabling it does not enable the other.
+
+| Direction | What it is | Where to read |
+|---|---|---|
+| Repowise calls Claude Code | The `claude_cli` **LLM provider**. Repowise runs your local Claude Code CLI to write wiki pages, so a Claude subscription works instead of an API key. | This page |
+| Claude Code calls Repowise | The `claude-code` **agent target**. Claude Code gets the repowise MCP tools, so you can ask it about your codebase. | [Integrations](INTEGRATIONS.md) |
+
+## `claude_cli` Provider
+
+Use `claude_cli` when you want page generation to run through your local Claude
+Code CLI instead of an `ANTHROPIC_API_KEY`:
+
+```bash
+repowise init --provider claude_cli --yes
+```
+
+Or for an existing index:
+
+```bash
+REPOWISE_PROVIDER=claude_cli repowise generate --unwritten
+```
+
+To persist the choice, put it in `.repowise/config.yaml`:
+
+```yaml
+provider: claude_cli
+model: claude_cli/claude-haiku-4-5
+```
+
+### Prerequisites
+
+```bash
+# Install Claude Code, then authenticate once:
+claude login
+```
+
+Any plan that can run `claude -p` works — Pro, Max, Team or Enterprise. Repowise
+never sees your credentials: the CLI holds them, and authentication happens out
+of band exactly as it does for `codex_cli` and `opencode`.
+
+### What the provider runs
+
+```bash
+claude -p --output-format json --model <model> --max-turns 1 \
+  --strict-mcp-config --tools "" --system-prompt <system> [--effort <level>]
+```
+
+Repowise sends the **user** prompt on stdin (prompts carry file context and can
+be large) and passes the **system** prompt via `--system-prompt`, which
+*replaces* Claude Code's agent preamble rather than appending to it — repowise's
+prompt is the whole instruction set, and the coding-agent framing only competes
+with it. Repowise removes the tool catalog and adds a final system instruction
+to answer directly in one response. The instruction is necessary even with a
+single-turn budget: denying individual tools still lets the model attempt one,
+consume that turn, and return `error_max_turns` without any page prose.
+
+It parses the single JSON object from stdout, taking `result` as the page
+content and mapping `usage`:
+
+| Claude Code field | Repowise field |
+|---|---|
+| `usage.input_tokens` | `input_tokens` |
+| `usage.output_tokens` | `output_tokens` |
+| `usage.cache_read_input_tokens` | `cached_tokens` |
+| `usage.cache_creation_input_tokens` | recorded in `usage`, not counted as cached |
+| `stop_reason` | normalised `stop_reason` |
+
+`claude_cli/*` models are priced at **$0.00**, because billing is handled by the
+subscription rather than per-token API spend. The CLI's own `total_cost_usd` is
+still recorded under `usage.reported_cost_usd` for auditing.
+
+### Default model
+
+`claude_cli/claude-haiku-4-5` is the default, matching the `anthropic` provider,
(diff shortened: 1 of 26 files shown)


RaghavChamadiya (approved):
Green across 3.11, 3.12, 3.13, ruff, web lint and type check, and the VS Code extension, on `3de23a1f`.

I verified the remediation independently rather than merging on the summary, running your branch through a bridge venv and a live Claude Code 2.1.232:

- 175 focused tests pass (provider, registry, cost pricing, synthesis timeout, init UX) and `ruff check .` is clean.
- The argv carries `--tools ""` alongside `--max-turns 1` and `--strict-mcp-config`, with the tool-free instruction appended to the system prompt.
- A prompt that explicitly says "Investigate the repository with your tools" returns `end_turn` with real prose. That is the exact shape that used to come back `error_max_turns` with `result: null`.
- `reasoning="low"` reaches the CLI and returns in one turn.
- A nonexistent model raises `claude -p reported failure (404: There's an issue with the selected model ...)` rather than `exited with 1`.
- `is_local_model("claude_cli/claude-haiku-4-5")` is now `True`, so the runtime ledger agrees with the estimator.
- A full `repowise init` on `microdot` through your provider: 188 pages, **zero failures**, zero structural placeholders, `repo_overview` written as prose, shortest page 872 characters. The run that prompted my August review lost 3 pages including `repo_overview`.

Three notes on the work itself.

`--tools ""` is the better fix and mine was not. I had reached for the system-prompt instruction alone, which works by persuading the model rather than by removing the capability, and leaves the tool catalog in context. You did both, which is right for a failure that is silent when it happens.

The scratch-directory change is an upgrade over what I proposed. `atexit` handles one leaked directory and nothing else. A per-call `TemporaryDirectory` cleaned up on success, failure, timeout and cancellation, with the subprocess killed and awaited on cancel, closes something I had missed entirely: the 180 second interactive timeout returns while the subprocess runs on toward its own 600 second ceiling, holding a subscription slot nobody is waiting for.

The cost-ledger bug is yours alone and neither the review nor my branch caught it. `is_local_model` covered `codex_cli/` and `opencode/` and not `claude_cli/`, so the estimator said $0 while the ledger stripped the prefix and priced subscription calls at metered Anthropic rates. That is the worst shape a billing bug can take, because the number people check first is the one that was right.

Thank you for the patience through a stale rebase, a close I should not have made, and a 422 on the reopen. Merging as yours, which is how it should have gone in August. This closes #1513.

PR #1517, opened 2026-08-14
fix(providers): title OpenCode sessions from Repowise

Mushst:
Adds a distinct generated title to each Repowise OpenCode CLI session. Uses OpenCode's supported `--title` option; no server integration or session archival logic.

Current generation with opencode run creates a bunch of orphan chats that clutter everything up. Minimal change is to tag them so users know they came from repowise generation. Further auto-archival or auto-deletion of those chats might be useful.

### DIFF
--- a/packages/core/src/repowise/core/providers/llm/opencode.py
+++ b/packages/core/src/repowise/core/providers/llm/opencode.py
@@ -19,6 +19,7 @@ import os
 import re
 import shutil
 import subprocess
+import uuid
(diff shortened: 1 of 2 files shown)


RaghavChamadiya (approved):
Thanks for this. The orphan-chat clutter is a real annoyance and tagging the sessions is the right minimal fix. The uuid suffix keeps each run distinguishable and the test pins the format, so nothing can quietly drop the flag later. Merging.

PR #1521, opened 2026-08-14
fix(cli): preserve sync baseline when HEAD is unavailable

### LINKED ISSUE #1507: [Bug] repowise generate overwrites last_sync_commit with None when git fails, breaking the next update
## Describe the Bug

`_write_state` unconditionally does `state["last_sync_commit"] = get_head_commit(repo_path)` (generate_cmd/command.py:359). `get_head_commit` returns `None` whenever `git rev-parse HEAD` fails (non-git repo, subprocess error, timeout — helpers.py:556-557; workspace/update.py:106-120). This overwrites a previously valid sync pointer with `None`. The next `repowise update` then reads `base_ref = state.get("last_sync_commit")` (update_cmd/command.py:708) → `None` → hard-fails "No previous sync found for {repo_path}. Run 'repowise init' there first" (714-732). Notably, `update` itself guards the same situation (`head and head == base_ref`, 820-826).

## Steps to Reproduce

1. Index a repo (state.json has a valid `last_sync_commit`).
2. Run `repowise generate` in an environment where `git rev-parse HEAD` fails (e.g. repo no longer a git checkout, or a git subprocess error).
3. `_write_state` sets `last_sync_commit` to `None`.
4. Run `repowise update` — it errors out with "No previous sync found".

## Expected Behavior

A failed/unavailable HEAD read should not erase the persisted baseline (update preserves the pointer).

## Actual Behavior

`generate` on a repo where HEAD cannot be resolved clobbers `last_sync_commit`, and `update` afterwards refuses to run.

## Environment

- OS: macOS 14
- Python version: 3.13
- Repowise version: 0.41.0

## Additional Context

Code: `packages/cli/src/repowise/cli/commands/generate_cmd/command.py:359` (unconditional clobber); `packages/core/src/repowise/core/workspace/helpers.py:556-557` (`get_head_commit` returns None on git failure); `update_cmd/command.py:708, 714-732` (base-ref read + hard-fail).

Suggested fix: guard the write — `head = get_head_commit(repo_path); if head: state["last_sync_commit"] = head` (or leave the pointer untouched when HEAD is None), matching update's own None-tolerant handling.

Sanjays2402:
Closes #1507

## Summary

- Preserve the existing `last_sync_commit` when `git rev-parse HEAD` is unavailable during `generate`.
- Add a regression test for the failed HEAD lookup path.

## Related Issues

Closes #1507

## Test Plan

- [x] Tests pass (`pytest`)
- [ ] Web build passes (`npm run build`) (N/A: no frontend changes)

This change was prepared with AI assistance; the regression test was run locally and fails without the fix.

## Checklist

### DIFF
--- a/packages/cli/src/repowise/cli/commands/generate_cmd/command.py
+++ b/packages/cli/src/repowise/cli/commands/generate_cmd/command.py
@@ -356,7 +356,9 @@ def _write_state(repo_path: Path, state: dict, provider: Any, outcome: Any) -> N
     upgrade leaves a mixed wiki, so it keeps its current mode rather than
     over-claiming that everything is written.
     """
-    state["last_sync_commit"] = get_head_commit(repo_path)
+    head = get_head_commit(repo_path)
+    if head is not None:
+        state["last_sync_commit"] = head
     state["total_pages"] = outcome.total_pages
     state["provider"] = provider.provider_name
     state["model"] = provider.model_name
--- a/tests/unit/cli/test_commands.py
+++ b/tests/unit/cli/test_commands.py
@@ -2,16 +2,38 @@
 
 from __future__ import annotations
 
+from pathlib import Path
+from types import SimpleNamespace
+from unittest.mock import patch
+
 import pytest
 from click.testing import CliRunner
 
 from repowise.cli import __version__
+from repowise.cli.commands.generate_cmd.command import _write_state
 from repowise.cli.main import cli
 
 
 @pytest.fixture
 def runner():
     return CliRunner()
+
+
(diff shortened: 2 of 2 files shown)


RaghavChamadiya (approved):
Thanks for the fix, and the regression test is the right shape: it fails without the change and it checks the real symptom rather than the call.

Heads up that #1527 landed the same fix about an hour later. Yours was first so I'm going with this one. @Ayush7614, would you mind rebasing the extra "git succeeds, new SHA is stamped" case from your test file on top? That direction isn't covered here and it's worth keeping.

PR #1522, opened 2026-08-14
fix(security-scan): flag multi-line subprocess.*(... shell=True) calls

### LINKED ISSUE #1506: [Bug] Security scanner misses multi-line subprocess.*(... shell=True) — the command-injection sink escapes detection
## Describe the Bug

The pattern `subprocess\..*shell\s*=\s*True` is matched **per physical line** (`for lineno, line in enumerate(source.splitlines(), start=1)`, security_scan.py:94-109). `.*` cannot cross a line break, so idiomatic multi-line calls produce no finding on any line (the `subprocess\.` prefix and `shell=` land on different lines). The prefilter `_ANY_PATTERN` (line 51) has the same shape. `shell=True` is exactly the sink the scanner claims to flag ("subprocess_shell_true", high), and multi-line formatting is how it appears in real code.

## Steps to Reproduce

Scan a file containing:

```python
subprocess.run(
    "rm -rf " + path,
    shell=True,
)
```

Result: zero `subprocess_shell_true` findings (each line fails the anchored pattern).

## Expected Behavior

`subprocess.run(cmd, shell=True)` / `subprocess.Popen(..., shell=True)` should be flagged regardless of line breaks within the call.

## Actual Behavior

Only single-line `subprocess.* ... shell=True` is detected; the multi-line form is silently missed.

## Environment

- OS: macOS 14
- Python version: 3.13
- Repowise version: 0.41.0

## Additional Context

Code: `packages/core/src/repowise/core/analysis/security_scan.py:38` (pattern), `:51` (`_ANY_PATTERN`), `:94-109` (per-line loop). Existing `tests/unit/analysis/test_security_scan.py` has no multi-line case.

Suggested fix: scan with `re.DOTALL` across whole-source regexes (with start-of-match anchor `^`), or normalize whitespace (e.g. match `subprocess\.[A-Za-z]+\([\s\S]*?shell\s*=\s*True`) on a per-file basis while keeping the per-line snippet/line reporting.

Ayush7614:
Fixes #1506

## Problem

The pattern `subprocess\..*shell\s*=\s*True` is matched per physical line (`security_scan.py` line loop), and `.*` cannot cross a newline. Idiomatic multi-line calls:

```python
subprocess.run(
    "rm -rf " + path,
    shell=True,
)
```

produce zero `subprocess_shell_true` findings — the high-severity command-injection sink escapes detection.

## Fix

Add a whole-source pass for patterns that legitimately span physical lines:
- `subprocess\.[A-Za-z]+\([\s\S]{0,500}?shell\s*=\s*True` scanned over the full source once.
- The span is bounded (500 chars) so a `subprocess` call without `shell=` cannot swallow an unrelated `shell=True` later in the file.
- The finding is reported on the call's start line, and a match the per-line pass already caught on that line is not duplicated.

## Tests

Added 4 cases to `test_security_scan.py`: single-line `subprocess.run`, multi-line `subprocess.run`, multi-line `subprocess.Popen`, and a negative case (multi-line call without `shell=`). All pass.

### DIFF
--- a/packages/core/src/repowise/core/analysis/security_scan.py
+++ b/packages/core/src/repowise/core/analysis/security_scan.py
@@ -40,7 +40,7 @@ _PATTERNS: list[tuple[re.Pattern, str, str]] = [
     (re.compile(r"password\s*=\s*['\"]"), "hardcoded_password", "high"),
     (re.compile(r"(?:api_?key|secret)\s*=\s*['\"]"), "hardcoded_secret", "high"),
     (re.compile(r'f[\'"].*SELECT.*\{.*\}'), "fstring_sql", "med"),
-    (re.compile(r"\.execute\(\s*[\'\"]\s*SELECT.*\+"), "concat_sql", "med"),
+    (re.compile(r'\.execute\(\s*[\'\"]\s*SELECT.*\+'), "concat_sql", "med"),
     (re.compile(r"verify\s*=\s*False"), "tls_verify_false", "med"),
     (re.compile(r"\bmd5\b|\bsha1\b"), "weak_hash", "low"),
 ]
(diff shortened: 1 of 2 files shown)


RaghavChamadiya (approved):
Span tightened so it cannot cross a statement boundary, and the format churn is dropped.

PR #1523, opened 2026-08-14
fix(security-scan): stop replace_findings from silently losing rows on duplicate keys

### LINKED ISSUE #1483: [Bug] SecurityScanner.replace_findings silently truncates the findings batch on duplicate provenance keys
## Describe the Bug

`SecurityScanner.replace_findings` is a delete-then-bulk-insert with **no conflict clause** on the bulk INSERT and a blanket `except Exception: return` around the whole method. When two findings in one batch collide on the unique provenance key `(repository_id, file_path, kind, line_number, commit_sha)`, the INSERT raises `IntegrityError`, which is swallowed, and the batch is inserted only up to the offending row. Because the DELETE already ran, the file loses both its previously stored findings **and** the new ones — silently, with no log line.

## Steps to Reproduce

Reproduced against the real module with an in-memory SQLite DB. Two `security_sensitive_symbol` findings at the same `start_line` (or symbols without `start_line` collapsing to line 0 via `getattr(sym, "start_line", 0) or 0` in security_scan.py:120) produce identical provenance keys:

- findings: `[{kind:'hardcoded_password',line:1},{kind:'security_sensitive_symbol',line:7},{kind:'security_sensitive_symbol',line:7}]`
- rows after `replace_findings`: `[('hardcoded_password', 1), ('security_sensitive_symbol', 7)]` — **one finding silently lost**, no exception raised.

A direct duplicate insert in the same session raises `IntegrityError` (verified), confirming the collision is real and only the method's error handling hides it.

## Expected Behavior

Re-running the scan must be idempotent and must never lose findings. The sibling `persist()` method (security_scan.py:134-210) does exactly this: per-row `ON CONFLICT ... DO NOTHING` / `INSERT OR IGNORE` plus a per-row `try/except continue`.

## Actual Behavior

On a duplicate key, the batch is partially inserted (rows before the collision survive, rows after are lost), no error is surfaced, and the DELETE has already removed the file's prior findings.

## Environment

- OS: macOS 14
- Python version: 3.13
- Repowise version: 0.41.0

## Additional Context

Code: `packages/core/src/repowise/core/analysis/security_scan.py:261-271` (bulk INSERT, no conflict handling), `:272` (`except Exception: return`). Unique constraint: `packages/core/src/repowise/core/persistence/models.py:1069-1076`. Entry point: `pipeline/persist.py:persist_security_findings` (lines 1345-1370).

Impact: silent data loss in the security-findings surface (dashboard, history, MCP tools) for any file with two same-line keyword symbols (common in minified/compressed sources) or symbols lacking `start_line`.

Existing tests (`tests/unit/analysis/test_security_scan.py`, 6 passing) do not cover two same-key findings.

Suggested fix: make the bulk INSERT conflict-tolerant (`INSERT OR IGNORE` / `ON CONFLICT ... DO NOTHING`) mirroring `persist()`, and replace the blanket `except Exception: return` with per-row handling so genuine failures are logged.

Ayush7614:
Fixes #1483

## Problem

`replace_findings` is a delete-then-bulk-insert with no conflict clause on the bulk INSERT and a blanket `except Exception: return` around the whole method. When two findings in one batch collide on the unique provenance key `(repository_id, file_path, kind, line_number, commit_sha)` (e.g. two `security_sensitive_symbol` findings on the same line, or symbols without `start_line` collapsing to line 0), the INSERT raises `IntegrityError`, which is swallowed, and the batch inserts only up to the offending row. Because the DELETE already ran, the file loses both its previously stored findings and most of the new ones — silently.

## Fix

- Deduplicate rows in Python by provenance key before the INSERT (duplicates are redundant signals, not distinct rows).
- Use the same conflict-tolerant clauses as the sibling `persist()`: `INSERT OR IGNORE` on SQLite, `ON CONFLICT ... DO NOTHING` on Postgres.
- Narrow the blanket `except Exception: return`: pre-migration `no such table` failures still skip silently (the historical contract); any other failure is now logged via `logger.exception` instead of dropping the batch.

## Tests

Added two cases to `test_security_scan.py`:
- duplicate-key batch inserts exactly one row per provenance key (no partial insert),
- re-running replace_findings over the same file yields the identical full row set.

All 8 scanner tests + 10 history tests pass.

### DIFF
--- a/packages/core/src/repowise/core/analysis/security_scan.py
+++ b/packages/core/src/repowise/core/analysis/security_scan.py
@@ -40,7 +40,7 @@ _PATTERNS: list[tuple[re.Pattern, str, str]] = [
     (re.compile(r"password\s*=\s*['\"]"), "hardcoded_password", "high"),
     (re.compile(r"(?:api_?key|secret)\s*=\s*['\"]"), "hardcoded_secret", "high"),
     (re.compile(r'f[\'"].*SELECT.*\{.*\}'), "fstring_sql", "med"),
-    (re.compile(r"\.execute\(\s*[\'\"]\s*SELECT.*\+"), "concat_sql", "med"),
+    (re.compile(r'\.execute\(\s*[\'\"]\s*SELECT.*\+'), "concat_sql", "med"),
     (re.compile(r"verify\s*=\s*False"), "tls_verify_false", "med"),
     (re.compile(r"\bmd5\b|\bsha1\b"), "weak_hash", "low"),
 ]
@@ -51,7 +51,9 @@ _PATTERNS: list[tuple[re.Pattern, str, str]] = [
 _ANY_PATTERN = re.compile("|".join(f"(?:{p.pattern})" for p, _, _ in _PATTERNS))
 
 # Symbol names that are informational security hotspots
-_SYMBOL_KEYWORDS = re.compile(r"\b(auth|token|password|jwt|session|crypto)\b", re.IGNORECASE)
+_SYMBOL_KEYWORDS = re.compile(
+    r"\b(auth|token|password|jwt|session|crypto)\b", re.IGNORECASE
+)
 
 # Patterns whose matches are genuine leaked credentials (as opposed to the
 # broader "code smell" patterns like os.system/eval). Full-history scans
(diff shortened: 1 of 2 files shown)


RaghavChamadiya (approved):
Dedupe plus conflict-tolerant insert, and the bare except is split so only the pre-migration case stays silent.

PR #1524, opened 2026-08-14
fix(update-lock): never report 'acquired' with no lock file on disk

### LINKED ISSUE #1503: [Bug] try_acquire_update_lock can return 'acquired' (None) with no lock file on disk
## Describe the Bug

`try_acquire_update_lock` retries the exclusive `os.link` twice. If both attempts hit `FileExistsError` and the existing lock reads as stale each time, the stale lock is unlinked and the loop falls through to `return read_update_lock(repo_path)` (update_lock.py:102). If no other process re-creates the lock in that window, this returns `None` — which every caller treats as **"lock acquired"** — while **no lock file exists on disk**. A third process then also acquires, and the exclusivity guarantee is broken. The docstring explicitly says `None` means acquired (line 46).

## Steps to Reproduce

Simulated the interleaving by forcing both `os.link` calls to raise `FileExistsError` against a stale (dead-PID) lock while letting nothing re-create it:

```
result: None
lock file exists after call: False
RACE CONFIRMED: caller believes it owns the lock, but no lock file exists
```

## Expected Behavior

`None` must imply a lock file exists (or that the caller created one). The atomic-create design (PR #720) promises exactly one winner.

## Actual Behavior

After two consecutive lost create-races with a stale lock, the function returns `None` with no lock created; callers proceed without holding the lock, and a concurrent update can race on one index.

## Environment

- OS: macOS 14
- Python version: 3.13
- Repowise version: 0.41.0

## Additional Context

Code: `packages/core/src/repowise/core/update_lock.py:78-102`; fallthrough `return read_update_lock(repo_path)` at line 102; caller semantics at `update_cmd/command.py:907-908`. All 19 existing `tests/unit/cli/test_update_lock.py` tests pass — none covers the loop-exhaustion fallback.

Suggested fix: before the fallback return, make one final `os.link` attempt; if it succeeds return `None` (owned), if it loses return the winner's payload.

Ayush7614:
Fixes #1503

## Problem

`try_acquire_update_lock` retries the exclusive `os.link` twice. If both attempts hit `FileExistsError` and the existing lock reads as stale each time, the stale lock is unlinked and the loop falls through to `return read_update_lock(repo_path)`. If no other process re-creates the lock in that window, this returns `None` — which every caller treats as **"lock acquired"** — while **no lock file exists on disk**. A concurrent update then also "acquires", and two updates race on one index (the exact interleaving the lock exists to serialize).

## Fix

After both retries lose against a stale lock, make one final exclusive create:
- winner returns `None` (owned, lock file on disk),
- loser reports the fresh winner's payload,
- still-unreadable degrades to acquired, as everywhere.

`None` now always implies the caller owns a lock file.

## Tests

Added `test_loop_exhaustion_never_returns_acquired_without_a_lock_file`: forces the first two `os.link` calls to raise `FileExistsError` against a stale lock, then asserts the acquire returns `None` *and* a lock file with our PID exists on disk. All 20 lock tests pass.

### DIFF
--- a/packages/core/src/repowise/core/update_lock.py
+++ b/packages/core/src/repowise/core/update_lock.py
@@ -97,9 +97,24 @@ def try_acquire_update_lock(repo_path: Path, target_commit: str | None) -> dict[
             with contextlib.suppress(OSError):
                 tmp_path.unlink(missing_ok=True)
         return None
-    # Lost the create race twice in a row: someone else just acquired a
-    # fresh lock — report it. A still-unreadable lock degrades to acquired.
-    return read_update_lock(repo_path)
+    # Both create attempts lost a race against a stale lock that was then
+    # unlinked. Rather than falling through to a bare ``read_update_lock``
+    # — which can return ``None`` ("acquired") with *no lock file on disk*,
+    # letting a caller proceed without holding the lock — make one final
+    # exclusive create. Winner: return ``None`` (owned). Loser: report the
+    # fresh winner. Still-unreadable degrades to acquired, as everywhere.
+    try:
+        lock_path.parent.mkdir(parents=True, exist_ok=True)
+        tmp_path.write_text(data, encoding="utf-8")
+        os.link(tmp_path, lock_path)
+    except FileExistsError:
+        return read_update_lock(repo_path)
+    except OSError:
+        return None
+    finally:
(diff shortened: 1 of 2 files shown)


RaghavChamadiya (approved):
Reviewed above. Merging.

PR #1525, opened 2026-08-14
fix(init): make --test-run actually limit generation to top 10 files

### LINKED ISSUE #1505: [Bug] init --test-run is a silent no-op — the 'limit to top 10 files by PageRank' flag never takes effect
## Describe the Bug

`--test-run` help: "Limit generation to top 10 files by PageRank for quick validation." (init_cmd/command.py:446). The **only** consumer in the codebase is orchestrator.py:444-457, gated by `if test_run and generate_docs:`. `init` always invokes `run_pipeline(... generate_docs=False ...)` (command.py:1296; workspace.py:327), and generation runs through `run_repo_generation`, which never receives `test_run`. The interactive advanced-config flow collects it and even prints "Test run: yes" (ui/mode_selection.py:467, 553).

## Steps to Reproduce

1. Run `repowise init --test-run` (or select "Test run: yes" in the interactive advanced config).
2. Observe that generation is NOT limited to 10 files — the full generation runs.

## Expected Behavior

With `--test-run`, generation should be limited to the top-10 files by PageRank for a fast validation run.

## Actual Behavior

The flag is forwarded but never consumed: `generate_docs` is always False, so the `test_run` branch is dead. A user selecting "test run" gets a full generation.

## Environment

- OS: macOS 14
- Python version: 3.13
- Repowise version: 0.41.0

## Additional Context

Exhaustive grep: `test_run` appears only in orchestrator.py:444 (gated on `generate_docs`), init_cmd (forwarding), mode_selection (UI), and agent prompt docs — never in any generation path with `generate_docs=True`.

Suggested fix: either wire `test_run` into the generation entry (pass it to `run_repo_generation`/`run_generation` and apply the top-10 truncation there), or gate the flag on `generate_docs=True` paths and remove it from init.

Ayush7614:
Fixes #1505

## Problem

`--test-run` is documented as "Limit generation to top 10 files by PageRank for quick validation" but was a silent no-op. The only consumer was gated on `test_run and generate_docs` in the orchestrator (`orchestrator.py:444`), and init always runs the pipeline with `generate_docs=False` (generation happens in a separate phase afterwards). The flag was forwarded end-to-end and never consumed; a user selecting "test run" got a full generation.

## Fix

Move the top-10 PageRank cap into `run_generation` itself — the single entry both init paths converge on:
- `run_generation` gains a `test_run` parameter and truncates `parsed_files` to the top 10 by PageRank before generating.
- Threaded through `run_repo_generation` (single-repo `_run_generation_phase` and workspace `_run_workspace_generation`).
- The deterministic template renderers keep `test_run=False` — they're free and fast, so the cap only matters on the paid LLM path.

## Tests

New `tests/unit/pipeline/test_run_generation_test_run.py` stubs `PageGenerator` and asserts `test_run=True` hands it exactly 10 files (from 15), and the default passes all 15. All related init/generation tests pass (67).

### DIFF
--- a/packages/cli/src/repowise/cli/commands/init_cmd/command.py
+++ b/packages/cli/src/repowise/cli/commands/init_cmd/command.py
@@ -263,6 +263,7 @@ def _run_generation_phase(
     skip_infra: bool,
     embedder_name_resolved: str,
     resume: bool,
+    test_run: bool,
 ) -> tuple[bool, bool]:
     """Run the LLM generation phase for a single-repo init.
 
@@ -391,6 +392,7 @@ def _run_generation_phase(
         embedder_name_resolved=embedder_name_resolved,
         resume=resume,
         verbose=True,
+        test_run=test_run,
     )
     return False, False
 
@@ -1396,6 +1398,7 @@ def init_command(
             # file. Resuming against it would skip the entire model run and
             # say nothing, so a template wiki is never a run to continue.
             resume=resume and _prior_docs_mode != "deterministic",
+            test_run=test_run,
         )
         if gen_stop:
             return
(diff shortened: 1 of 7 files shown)


RaghavChamadiya (approved):
One helper, orchestrator copy removed, cap in a shared constant.

PR #1528, opened 2026-08-14
fix: up-to-date self-heal writes under the single-flight lock

### LINKED ISSUE #1486: [Bug] 'Already up to date' self-heal path writes state/DB outside the single-flight lock
## Describe the Bug

When `head == base_ref` and nothing changed, the early return in `run_update` (command.py:820-894) performs `save_state(...)` (844-852), `stamp_head_commit` (856), `heal_commit_offsets` (859), `_refresh_editor_stamp` (860), and `consume_update_pending` (865) — all **before** the lock is acquired at line 907.

The codebase's own comment (command.py:754-758) declares exactly this hazard: "Writing it here would persist state on a --dry-run, and would do it before the single-flight lock below, where a concurrent update is exactly what the lock exists to stop."

## Steps to Reproduce

1. Have a repo where `update` is "already up to date".
2. Start one `repowise update` that is mid-flight and holds the lock (e.g. a large incremental re-index).
3. Start a second `repowise update` in another process while the first is running. The second observes `head == base_ref` and takes the self-heal path, writing `state.json`, stamping HEAD, and consuming the `.update.pending` marker **concurrently** with the locked run.

## Expected Behavior

All state/DB writes happen under the lock.

## Actual Behavior

A second process observing `head == base_ref` writes state concurrently with the locked run — the exact interleaving the lock exists to serialize.

## Environment

- OS: macOS 14
- Python version: 3.13
- Repowise version: 0.41.0

## Additional Context

Code: `packages/cli/src/repowise/cli/commands/update_cmd/command.py:844-865` (self-heal writes), `:907` (lock acquisition), `:754-758` (comment acknowledging the hazard). `dry_run` guards some writes (844/855) but not the lock.

Suggested fix: move the self-heal writes after lock acquisition (or gate the whole early-return block on holding the lock), keeping the fast no-op path read-only.

Ayush7614:
## Summary

On the "already up to date" shortcut, `repowise update` performed self-heal writes **before** acquiring the single-flight lock:

- backfilled `config_fingerprint` / `renderer_fingerprint` via `save_state`
- stamped the DB head commit (`stamp_head_commit`), healed commit offsets, refreshed the editor stamp, and consumed the pending marker

Two updates racing on the no-op path could therefore `save_state` out of order — the exact race the single-flight lock exists to prevent (see the lock's comment about "wiki keeps going stale").

## Fix

The heal writes now run inside `try_acquire_update_lock`. When another update holds the lock, the heal is skipped (the running update will perform it); otherwise it runs and the lock is released in a `finally`.

## Tests

- New `tests/unit/cli/test_update_up_to_date_lock.py` drives the real `repowise update --no-workspace` command on an up-to-date repo:
  - lock held → acquire is attempted, but `stamp_head_commit`/`heal_commit_offsets`/`consume_update_pending`/`save_state` are never called and the lock is never released;
  - lock free → heal runs once and the lock is released once.
- Related update suites still pass (63 tests).

Closes #1486

### DIFF
--- a/packages/cli/src/repowise/cli/commands/update_cmd/command.py
+++ b/packages/cli/src/repowise/cli/commands/update_cmd/command.py
@@ -841,28 +841,42 @@ def run_update(
         # rather than a change, silently skipping the regeneration it should
         # trigger.
         needs_backfill = prev_config_fp is None or prev_renderer_fp is None
-        if needs_backfill and not dry_run:
-            save_state(
-                repo_path,
-                {
-                    **state,
-                    "config_fingerprint": curr_config_fp,
-                    "renderer_fingerprint": curr_renderer_fp,
-                },
-            )
         # Self-heal a row a pre-fix run left behind: state.json can already be
         # current here while the DB head_commit is still the last full index.
+        #
+        # These state/DB writes must sit under the single-flight lock below
+        # just like a content-changing update, or two updates racing on the
+        # "already up to date" path could stomp each other's save_state — the
+        # exact race the lock exists to prevent. If another update holds the
+        # lock, its run will perform this heal; skip ours rather than write
+        # from outside the lock.
         if not dry_run:
-            stamp_head_commit(repo_path, head)
-            # A capture added after this repo was indexed would otherwise wait
-            # for the next commit to land, which on a quiet repo is never.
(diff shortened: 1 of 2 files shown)


RaghavChamadiya (approved):
Reviewed above. Merging.

PR #1529, opened 2026-08-14
fix: update --full runs under the single-flight lock

### LINKED ISSUE #1485: [Bug] update --full bypasses the single-flight update lock
## Describe the Bug

The single-flight lock exists precisely because "two updates racing on save_state was the actual root cause of 'wiki keeps going stale'" (update_cmd/command.py:896-906). But the `if full:` branch (command.py:666-692) runs `upgrade_to_full(...)` and **returns before** `try_acquire_update_lock` at line 907. `upgrade_to_full` writes page rows and `save_state` (upgrade_flow.py:506-527) — the exact class of state/DB write the lock protects.

## Steps to Reproduce

1. Configure an LLM provider.
2. Start `repowise update --full` on a repo (runs `upgrade_to_full` on the `full` path, no lock).
3. While it is running, trigger a second update (e.g. the post-commit auto-sync hook, or a second terminal) — it acquires the lock and writes the same index concurrently.

## Expected Behavior

Every state/page-writing update path is single-flight under the lock.

## Actual Behavior

A concurrent `update --full` and a locked incremental update can both write page rows and `state.json` on one index, re-creating the stale-wiki race the lock was built to end.

## Environment

- OS: macOS 14
- Python version: 3.13
- Repowise version: 0.41.0

## Additional Context

Code: `packages/cli/src/repowise/cli/commands/update_cmd/command.py:666-692` (full branch, returns before lock at :907); lock rationale comment at :896-906; `upgrade_flow.py:506-527` (page writes + `save_state`).

Suggested fix: acquire the update lock before dispatching the `--full` branch (release in `finally`), so `--full` participates in single-flight.

Ayush7614:
## Summary

`repowise update --full` (the fast → full upgrade) dispatched in its own early branch **before** the incremental update's single-flight lock, yet it writes state — `save_state` at upgrade_flow.py, plus the DB writes inside `_run_upgrade` — without ever acquiring the lock. A concurrent `repowise update` could race the full upgrade's state writes, the exact corruption the lock exists to prevent.

## Fix

`upgrade_to_full` now acquires the single-flight lock itself before doing any work:

- lock free → proceeds; lock is released on every exit path via `atexit.register(release_update_lock, ...)`, mirroring the incremental update;
- lock held by another update → defers: raises a clear `ClickException` and writes `.update.pending` so the running update rolls forward to HEAD, instead of writing state from outside the lock.

## Tests

- New `tests/unit/cli/test_full_upgrade_lock.py` drives `upgrade_to_full` directly with a stubbed provider/`_run_upgrade`:
  - lock free → the upgrade acquires the lock;
  - lock held → the upgrade raises `ClickException`, never runs `_run_upgrade`, and leaves a pending marker.
- Related update suites still pass (48 tests).

Closes #1485

### DIFF
--- a/packages/cli/src/repowise/cli/commands/upgrade_flow.py
+++ b/packages/cli/src/repowise/cli/commands/upgrade_flow.py
@@ -46,6 +46,7 @@ from repowise.cli.helpers import (
 )
 from repowise.core.analysis.health import HEALTH_ANALYZER_VERSION
 from repowise.core.docs_mode import docs_mode_state_fields
+from repowise.core.update_lock import try_acquire_update_lock
 
 
 def _gate_cost(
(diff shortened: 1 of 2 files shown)


RaghavChamadiya (approved):
`try/finally` with the module-level import reads much better, and the test asserts the release now instead of collecting it and dropping it. The deferral path raises before acquiring, so there's no unbalanced release. Approving.

PR #1532, opened 2026-08-14
fix(update): reuse persisted function-mod p80 in incremental hotspot gate

### LINKED ISSUE #1484: [Bug] Incremental runs compute repo_function_mod_p80 over the changed-files subset, biasing the Function Hotspot gate
## Describe the Bug

On an incremental run, `walked`/`target_files` contains **only the changed files** (health/engine.py:427-428 and 569-573). `repo_fn_mod_p80 = _compute_repo_function_mod_p80(walked, ...)` therefore computes the "repo-wide" 80th percentile of per-function modification counts over the churn-heavy changed subset, not the whole repo. Sibling repo-wide stats are computed from full data: `_compute_repo_dependents_p80` uses `self.parsed_files` (engine.py:441/616) and active-contributors uses the full `git_meta_map`. The code itself acknowledges this bias class for KPIs ("on an incremental run they would be biased by the changed-files subset", engine.py:496-499) but the hotspot gate percentile is passed live to every file.

## Steps to Reproduce

Full-repo mod counts `[1,2,2,3,4,4,5]` → p80 = 4; changed subset `[4,4,5]` → p80 = 5. A function with mod_count=4 fires on a full run (`4 >= 4`) and is silent on incremental (`4 < 5`). Because `save_health_findings` (crud/analysis/health.py:29-82) deletes-and-reinserts all open findings, the incremental result **replaces** stored state.

1. Run `repowise init` (full) — a hotspot is reported.
2. Touch one file that is at the true repo p80 and run `repowise update` (incremental) — the hotspot disappears, dashboard differs.

## Expected Behavior

`repo_function_mod_p80` must be the same value in a full and an incremental run so a file's hotspot verdict is run-type-independent.

## Actual Behavior

A changed file's `mod_count` is compared against the percentile of a few churned files. The health dashboard silently differs between `repowise update` and `repowise init`; `update` (the frequent path) can permanently drop a hotspot a full re-index would report.

## Environment

- OS: macOS 14
- Python version: 3.13
- Repowise version: 0.41.0

## Additional Context

Code: `packages/core/src/repowise/core/analysis/health/engine.py:171-192` (helper), `:440` / `:615` (walked vs parsed_files), `:496-499` (KPI bias acknowledged). Consumer: `packages/core/src/repowise/core/analysis/health/biomarkers/function_hotspot.py:55-91` (gate at :66).

Suggested fix: compute the percentile from full-repo data (persist it in the DB and reuse on incremental runs), or skip the hotspot gate on incremental runs like KPIs, rather than deriving a "repo-wide" percentile from the changed subset.

Ayush7614:
Closes #1484

## Problem

The Function Hotspot biomarker's gate is the repo-wide 80th percentile of per-function modification counts. On an incremental update the walked set holds only the changed files, so the percentile was derived from a churn-heavy subset. A function near the full-repo threshold could flip in or out of the hotspot verdict purely because an unrelated file changed — and a full `init` that surfaced the hotspot would silently lose it after the first incremental `update`.

## Fix

The persisted `git_function_blame` rollup already carries repo-wide per-function mod counts (a full index writes all modified functions; an incremental persists only the changed subset). 

- New `get_git_function_mod_counts` CRUD returns stored `mod_count > 0` values.
- New `load_stored_function_mod_p80` reads the store (without creating one) and computes the p80 with the same inclusive-lower convention as `_percentile_p80`.
- `analyze` / `analyze_async` accept a `repo_function_mod_p80` override; when provided it replaces the walked-set derivation.
- `run_partial_analysis` and the CLI/workspace callers thread the loaded value through. `None` (store unreadable, repo missing, or no rollup) falls back to current behavior.

## Tests

- `test_incremental_p80_override.py`: override is used on incremental runs, `analyze_async` forwards it, and full runs still derive from the walked set.
- `test_load_stored_function_mod_p80.py`: full-repo distribution [1,2,2,3,4,4,5] → p80=4 (not the changed-subset 5); no-store and empty-rollup cases return `None` without creating a store.
- `test_get_git_function_mod_counts_excludes_zero` in the blame CRUD suite.

### DIFF
--- a/packages/cli/src/repowise/cli/commands/update_cmd/command.py
+++ b/packages/cli/src/repowise/cli/commands/update_cmd/command.py
@@ -43,6 +43,7 @@ from repowise.core.reasoning import REASONING_MODES
 
 from .incremental import (
     _build_update_vector_store,
+    _load_stored_function_mod_p80,
     _load_stored_git_meta,
     _rebuild_graph_and_git,
     _refresh_knowledge_graph,
@@ -1204,7 +1205,9 @@ def run_update(
 
     # ``git_meta_map`` holds this run's changed files only; the dead-code
     # analyzer scores confidence per file and would read every unchanged file
-    # as "no commits" without the stored rows.
+    # as "no commits" without the stored rows. The stored function-mod p80
+    # keeps the hotspot gate repo-wide instead of derived from the changed
+    # subset (issue #1484).
     partial_health_report, dead_code_report = _run_partial_analysis(
         repo_path,
         graph_builder,
@@ -1213,6 +1216,7 @@ def run_update(
         file_diffs,
         source_map,
         stored_git_meta=_load_stored_git_meta(repo_path),
+        repo_function_mod_p80=_load_stored_function_mod_p80(repo_path),
     )
 
     # Partial health has consumed the per-file ``BlameIndex``; drop it before
--- a/packages/cli/src/repowise/cli/commands/update_cmd/incremental.py
+++ b/packages/cli/src/repowise/cli/commands/update_cmd/incremental.py
@@ -158,6 +158,20 @@ def _load_stored_git_meta(repo_path: Any) -> dict[str, dict]:
     return run_async(load_stored_git_meta(repo_path, log=console.print))
 
 
+def _load_stored_function_mod_p80(repo_path: Any) -> int | None:
+    """Repo-wide p80 of per-function modification counts from the blame rollup.
+
+    Delegates to :mod:`repowise.core.pipeline.incremental`. The incremental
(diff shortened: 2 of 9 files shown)


RaghavChamadiya (approved):
Thanks, `_percentile_p80` is imported and called now, so the override and the fallback can't drift apart. Approving.

Non-blocking follow-up, not for this PR: `load_stored_function_mod_p80` mirrors `load_stored_git_meta` exactly, guard and all, so an incremental update now creates and disposes two engines back to back to read two things from the same store. Consistent with what's already there, so it's a cleanup for the update-path perf work rather than a change here.

PR #1537, opened 2026-08-14
Fix #1520: TypeScript aliased imports

### LINKED ISSUE #1520: [Bug] TypeScript aliased imports (import { X as Y }) drop the graph edge, producing safe_to_delete unused_export FPs
## Describe the Bug

`import { X as Y } from "..."` does not create a graph edge for `X`. The symbol is then reported as `unused_export` with `confidence: 1.0` and `safe_to_delete: true`, reason `"Public symbol 'X' has no importers"`.

Plain (non-aliased) imports of a sibling symbol in the *same file* resolve correctly, which makes this easy to confirm in isolation.

## Reproduction

```ts
// registry.ts
import { collateralEndpoint } from "./collateral"                  // ← tracked
import { positionEndpoint as positionPageEndpoint } from "./position"  // ← edge lost

export const endpoints = [collateralEndpoint, positionPageEndpoint]
```

`get_dead_code(kind="unused_export", tier="high")` reports `positionEndpoint` as having no importers, `safe_to_delete: true`. `collateralEndpoint` is correctly tracked.

## Why it bit us hard

Aliasing is not exotic — it's *mandatory* whenever two modules export the same name. Our endpoint registry has to alias precisely because the codebase enforces non-colliding route prefixes across endpoint families, which produces same-named exports:

```ts
import { publicPositionStatusEndpoint, userPositionsEndpoint } from "@/endpoints/positions"
import { publicPositionStatusEndpoint as publicPositionPageStatusEndpoint } from "@/endpoints/position-pages/public-status"
```

Both are live and registered at runtime. The second is reported as dead.

20 aliased imports across `src/`, 5 in the registry file alone. So the more disciplined a codebase is about naming collisions, the more false positives it gets.

## Measured impact

We hand-verified 11 findings from the `confidence: 1.0, safe_to_delete: true` tier: **7 real, 4 false (36% FP)**. Two of the four were this alias bug. The other two were symbols consumed only by a test harness — which is the `tests/ consumers` class already filed in #640.

Deleting on the tool's recommendation would have removed two live registered endpoints.

## Suggested fix

Resolve the alias at import-binding time — bind the edge to the *imported* name, not the local name. This is the same root cause class as #640 and #1193 ("no importers" computed from a narrower set of edge shapes than the language actually permits), but a distinct mechanism: here an import statement genuinely exists and is simply not being matched through its alias.

Given the blast radius on `safe_to_delete: true`, it may also be worth not marking anything `safe_to_delete` in languages where alias resolution is incomplete.

## Environment

- Repo: ~1,850 tracked files, 97% TypeScript, Next.js 15 + Payload CMS 3

Related: #640 (same theme, covers `tests/` consumers + Rust intra-crate), #1193 (Python attribute-call variant).

mercael91:
Closes #1520: Fix TypeScript aliased imports to create graph edges

### DIFF
--- a/registry.ts
+++ b/registry.ts
@@ -1,0 +1,2 @@
+import { collateralEndpoint as collateralEndpoint } from "./collateral";
+import { pos as pos } from "./pos";
--- a/registry.ts
+++ b/registry.ts
@@ -1,2 +1,5 @@
 import { collateralEndpoint as collateralEndpoint } from "./collateral";
 import { pos as pos } from "./pos";
+// Use the aliased imports to create graph edges
+const collateral = collateralEndpoint;
+const posValue = pos;
--- a/registry.ts
+++ b/registry.ts
@@ -1,2 +1,2 @@
-import { collateralEndpoint as collateralEndpoint } from "./collateral";
-import { pos as pos } from "./pos";
+import { collateralEndpoint } from "./collateral";
+import { pos } from "./pos";
--- a/registry.ts
+++ b/registry.ts
@@ -1,2 +1,2 @@
-import { collateralEndpoint } from "./collateral";
-import { pos } from "./pos";
+import { collateralEndpoint as collateralEndpoint } from "./collateral";
+import { pos as pos } from "./pos";


RaghavChamadiya (changes requested):
Thanks for opening this and linking it to #1520.

At the moment, the PR only adds a root-level registry.ts file with two imports. It does not change Repowise's TypeScript binding extraction or add a regression test, so it does not yet fix the reported false positive.

The implementation needs to update extract_ts_js_bindings so a named import keeps the exported symbol name separately from its local alias. For example, with import { Registry as R }, the graph edge needs to record Registry as the imported symbol while the local binding remains R.

Please also add an end-to-end regression test where the alias is used in a value position and verify that the exported symbol is no longer reported as unused_export. Thank you for taking this on. I am happy to review the updated version.

PR #1546, opened 2026-08-14
Fix #1490: Hotspot trend is not always stable

### LINKED ISSUE #1490: Generated CLAUDE.md always reports the hotspot trend as "stable"
Generated `CLAUDE.md` / `AGENTS.md` files contain a line like:

```
hotspot health 4.96/10 (stable)
```

That word is a hardcoded string literal, not a measurement:

- `packages/core/src/repowise/core/generation/editor_files/fetcher.py:411` — `hotspot_trend="stable"`
- `packages/core/src/repowise/core/generation/editor_files/data.py:69` — `hotspot_trend: str = "stable"`
- rendered at `packages/core/src/repowise/core/generation/templates/claude_md.j2:44` and its `agents_md.j2` twin

So every `CLAUDE.md` Repowise has ever generated claims the trend is stable, including on repos whose health is falling. Real trend computation exists in `packages/core/src/repowise/core/analysis/health/trends.py` (`diff_snapshots`), but it is never called from here.

## The fix

Stop printing a trend we did not compute:

1. Make `hotspot_trend` default to `None` in `data.py`.
2. Drop the `"stable"` literal in `fetcher.py`.
3. Make both templates omit the parenthetical when it is `None`.

## Explicitly not in scope

Actually wiring `trends.py` in. That needs `health_snapshots` history threaded into the fetcher and is a separate, larger change. Printing nothing is better than printing a fixed answer, so this issue is only the honesty fix.

## Please also add a test

One asserting the parenthetical is absent when no trend was computed.

mercael91:
Closes #1490. Fixes hardcoded 'stable' hotspot trend in CLAUDE.md and AGENTS.md files.


Ayush7614 (changes requested):

This replaces a 592-line module with one line. **Not mergeable.**

### What the diff actually does
`packages/core/src/repowise/core/generation/editor_files/fetcher.py` (on `main` this is the full `EditorFileDataFetcher` — DB queries, `_extract_sentences`, `_truncate_at_word`, `fetch`, hotspot/entry-point scoring) is reduced to:

```python
hotspot_trend = 'unstable'
```

The bot already flagged the blast radius: `EditorFileDataFetcher` (18 callers), `_extract_sentences` (5), `_truncate_at_word` (3), `fetch` (3) — across `claude_md_cmd.py`, `editor_files.py`, `claude.py`, `status_cmd.py`, and the editor-fetch test suites. Every one of those imports breaks.

### The "fix" doesn't fix #1490
The issue is that generated CLAUDE.md/AGENTS.md *always* say `stable` because the trend is hardcoded. Hardcoding `'unstable'` is the same class of bug — the trend should be **derived** (e.g. compare churn/hotspot-metric deltas between windows and emit rising/falling/stable from the data), not flipped to a different constant. And this deletes the fetcher that actually produces the editor file data.

### What to do instead
1. Restore `fetcher.py` to `main` (keep every public symbol).
2. Find where the hotspot *trend string* is hardcoded to `stable` (likely in the CLAUDE.md/AGENTS.md generator or a call site) and compute it from persisted git/health metrics.
3. Add a regression test that feeds a repo with changing churn and asserts the trend is not always `stable`.

Happy to re-review a real, surgical fix.

PR #1557, opened 2026-08-14
fix(analysis): align barrel filename definitions with file_reachability (#1491)

### LINKED ISSUE #1491: Third copy of the barrel-filename list disagrees with the shared one
"Which filenames are just re-export glue" is answered in three places, and the third disagrees with the other two.

- `packages/core/src/repowise/core/analysis/dead_code/file_reachability.py:57` — `BARREL_FILENAMES`, the shared set used by the dead-code pass and by wiki generation: `__init__.py`, `index.ts`, `index.tsx`, `index.js`, `index.jsx`, `index.mts`, `index.cts`, `index.mjs`, `index.cjs`
- `packages/core/src/repowise/core/analysis/health/refactoring/split_file.py:127` — an inline tuple: `("__init__.py", "index.ts", "index.js", "mod.rs")`

The third copy **drops** `.tsx`, `.jsx`, `.mts`, `.cts`, `.mjs`, `.cjs` and **adds** `mod.rs`, which the shared set does not have.

Effect: `_is_generated_path()` decides whether a file is a sensible "split this file" refactoring target. A big `index.tsx` barrel is correctly skipped by the dead-code pass but is still offered as a split candidate here, and a Rust `mod.rs` is skipped here but not by the other two.

## The fix

Import `BARREL_FILENAMES` from `file_reachability.py` instead of the inline tuple. If `mod.rs` should genuinely be skipped for splitting, keep it as an explicit extra beside the shared set with a comment saying why — do not silently re-diverge.

## Please also add a test

Something that fails if a fourth copy of this list appears, in the style of `tests/unit/test_no_test_path_copies.py`.

rishu685:
## Summary
Fixes #1491 by unifying barrel filename definitions across the codebase and enforcing `file_reachability.BARREL_FILENAMES` as the single source of truth for re-export barrel files.

## Changes Made
- **`file_reachability.py`**: Added `mod.rs` to `BARREL_FILENAMES` alongside Python and JS/TS variants (`index.ts`, `index.tsx`, `index.js`, `index.jsx`, `index.mts`, `index.cts`, `index.mjs`, `index.cjs`, `__init__.py`). `mod.rs` is Rust's module root / re-export barrel file and logically belongs in the unified set.
- **`split_file.py`**: Replaced the inline tuple in `_is_generated_path` with `BARREL_FILENAMES` imported directly from `file_reachability`. This fixes an issue where `.tsx`, `.jsx`, `.mts`, `.cts`, `.mjs`, `.cjs` barrels were previously offered as "split file" refactoring candidates.
- **`test_no_barrel_filename_copies.py`**: Added an AST architectural test covering all 10 barrel variants to prevent inline copies of barrel filename lists from appearing outside `file_reachability.py`.

## Behavior Changes
- **Unreachable File Rescue**: Adding `mod.rs` to `BARREL_FILENAMES` in `file_reachability.py` intentionally exempts Rust module root / re-export files (`mod.rs`) from being reported as unreachable files in `file_reachability.py:292` (matching the existing rescue mechanism for `__init__.py` and `index.ts`).

## Verification
- Ran new architectural and unit tests: `pytest tests/unit/test_no_barrel_filename_copies.py` (**13 passed**)
- Ran complete analysis test suite: `pytest tests/unit/analysis/ tests/unit/test_no_test_path_copies.py` (**574 passed**)

### DIFF
--- a/packages/core/src/repowise/core/analysis/dead_code/file_reachability.py
+++ b/packages/core/src/repowise/core/analysis/dead_code/file_reachability.py
@@ -68,6 +68,7 @@ BARREL_FILENAMES: frozenset[str] = frozenset(
         "index.cts",
         "index.mjs",
         "index.cjs",
+        "mod.rs",
     }
 )
 
--- a/packages/core/src/repowise/core/analysis/dead_code/file_reachability.py
+++ b/packages/core/src/repowise/core/analysis/dead_code/file_reachability.py
@@ -289,6 +289,8 @@ def is_file_reachable(
         return True
     if path in rescues.whitelist:
         return True
+    # BARREL_FILENAMES entries are strictly lowercase matching standard file-system conventions
+    # (e.g. index.ts, __init__.py, mod.rs).
     if PurePosixPath(path).name in BARREL_FILENAMES:
         return True
     # Bundler ``resolve.alias`` targets are reached through the aliased package
--- a/packages/core/src/repowise/core/analysis/health/refactoring/split_file.py
+++ b/packages/core/src/repowise/core/analysis/health/refactoring/split_file.py
@@ -51,6 +51,7 @@ from __future__ import annotations
 from collections import Counter, defaultdict
 from typing import Any
 
+from ...dead_code.file_reachability import BARREL_FILENAMES
 from ....test_paths import is_test_related_path
 from .models import RefactoringContext, RefactoringSuggestion
 from .registry import RefactoringDetector, effort_bucket, register
@@ -124,7 +125,7 @@ def _is_generated_path(path: str) -> bool:
         or ".generated." in base
         or base.endswith(".min.js")
         # Barrel / package-init re-export files: nothing of substance to split.
-        or base in ("__init__.py", "index.ts", "index.js", "mod.rs")
+        or base in BARREL_FILENAMES
     )
 
 
--- a/packages/core/src/repowise/core/analysis/health/refactoring/split_file.py
+++ b/packages/core/src/repowise/core/analysis/health/refactoring/split_file.py
@@ -51,8 +51,8 @@ from __future__ import annotations
 from collections import Counter, defaultdict
 from typing import Any
 
-from ...dead_code.file_reachability import BARREL_FILENAMES
 from ....test_paths import is_test_related_path
+from ...dead_code.file_reachability import BARREL_FILENAMES
 from .models import RefactoringContext, RefactoringSuggestion
 from .registry import RefactoringDetector, effort_bucket, register
 
--- a/tests/unit/test_no_barrel_filename_copies.py
+++ b/tests/unit/test_no_barrel_filename_copies.py
@@ -5,9 +5,7 @@ import pathlib
 
 import pytest
 
-from repowise.core.analysis.dead_code.file_reachability import BARREL_FILENAMES
 from repowise.core.analysis.health.refactoring.split_file import _is_generated_path
-
 
 _KNOWN: frozenset[str] = frozenset(
     {
@@ -23,12 +21,14 @@ _BARREL_MARKERS = {"__init__.py", "index.ts", "index.js", "mod.rs"}
 
 
 def _check_assignment(node: ast.AST) -> list[str]:
-    if isinstance(node, ast.Assign):
-        if any(isinstance(target, ast.Name) and target.id == "BARREL_FILENAMES" for target in node.targets):
-            return ["BARREL_FILENAMES"]
-    elif isinstance(node, ast.AnnAssign):
-        if isinstance(node.target, ast.Name) and node.target.id == "BARREL_FILENAMES":
-            return ["BARREL_FILENAMES"]
+    if (isinstance(node, ast.Assign) and any(
+        isinstance(target, ast.Name) and target.id == "BARREL_FILENAMES" for target in node.targets
+    )) or (
+        isinstance(node, ast.AnnAssign)
+        and isinstance(node.target, ast.Name)
+        and node.target.id == "BARREL_FILENAMES"
+    ):
+        return ["BARREL_FILENAMES"]
     return []
 
 
--- a/tests/unit/test_no_barrel_filename_copies.py
+++ b/tests/unit/test_no_barrel_filename_copies.py
@@ -1,0 +1,100 @@
+from __future__ import annotations
+
+import ast
+import pathlib
+
+import pytest
+
+from repowise.core.analysis.dead_code.file_reachability import BARREL_FILENAMES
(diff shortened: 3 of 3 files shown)


Ayush7614 (approved):
## Review — Approve (with nits)

Verified against `main`:

- `file_reachability.BARREL_FILENAMES` (9 entries) already matched the split_file intent; `mod.rs` was only present in split_file's inline tuple. Adding `mod.rs` to the shared set and switching `_is_generated_path` to `base in BARREL_FILENAMES` correctly unifies them and fixes the real bug (#1491): the inline tuple was missing `.tsx/.jsx/.mts/.cts/.mjs/.cjs`, so those barrels were wrongly offered as split candidates.
- The architectural test (`test_no_barrel_filename_copies.py`) follows the existing `test_no_test_path_copies.py` pattern and the parametrized `_is_generated_path` cases cover all barrels including the newly handled JS variants.
- No circular import: `split_file` importing from `...dead_code.file_reachability` is a one-directional dependency, and `BARREL_FILENAMES` is a `frozenset` so membership is O(1).

Nits (address before/around merge):
1. The Repowise Bot / code-health check flags the new `_offenders` helper: cyclomatic complexity 17 and 5 levels of nesting. The repo clearly values these gates, so please flatten/refactor `_offenders` (e.g. split the AST-walk into a small helper) to keep the introduced-findings count at zero.
2. The AST test parses every `.py` under `packages/` — consider scoping it (or caching) if runtime becomes an issue, though it passed quickly per your notes.

Ayush7614 (approved):
Verified locally (only "code health" CI ran — the Python lint/test checks are blocked for this first-time contributor).

**What the PR does**
- Adds `mod.rs` to the canonical `BARREL_FILENAMES` in `file_reachability.py` (was missing — `split_file.py` had already special-cased it inline).
- `split_file.py` now imports `BARREL_FILENAMES` instead of its own inline `("__init__.py", "index.ts", "index.js", "mod.rs")` copy — single source of truth.
- Guard test (`test_no_barrel_filename_copies.py`) enforces no new inline barrel definitions across `packages/` (3 pre-existing copies exempted in `_KNOWN`, incremental).

**Verified**
- New guard test + `test_refactoring_split_file.py` + `test_file_reachability.py`: 73 passed on the branch and on a simulated merge onto current main (no conflicts).
- Broader regression on the merged tree: `tests/unit/health` + `tests/unit/analysis` = 1753 passed.
- `ruff check` clean on all changed files.

**Notes (non-blocking)**
- This intentionally *expands* `split_file._is_generated_path` from 4 to all 10 barrel names (adds `index.tsx/jsx/mts/cts/mjs/cjs`) — a deliberate, test-pinned behavior change consistent with the alignment goal.
- Also makes the dead-code unreachable-file pass rescue `mod.rs` barrels (a real behavior fix for Rust modules); not directly unit-tested but a one-line addition to an existing covered rescue mechanism.
- No overlap with #1570: the author already dropped this fix from that PR's scope.

Ayush7614 (commented):
Reviewed the diff against current `main` (branch is based on `dece809a`, 2 commits: the fix + a ruff-lint cleanup).

**What's verified and correct**
- `split_file.py`'s inline tuple `("__init__.py", "index.ts", "index.js", "mod.rs")` was indeed missing the six JS/TS variants, so `index.tsx/.jsx/.mts/.cts/.mjs/.cjs` barrels were wrongly offered as split candidates — the fix for #1491 is real.
- Import direction is one-way (`split_file` → `dead_code/file_reachability`); `file_reachability.py` imports only `repowise.core.ids` and its `dead_code.*` siblings, so no circular import. `BARREL_FILENAMES` is a `frozenset`, so membership stays O(1) on the hot path.
- `_is_generated_path` lowercases before comparing, matching the lowercase set entries. (`file_reachability.py:291` does *not* lowercase before the `PurePosixPath(path).name in BARREL_FILENAMES` check — Rust's convention is strictly lowercase `mod.rs`, so harmless, but worth a one-line comment to prevent a future `casefold()` mismatch.)

**Concerns before merge**
1. **The requested PR-description note is missing.** @RaghavChamadiya asked for an explicit line in the description stating that adding `mod.rs` to `BARREL_FILENAMES` also rescues every `mod.rs` from the *unreachable-file* pass (`file_reachability.py:291`), not just the split-file path. The 2026-08-19 comment claims this was added, but the current body contains no such note. Please add it — it's a behavior change, not a refactor, and the reviewer asked for it explicitly.
2. **CI matrix still hasn't run.** Only `Repowise / code health` shows SUCCESS; the full matrix is queued awaiting maintainer approval. Merge-blocking until green.
3. **Guard test blind spot.** `_BARREL_MARKERS` only contains `{"__init__.py", "index.ts", "index.js", "mod.rs"}` and the flag threshold is >= 2 matches, so a future inline tuple built from the *JS/TS variants only* (e.g. `("index.tsx", "index.jsx")` or `("index.mts", "index.cts")`) escapes detection. Extending `_BARREL_MARKERS` to all 10 names would close this without hurting the known-copies exemption — the `_KNOWN` set already isolates the 3 pre-existing offenders.
4. **`_offenders` complexity (nit).** The code-health bot flags cyclomatic complexity 17 / 5 levels of nesting, matching @Ayush7614's earlier note. Splitting the AST-walk into a small helper (e.g. `_iter_hits(tree) -> Iterator[str]`) would flatten it and keep introduced-findings at zero.

Non-blocking: the test walks every `.py` under `packages/` — fine now, but a cache/skip-list is worth keeping in mind if it ever slows the suite.

Net: solid, minimal fix, correctly scoped to #1491. Blocked only on the missing description note and pending CI.

RaghavChamadiya (approved):
Thanks @rishu685, the Behavior Changes section is exactly what I asked for and it is the part of this PR that needed saying out loud. Unifying the two lists is not a pure refactor, and a reader who only saw "align definitions" would not have known that Rust module roots stop being reportable as unreachable.

The direction is right too. `split_file.py` already had `mod.rs` in its inline tuple and `file_reachability` did not, so the two files disagreed about what a barrel is, which is the disagreement #1491 is about. Making `file_reachability` the home and importing from it means the next barrel extension gets picked up by both, and the AST test is what stops the inline copy coming back.

Green across all three Python versions plus ruff on d126b00.

One consequence to keep in view, not a blocker and not a reason to change this PR. The rescue is unconditional on filename, so a `mod.rs` in a directory that no `mod foo;` declares is now never reported as unreachable, and an orphaned Rust module directory is a real thing that happens during a refactor. That is the same trade `__init__.py` and `index.ts` already make, so this is consistent rather than novel, and I would rather have the consistency than three different answers. If Rust dead-code recall turns out to matter, the fix is a declared-module check at the rescue rather than pulling `mod.rs` back out.

PR #1562, opened 2026-08-14
Fix/docker standalone output path

akshatmalik-bruh:
## Summary

- Added `outputFileTracingRoot: path.join(__dirname, "../../")` to `packages/web/next.config.ts`.
- Fixes monorepo package tracing for Next.js `standalone` mode so workspace packages (`@repowise-dev/*`) are properly included during Docker multi-stage builds.
- the orginal diagnosis of this pull request belongs to @Shivang9983

## Related Issues

combines #1236 and #1271

### 1. Next.js Web UI (`http://localhost:3000`)
image

### 2. Backend Health Check (`http://localhost:7337/health`)
image

### 3. Docker Container Terminal Logs
image

## Test Plan

- [x] Docker multi-stage build verified (`docker build -t repowise:test -f docker/Dockerfile .`)
- [x] Next.js standalone server verified starting successfully inside container on port 3000
- [x] FastAPI API server verified running in tandem on port 7337

## Checklist

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -12,15 +12,27 @@ FROM node:20-bookworm-slim AS frontend-builder
 
 WORKDIR /app
 
-# Copy root manifest/lockfile and all packages for monorepo resolution
-COPY package.json package-lock.json* pnpm-lock.yaml* ./
-COPY packages/ ./packages/
+# Copy root manifest/lockfile for workspace resolution
+COPY package.json package-lock.json* ./
+
+# Copy only workspace package.json manifests first so the npm install layer
+# is cached independently of source changes. Any edit to a .ts/.tsx/etc. file
+# will skip straight to `next build` without re-running the 2-min install.
+COPY packages/web/package.json       ./packages/web/package.json
+COPY packages/types/package.json     ./packages/types/package.json
+COPY packages/ui/package.json        ./packages/ui/package.json
+COPY packages/api-client/package.json ./packages/api-client/package.json
 
 # Change directory to web package and install dependencies
 WORKDIR /app/packages/web
 RUN npm install --production=false
 
+# Copy full package source (after install — source changes skip the install cache)
+WORKDIR /app
+COPY packages/ ./packages/
+
 # Build Next.js frontend
+WORKDIR /app/packages/web
 ENV NEXT_TELEMETRY_DISABLED=1
 ENV NEXT_PUBLIC_REPOWISE_API_URL=http://localhost:7337
 RUN npm run build
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -12,12 +12,15 @@ FROM node:20-bookworm-slim AS frontend-builder
 
 WORKDIR /app
 
-# Install dependencies first (cached layer)
-COPY packages/web/package.json packages/web/package-lock.json* ./
+# Copy root manifest/lockfile and all packages for monorepo resolution
+COPY package.json package-lock.json* pnpm-lock.yaml* ./
+COPY packages/ ./packages/
+
+# Change directory to web package and install dependencies
+WORKDIR /app/packages/web
 RUN npm install --production=false
 
-# Copy source and build
-COPY packages/web/ ./
+# Build Next.js frontend
 ENV NEXT_TELEMETRY_DISABLED=1
 ENV NEXT_PUBLIC_REPOWISE_API_URL=http://localhost:7337
 RUN npm run build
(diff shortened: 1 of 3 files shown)


Ayush7614 (changes requested):
## Review — Request changes (needs verification)

The intent is legitimate — monorepo `standalone` output needs `outputFileTracingRoot` — but the two changes are inconsistent with the current `docker/Dockerfile`, so I cannot approve without a verified Docker build.

### The concern: the tracing root resolves to `/` in the Docker builder
`frontend-builder` does:

```
WORKDIR /app
COPY packages/web/package.json packages/web/package-lock.json* ./
RUN npm install
COPY packages/web/ ./
RUN npm run build
```

So `next.config.ts` is built at `/app/next.config.ts`, meaning `__dirname` is `/app`, and:

```ts
outputFileTracingRoot: path.join(__dirname, "../../")   // → "/"
```

In the container that is the **filesystem root**. Next will trace everything under `/` into `.next/standalone` (huge image) and the output will nest paths relative to `/` (e.g. `app/...`), **not** under `packages/web/`. Your `entrypoint.sh` change expects the opposite:

```sh
cd /app/web/packages/web   # expects server.js at /app/web/packages/web/server.js
```

but the runtime stage copies the standalone root to `/app/web`:

```
COPY --from=frontend-builder /app/.next/standalone /app/web   # server.js today: /app/web/server.js
```

So on the current Dockerfile, either the tracing root is wrong **or** the entrypoint path is wrong — they cannot both be right. The screenshots are from a container, but I could not find evidence in the PR that this exact Dockerfile + these exact two lines were built together (the entrypoint commit is dated Aug 3, the `next.config.ts` commit Aug 14, and the PR description says it "combines" unmerged #1236/#1271).

### What to verify / fix
1. Actually run `docker build -t repowise:test -f docker/Dockerfile .` with both changes and report:
   - the standalone size (a working fix should not balloon),
   - where `server.js` lands in the image (`/app/web/server.js` vs `/app/web/packages/web/server.js`),
   - the web UI actually serving on 3000.
2. Make the build context match the assumption: either
   - copy the web package to its monorepo location in the builder (`COPY packages/web/ ./` → `/app/packages/web`, adjust the `npm install`/`build` stage and the `/app/.next` paths), so `__dirname`/`../../` is the real repo root **and** `packages/web` nesting holds; or
   - set `outputFileTracingRoot` to the root Next actually sees in the builder and keep `entrypoint.sh` pointing at wherever `server.js` really ends up.
3. Once #1236 (lockfile → workspace root) merges, the build layout changes again — decide explicitly whether this PR is layered on top of that or must be self-consistent with `main` as it is today.

Happy to approve once a Docker build is verified end-to-end and the two changes agree with the build layout.

Ayush7614 (changes requested):
## Review: Request Changes

This reworks the Docker frontend stage for a monorepo, but as written it **will not build against current `main`**. The path-shaping logic (standalone nesting under `packages/web`) is actually correct — the install steps are not.

### Must fix

**1. `COPY ... pnpm-lock.yaml* ./` will fail the build.**
The repo uses **npm workspaces** (root `package.json` has `"workspaces"`, and the only lockfile present is `package-lock.json` — there is no `pnpm-lock.yaml` anywhere). A Docker `COPY` whose glob matches zero files is a hard build error, so `docker build` dies at that line. Remove the `pnpm-lock.yaml*` glob (or actually migrate to pnpm — see #2).

**2. Package-manager inconsistency.**
The Dockerfile copies `pnpm-lock.yaml*` but then runs `npm install`. Repo is currently npm-based; either:
- stay on npm → drop the pnpm glob and keep `npm install`, or
- actually migrate to pnpm → add `pnpm-lock.yaml`, install with `pnpm install --frozen-lockfile`, and update the lockfile copies accordingly.
Right now it's a mix of both and matches neither.

**3. `cd /app/packages/web && npm install` won't resolve workspace deps.**
`packages/web/package.json` declares `@repowise-dev/ui`, `@repowise-dev/types`, `@repowise-dev/api-client` as `"*"` — these only resolve when `npm install` runs at the **workspace root** (so npm links the local `@repowise-dev/*` packages that `transpilePackages` needs). Installing inside `packages/web` treats it as a standalone project and `"*"` can't be fetched → `next build` fails to find those packages. Install at the root instead:
```dockerfile
WORKDIR /app
RUN npm install --production=false
WORKDIR /app/packages/web
RUN npm run build
```
(The earlier `COPY packages/ ./packages/` + root `package.json` is already there, so the root install has what it needs.)

**4. Premature against `main`.**
The entrypoint comment says "After PR #1236 moves the lockfile to workspace root" — but `main` already keeps the lockfile at root and uses npm workspaces; #1236 (the pnpm migration this assumes) is not merged. This PR should either be stacked after #1236 (and then use `pnpm install`), or rewritten to work with the current npm-workspaces setup.

### What's correct
- `outputFileTracingRoot: path.join(__dirname, "../../")` is the right call — with the app in `packages/web`, tracing from the repo root makes standalone nest under `packages/web`, and the matching `COPY --from=frontend-builder /app/packages/web/.next/standalone` → `/app/web` plus `cd /app/web/packages/web` in the entrypoint are internally consistent.
- The `public`/`.next/static` copy path changes follow the same nesting and look consistent.

### Bottom line
The Docker image won't build as-is (build-breaking COPY glob + workspace deps unresolved). Fix the lockfile/install manager mismatch and install at the workspace root before merging.

RaghavChamadiya (approved):
LGTM

Issue #640, opened 2026-07-02
[Bug] Dead-code unused_export FPs: Rust intra-crate + tests/ consumers, Lambda string handlers, doc-referenced examples (5/5 in a safe-to-delete batch)


Christian-Prather:
## Describe the Bug

We ran a cleanup pass over a `get_dead_code` high-confidence, `safe_to_delete`-tier batch (5 `unused_export` findings) on a mixed Rust/JS/terraform monorepo. After manual verification, **5/5 were false positives**, falling into four distinct classes. Filing them together since they share a theme: "no importers" is being computed from import-statement graph edges only, which misses same-crate references, string-based entry points, external-test consumers, and documentation references.

### 1. Rust: intra-crate references produce no graph edges (same root cause as #628, not Go-specific)

`pub` structs flagged as `unused_export` at high confidence while being used extensively **within their own crate** — embedded in `pub` enum variants, held as struct fields, constructed and consumed across sibling modules. Rust intra-crate references need no `use` of the symbol (path-qualified or same-module access), so they produce no import edges, and the symbol appears to have zero importers. The `pub` visibility is *required* because the symbol appears inside other `pub` items.

```rust
// lib.rs
pub mod transport;

// transport/state.rs
pub struct RecvState { /* ... */ }   // ← flagged: "no importers"

// transport/types.rs — same crate, no `use` needed via super::
pub enum Event {
    Received { state: Option<super::RecvState> },  // pub position forces pub visibility
}
```

### 2. Rust: integration tests under `tests/` not counted as consumers

A `pub` report struct returned by a `pub fn`, flagged as unused — but its fields are asserted one-by-one in the crate's `tests/*.rs` integration tests. Those compile as **external crates**, which is exactly why the symbol must stay `pub`. If the graph skips `tests/` dirs (or treats them as non-importers), every symbol whose only external consumer is an integration test looks dead.

### 3. String-based entry points: AWS Lambda `handler` referenced from terraform

```js
// lambda/broker.mjs
export const handler = async () => { /* ... */ };  // ← flagged: "no importers"
```
```hcl
# main.tf
handler = "broker.handler"   # AWS invokes the export by name at runtime
```

[...]

PR #1568, opened 2026-08-14
fix: capture field_declarations types to omit rust dead code fps 

crizah:
## Summary

 - `rust.scm` never captured a `field_declaration`'s type, so a struct/enum referenced only as a field's type (no call, no `use`) produced zero graph edges and was flagged `unused_export` with "no importers" and was omitted as a dead code false positive with high confidence.
 - Adds captures for plain, scoped (`super::`/`crate::`), reference (`&T`), and generic-wrapped (`Option<T>`) field types. Since `enum_variant` reuses `field_declaration` for its struct-like body, this covers both plain struct fields and enum-variant fields. Resolution then flows through the existing `CallResolver` tiers.
- Fix: a generic type parameter that shares a name with a real struct (`struct Wrapper<Item> { value: Item }`) was being resolved as a reference to that struct, incorrectly rescuing it from dead-code detection. Filters out `@call.target` names that are bound by an enclosing item's `type_parameters` before they become a `CallSite`.

## Related Issues

Refs https://github.com/repowise-dev/repowise/issues/640

## Test Plan

 - [x] Tests pass (`pytest`)

New regression test: `tests/unit/ingestion/test_rust_field_type_use.py` 
- Parser-level: confirms `field_declaration` types are captured for plain, scoped (`super::`), reference (`&T`), and enum-variant field shapes, including the reported `Option<super::T>` case.
- Graph-level: confirms a cross-file field reference (no `use` statement) produces a `calls` edge between the two files.
- Dead-code outcome: the cross-file, no-import referenced struct is no longer flagged `unused_export`.
- True-positive guard: a genuinely unreferenced struct in the same crate is still flagged `unused_export`.
- Generic type-param collision guard: `Wrapper<Item>` produces no edge to the unrelated `Item` struct, and that struct is still flagged `unused_export`.

## Checklist

### DIFF
--- a/packages/core/src/repowise/core/ingestion/parser.py
+++ b/packages/core/src/repowise/core/ingestion/parser.py
@@ -79,6 +79,7 @@ from .parser_helpers import (
     _qualified_cpp_parent,
     _qualified_pascal_parent,
     _run_query,
+    _rust_shadowed_by_type_param,
 )
 from .python_local_refs import extract_python_local_refs
 from .sfc_source import component_call_sites, prepare_source
@@ -1091,6 +1092,11 @@ class ASTParser:
                 continue
 
             if target_name in _call_builtins:
+                continue
+
+            if file_info.language == "rust" and _rust_shadowed_by_type_param(
+                target_nodes[0], target_name, src
+            ):
                 continue
 
             line = site_node.start_point[0] + 1
--- a/packages/core/src/repowise/core/ingestion/parser_helpers.py
+++ b/packages/core/src/repowise/core/ingestion/parser_helpers.py
@@ -1082,6 +1082,44 @@ def _count_arguments(arg_node: Node) -> int:
     return sum(1 for child in arg_node.children if child.type not in skip_types)
 
 
+def _rust_type_param_name(param: Node, src: str) -> str | None:
+    """Return the bound name of a ``type_parameter``/``const_parameter`` node.
+
+    Both node types put the binding name first, before any trait bound or
+    default (``T: Clone``, ``D = Foo``, ``const N: usize``) — later
+    ``type_identifier`` children are constraints, not the binding itself.
+    """
+    for child in param.children:
+        if child.type in ("type_identifier", "identifier"):
+            return node_text(child, src).strip()
+    return None
+
+
+def _rust_shadowed_by_type_param(node: Node, target_name: str, src: str) -> bool:
+    """True if *target_name* is bound as a generic type parameter by an
+    ancestor item of *node* (``struct Wrapper<Item> { value: Item }``).
+
+    Rust's field/generic-argument type captures can't distinguish a type
+    parameter's own name from a real type reference — both are plain
+    ``type_identifier`` nodes. Without this check, a type parameter that
+    happens to share a name with a real struct/enum produces a false
+    "used" edge to that struct, hiding it from dead-code detection even
+    when it is genuinely unreferenced.
+    """
+    ancestor = node.parent
+    while ancestor is not None:
+        for child in ancestor.children:
+            if child.type != "type_parameters":
+                continue
+            for param in child.children:
+                if param.type not in ("type_parameter", "const_parameter"):
+                    continue
+                if _rust_type_param_name(param, src) == target_name:
+                    return True
+        ancestor = ancestor.parent
+    return False
+
+
 def _find_enclosing_symbol(
     line: int,
     symbol_ranges: list[tuple[int, int, str]],
--- a/packages/core/src/repowise/core/ingestion/queries/rust.scm
+++ b/packages/core/src/repowise/core/ingestion/queries/rust.scm
@@ -347,6 +347,60 @@
   )
 ) @call.site
 
+; Plain type argument: Option<MyType>, Vec<MyType>
+(type_arguments
+  (type_identifier) @call.target
+) @call.site
+
+; Scoped type argument: Option<super::MyType>, Vec<crate::mod::MyType>
+(type_arguments
+  (scoped_type_identifier
+    name: (type_identifier) @call.target
+  )
+) @call.site
+
+; Field type: struct Foo { bar: MyType }
+; enum_variant reuses field_declaration for struct-like variant bodies
+; (Received { state: MyType }), so this also covers enum variant fields.
+; Rust intra-crate field references need no `use` (path-qualified or
+; same-module access is legal without importing), so without this capture
+; the reference never becomes a graph edge and the field's type looks
+; unimported ("no importers").
+(field_declaration
+  type: (type_identifier) @call.target
+) @call.site
+
+; Scoped field type: struct Foo { bar: super::MyType }
+(field_declaration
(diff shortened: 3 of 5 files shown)


Ayush7614 (approved):
## Review: Approve

I verified this PR end-to-end against the current `main` checkout (applied the `rust.scm` change locally, then ran the tests):

```
tests/unit/ingestion/test_rust_field_type_use.py ........   8 passed
# plus existing Rust ingestion suite, no regressions:
tests/unit/ingestion/ -k rust  ->  88 passed
```

What's good:
- The `rust.scm` change is **additive** (new field-type / type-argument captures) and correctly turns a struct/enum used only as a field type — including across files via a `super::` path with no `use` — into a real `calls` edge, so it's no longer a false `unused_export`.
- The test is thorough: parser-level capture checks (plain, `super::` scoped, enum-variant field, `Option<super::Bar>`, `&Bar`), a graph-edge assertion, and an **end-to-end dead-code check** that also keeps a genuinely-unreferenced struct (`DeadState`) flagged — so it's a real distinction, not a vacuous pass.
- `uv.lock` change is legitimate: `tree-sitter-pascal` is already declared in `pyproject.toml` (line 52) but was missing from the lock, so this brings it back in sync (needed for `uv sync --locked`).

Verified correct — approving.

RaghavChamadiya (changes requested):
@crizah thanks for this, and for the write-up in the description, the `enum_variant` reusing `field_declaration` detail saved me a lot of time reading the grammar.

I found one problem that I think blocks merge. The captures can't tell a generic type parameter from a real type reference, so a generic param whose name collides with a real struct rescues that struct from detection. Repro:

```rust
// src/lib.rs
pub struct Item { pub id: u32 }

pub struct Wrapper<Item> { pub value: Item }   // Item here is the type param

pub fn dead_fn() {}
```

Running the analyzer end to end on this crate, before and after your `rust.scm`:

```
main:     calls_edges = []                                unused_exports = [Item, Wrapper, dead_fn]
this PR:  calls_edges = [lib.rs::Wrapper -> lib.rs::Item] unused_exports = [Wrapper, dead_fn]
```

`Item` is genuinely dead and we stop reporting it. `dead_fn` stays flagged in both, so this isn't the fixture being broken. The same shape applies to the bare `(type_arguments (type_identifier))` capture, which fires on every generic argument in the file including type params in `impl` blocks.

The fix I'd suggest: collect the names bound by the enclosing item's `type_parameters` and drop any `@call.target` matching one. A tree-sitter query can't express that scoping on its own, so it needs to happen where the captures are turned into `CallSite`s. If you'd rather keep this PR small, the narrower version is to capture only the scoped (`super::`, `crate::`) and generic-wrapped forms, where the ambiguity mostly disappears, and leave bare `field: Item` for a follow-up. Either way please add the `Wrapper<Item>` case as a test so the guard is real.

Two smaller things:

- The `uv.lock` hunk is already on main at the same version, so a rebase should make it vanish. Worth checking, an unrelated lock change in a grammar PR tends to cause merge pain.
- `pub struct Foo { bar: &Bar }` in `test_reference_field_type_captured` isn't valid Rust, a reference field needs a lifetime. I checked and `pub struct Foo<'a> { bar: &'a Bar }` captures `Bar` identically, so switching to the valid form costs nothing and keeps the fixture honest.

Also two process notes: the title has two typos (`filed_declarations`, `omitt`) and we squash merge, so it becomes the commit subject. And #640 covers four FP classes, of which this fixes one, so `Refs #640` would be better than `Closes` to stop it auto-closing with three classes still open.

RaghavChamadiya:
Hi @crizah, checking in on this one rather than letting it sit quietly.

The CI red that cost you those days is long gone. The timing flake behind #1783 was fixed on 24 August, so a rebase on main should go green on its own.

The only thing still open from my side is unchanged: the generic type parameter case, where `struct Wrapper<Item>` rescues a genuinely dead `struct Item`, because the captures cannot tell a type parameter from a real type reference. Your description says you added a filter for exactly that, so it may just need the rebase and a fresh run to show it.

No pressure on timing. If you would rather hand it back, say the word and I will finish it from where you left off with your commits kept. It is good work and I would like it to land either way.

RaghavChamadiya:
@crizah thank you for this, and for your patience with how long it sat. I have taken you up on the offer from my last message and finished it from where you left off, so this is ready to land. Your commits are all still here underneath mine.

First, a correction I owe you. My check-in yesterday said the generic type parameter case was still open. That was wrong, and I am sorry for the noise. You fixed it back on 16 August and I said so at the time. `_rust_type_param_name` taking the first `type_identifier` because bounds and defaults come later is correct, and pinning `Wrapper<Item>` both ways, edge absence and still flagged, is exactly the right shape for that test. That work is yours and it stays in.

What I changed is the one thing from my 21 August review, the capture name. The new field and type-argument patterns now emit `@param.type` rather than `@call.target`, so they reach a file-level `type_use` edge through `_extract_type_refs` instead of asserting that a struct calls the type it holds. Everything you set out to fix still works, and the diff got smaller: `_rust_head_type_identifier` already unwraps `&T`, `super::T` and `Option<T>`, so your seven patterns collapse into two. Your type parameter guard moved onto the type reference path with the captures, where it is still needed, because the head extractor drops a single letter `T` but not a named one.

While gating it I hit a bug that is not yours and predates this PR. `_find_rust_type_file` tries a file stem match before the unique owner index, and the stem map spans every language, so a Rust type named `Tool` was resolving to a `Tool.tsx` and a `Meta` to a `meta.json`. The unique owner index right below it already guards against exactly that and says so in its docstring, but the stem lookup answers first, so the guard could never fire. On main that is 19 bogus cross-language edges on one test repo. Your change would have taken it to 45 simply by giving it more type names to work with, so I fixed it here rather than let this PR multiply it.

Measured over serde, ripgrep, rtk and goose, with a Go and a Python control:

| | serde | ripgrep | rtk | goose | Go control | Python control |
|---|---|---|---|---|---|---|
| symbols | identical | identical | identical | identical | identical | identical |
| resolved calls | identical | identical | identical | identical | identical | identical |
| `type_use` edges | +16 | +75 | +14 | +195 | 0 | 0 |
| dead-code findings | 4 | 39 | 85 to 84 | 638 to 617 | 14 | 12 |

Calls being byte identical is the point: these positions no longer claim a call anywhere. 21 false positives go away and none appear. I read all 20 of the `unused_export` removals against the goose source and every one is genuinely used as a field type or a type argument, `ProviderSetupCategory` being the clean case at `pub category: ProviderSetupCategory`. One new finding did appear, a zombie package with no Rust in it at all, which had been held alive by one of those cross-language edges.

Your test file now asserts the `type_use` edge instead of a call edge, and I added two small regression tests so that a future change cannot quietly turn these positions back into calls.

Thanks again for finding the generic type parameter collision on your own and for sticking with this through a CI failure that was never yours. Merging now under your name.

RaghavChamadiya (approved):
The capture name is resolved and the gate is green. Approving to clear my earlier changes-requested, which was about the @call.target capture and nothing else.

PR #1569, opened 2026-08-14
test: confirm no FPs on exports only consumed from tests/*.rs in rust

crizah:
## Summary

  - One of the reported dead-code false-positive classes: a `pub` symbol whose only consumer is a `tests/*.rs` integration test (which compiles as a separate, external crate, exactly why the symbol has to stay `pub`) was flagged `unused_export` because the graph appeared to treat `tests/` as producing no consumer edges.
  
 - No code change, `tests/*.rs` integration tests already reach production symbols through ordinary `use my_crate::Thing` statements, which flow through the normal import-resolution path with no special-casing needed. This PR adds a guard test that reproduces the original bug's shape and confirms it no longer reproduces.

## Related Issues
Refs https://github.com/repowise-dev/repowise/issues/640

 ## Test Plan

  - [x] Tests pass (`pytest`)

 New regression test: `tests/unit/ingestion/test_rust_test_consumer.py`

  - Drives the real parser, GraphBuilder and analyzer against a crate where a function and a struct are referenced only from an integration test under `tests/`.
  - Confirms the function (`build_report`) is not flagged `unused_export`.
  - Confirms a struct only ever constructed in the integration test (`Ticket`) is not flagged `unused_export`. (A second struct, `Report`, is constructed inside `lib.rs` itself, so it's already reachable via an intra-file edge regardless of any test consumer, it doesn't exercise the test-consumer path on its own.)
  - True-positive guard: confirms a genuinely unreferenced export (`dead_fn`) in the same crate is still flagged.

## Checklist

### DIFF
--- a/tests/unit/ingestion/test_rust_test_consumer.py
+++ b/tests/unit/ingestion/test_rust_test_consumer.py
@@ -2,8 +2,14 @@
 (which compiles as a separate, external crate — exactly why the symbol
 must stay `pub`) must not be flagged `unused_export`. Drives the real
 parser, GraphBuilder and analyzer end-to-end against a crate where a
-struct and the function returning it are only referenced via
-`use net::build_report` from an integration test.
+function and a struct are only referenced from an integration test.
+
+`Report` is constructed inside `lib.rs` by `build_report`, so it is
+already reachable via an intra-file edge regardless of any test consumer —
+asserting `Report` isn't flagged doesn't exercise the test-consumer path.
+`Ticket` is declared in `lib.rs` but never constructed there; the
+integration test is its only consumer, so it is the type that actually
+pins the behavior under test.
 """
 
 from __future__ import annotations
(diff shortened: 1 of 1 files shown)


Ayush7614 (approved):
## Review: Approve

This PR adds a single, self-contained regression test (`test_rust_test_consumer.py`) with no production-code changes. I ran it against the current `main` checkout:

```
tests/unit/ingestion/test_rust_test_consumer.py ...   [100%]  3 passed in 0.38s
```

Why it's a good test:
- It drives the **real** parser, `GraphBuilder`, and `DeadCodeAnalyzer` end-to-end (no mocks), so it actually exercises the `pub`-symbol / integration-test-consumer behavior.
- It includes a **true-positive guard** (`dead_fn`, never referenced anywhere, is still flagged as `UNUSED_EXPORT`), which proves the unused-export detection is genuinely running and the non-flagging of `build_report`/`Report` is a real distinction — not the test passing vacuously.
- `is_test` handling already exists in `_detect_unused_exports` (analyzer.py:736), so the behavior it pins is real and stable.

No issues found. Approving.

RaghavChamadiya (commented):
@crizah thanks for this. A guard test for a bug that no longer reproduces is genuinely useful, and including `dead_fn` as a true-positive control is the right instinct, it's what let me check the rest of the file quickly.

One assertion isn't testing what it says. `test_struct_only_used_via_integration_test_not_flagged` passes with the `tests/` file removed entirely:

```
WITH    tests/report_test.rs -> unused_exports = ['dead_fn']
WITHOUT tests/report_test.rs -> unused_exports = ['build_report', 'dead_fn']
```

`build_report` flips, so that assertion is real and pins the behavior you're after. `Report` doesn't, because `build_report` constructs it inside `lib.rs` and that intra-file edge rescues it regardless of any test consumer.

To make it bite, `Report` needs to be reachable only from the test. Simplest version is a second type that `lib.rs` names but never constructs, with the test doing the constructing, so removing the test file flips it. Worth adding the deleted-consumer case as an explicit check too, since that's the property you actually want and it's cheap to assert directly.

Everything else looks right to me, and the `use net::build_report` shape matches the report in #640. Same note as on #1568: since #640 lists four FP classes and this covers one, `Refs #640` avoids auto-closing it early.

RaghavChamadiya (approved):
Thanks @crizah, swapping the assertion onto `Ticket` is exactly right, and the deleted-consumer check is what turns this from a test that passes into a test that pins something. Without it the whole file could have gone on passing for a reason unrelated to test-consumer edges, which is what `Report` was quietly doing.

Confirming it still holds, since main has moved since your run:

```
tests/unit/ingestion/test_rust_test_consumer.py
5 passed
```

The three-way shape here is the one I would point other people at: the property, a true-positive guard so the analyzer has not simply stopped flagging anything, and the inversion showing the property flips when the consumer goes away.

PR #1570, opened 2026-08-14
fix(analysis): detect JSX prop-guard dead code (#1554)

### LINKED ISSUE #1554: [Bug] Dead-code misses components rendered behind a prop guard that is never satisfied

rishu685:
## Summary
Fixes #1554 by introducing AST-based JSX prop-guard dead code detection.

### Key Changes
- **Ingestion & Graph Schema**: Added `supplied_props` to `CallSite` and `ResolvedCall` (as immutable `frozenset[str]`), propagated through all call resolvers including member calls, and merged on `CALLS` graph edges.
- **Prop Guard Extraction (`jsx_prop_guards.py`)**: AST helper extracting JSX components rendered behind binary `&&` and ternary prop guards.
- **Dead Code Analysis (`analyzer.py`)**: Integrated `_get_unsatisfied_prop_guard` into unused-export detection with file-level guard caching (`_guard_cache`) and language gating.
- **Tests**: Added regression tests in `test_jsx_prop_guard_dead_code.py` covering satisfied/unsatisfied guards and spread props.

### DIFF
--- a/packages/core/src/repowise/core/analysis/dead_code/analyzer.py
+++ b/packages/core/src/repowise/core/analysis/dead_code/analyzer.py
@@ -678,6 +678,7 @@ class DeadCodeAnalyzer:
         # along so the evidence line can say why.
         self._unindexed_source_files = list(unindexed_source_files or [])
         self._repo_root = repo_root
+        self._source_map = source_map
         self._unindexed_tokens: frozenset[str] | None = None
         # Kept for the absence check below, which is the one pass that needs
         # the raw text of *every* indexed file rather than of a suffix-filtered
@@ -733,6 +734,11 @@ class DeadCodeAnalyzer:
         # because the whitelist arrives per ``analyze()`` call and the map does
         # not.
         self._package_files: PackageFileMap | None = None
+        # Cache for JSX prop-guard extraction: maps pred_file → list of
+        # (parent_fn, child_component, guard_prop) triples. Populated lazily
+        # by _get_unsatisfied_prop_guard so each TSX/JSX file is read from
+        # disk and parsed by tree-sitter at most once per analysis run.
+        self._guard_cache: dict[str, list[tuple[str, str, str]]] = {}
 
     def _reachability_rescues(self, whitelist: AbstractSet[str]) -> ReachabilityRescues:
         """Assemble the rescue state the shared predicate reads."""
@@ -1329,6 +1335,8 @@ class DeadCodeAnalyzer:
                 # alias; importers carry the alias in ``imported_names``.
                 export_alias = self._ts_export_aliases.get(str(node), {}).get(sym_name)
 
+                unsatisfied_guard: tuple[str, str] | None = None
+
                 has_importers = False
                 for pred in self.graph.predecessors(node):
                     edge_data = self.graph[pred][node]
@@ -1342,13 +1350,17 @@ class DeadCodeAnalyzer:
                         break
 
                 if has_importers:
-                    continue
+                    node_lang = node_data.get("language", "")
+                    if node_lang in ("typescript", "javascript"):
+                        unsatisfied_guard = self._get_unsatisfied_prop_guard(sym_id, sym_name)
+                    if unsatisfied_guard is None:
+                        continue
 
                 # Namespace-import rescue: see ``file_imported_as_namespace``
                 # computation above. Any public symbol in a file pulled by
                 # module name could be the dispatch target for
                 # ``<modname>.<attr>(...)``.
-                if file_imported_as_namespace:
+                if file_imported_as_namespace and unsatisfied_guard is None:
                     continue
 
                 # Symbol-level usage signal: any incoming ``calls`` /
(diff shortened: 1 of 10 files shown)


Ayush7614 (changes requested):
## Review: Request Changes

I reviewed the full diff and verified the core logic against the real `tree-sitter-typescript` grammar (not just by reading). **The feature is not correct/verified as-is — several common JSX guard patterns are either silently missed or produce false positives, and the `supplied_props` edge-merge is buggy.**

###  fix

**1. `jsx_prop_guards.py` misses/garbles common guards (verified by running the extracted module):**
```
props.show && <Foo/>        => [('A','Foo','show')]   ✅
props.show && <UI.Button/>  => []                      ❌ missed
this.props.show && <Foo/>   => []                      ❌ missed
!show && <Foo/>             => [('A','Foo','show')]    ⚠️ wrong direction (false positive)
a && b && <Foo/>            => []                      ❌ missed
```
- `_find_jsx_targets` only matches `identifier`/`property_identifier` element names; `<UI.Button/>` has a `member_expression` name → namespaced guarded renders are never detected.
- `_extract_prop_name` only inspects direct children of a `member_expression`, so `this.props.show` (nested member) returns `None` → missed.
- `!show && <Foo/>` records guard prop `show` as if truthiness were required, but it renders `Foo` when `show` is *false/absent* → will wrongly flag real components as dead.
- Compound AND `a && b && <Foo/>` parses as `(a && b) && <Foo/>`; only the immediate left operand is read (itself a binary), so `pn` is `None` → entirely missed.

**2. Buggy `supplied_props` merge in `ingestion/graph/_resolvers.py`:**
```python
ex_props = existing.get("supplied_props")
if ex_props is not None and rc.supplied_props is not None:
    existing["supplied_props"] = frozenset(ex_props | rc.supplied_props)
elif rc.supplied_props is None:
    existing["supplied_props"] = None
```
Order-dependent and lossy: a later non-JSX call site (`None`) overwrites an earlier known `{a}`; and if `None` arrives first, a later `{a}` is silently ignored. `None` means "unknown", so the whole union should become `None`:
```python
if ex_props is None or rc.supplied_props is None:
    existing["supplied_props"] = None
else:
    existing["supplied_props"] = frozenset(ex_props | rc.supplied_props)
```

###  Should fix

**3. False-positive risk for root/top-level components.** In `DeadCodeAnalyzer._get_unsatisfied_prop_guard`, `if not incoming: return (prop_name, parent_name)` flags the child dead whenever the parent component has no callers (e.g., the app root). Root components receive guard props at runtime (router/context/query params), not as statically-supplied JSX props. Don't auto-conclude "unsatisfied" from `not incoming`.

**4. Missing regression test.** The PR description says `test_jsx_prop_guard_dead_code.py` was added, but it is **not in the diff and does not exist anywhere in the repo**. The new detection path has zero test coverage. Please add it (satisfied guard, unsatisfied guard, spread props, namespaced, `this.props`, negation, compound) before merge.

### Minor / ok

- `split_file.py` → `BARREL_FILENAMES` refactor and the new `test_no_barrel_filename_copies.py` look correct (`index.tsx`/`index.jsx` are in the set, so the parametrized cases pass).
- `RISK_CAP_CONFIDENCE = 0.4` means flagged findings are "needs review" (not auto-delete), which is the right severity for a heuristic.

### Bottom line
Not verified-correct. Fix the merge bug, decide whether negation/compound/namespaced/`this.props` guards must be supported (today they silently return nothing or false-positive), and add the missing regression test. Happy to draft the `_resolvers` fix and the test if useful.

packages/core/src/repowise/core/analysis/dead_code/jsx_prop_guards.py
```diff
@@ -0,0 +1,195 @@
...
+                if c.type in ("&&", "===", "==", "!==", "!="):
+                    op = c.type
+                elif left is None:
+                    left = c
+                else:
+                    right = c
+            if op == "&&" and left and right:
+                return _extract_prop_names(left) + _extract_prop_names(right)
+            elif op in ("===", "==") and left and right:
+                return _extract_prop_names(left) + _extract_prop_names(right)
```
RaghavChamadiya: Both branches have the same body, which is what ruff is failing on (SIM114).

```suggestion
            if op in ("&&", "===", "==") and left and right:
                return _extract_prop_names(left) + _extract_prop_names(right)
```

packages/core/src/repowise/core/analysis/dead_code/analyzer.py
```diff
@@ -1317,6 +1333,81 @@ def _detect_unused_exports(
...
+                return parent_guard
+
+        return None
+
+    def _read_file_text(self, file_path: str) -> str | None:
+        """Safely read the content of a file."""
+        try:
+            p = Path(file_path)
+            if not p.is_absolute() and getattr(self, "repo_path", None):
+                p = Path(self.repo_path) / p
```
RaghavChamadiya: There's no `repo_path` attribute on `DeadCodeAnalyzer`. `__init__` takes `repo_root: Path | None` and stores it as `self._repo_root`, so this `getattr` always returns `None` and the branch never runs. Graph node keys are repo-relative, so `p.exists()` only succeeds when the process happens to have cwd at the repo root. Under the CLI it usually does, which is why nothing fails loudly, but from the server or any other entry point every guard lookup silently returns `[]` and the feature is a no-op.

```suggestion
            if not p.is_absolute() and self._repo_root:
                p = Path(self._repo_root) / p
```

Worth a test that builds the analyzer with `repo_root=tmp_path` and runs from a different cwd, otherwise this exact regression comes back the next time someone touches the path handling.

RaghavChamadiya (commented):
@rishu685 thanks for turning this around so quickly, and for going through Ayush's list point by point. I checked the new diff against his review and the merge fix, the member-expression targets, `this.props`, compound `&&`, the negation case and the missing test file are all there now. Nice work.

Two things before this can go in. I've left both as suggestions inline, so you can apply them from the Files tab and CI will rerun itself.

1. CI is red on ruff, SIM114 in `jsx_prop_guards.py:95`. `uv run ruff check .` reproduces it locally. Please don't run `ruff format`, CI only gates on `ruff check`.
2. `_read_file_text` looks for `self.repo_path`, but the analyzer stores the root as `self._repo_root`, so the `getattr` always returns `None` and the relative-path branch never fires. Details in the inline comment, this one is worth a test.

Smaller things, take or leave:

- `parser.py`: `supplied_props: set[str] | None = None` is annotated `set` but assigned a `frozenset`. Harmless at runtime, mypy will complain, just annotate it `frozenset[str] | None`.
- `_get_unsatisfied_prop_guard` reads and tree-sitter-parses TSX files off disk. `DeadCodeAnalyzer` already receives `source_map` (`{repo_relative_path: raw bytes}`) for exactly this reason, the other prepasses use it to avoid disk passes. Reusing it would keep the pattern consistent and dodge the path problem entirely. Fine as a follow-up if you want to keep this PR tight.

Ping me once CI is green and I'll do another pass. Happy to push the two fixes to your branch myself if you're short on time.

Ayush7614 (changes requested):
Verified the JSX prop-guard logic locally (CI is `action_required` for this author, so it has never run). The feature itself is sound, but the branch needs a rebase and one persistence gap is worth fixing.

**Core logic: correct (verified)**
- `jsx_prop_guards.py` AST extraction is reasonable: filters local state (`useState` vars, array/object patterns) and only treats destructured params as props; negated guards (`!show && <B/>`) and spread props (`{...props}`) are conservative and never flagged.
- `supplied_props` propagation (parser → call_resolver → `_resolvers.py` edge merge) is None-aware: spread/unknown → `None` → treated as "maybe supplied", so no false positives.
- Findings are review-only: confidence 0.4 (`RISK_CAP_CONFIDENCE`), `safe_to_delete=False`.
- All 9 new tests pass; `tests/unit/dead_code` + `tests/unit/ingestion` pass (2784 on the branch, 2863 on a simulated merge onto current main — the extra 79 include upstream's newer C++ forward-declaration tests). `ruff check` clean on all changed files. Simulated `git merge` onto `main` has no conflicts, and the merged `ResolvedCall` correctly keeps `origin` alongside the new defaulted `supplied_props` field (no #1720-style breakage).

**Required before merge**

1. **Rebase onto current upstream main.** The branch is based on an old main. The diff vs current main misleadingly shows the *removal* of the C++ forward-declaration guard (`_CPP_TYPE_DECLARATION_KINDS` block in `_detect_unused_exports`) and *re-adds* the interface / contract-method / risk-factor confidence caps that already landed on main (via the recent analyzer PRs). Because of this, an auto-merge produces a **duplicated confidence block**: `_detect_unused_exports` computes `confidence` + the caps at lines 1396-1428, then your new `else` branch recomputes the identical thing at lines 1436-1477. Rebasing collapses the diff to just the JSX-guard logic and eliminates the redundancy (merge is currently correct, but redundant).

2. **Persist `supplied_props` on graph edges (or document the limitation).** `_rehydrate.py` only restores `edge_type` and `imported_names` on rehydrated edges, so `supplied_props` is dropped. A graph rehydrated from a `--mode fast` index (via `repowise update --full`) will see `None` on every edge, which is treated as "maybe supplied" — the new guard detection silently no-ops in that path. Either persist the field next to `imported_names` or note the limitation in the PR description.

**Minor / optional**
- In `jsx_prop_guards._extract_prop_names`, a non-destructured `props` component reads *any* bare identifier in a guard position as a prop — including module-level constants (`const SHOW_DEBUG = ...; {SHOW_DEBUG && <B/>}`). Low confidence (0.4, not safe-to-delete) keeps this review-only, but a module-scope check would reduce noise.

RaghavChamadiya (approved):
Rebased clean and the full matrix is green on 433f39d7, so this is the real pass I said I owed you.

The prop-extraction concern from 16 August is closed, and I checked it by re-running the same measurement rather than reading the diff. `_extract_prop_names` now takes `local_vars` and `param_props` and gates the bare-identifier branch on them, with `_collect_function_local_vars` walking the function body's `variable_declarator` nodes and stopping at nested function boundaries. Over `packages/ui/src`, 412 TSX files:

```
                                   before      now
guard triples                        229        84
guard name is a local const/let/var  127 (55%)   7 (8%)
```

All three of the components I named are gone: `FileExplorer.tsx` now yields no triples at all, same for `ArchLegend.tsx`, and `C4Diagram.tsx` no longer produces `isLoading`. The array-pattern path in `_extract_bound_names` is what catches `const [isExpanded, setIsExpanded] = useState(...)`, which was the shape doing most of the damage.

The `_parent_fn` collision is closed too. `_get_unsatisfied_prop_guard` now derives `pred_parent` and filters on `parent_fn == pred_parent`, so a guard can no longer be borrowed from a different component in the same file. `owner-profile.tsx` was the file with the real collision and it is down to one triple from one parent.

Two things noted for later, neither blocking. The `parent_fn == "Anonymous"` escape hatch on that same filter still lets a guard from an anonymous function match any predecessor, which is a narrower version of the collision I just described. And the graph walk in the recursion is still unbounded while `_guard_cache` only bounds the parsing, so the timing question on a large TS repo is unanswered. If either turns into a real finding on a big repo I would rather fix it with that repo in hand than guess now.

@Ayush7614's two required items are both in: the branch is rebased, and `_rehydrate.py` restores `supplied_props` as a frozenset with `test_supplied_props_survives_rehydration` covering it.

Thanks @rishu685 for staying with this across five rounds and two months, and for taking the barrel-filename work back out of scope when I asked. Merging.

PR #1629, opened 2026-08-16
feat(health): register Pascal in complexity, duplication, and dataflow dialects

Laex:
## Summary

Object Pascal support (ingestion/AST parsing, symbols, imports) already landed and sits at "Good" tier per `docs/layers/LANGUAGE_SUPPORT.md`, but the three health-marker dialects (complexity, duplication, dataflow) had no Pascal registration, so Pascal files produced no complexity/nesting scores, no clone-detection tokens, and no CFG-based dataflow signal.

- Adds a Pascal `LanguageNodeMap` to `complexity/languages.py` — branch/loop/case/try/catch/boolean-operator/CFG node kinds, derived from actually parsing sample Delphi source with `tree_sitter_pascal` rather than guessed. Class-level metrics (LCOM4 / god-class) are deliberately left unmapped: Pascal splits a class into a signature-only declaration and a fully separate set of top-level qualified-method implementations, so there's no single AST node grouping a method's body the way class analysis expects — same posture the codebase already takes for Go's external-receiver methods.
- Registers Pascal's literal node kinds (`literalNumber` / `literalString`) in `duplication/tokenizer.py` so clone detection normalizes literals correctly.
- Fixes name/parameter extraction in `complexity/ast_utils.py`, which assumed AST shapes Pascal's grammar doesn't use: the function name sits two hops down under a `header` field rather than directly on the node, the parameter-list field is named `args` not `parameters`, and Pascal groups same-typed parameters under one node (`A, B, C: Boolean`) rather than one node per parameter. Without this fix every Pascal function showed up as `<anonymous>` with 0 parameters.
- Dataflow/CFG needed no separate registration — the dataflow pass reuses the same `LanguageNodeMap`, confirmed by building a real (non-degenerate) CFG with branches, loops, and join blocks for a test function.
- Updates `docs/layers/LANGUAGE_SUPPORT.md`: tier table, the detailed code-health coverage table (new Pascal row with a footnote explaining the class-metrics gap), and the "known gaps" section.

## Test plan

- [x] New unit tests in `tests/unit/health/test_complexity_walker.py` (nesting/CCN on new `pascal/nested.pas` and `pascal/complex.pas` fixtures, plus a test confirming class metrics stay empty rather than reporting a misleading zero-method class) and `tests/unit/health/test_duplication.py` (literal/identifier normalization)
- [x] `pytest tests/unit/health/ tests/unit/ingestion/parser/test_pascal.py` — 1161 passed
- [x] Validated against a real ~150-file Delphi codebase via `repowise update --index-only` + `repowise health`: correct qualified method names, plausible CCN/nesting numbers, and duplication findings firing, with zero crashes across 233 files / 1500+ findings


RaghavChamadiya (approved):
CI is green on all three Python versions, and the grammar-missing test nit I raised is a local-environment thing that does not affect CI, so I am not holding the PR on it. Merging this as the foundation of the Pascal stack.

PR #1633, opened 2026-08-16
fix(risk): ignore shallow boundary in fix history

### LINKED ISSUE #1623: [Bug] fix_pressure credits every file in the repo on a shallow clone
## Describe the Bug

On a shallow clone, `fix_pressure` credits bug-fix history to files that have never
been fixed. In the worst case it credits nearly every file in the repository.

The oldest commit in a shallow clone is grafted, so git reports it as parentless.
`git log --name-only` then lists it as touching the entire tree. The walk in
`fix_pressure` only checks whether a commit's subject looks like a fix
(`_walk`, `packages/core/src/repowise/core/analysis/change_risk/fix_history.py:103`),
so if that boundary commit happens to have a fix-shaped subject, every file at that
point in history gets a fix credit. The credit is also recent rather than old, so the
365 day decay barely discounts it.

Whether it fires is down to one commit's subject line, which makes it look intermittent.
react at depth 2000 is affected, react at depth 500 is not, django at neither.

`repowise risk` leads with fix history, so this lands on the first line of its output
rather than somewhere a user has to go looking.

## Steps to Reproduce

1. Shallow clone any active repo:

   ```
   git clone --single-branch --depth=2000 https://github.com/facebook/react react-shallow
   ```

2. Look at the boundary commit. Whether the bug fires depends on its subject, so this
   step tells you whether the clone you just made is affected:

   ```
   cd react-shallow
   BOUNDARY=$(git rev-list --max-parents=0 HEAD)
   git log -n1 --format=%s $BOUNDARY              # fix-shaped subject?
   git log -n1 --name-only --format= $BOUNDARY | grep -c .   # near the whole tree
   ```

   Since the boundary moves with the clone date, pick a depth where the subject matches.
   In the clone I measured (depth 2000, boundary `4a86ec5a66`) the subject was
   `fix[react-devtools]: removed redundant startProfiling call (#31131)` and it reported
   6861 files of a 7243 file tree.

3. Make a full clone of the same repo, then edit one file that has obviously never been
   bug fixed, in each clone, and run `repowise risk` on the uncommitted change:

   ```
   echo "# touch" >> .editorconfig
   repowise risk
   ```

4. Compare the fix maps directly if you want the size of it:

   ```python
   from repowise.core.analysis.change_risk.fix_history import fix_pressure
   shallow = fix_pressure("react-shallow")
   full = fix_pressure("react-full")
   print(len(shallow), len(full))
   print(len([p for p in shallow if p not in full]))
   ```

## Expected Behavior

Files with no bug-fix history stay out of the map. A shallow clone should give a
smaller map than a full clone, not a larger one.

## Actual Behavior

The clearest way to see it. Same one line edit to `.editorconfig`, one dirty file in
both clones, `repowise risk` on the uncommitted change.

Shallow clone at depth 2000:

```
Change risk for working tree: touches files that have broken before · 0th percentile of this repo's fix-bearing files
┌───────────────┬───────┬─────────────┐
│ File          │ Lines │ Prior fixes │
├───────────────┼───────┼─────────────┤
│ .editorconfig │     1 │         0.3 │
└───────────────┴───────┴─────────────┘
```

Full clone, same edit:

```
Change risk for working tree: no bug-fix history in the files it touches
```

`.editorconfig` has never been bug fixed. The two runs give opposite headline sentences.

The size of it, measured on react, shallow depth 2000 against full history:

```
fix-bearing files, full history:    3795
fix-bearing files, shallow clone:   7249
files credited that have never been fixed: 5487
```

The falsely credited files are what you would expect from a whole-tree commit:

```
 0.305  .editorconfig
 0.305  .gitattributes
 0.305  .github/dependabot.yml
 0.305  .github/ISSUE_TEMPLATE/bug_report.md
 0.305  .github/workflows/compiler_rust.yml
```

All 5487 sit at exactly 0.305, which is `0.5 ** (625 / 365)` for the 625 day span of that
clone. They are all the one boundary commit.

Knock-on effects:

- `hot_files` can name a file that has never been fixed, which is the case above.
- `change_fix_density` is inflated for any change touching never-fixed files.
- `fix_density_percentile` ranks against `pressure.values()`, so the denominator is
  mostly invented entries.

The percentile effect is smaller than I first assumed and worth stating accurately. On a
change touching a file that does have real history, scoring the same commit in both clones
gave 96th percentile shallow against 94th full, so the ranking held up. The damage is
concentrated in the files that should not be in the map at all.

## Environment

- OS: Windows 11 (10.0.26200)
- Python version: 3.11.15
- Repowise version: 0.43.0
- Installation method: pip

## Additional Context

Worth noting that the module already takes care over this class of problem. `fix_pressure`
raises `FixHistoryUnavailableError` rather than returning an empty map, with a comment
explaining that empty is indistinguishable from "this repository has never had a bug fix".
The shallow case slips past that guard, because it is neither empty nor an error.

CI checkouts are shallow by default (`actions/checkout` uses depth 1), so anything running
`repowise risk` in CI is likely hitting this.

## Notes for anyone picking this up

The fix is small and contained. Everything below is in
`packages/core/src/repowise/core/analysis/change_risk/fix_history.py`.

Two options, either of which would close it:

- Skip parentless commits in `_walk` when `git rev-parse --is-shallow-repository` returns
  true. Keeps a windowed but honest map, which seems better than nothing for CI users who
  cannot control clone depth. Note a real root commit in a full clone is also parentless
  and legitimately adds every file, so the shallow check has to gate it rather than
  skipping parentless commits unconditionally.
- Raise `FixHistoryUnavailableError` on a shallow repository. Callers already handle it
  (`ChangeRiskResult.fix_history_available` and the CLI branch at `risk_cmd.py:490`), so
  the surfaces would report a gap instead of a wrong number.

The first is probably the better default, but the second is more in keeping with the
module's existing stance that a number it cannot stand behind should not be served.

Tests live in `tests/unit/analysis/test_change_risk_fix_history.py`. A regression test
wants a shallow fixture, so it needs a real `git clone --depth` in a tmpdir rather than the
commit-building helpers already in that file.

Roovelrz:
## Summary

- detect shallow repositories before walking fix history
- skip parentless boundary commits only when the repository is shallow
- preserve normal root-commit behavior for full clones
- add regression coverage using a real depth-limited clone

## Why

In a shallow clone, Git treats the oldest available commit as parentless. When that boundary commit has a fix-shaped subject, `git log --name-only` reports its tree as if the commit introduced all of those files.

`fix_pressure` then incorrectly credits those files with bug-fix history.

This change ignores that untrustworthy boundary commit while keeping the valid visible history after it.

Fixes #1623

## Test plan

- [x] `uv run pytest tests/unit/analysis/test_change_risk_fix_history.py -q`
- [x] `uv run pytest tests/unit/` — existing 12 unrelated failures
- [x] `npm run lint`
- [x] `npm run type-check`
- [x] `uv run repowise impacted-tests --staged`
- [x] `uv run repowise risk main..HEAD`
- [x] `uv run repowise health --file packages/core/src/repowise/core/analysis/change_risk/fix_history.py`

### DIFF
--- a/packages/core/src/repowise/core/analysis/change_risk/fix_history.py
+++ b/packages/core/src/repowise/core/analysis/change_risk/fix_history.py
@@ -65,11 +65,20 @@ def clear_fix_pressure_cache() -> None:
 
 def _walk(repo_path: str, upto_ref: str, depth: int) -> dict[str, list[float]]:
     """Commit timestamps of every bug-fix touching each file, up to *upto_ref*."""
+    is_shallow = (
+        _git(["rev-parse", "--is-shallow-repository"], repo_path, check=False).strip() == "true"
+    )
     try:
         proc = subprocess.run(
             [
-                "git", "log", f"-n{depth}", "--no-merges",
-                "--format=%x1e%ct%x1f%s", "--name-only", "--end-of-options", upto_ref,
+                "git",
+                "log",
+                f"-n{depth}",
+                "--no-merges",
+                "--format=%x1e%P%x1f%ct%x1f%s",
+                "--name-only",
+                "--end-of-options",
+                upto_ref,
             ],
             cwd=repo_path,
             capture_output=True,
@@ -93,14 +102,17 @@ def _walk(repo_path: str, upto_ref: str, depth: int) -> dict[str, list[float]]:
         if not block:
             continue
         lines = block.split("\n")
-        head = lines[0].split("\x1f")
-        if len(head) != 2:
+        head = lines[0].split("\x1f", 2)
+        if len(head) != 3:
+            continue
+        parents, timestamp_raw, subject = head
+        if is_shallow and not parents:
             continue
         try:
-            timestamp = float(head[0])
+            timestamp = float(timestamp_raw)
         except ValueError:
             continue
-        if not is_fix_commit(head[1]):
+        if not is_fix_commit(subject):
             continue
         for path in lines[1:]:
             path = path.strip()
--- a/tests/unit/analysis/test_change_risk_fix_history.py
+++ b/tests/unit/analysis/test_change_risk_fix_history.py
@@ -64,6 +64,60 @@ def repo_with_history(tmp_path: Path) -> Path:
     return repo
 
 
+def test_fix_pressure_skips_fix_shaped_shallow_boundary(tmp_path: Path) -> None:
+    source = tmp_path / "source"
+    source.mkdir()
+    _git(["init", "-q"], source)
+    _git(["config", "user.name", "Dev"], source)
+    _git(["config", "user.email", "d@e.com"], source)
+
+    _commit(
+        source,
+        {"hot.py": "v = 0\n", "cold.py": "c = 0\n"},
+        "feat: seed",
+    )
+    _commit(source, {"hot.py": "v = 1\n"}, "feat: prepare")
+    _commit(source, {"hot.py": "v = 2\n"}, "fix: boundary bug")
+    _commit(source, {"hot.py": "v = 3\n"}, "fix: later bug")
+    _commit(source, {"cold.py": "c = 1\n"}, "feat: latest")
+
+    shallow = tmp_path / "shallow"
+    subprocess.run(
+        [
+            "git",
+            "clone",
+            "-q",
+            "--depth=3",
+            source.resolve().as_uri(),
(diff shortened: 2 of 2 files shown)


RaghavChamadiya (approved):
Verified the root cause and ran the change-risk suites locally, all green. Merging.

PR #1675, opened 2026-08-17
fix(ingestion): resolve Pascal dead-code false positives and disambiguate type kinds

Laex:
## Summary

A 5-symbol sample of "confidence 1.0, no importers" dead-code findings on a real ~150-file Delphi codebase was 5/5 false positives, from three root causes in the Pascal ingestion queries:

- The parenless-call pattern missed the dominant Delphi idiom of a niladic call on an assignment's RHS (`Result := GetDefaultNDNProfile;`).
- There was no type-reference capture at all, so a class named only in a `var`/field/parameter/return-type declaration, or as `Application.CreateForm`'s class argument, minted no usage edge.
- `uses` carried no per-symbol `imported_names`, so the file-level unused-export rescue could never fire for Pascal specifically.

All three are fixed: `pascal.scm` gained assignment-RHS call patterns and `@param.type` captures, and `uses` clauses now emit `imported_names=["*"]` wildcard imports (accurate, since `uses` really does expose a whole unit's public interface). A temporary blanket confidence cap was added while these were being fixed and then lifted after re-validating against the same real repo: 603 findings dropped to 1, and that one was already a known, correctly-low-confidence candidate unrelated to Pascal.

Also disambiguates `declType`'s previously-collapsed symbol kind: `record` / `interface` / `enum` / `set` / `array` / plain-alias now report their real `SymbolKind` via `refine_pascal_type_kind` instead of all eight forms reporting `kind="class"`. `object` and class-helper stay `"class"` (no closer match in the vocabulary).

Docs updated: `docs/layers/LANGUAGE_SUPPORT.md`'s "Object Pascal known gaps" section.

## Test plan

- [x] New/updated tests in `tests/unit/ingestion/parser/test_pascal.py` (type-kind disambiguation for each declType shape) and `tests/unit/dead_code/test_interfaces.py` (the confidence-cap and rescue-path fixes)
- [x] `pytest tests/unit/ingestion/ tests/unit/dead_code/` — 2356 passed
- [x] Re-validated against the same real ~150-file Delphi codebase used to find the original false positives: 603 findings → 1


RaghavChamadiya (approved):
Thanks @Laex. Sampling five confidence-1.0 findings on a real Delphi codebase and hand-adjudicating all five before writing a line of query is the right way round, and it is why this PR names three distinct root causes instead of one symptom.

I want to record two checks I made, because both are places where a query change like this usually goes wrong and this one does not.

**The `@param.type` captures cannot widen call scope here.** Type-use edges are normally dangerous: `graph/builder.py` folds every `type_use` edge into `import_targets`, which opens the target file to bare-name call matching from the source, and on Rust an inferred target measured 25% precision on the calls it gained. That does not reach Pascal, because `resolve_type_refs` dispatches through `_STRATEGIES` and there is no `pascal` entry, so these captures stay local type references feeding the unused-export rescue and never mint a cross-file edge. That is the correct side of the line to be on: evidence a type is used, not evidence that this file's bare names live in the target.

**The over-eager assignment-RHS pattern is safer than your comment claims.** You argue a stray edge can only add usage evidence, so the failure direction is a false negative rather than a false positive. That is right, and there is a stronger reason you did not give: Pascal declares symbols only from `declType`, `declProc`, `defProc` and `declProp`. A variable is never a symbol in this graph, so `Result := SomeLocal;` has nothing to resolve to and mints no edge at all. The residual collision is a variable sharing a name with a reachable procedure or property, and in the property case a Delphi property read really does run a getter, so even the "wrong" edge points at something being invoked. I was ready to hold the PR on the call-graph blast radius of that pattern, since `calls` edges feed callers, blast radius and PageRank rather than only dead code, and it does not survive contact with what Pascal actually defines.

The rest reads well. `imported_names=["*"]` for `uses` is the semantically right value rather than a workaround, and you are correct that the empty list made the file-level rescue structurally unable to fire for Pascal. Scoping the `CreateForm` capture to the literal target with a case-insensitive match, and anchoring to the first argument only, is a much better instinct than treating every bare-identifier argument as a type reference.

One note for whoever picks this up next, not a change request: I could not reproduce any of the Pascal numbers locally, because `tree_sitter_pascal` is not installed in my environment and every Pascal file parses to nothing here. CI has the grammar and your 364 lines of parser tests pass on all three Pythons, so the evidence is CI's and yours, not mine. Worth knowing that the local story and the CI story differ for this language.

PR #1679, opened 2026-08-17
fix(health): flatten Pascal else-if chains instead of nesting each arm

Laex:
## Summary

Found while validating Pascal's health markers against a real Delphi codebase: `StringToVK`, a ~40-arm `else if` VK-code lookup table, reported `max_nesting=33` and `cognitive=583`. That's a plain flat dispatch table, not 33 levels of logic.

Root cause: Pascal's grammar nests each `else if` arm one level inside the previous arm's `else` field (`ifElse -> ... -> kElse -> ifElse -> ... -> kElse -> ifElse -> ...`), so an unflattened tree walk charges a fresh nesting level per arm. This is exactly the distortion `_is_elif_continuation` in `cyclomatic.py` already solves for Python (`elif_clause`), Ruby (`elsif`), and the else-if shapes of TypeScript/Rust/C++/Java/Go/C#/Dart/Kotlin — Pascal's node types (`if` / `ifElse`) and else-token spelling (`kElse` vs. `else`) just weren't registered with it.

Fix: add `"if"` / `"ifElse"` to `_ELSE_IF_NODE_KINDS` and `"kElse"` to the else-token check. A continuation arm still charges CCN (it's a real decision point — McCabe complexity is unaffected), it just no longer opens a fresh nesting level or compounds cognitive complexity.

`StringToVK`: nesting 33 → 1, cognitive 583 → 55, CCN unchanged at 56.

## Test plan

- [x] New regression test `test_pascal_else_if_chain_is_flat_not_nested` in `tests/unit/health/test_complexity_walker.py` (a 5-arm chain: CCN 6, nesting 1)
- [x] `pytest tests/unit/health/ tests/unit/ingestion/` — 3563 passed, 5 skipped
- [x] Re-verified against the real `StringToVK` function that surfaced this


RaghavChamadiya (approved):
Thanks @Laex, the 40-arm `StringToVK` dispatch is exactly the evidence this needed. Nesting 33 and cognitive 583 for a flat lookup table is the kind of number that makes a whole language's health scores unusable, and it is much easier to argue from than a synthetic case.

The fix is in the right place. `_ELSE_IF_NODE_KINDS` and the else-token check are the only two things `_is_elif_continuation` looks at, and it is the sole consumer of both, so widening them cannot reach anything beyond the nesting gate at `cyclomatic.py:304`. I checked the generic-looking `"if"` entry specifically, since a bare `if` is also the anonymous keyword token in most grammars: that token is always the first child of its own `if_statement`, so its `prev_sibling` is None and the else-token test rejects it. The widening stays Pascal-shaped in practice.

Splitting `_ELSE_TOKEN_KINDS` out instead of special-casing `kElse` inline is the right call too, since the next grammar with its own spelling is a one-line change now.

The test pins both halves, CCN 6 because five arms are five real decisions, nesting 1 because they are not nested. That distinction is the whole point and it would have been easy to assert only the second.

PR #1691, opened 2026-08-18
fix(ingestion): record the exported name for aliased TS/JS imports

### LINKED ISSUE #1520: [Bug] TypeScript aliased imports (import { X as Y }) drop the graph edge, producing safe_to_delete unused_export FPs

ghostiee-11:
## Summary

- `import { beta as renamedBeta }` stamped the local alias on the edge's `imported_names`, so the unused-export pass never matched and reported `beta` as `safe_to_delete`. Named imports and `require()` destructures now record the exported name, matching the re-export branch above them and the Python extractor.
- The two call sites that want the name *this* file writes — `CallResolver._externally_bound_names` and the router-DSL map in `framework_edges._build_ts_var_to_file` — now read a new `Import.local_names`. Without that, an aliased Express/Hono handler would have stopped resolving.
- Default and namespace imports keep the local name (their source-side name is `default` / the whole module).

## Related Issues

Fixes #1520

## Test Plan

New `tests/unit/ingestion/test_ts_import_alias_dead_code.py`: binding-level assertions, the value-position repro from the issue end to end through parser → GraphBuilder → DeadCodeAnalyzer, and the inverse case (a genuinely unimported sibling is still flagged). The end-to-end assertions fail on `main`.

- [x] Tests pass (`pytest tests/unit/{ingestion,dead_code,pipeline,analysis}` — 3516 passed)
- [ ] Web build passes *(no frontend changes)*

## Checklist

- [ ] I have updated documentation if needed *(no user-facing change)*

### DIFF
--- a/packages/core/src/repowise/core/ingestion/call_resolver.py
+++ b/packages/core/src/repowise/core/ingestion/call_resolver.py
@@ -1348,6 +1348,10 @@ class CallResolver:
         Read off the raw import statements rather than ``_import_names``, which
         only carries bindings that resolved to a file — precisely the ones this
         needs to exclude.
+
+        The names wanted here are the ones *this file writes*, which is what
+        ``Import.local_names`` answers: ``imported_names`` carries the source
+        module's name, and under an alias the two differ.
         """
         names = self._external_names.get(file_path)
         if names is not None:
@@ -1358,7 +1362,7 @@ class CallResolver:
         for imp in parsed.imports if parsed else ():
             if imp.resolved_file and not imp.resolved_file.startswith("external:"):
                 continue
-            bound = (*imp.imported_names, imp.module_path.rsplit(".", 1)[-1])
+            bound = (*imp.local_names, imp.module_path.rsplit(".", 1)[-1])
             found.update(name for name in bound if name and name != "*")
 
         names = frozenset(found)
--- a/packages/core/src/repowise/core/ingestion/extractors/bindings/ts_js.py
+++ b/packages/core/src/repowise/core/ingestion/extractors/bindings/ts_js.py
@@ -59,7 +59,7 @@ def _extract_require_bindings(
                 if key is not None and val is not None:
                     exported = node_text(key, src)
                     local = node_text(val, src)
-                    names.append(local)
+                    names.append(exported)
                     bindings.append(
                         NamedBinding(local_name=local, exported_name=exported, source_file=None)
                     )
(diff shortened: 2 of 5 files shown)


Ayush7614 (approved):
## Review: APPROVED

- **Root cause is correct**: aliased named imports (`import { beta as renamedBeta }`) stamped the *local* alias onto `imported_names`, so the unused-export pass (which matches the *source* module's symbol name) never matched and flagged `beta` as `safe_to_delete`. Recording the exported name now matches the re-export branch and the Python extractor's documented rule.
- **Audited all `imported_names` consumers** to confirm the two call sites the PR switched to `local_names` are the right ones:
  - `call_resolver._externally_bound_names` and `framework_edges/base._build_ts_var_to_file` genuinely want the name *this file writes* → correctly moved to `Import.local_names`.
  - `import_index.py` already prefers `imp.bindings` (keyed by `local_name`) and only falls back to `imported_names` when no bindings exist → unaffected.
  - Dead-code namespace rescue and dispatch-table paths (`analyzer.py`) read edge `imported_names` against the *target* file's symbol names / stem → they benefit from source names, no regression.
  - `page_generator/validation.py` adds both binding names → superset, no regression. Flask/FastAPI edges are Python-only.
  - `local_names` falls back to `imported_names` for extractors that emit no bindings, so other languages are unchanged by construction.
- **Tests are strong**: binding-level assertions, `local_names` property tests, and end-to-end parser → GraphBuilder → DeadCodeAnalyzer runs including the inverse case (a genuinely unimported sibling is still flagged), which proves the fix doesn't blanket-rescue.

All CI green. Approved.

swati510 (approved):
LGTM

PR #1701, opened 2026-08-18
test(health): guard Pascal-only tests that skip the shared _walk() helper

Laex:
## Summary

Review feedback on #1629 (https://github.com/repowise-dev/repowise/pull/1629#issuecomment-5324140617): `test_pascal_no_class_metrics` (`test_complexity_walker.py`) and `test_tokenize_file_pascal_normalizes_identifiers_and_literals` (`test_duplication.py`) call `walk_file()` / `tokenize_file()` directly with inline byte literals rather than through the file's `_walk()` / `_walk_classes()` helpers, so they never picked up the module's documented best-effort `pytest.skip` guard for an environment without `tree-sitter-pascal` synced — they'd hard-fail instead.

`_walk_classes()` isn't a fit for the class-metrics test specifically: it skips on `not fcx.classes`, which is exactly the condition that test asserts.

Adds an explicit `pytest.importorskip("tree_sitter_pascal", reason=...)` to both, matching the guard `test_pascal.py` already uses.

## Test plan

- [x] `pytest tests/unit/health/test_complexity_walker.py tests/unit/health/test_duplication.py` — 72 passed


Ayush7614 (approved):
## Review: APPROVED

Verified the change:

- Both `test_complexity_walker.py` and `test_duplication.py` already `import pytest` at module level, so the added `pytest.importorskip("tree_sitter_pascal", ...)` calls work correctly.
- The reasoning is sound: these two tests call `walk_file()` / `tokenize_file()` directly with inline byte literals, so they bypass the shared `_walk()`/`_walk_classes()` guards; and `_walk_classes()` can't be used for the class-metrics test because it skips on `not fcx.classes` — exactly the condition that test asserts.
- Matches the established guard pattern in `test_pascal.py` and the skip reason is helpful (`run uv sync --all-packages`).

All CI green. Approved.

PR #1705, opened 2026-08-18
fix(persistence): store symbol names as Text, not VARCHAR(255)

### LINKED ISSUE #1565: [Bug] Postgres SQL Schema doesn't fit data for big projects, Phase 4 of 4 · Persistence fails
## Describe the Bug

Postgres SQL Schema doesn't fit data for big projects,  Phase 4 of 4 · Persistence fails

## Steps to Reproduce

docker run --name pg_repowise -e POSTGRES_DB=repowise -e POSTGRES_USER=repowise -e POSTGRES_PASSWORD=repowise -p 5432:5432 -d postgres

export REPOWISE_DB_URL="postgresql+asyncpg://repowise:repowise@localhost:5432/repowise"

Run `repowise init --no-prose --yes` 

## Expected Behavior
Ingestion persisted to POSTGRES DB

## Actual Behavior
Errors appear similar to the following one.

sqlalchemy.exc.DBAPIError: (sqlalchemy.dialects.postgresql.asyncpg.Error) <class 'asyncpg.exceptions.StringDataRightTruncationError'>: value too long for type character varying(255)
[SQL: INSERT INTO health_findings 

## Environment

- OS: Ubuntu 22.04
- Python version: [3.12.1]
- Repowise version: 0.42
- Installation method: [uv pip install]

## Additional Context

Here is the list of tables and columns I've modified to pass the ingestion:

ALTER TABLE refactoring_suggestions ALTER COLUMN target_symbol TYPE VARCHAR(512);
ALTER TABLE health_findings ALTER COLUMN file_path TYPE VARCHAR(512);
ALTER TABLE health_findings ALTER COLUMN function_name TYPE varchar(512);
ALTER TABLE git_function_blame ALTER COLUMN symbol_id TYPE TEXT;

ghostiee-11:
## Summary

- Six columns hold a symbol name under `VARCHAR(255)`. SQLite ignores the width, Postgres enforces it, so the run dies in the persistence phase after the whole index is computed. A one-file repo with a 365-character function name reproduces it on `main`: `StringDataRightTruncationError` on `INSERT INTO graph_nodes`, which is a different table than the one reported, so widening only the reported columns moves the crash.
- Columns widened: `graph_nodes.name`, `wiki_symbols.name`, `wiki_symbols.parent_name`, `dead_code_findings.symbol_name`, `health_findings.function_name`, `refactoring_suggestions.target_symbol`. With them as `Text` the same run completes and the stored values measure 365 to 372.
- Left narrow on measurement: `health_file_metrics.module` (holds a package path, longest producible is 142 on an 18-component tree) and `git_function_blame.symbol_id` at 512 (max 114 across a real 4.5k-row index). Bounded vocabularies keep `String(n)`.
- Migration 0055 is Postgres-only. SQLite does not enforce the width and local stores never run Alembic per 0054, so a batch rebuild of six tables would change nothing.

## Related Issues

Fixes #1565

## Test Plan

Verified against PostgreSQL 16:

- the init that died in Phase 4 now exits 0
- `alembic upgrade head` flips all six columns to `text`; `downgrade -1` returns them to `varchar(255)`
- downgrade with a 400-character row present fails with `StringDataRightTruncationError` and leaves the row intact, rather than truncating it clean
- `alembic upgrade head` then `downgrade -1` on a fresh SQLite file is a no-op that does not error

Two new tests, both confirmed to fail when a column is reverted:

- `tests/unit/persistence/test_models.py` asserts the six columns are unbounded, so the next one cannot ship narrow
- `tests/integration/test_postgres_symbol_lengths.py` round-trips a 365-character name through all six. It skips unless `REPOWISE_TEST_PG_URL` is set, and refuses a database whose name does not look like a scratch one, since it drops tables

- [x] Tests pass (`pytest tests/unit/persistence tests/unit/analysis tests/unit/pipeline tests/integration/test_persistence.py`, 1360 passed)
- [ ] Web build passes (no frontend changes)

Nothing in CI runs Postgres, so this class is invisible there. Happy to open that separately rather than widen this PR.

## Checklist

- [ ] I have updated documentation if needed (no user-facing change)

### DIFF
--- a/packages/core/alembic/versions/0062_widen_symbol_name_columns.py
+++ b/packages/core/alembic/versions/0062_widen_symbol_name_columns.py
@@ -1,0 +1,78 @@
+"""Widen the symbol-name columns to ``Text`` (PostgreSQL only).
+
+Six columns store a symbol name, or a label built from one, under
+``VARCHAR(255)``. A symbol name has no length bound — generated code
+(protobuf/OpenAPI bindings, minified bundles) routinely emits identifiers past
+255 characters — and PostgreSQL enforces the declared length where SQLite
+ignores it. So the narrow type is invisible on the default backend and aborts
+the run on the one the architecture docs recommend for production, in the
+persistence phase, after the entire index has been computed (issue #1565).
+
+Reproduced on a repository with one 365-character function name:
+``StringDataRightTruncationError`` on ``INSERT INTO graph_nodes``.
+
+Only PostgreSQL is altered. SQLite does not enforce ``VARCHAR`` length — the
+columns already behave as ``TEXT`` there — and local SQLite stores never run
+Alembic anyway (``init_db``'s reconciler is additive-only), so a
+``batch_alter_table`` would rebuild six tables to change nothing.
+
+Revision ID: 0062
+Revises: 0061
+Create Date: 2026-08-18
+"""
+
+from __future__ import annotations
+
+from collections.abc import Sequence
+
+import sqlalchemy as sa
+from alembic import op
+
+# revision identifiers
+revision: str = "0062"
+down_revision: str | None = "0061"
+branch_labels: str | Sequence[str] | None = None
+depends_on: str | Sequence[str] | None = None
+
+# (table, column, nullable) — nullability is preserved, only the type changes.
+_COLUMNS: tuple[tuple[str, str, bool], ...] = (
+    ("graph_nodes", "name", True),
+    ("wiki_symbols", "name", False),
+    ("wiki_symbols", "parent_name", True),
+    ("dead_code_findings", "symbol_name", True),
+    ("health_findings", "function_name", True),
+    ("refactoring_suggestions", "target_symbol", False),
+)
+
+
+def upgrade() -> None:
+    if op.get_bind().dialect.name != "postgresql":
+        return
+    for table, column, nullable in _COLUMNS:
+        op.alter_column(
+            table,
+            column,
+            existing_type=sa.String(length=255),
+            type_=sa.Text(),
+            existing_nullable=nullable,
+        )
+
+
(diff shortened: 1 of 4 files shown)


Ayush7614 (approved):
## Review: APPROVED (with one note about CI)

The code is valid and the fix is correct:

- **Root cause is right**: SQLite ignores VARCHAR width while PostgreSQL enforces it, so a >255-char symbol name aborts the Postgres run in the persistence phase. I confirmed the fix widens the *actual* crash site (`graph_nodes.name`) and not just the reported columns.
- **Migration `0055` is sound**: correctly chains from `0054` (verified current head on upstream main — it only drops `dead_code_findings.package`, the table still exists), is PostgreSQL-only guarded, preserves nullability, and the downgrade deliberately fails rather than silently truncating a 400-char row. Follows the repo's existing migration conventions.
- **Models ↔ migration ↔ tests all aligned**: all six columns listed in `_COLUMNS` match the `Text` model changes, and the new unit test pins them to unbounded `Text` so a future narrow type can't ship unnoticed. The Postgres integration test is env-guarded and refuses non-`*test*`/`*scratch*` database names before dropping tables.
- Leave-as-narrow justification (`health_file_metrics.module`, `git_function_blame.symbol_id`) is reasonable and documented.

**One note**: the Python 3.13 unit-test job is red, but it failed on `tests/unit/distill/test_rewrite_perf.py::test_a_ledgered_invocation_stays_under_budget` — a wall-clock p95 assertion (234ms vs 200ms budget) that passed on Python 3.11 and 3.12 in the same run and exercises code this PR doesn't touch. That's a timing flake, not a code failure. Recommend re-running the 3.13 job to confirm green before merge.

RaghavChamadiya (approved):

PR #1706, opened 2026-08-18
fix(health): recognize Delphi's Test<Stem>.dpr test-file convention

Laex:
## Summary

Found while validating Pascal health markers against a real Delphi codebase: `has_test_file` (and downstream, `untested_hotspot`) reported false for `uKeymap.pas`, `uANSIParser.pas`, `uConsoleBuffer.pas`, `uVfsRegistry.pas`, `uPanelColumns.pas`, `uShellProfiles.pas`, `uDualPanelInput.pas` — every one of which has a real, exercising test.

Root cause: `_has_paired_test_file`'s candidate list is suffix-only (`test_X.py`, `X_test.go`, `X.test.ts`, `X.spec.ts`, ...) — there was no Pascal entry at all. Real-world Delphi/FPC projects don't use a same-directory same-stem test file convention; they pair a unit `uFoo.pas` with a standalone console test *program* named `TestFoo.dpr`, typically off in its own directory (`src/tools/` in this codebase, not next to the unit).

Fix: when the evaluated file is a Pascal unit (`.pas`/`.pp`/`.dpr`/`.dpk`/`.lpr`), also check for `Test<Stem>.dpr` in the basename set, stripping a **lowercase** `u` unit-prefix only (`uKeymap` → `Keymap`) — a stem that merely starts with capital `U` (`Utils.pas`) is a different word, not this naming convention, and is left alone.

This is intentionally exact-match, so it stays conservative: `uConPty.pas`'s real test is `TestConPtyEncoding.dpr` (an extra suffix), which still won't match — under-detection, not a new false positive. Re-verified against the real corpus: 7 of 9 sampled files now correctly report `has_test_file=true`; the 2 that don't (`uConPty.pas`, and `uDualPanelWindow.pas` which genuinely has no dedicated test) are honest.

## Test plan

- [x] New `TestPascalPairing` class in `tests/unit/health/test_paired_test_basenames.py` (positive pairs, an orphan unit, a same-stem-wrong-extension near-miss, and a non-Pascal file with the same stem staying unaffected)
- [x] `pytest tests/unit/health/` — 1138 passed
- [x] Re-verified against the real ~150-file Delphi codebase that surfaced this: 7/9 sampled files flip from a false `has_test_file=false` to the correct `true`


RaghavChamadiya (approved):
Thanks @Laex for checking the convention against a real 150-file Delphi tree rather than inferring it. `uKeymap.pas` pairing with `src/tools/TestKeymap.dpr` is not something you would guess from the language spec, and listing three confirmed pairs in the comment means the next person can tell whether their codebase follows it.

Two details make this safe rather than just useful. Gating on `_PASCAL_UNIT_SUFFIXES` keeps the candidate out of every other language's set, and `test_non_pascal_file_is_unaffected_by_pascal_candidate` is the test that pins it, which matters because a `uKeymap.ts` picking up a `TestKeymap.dpr` would be a silently wrong answer rather than a visible one. And stripping only a lowercase `u` is the right line to draw: `Utils.pas` is a word that starts with U, not a unit-prefixed `tils`, and you tested that too.

The near-miss case (`TestOther.pp` must not pair with `uOther.pas`) is the one I would have forgotten. Good that the extension is pinned to `.dpr`, since a runnable test program is a project file.

PR #1707, opened 2026-08-18
fix(health): recognize test coverage a name-pairing heuristic can't see

Laex:
## Summary

`is_test` (`test_paths.py` — the single, globally-consumed answer to "is this path test material": health, dead-code, generation/layers, the tour) had no mechanism for a camel-boundary *prefix* test convention. It already has one for suffixes (`test_camel_suffixes`: `FooTest.java` matches, `latest.java`/bare `Test.java` don't), but Delphi/FPC's convention is the mirror shape: a standalone console test program is named `Test<Unit>.dpr`, not `<Unit>Test.dpr`.

Adds `test_camel_prefixes` as the symmetric field (`LanguageSpec` + a `camel_test_prefix_res_by_extension()` registry compiler + `test_paths.py` wiring) and declares it for Pascal: `TestKeymap.dpr` matches, `Testing.dpr` and bare `Test.dpr` don't (needs an uppercase boundary right after `Test`, exactly mirroring the suffix rule's lowercase-boundary requirement).

Without this, Pascal's `Test*.dpr` files were never recognized as `is_test` at all — which also means #1764's newly-landed `reached_by_tests` (call-graph reachability from test files, now the primary "is this covered" signal) silently produced nothing for Pascal either, since it seeds its walk from `is_test` files. This fix is the prerequisite for that mechanism — and for any other `is_test`-consuming pass — to work on Pascal at all.

(An earlier revision of this PR also added a hand-rolled import-graph reachability check for the "one test program exercises several unrelated-stem units" case. Dropped after #1764 landed a call-graph-based version of the same idea with measured precision/recall backing the design choice not to union the import graph in — no need to carry a weaker duplicate.)

## Test plan

- [x] `TestParityGoldens::test_camel_prefix_extension_map` in `tests/unit/ingestion/test_language_capabilities.py` (registry compiler output)
- [x] `test_camel_prefix_rule_mirrors_the_suffix_rule` in `tests/unit/test_test_paths.py` (positive/negative/case-sensitivity/extension-scoping)
- [x] `pytest tests/unit/health/ tests/unit/ingestion/ tests/unit/analysis/ tests/unit/persistence/ tests/unit/server/mcp/ tests/unit/test_test_paths.py` — 6436 passed, 5 skipped
- [x] Re-verified against a real ~150-file Delphi codebase after rebasing onto #1764: `reached_by_tests` now correctly fires for Pascal files a test's calls reach (`uKeymap.pas`, `uDualPanelInput.pas`, `uFunctionBar.pas`, `uVfsTypes.pas`, `uConPty.pas`, and others in the lowest-scoring list)


Ayush7614 (approved):
## Review: APPROVED

- **Prefix rule mirrors the suffix rule correctly**: `^(?:Test)(?=[A-Z])` requires an uppercase boundary right after the prefix, so `TestKeymap.dpr` matches while `Testing.dpr` and bare `Test.dpr` don't — symmetric with the existing lowercase-boundary suffix rule. Extension-scoping and case-sensitivity are preserved (uses the original-case stem).
- **`_has_importing_test_file` is wired correctly**: I confirmed graph file nodes carry `is_test` (`graph/builder.py:217` → `parsed.file_info.is_test`), so reading `graph.nodes[pred].get("is_test", False)` from the health engine's graph is valid, not a silent no-op.
- **Correct integration point**: added to the `has_test_file` OR-chain in `_evaluate_file` after the name-pairing and inline-test checks, and the docstring correctly notes it's language-agnostic.
- **Tests**: thorough — positive/negative/case-sensitivity/extension-scoping for the prefix rule, and 6 unit cases for the import-graph fallback (direct, transitive, orphan, missing node, `None` graph, one-of-several-importers). Pascal spec on upstream main has exactly the 5 extensions the registry-map test asserts.
- Minor (non-blocking): `except Exception: return False` is broad, but acceptable defensive coding given the graph is typed `Any` and the alternative is a hard crash in health analysis.

Approved.

RaghavChamadiya (approved):
Thanks @Laex. Framing this as the mirror of `test_camel_suffixes` rather than a Pascal special case is what makes it worth taking. `FooTest.java` and `TestFoo.dpr` are the same convention pointing opposite directions, and now the registry says so in one place instead of Pascal getting a bespoke branch in `_is_test_name`.

The uppercase-boundary lookahead is the load-bearing part. `^(?:Test)(?=[A-Z])` is precisely the mirror of the suffix rule's `(?<=[a-z0-9])`, so `Testing.dpr` and bare `Test.dpr` stay out for the same structural reason `latest.java` and `Test.java` do. Pinning the compiled pattern string in `test_camel_prefix_extension_map` means a future edit to the alternation cannot loosen that silently.

Scoping by extension is what keeps the blast radius at zero here: Pascal is the only spec declaring prefixes, and `.pas/.pp/.dpr/.dpk/.lpr` belong to nobody else, so no other language can pick this up by accident. `test_TestKeymap.txt` covers that.

PR #1713, opened 2026-08-19
fix: preserve Java and C# method return types

### LINKED ISSUE #1696: [Bug] build_signature drops the return type for Java and C# and stamps them with Go's func
**Summary**

`method_declaration` is the node name for Go, Java and C# alike, but the branch
handling it only asks for Go's return type field and hardcodes Go's `func`
keyword. A Java method is therefore stored as `func bar()`, with the return type
gone and a Go keyword stamped on a Java symbol.

**Where**

`packages/core/src/repowise/core/ingestion/extractors/signatures.py:45-50`

```python
if node_type == "method_declaration":
    # Go method: include receiver text and result type
    recv_node = def_node.child_by_field_name("receiver")
    recv_text = node_text(recv_node, src) if recv_node else ""
    recv_prefix = f"{recv_text} " if recv_text else ""
    return f"func {recv_prefix}{name}{params_text}{_ret(('result',))}"
```

Java carries its return type on the `type` field, which this branch never asks
for. C# is the same node type and is equally unhandled.

**Repro**

Index a Java repository such as caffeine and read any method symbol's stored
signature. A method declared `V get(K key)` is stored as `func get(K key)`.

**Expected**

The Java signature keeps its return type and does not say `func`. Go keeps
behaving exactly as it does now.

**Impact**

Anything reading return types cannot use the symbol index for Java or C# at all
and has to go back and scan source text instead. That is what held one of our
measurements to 54% recall.

**Done looks like**

Java and C# signatures carry their declared return type and use the right
keyword, Go is unchanged and its existing tests still pass. Note that simply
widening the field list to `('result', 'type')` gets the return type back but
still prints `func` for Java, so the language needs to be considered too. Tests
for a Java method, a C# method, and a Go method with a receiver.

**If you are also looking at #1695**

Same function, same file. This can be one pull request or two. Say which you are
taking so two people do not collide.

nightcityblade:
## Summary

- Preserve Java and C# method return types without adding Go's `func` prefix.
- Keep Go receiver and result signatures unchanged.
- Add regression coverage for Java, C#, and Go method signatures.

## Related Issues

Fixes #1696

## Test Plan

- [x] Tests pass (`uv run pytest -q tests/unit/ingestion/parser/test_method_signatures.py tests/unit/ingestion/parser/test_java.py tests/unit/ingestion/parser/test_csharp.py tests/unit/ingestion/parser/test_go.py`) — 36 passed
- [x] Lint passes (`uv run ruff check` on both changed files)
- [x] Format check passes (`uv run ruff format --check` on both changed files)
- [x] Type check passes (`uv run mypy --follow-imports=skip packages/core/src/repowise/core/ingestion/extractors/signatures.py`)
- [ ] Web build passes (`npm run build`) *(N/A — no frontend changes)*

## Checklist

- [ ] All existing tests still pass *(full suite not run; all relevant parser tests pass)*
- [x] I have updated documentation if needed *(N/A — behavior fix covered by tests)*

### DIFF
--- a/packages/core/src/repowise/core/ingestion/extractors/signatures.py
+++ b/packages/core/src/repowise/core/ingestion/extractors/signatures.py
@@ -43,8 +43,14 @@ def build_signature(node_type: str, name: str, params_text: str, def_node: Node,
         # TypeScript/JavaScript class method
         return f"{name}{params_text}{_ret(('return_type',))}"
     if node_type == "method_declaration":
-        # Go method: include receiver text and result type
         recv_node = def_node.child_by_field_name("receiver")
+        declared_type_fields = ("type", "returns")
+        if recv_node is None and any(
+            def_node.child_by_field_name(field) is not None for field in declared_type_fields
+        ):
+            # Java/C#: the declared type is the return type; there is no method keyword.
+            return f"{name}{params_text}{_ret(declared_type_fields)}"
+        # Go method: include receiver text and result type.
         recv_text = node_text(recv_node, src) if recv_node else ""
         recv_prefix = f"{recv_text} " if recv_text else ""
         return f"func {recv_prefix}{name}{params_text}{_ret(('result',))}"
--- a/tests/unit/ingestion/parser/test_method_signatures.py
+++ b/tests/unit/ingestion/parser/test_method_signatures.py
@@ -1,0 +1,40 @@
+from __future__ import annotations
+
+import pytest
+
+from repowise.core.ingestion.parser import ASTParser
+from tests.unit.ingestion.parser._helpers import _make_file_info
+
+
+@pytest.mark.parametrize(
+    ("path", "language", "source", "name", "expected"),
+    [
+        (
+            "Example.java",
+            "java",
+            b"class Example { V get(K key) { return null; } }",
+            "get",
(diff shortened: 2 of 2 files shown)


Ayush7614 (approved):
## Review: APPROVED

- **Correct branching** in `signatures.py`: the `method_declaration` branch now distinguishes Go (has `receiver` field) from Java/`type` and C#/`returns` fields. The `_ret` helper's multi-field fallback means Java (uses `type`) and C# (uses `returns`) each get their declared return type without a `func` prefix.
- **No regression for Go**: any `method_declaration` with a receiver takes the existing Go path unchanged (`func (e *Example) Get(...) -> ...`).
- **PHP unaffected**: PHP also emits `method_declaration`, but its return-type field is named `return_type`, so neither `type` nor `returns` matches and PHP keeps its existing behavior — no accidental format change.
- **Tests**: the new parametrized test covers Java, C#, and Go signatures end-to-end through `ASTParser`; CI is green across Python 3.11/3.12/3.13, lint, and the rest of the suite.

Small, focused, correct fix for #1696. Approved.

RaghavChamadiya (approved):
LGTM

PR #1714, opened 2026-08-19
fix(mcp): honor configured reasoning in get_answer

### LINKED ISSUE #1560: get_answer never passes the configured reasoning mode to the provider, so REPOWISE_REASONING=off has no effect on synthesis

linhaixin45-cmyk:
## Summary

- resolve the MCP answer synthesis reasoning mode through the shared `resolve_reasoning` helper
- thread the resolved mode through `get_answer` and `synthesize` into `provider.generate`
- add focused regression coverage for environment, repo-config, orchestration, and provider-call propagation

## Root cause

`get_answer` resolved the configured provider, but it never resolved the reasoning mode. `synthesize` also omitted the `reasoning` argument when calling `provider.generate`, so the provider's default `"auto"` applied on every synthesis call. As a result, `REPOWISE_REASONING=off` and the repo-local reasoning setting had no effect on MCP answer synthesis.

## Impact

Users can now disable or select reasoning for `get_answer` through the same environment and `.repowise/config.yaml` settings used by other generation paths. This prevents reasoning-capable models from silently ignoring `off` during the 1024-token answer synthesis budget.

## Validation

- `python -m pytest tests/unit/server/mcp/test_provider_resolution.py tests/unit/server/mcp/test_answer_synthesis_timeout.py tests/unit/server/mcp/test_answer_always_synthesize.py -q` — 73 passed
- `ruff check` on the five changed files
- `ruff format --check` on the five changed files

Closes #1560

### DIFF
--- a/packages/server/src/repowise/server/mcp_server/tool_answer/answer.py
+++ b/packages/server/src/repowise/server/mcp_server/tool_answer/answer.py
@@ -179,6 +179,7 @@ from repowise.server.mcp_server.tool_answer.symbols import (
 from repowise.server.mcp_server.tool_answer.synthesis import (
     _hash_question,
     _resolve_provider_for_answer,
+    _resolve_reasoning_for_answer,
     synthesize,
 )
 
@@ -669,6 +670,7 @@ async def get_answer(
         provider,
         _SYSTEM_PROMPT,
         user_prompt,
+        reasoning=_resolve_reasoning_for_answer(getattr(ctx, "path", None)),
         session_factory=getattr(ctx, "session_factory", None),
         repo_id=repo_id,
     )
--- a/packages/server/src/repowise/server/mcp_server/tool_answer/synthesis.py
+++ b/packages/server/src/repowise/server/mcp_server/tool_answer/synthesis.py
@@ -15,6 +15,8 @@ import math
 import os
 from pathlib import Path
 
+from repowise.core.reasoning import ReasoningMode, resolve_reasoning
+from repowise.core.repo_config import load_repo_config
 from repowise.server.mcp_server.tool_answer.config import (
     _SYNTHESIS_MAX_TOKENS,
     _SYNTHESIS_TEMPERATURE,
@@ -44,6 +46,12 @@ def _hash_question(question: str) -> str:
     """Stable SHA-256 of the normalized question. Lowercase + strip + collapse ws."""
     norm = " ".join(question.lower().strip().split())
     return hashlib.sha256(norm.encode("utf-8")).hexdigest()
+
+
+def _resolve_reasoning_for_answer(repo_path: Path | None) -> ReasoningMode:
+    """Resolve the synthesis reasoning mode from env and repo config."""
+    config = load_repo_config(repo_path) if repo_path is not None else None
+    return resolve_reasoning(config=config)
 
 
 def _load_repo_provider_config(
@@ -392,6 +400,7 @@ async def synthesize(
     system_prompt: str,
     user_prompt: str,
     *,
+    reasoning: ReasoningMode = "auto",
     session_factory=None,
     repo_id: str | None = None,
 ) -> tuple[str, str | None]:
@@ -424,6 +433,7 @@ async def synthesize(
                     user_prompt=user_prompt,
                     max_tokens=_SYNTHESIS_MAX_TOKENS,
                     temperature=_SYNTHESIS_TEMPERATURE,
+                    reasoning=reasoning,
                 ),
                 None,
             )
--- a/tests/unit/server/mcp/test_answer_always_synthesize.py
+++ b/tests/unit/server/mcp/test_answer_always_synthesize.py
@@ -83,6 +83,50 @@ async def test_non_dominant_returns_prose_at_medium_with_evidence(setup_mcp, mon
 
 
 @pytest.mark.asyncio
+async def test_configured_reasoning_reaches_synthesis(setup_mcp, monkeypatch):
+    import repowise.server.mcp_server.tool_answer.answer as answer_mod
+    from repowise.server.mcp_server import get_answer
+
+    _patch_pipeline(monkeypatch, answer_mod, scores=(2.0, 1.9))
+    _patch_provider(monkeypatch, answer_mod, "The go() function drives it (pkg/alpha/one.py).")
+    monkeypatch.setenv("REPOWISE_REASONING", "off")
+    seen: dict = {}
+
+    async def _capture_synthesis(provider, system_prompt, user_prompt, **kwargs):
+        seen.update(kwargs)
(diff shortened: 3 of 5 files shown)


Ayush7614 (approved):
## Review: APPROVED

Verified against the codebase:

- `resolve_reasoning(config=...)` matches the shared helper's actual signature (`reasoning.py:52` — env `REPOWISE_REASONING` correctly takes precedence over repo config), and it's the same call pattern `job_executor.py` already uses.
- `load_repo_config` returns a dict, so `config.get("reasoning")` works; `_resolve_reasoning_for_answer(None)` safely degrades to env-only when there's no repo path.
- `provider.generate` accepts a `reasoning` kwarg across the provider base (`providers/llm/base.py:264`) and the mock used in tests — so threading it through `synthesize` into the call is correct.
- The new keyword-only `reasoning: ReasoningMode = "auto"` default keeps all existing `synthesize` callers unchanged.
- Tests: 73 passed locally (incl. the 3 new reasoning tests); ruff check + format clean; full CI green on 3.11/3.12/3.13.

Fixes a real gap — reasoning settings now actually reach MCP answer synthesis. Approved.

RaghavChamadiya (approved):
Thanks @linhaixin45-cmyk for tracing this to the one caller that dropped the parameter rather than adding a new setting on top.

The base signature has carried `reasoning: ReasoningMode = "auto"` since it was introduced (`providers/llm/base.py:254`), and every other call site passes it through. Synthesis was the one place that took the default, so a repo pinning `REPOWISE_REASONING` got it honoured everywhere except the tool most likely to need it. Resolving from env and repo config through `resolve_reasoning` is the same path the generation phase uses, so the two agree by construction.

Both tests are the right shape: one asserts the env value reaches `synthesize`, one asserts the repo-config resolution does. Green across 3.11, 3.12, 3.13 and ruff. Merging.

PR #1719, opened 2026-08-19
fix(ingestion): strip leading colon from TS return type in signatures

### LINKED ISSUE #1695: [Bug] Every TypeScript signature carries a stray colon (`-> : string`)
**Summary**

Stored TypeScript signatures read `-> : string` instead of `-> string`. The code
appends the whole return type node, and in tree-sitter-typescript that node is a
`type_annotation` that already includes its own leading colon.

**Where**

`packages/core/src/repowise/core/ingestion/extractors/signatures.py:14-19`

```python
def _ret(fields: tuple[str, ...]) -> str:
    for f in fields:
        n = def_node.child_by_field_name(f)
        if n is not None:
            return f" -> {node_text(n, src)}"
    return ""
```

Reached for TypeScript and JavaScript from line 30 (`function_declaration`) and
line 42 (`method_definition`).

**Repro**

Index any TypeScript repository and read a stored function signature. A source
function written as:

```ts
function randomString(length: number): string
```

is stored as:

```
function randomString(length: number) -> : string
```

**Expected**

`function randomString(length: number) -> string`

**Impact**

It looks cosmetic and it is not. It already cost us one wrong measurement. A
consumer splitting on `"-> "` was handed `": string"`, which then failed a type
name check, so 78 of zod's callees appeared to carry a readable return type when
the real figure was 570.

**Done looks like**

The stored signature has no stray colon. Either descend to the annotation node's
type child, or strip one leading colon where the string is joined. Either is
fine, the test is what matters. Add a parser test for a TypeScript function with
a return type and one for a class method.

Signatures are rebuilt on re-index, so no migration is needed.

sloemo01:
Fixes #1695

## Problem
TypeScript/JavaScript function signatures rendered with a stray colon: `function foo() -> : string` instead of `function foo() -> string`.

## Root cause
In `signatures.py`, `_ret` prepends ` -> ` to the return-type node's text. For TypeScript/JavaScript, the `return_type` field is a `type_annotation` node whose text **already includes a leading colon** (`: string`). Python (`int`), Rust (`i32`), and Go (`string`) return-type nodes don't include the colon, so only TS/JS were affected.

## Fix
Strip a leading colon (and any following space) from the return-type text in `_ret`, so signatures render as `-> string`. Other languages are unaffected since their return-type text doesn't start with a colon.

## Tests
Added `test_ts_signature_colon.py` covering:
- TS function with return type → `function foo() -> string` (no `-> :`)
- TS method with return type → `bar() -> number`
- TS function with no return type → `function baz()`

All pass.

### DIFF
--- a/packages/core/src/repowise/core/ingestion/extractors/signatures.py
+++ b/packages/core/src/repowise/core/ingestion/extractors/signatures.py
@@ -15,7 +15,14 @@ def build_signature(node_type: str, name: str, params_text: str, def_node: Node,
         for f in fields:
             n = def_node.child_by_field_name(f)
             if n is not None:
-                return f" -> {node_text(n, src)}"
+                text = node_text(n, src)
+                # TypeScript/JavaScript's return_type is a type_annotation
+                # node whose text already carries a leading colon (": string"),
+                # unlike Python ("int"), Rust ("i32"), or Go ("string"). Strip
+                # it so we never emit a doubled " -> : type".
+                if text.startswith(":"):
+                    text = text[1:].lstrip()
+                return f" -> {text}"
         return ""
 
     if node_type == "function_definition":
--- a/tests/unit/ingestion/test_ts_signature_colon.py
+++ b/tests/unit/ingestion/test_ts_signature_colon.py
@@ -1,0 +1,56 @@
+"""Tests for TypeScript/JavaScript signature building (stray colon fix)."""
+
+from __future__ import annotations
+
+import pytest
+from tree_sitter import Language, Parser
+
+import tree_sitter_typescript
+
+from repowise.core.ingestion.extractors.signatures import build_signature
+
+
+def _sig(code: bytes, node_type: str) -> str:
+    parser = Parser(Language(tree_sitter_typescript.language_typescript()))
+    tree = parser.parse(code)
+
+    def find(node):
+        if node.type == node_type:
+            return node
+        for child in node.children:
+            found = find(child)
+            if found is not None:
+                return found
+        return None
+
+    def_node = find(tree.root_node)
+    assert def_node is not None
+    name = ""
+    for child in def_node.children:
+        if child.type in ("identifier", "property_identifier"):
+            name = child.text.decode()
+            break
+    return build_signature(node_type, name, "()", def_node, code.decode())
+
+
+def test_ts_function_signature_no_doubled_colon() -> None:
+    sig = _sig(
+        b'function foo(a: number): string { return "x"; }',
+        "function_declaration",
+    )
+    assert sig == "function foo() -> string"
+    assert "-> :" not in sig
+
+
+def test_ts_method_signature_no_doubled_colon() -> None:
+    sig = _sig(
+        b"class C { bar(): number { return 1 } }",
+        "method_definition",
+    )
+    assert sig == "bar() -> number"
+    assert "-> :" not in sig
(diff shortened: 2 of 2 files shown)


Ayush7614 (changes requested):
## Review: CHANGES REQUESTED

The fix itself is correct and I verified it locally:

- `_ret` now strips a leading colon, which only fires for TS/JS `type_annotation` nodes — Python (`int`), Rust (`i32`), Go (`string`), and the Java/C# fields don't start with `:`, so no other language is affected.
- All 3 new tests pass locally (`test_ts_signature_colon.py`).

**However, the new test file fails RUF036 [*] `None` not at the end of the type union.
   --> packages/cli/src/repowise/cli/commands/augment_cmd/search.py:415:69
    |
415 | def _fast_pagerank_file_order(repo_path: Path, paths: list[str]) -> list[str] | None | object:
    |                                                                     ^^^^^^^^^^^^^^^^^^^^^^^^^
416 |     """``_pagerank_file_order`` over stdlib sqlite3. ``_ORM`` to fall back."""
417 |     if not _wiki_db_exists(repo_path):
    |
help: Move `None` to the end of the type union
    |
414 |
    - def _fast_pagerank_file_order(repo_path: Path, paths: list[str]) -> list[str] | None | object:
415 + def _fast_pagerank_file_order(repo_path: Path, paths: list[str]) -> list[str] | object | None:
416 |     """``_pagerank_file_order`` over stdlib sqlite3. ``_ORM`` to fall back."""
    |

RUF036 [*] `None` not at the end of the type union.
   --> packages/cli/src/repowise/cli/commands/augment_cmd/search.py:441:6
    |
439 |     result_count: int,
440 |     matched: dict[str, int] | None,
441 | ) -> str | None | object:
    |      ^^^^^^^^^^^^^^^^^^^
442 |     """Triage and the widened rescue without the ORM. ``_ORM`` to fall back.
    |
help: Move `None` to the end of the type union
    |
440 |     matched: dict[str, int] | None,
    - ) -> str | None | object:
441 + ) -> str | object | None:
442 |     """Triage and the widened rescue without the ORM. ``_ORM`` to fall back.
    |

Found 2 errors.
[*] 2 fixable with the `--fix` option.** (which CI runs as `uv run ruff check .`):

1. `F401` — `pytest` is imported but unused in `tests/unit/ingestion/test_ts_signature_colon.py:5`.
2. `I001` — import block is unsorted: `import tree_sitter_typescript` should be grouped with `from tree_sitter import ...` in the same block.

Please remove the unused `pytest` import and sort the imports (or just run `uv run ruff check --fix`), then re-verify.

Also a heads-up: this PR and #1713 both touch `signatures.py` (different hunks — `_ret` here, the `method_declaration` branch there) and are based on the same `main`. They likely merge cleanly, but it's worth rebasing one onto the other before merge so the CI run reflects the combined state.

Ayush7614 (approved):
Lint issues are fixed — re-verified after the `dcf084bf` commit (removed unused `pytest` import, sorted imports).

Verified on a simulated merge onto current `main` (clean, no conflicts):
- `ruff check` clean on both changed files
- `test_ts_signature_colon.py`: 3 passed
- Full `tests/unit/ingestion`: 2507 passed, 5 skipped — no regressions

sloemo01:

RaghavChamadiya (approved):
LGTM

PR #1720, opened 2026-08-19
fix(cli): warn when cascade budget truncates regeneration

### LINKED ISSUE #850: [Bug] Cascade budget silently caps regeneration at 50 pages
As discovered in #847, the auto-scaled budget has a hard cap of 50 pages. We need to evaluate whether this cap should be raised or if a warning should be emitted when the budget is artificially truncated so users know why some of their files were skipped.

jyotirmya17:
Fixes #850

When the `--cascade-budget` is omitted, `repowise update` automatically sets a cascade budget of up to 50 pages to prevent runaway costs on large refactors. Previously, if this 50-page cap was reached, the regeneration would silently truncate, leaving the remaining dependent pages untouched but quietly marking them as stale (decayed) without notifying the user.

This PR adds a warning that fires **only** when the 50-page auto-cap is actively binding and truncation occurs.

### Changes
- Updated `AffectedPages` in `change_detector.py` to track `stale_due_to_budget`, which captures the exact number of pages skipped due to the cascade budget.
- Added a warning in the CLI (`update_cmd/command.py`) that informs the user how many dependent pages were left stale when the budget hits the cap.
- The warning explicitly provides the command needed to regenerate the skipped pages (`--cascade-budget N`).
- Added tests in `test_change_detector.py` to assert that `stale_due_to_budget` correctly tracks skipped pages.

### DIFF
--- a/packages/cli/src/repowise/cli/commands/update_cmd/command.py
+++ b/packages/cli/src/repowise/cli/commands/update_cmd/command.py
@@ -1203,9 +1203,9 @@ def run_update(
         affected.regenerate = list(dict.fromkeys([*affected.regenerate, *stale_renderer_paths]))
         console.print(f"Pages from an older renderer: [cyan]{len(stale_renderer_paths)}[/cyan]")
 
-    if cascade_budget_is_auto and cascade_budget == 50 and getattr(affected, "stale_due_to_budget", 0) > 0:
+    if affected.stale_due_to_budget > 0:
         console.print(
-            f"\n[yellow]⚠ Cascade budget capped at 50 pages. "
+            f"\n[yellow]⚠ Cascade budget of {cascade_budget} pages was reached. "
             f"{affected.stale_due_to_budget} dependent pages were skipped and marked stale.[/yellow]"
         )
         console.print(
--- a/packages/cli/src/repowise/cli/commands/update_cmd/command.py
+++ b/packages/cli/src/repowise/cli/commands/update_cmd/command.py
@@ -1177,7 +1177,8 @@ def run_update(
         emitter.stage("plan_pages")
 
     # Determine affected pages (auto-scale budget if not explicitly set)
-    if cascade_budget is None:
+    cascade_budget_is_auto = cascade_budget is None
+    if cascade_budget_is_auto:
         from repowise.core.ingestion.change_detector import compute_adaptive_budget
 
         cascade_budget = compute_adaptive_budget(file_diffs, file_count)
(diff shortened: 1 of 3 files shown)


Ayush7614 (changes requested):
## Review: CHANGES REQUESTED

Two problems, one of them blocking.

### 1. CRITICAL — the bundled Python submodule change is broken against current `main`

This PR carries two unrelated changes: the cascade-budget warning (the title/body, #850) and a Python call-resolver fix that isn't mentioned in the description at all. Since this branch forked, `main` refactored `ResolvedCall` in `call_resolver.py` to a **5-field** dataclass (added required `origin: ResolutionOrigin`), but the new fallback still constructs it with 4 args:

```
return ResolvedCall(caller_id, cand_syms[method_name], 0.88, call.line)
```

I simulated the merge (clean, per GitHub) and ran the PR's own test against the merged tree:

```
TypeError: ResolvedCall.__init__() missing 1 required positional argument: 'origin'
```

So if merged as-is, this line crashes at runtime and fails type checking. Fix: rebase on current `main` and pass an `origin=...` (e.g. `"receiver_import"`), **or better** — drop the Python change from this PR entirely and ship it separately. The cascade-budget work stands on its own and shouldn't carry an unrelated, now-stale resolver change.

### 2. Scope — PR bundles unrelated work

The Python submodule resolution (call_resolver.py + `test_python_submodules.py`) is absent from the PR description. Even after the break is fixed, it belongs in its own PR.

### What's good (verified)

- The cascade-budget warning itself is correct: `stale_due_to_budget` is computed accurately (`len(all_pages_needing_regen) - cascade_budget`), the CLI only warns when the auto cap is binding *and* truncation actually occurred (`stale_due_to_budget > 0`), and `--cascade-budget N` is suggested. `test_change_detector.py` (20 tests) passes, and the warning merges cleanly.
- The `[:-11]` slicing is correct (`__init__.py` is 11 chars → `pkg/`), and `test_python_submodules.py` passes on the branch itself.

### Minor

- The warning only fires when the auto budget equals the 50 hard cap; truncation at the smaller auto budgets (10/30) stays silent. Documented as intended, but worth a comment.
- `50` is duplicated between `compute_adaptive_budget` and the CLI check — a shared constant would be more robust.
- No CLI-level test for the warning message itself (only `change_detector` unit coverage).

Ayush7614 (changes requested):
Re-checked after the new commits (`f46795e7`, `83594a48`, `ad076483`). The cascade-budget warning part is fine, but the bundled Python submodule change **still breaks when merged onto current `main`**.

**Concrete failure (simulated merge onto `main`, clean no-conflict merge):**

`tests/unit/ingestion/test_python_submodules.py::test_python_submodule_resolution` FAILS:

```
packages/core/src/repowise/core/ingestion/call_resolver.py:1080: in _resolve_member_call
    return ResolvedCall(caller_id, cand_syms[method_name], 0.88, call.line)
TypeError: ResolvedCall.__init__() missing 1 required positional argument: 'origin'
```

`ResolvedCall` on current `main` is a 5-field dataclass — `origin: ResolutionOrigin` was added (via the MCP reasoning-propagation PR) after your branch forked. Your new fallback at `call_resolver.py:1080` still constructs it with 4 positional args. The test passes on your branch only because your branch predates the `origin` field; it fails on the merge, so CI would be red.

**Fix options**
1. Add the `origin` argument, e.g. `ResolvedCall(caller_id, cand_syms[method_name], 0.88, call.line, "receiver_same_package")` (matches the neighboring receiver-based resolutions).
2. Or split the Python submodule change into its own PR rebased on current `main` — the cascade-budget warning is otherwise good and merges cleanly (the `change_detector` tests pass).

RaghavChamadiya (approved):
Verified against current main; merging.

PR #1721, opened 2026-08-19
fix(ingestion): capture TypeScript #private class members and calls

### LINKED ISSUE #1715: [Bug] TypeScript #private class members are invisible at both ends
**Summary**

A class member declared with the ECMAScript private prefix, `#foo`, produces no
symbol, and a call to it produces no call site. Tree-sitter spells `#foo` as
`private_property_identifier`, and neither the symbol pattern nor the member
call pattern accepts that node type.

**Where**

`packages/core/src/repowise/core/ingestion/queries/typescript.scm:47-51`

```scheme
; Method inside class body
(method_definition
  name: (property_identifier) @symbol.name
  parameters: (formal_parameters) @symbol.params
) @symbol.def
```

`packages/core/src/repowise/core/ingestion/queries/typescript.scm:226-232`

```scheme
(call_expression
  function: (member_expression
    object: [(identifier) (this)] @call.receiver
    property: (property_identifier) @call.target
  )
  arguments: (arguments) @call.arguments
) @call.site
```

`queries/tsx.scm` carries the same two patterns and needs the same change.

`grep -rn private_property_identifier packages/core/src/repowise/core/ingestion/queries/`
returns nothing.

**Repro**

Parse this file:

```ts
export class Cache {
  #store = new Map();

  #evict(key: string): void {
    this.#store.delete(key);
  }

  set(key: string, value: string): void {
    this.#evict(key);
  }

  clear(): void {
    this.set("a", "b");
  }
}
```

Observed:

```
SYMBOLS:
   class      'Cache'  line 1
   method     'set'  line 8
   method     'clear'  line 12
CALL SITES:
   line 13  receiver='this'  target='set'
```

`#evict` is not in the symbol list, and the call to it on line 9 is not in the
call site list. Only the ordinary `this.set(...)` on line 13 survives.

**Expected**

`#evict` appears as a method symbol, and line 9 appears as a call site with
receiver `this` and target `#evict` or `evict`. Pick one spelling for the target
and keep it consistent between the two patterns, otherwise the call will not
resolve to the symbol.

**Impact**

A private method is dead to the graph in both directions. It never appears as a
caller and never as a callee, so it reads as unreferenced. On one benchmark
repository every `#`-prefixed callee in the missed-edge set was missed for this
reason, around 22 of 200 missed edges.

**Done looks like**

Both patterns accept an alternation, `[(property_identifier) (private_property_identifier)]`,
in `typescript.scm` and in `tsx.scm`. A parser test covers a class with a
`#private` method that is called from a public method, and asserts one symbol
and one call site.

Resolution needs no new code: a `#` name is private to the class by language
rule, so it resolves through the existing same-file scope path.

sloemo01:
Fixes #1715

## Problem
ECMAScript `#private` class members (`#foo`) produced **no symbol** and **no call site** — invisible at both ends.

## Root cause
tree-sitter-typescript spells `#foo` as `private_property_identifier`. The `method_definition` symbol pattern only accepts `property_identifier`, and the member-call pattern only accepts `property_identifier` for the property, so private methods were dropped entirely.

## Fix
In `queries/typescript.scm`:
- Add a `method_definition` symbol pattern that accepts `private_property_identifier`
- Widen the member-call `property:` alternation to include `private_property_identifier`

(`queries/tsx.scm` only contains JSX-specific additions and doesn't duplicate these patterns.)

## Tests
Added `TestTypeScriptPrivateMembers` to `test_typescript.py`:
- `#increment` appears as a method symbol
- `this.#increment()` produces a call site with target `#increment`

All 45 TS parser tests pass.

### DIFF
--- a/packages/core/src/repowise/core/ingestion/queries/typescript.scm
+++ b/packages/core/src/repowise/core/ingestion/queries/typescript.scm
@@ -50,6 +50,12 @@
   parameters: (formal_parameters) @symbol.params
 ) @symbol.def
 
+; Private method inside class body (ECMAScript #private members)
+(method_definition
+  name: (private_property_identifier) @symbol.name
+  parameters: (formal_parameters) @symbol.params
+) @symbol.def
+
 ; Arrow function assigned to const/let: const foo = (...) => { }
 (lexical_declaration
   (variable_declarator
@@ -226,7 +232,7 @@
 (call_expression
   function: (member_expression
     object: [(identifier) (this)] @call.receiver
-    property: (property_identifier) @call.target
+    property: [(property_identifier) (private_property_identifier)] @call.target
   )
   arguments: (arguments) @call.arguments
 ) @call.site
--- a/tests/unit/ingestion/parser/test_typescript.py
+++ b/tests/unit/ingestion/parser/test_typescript.py
@@ -630,3 +630,36 @@ export const handler = onCall(async () => 1);
 
     def test_class_and_function_expression_bindings_are_indexed(self, parser: ASTParser) -> None:
         assert {"KlassExpr", "fnExpr"} <= self._names(parser)
+
+
+class TestTypeScriptPrivateMembers:
+    """ECMAScript #private class members should produce a symbol and a call site.
+
+    Regression for #1715: tree-sitter spells ``#foo`` as
+    ``private_property_identifier``, which the symbol and member-call query
+    patterns did not accept, so private methods were invisible at both ends.
+    """
+
+    def test_private_method_produces_symbol(self, parser: ASTParser) -> None:
+        fi = _make_file_info("src/counter.ts", "typescript")
+        src = b"""\
+class Counter {
+  #count = 0;
+  #increment() { this.#count++; }
+  inc() { this.#increment(); return this.#count; }
+}"""
+        result = parser.parse_file(fi, src)
+        method_names = [s.name for s in result.symbols if s.kind == "method"]
+        assert "#increment" in method_names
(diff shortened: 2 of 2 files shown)


Ayush7614 (approved):
## Review: APPROVED

Verified against the codebase:

- `typescript.scm` change is correct: ECMAScript private members are `private_property_identifier` nodes in tree-sitter-typescript, and the two query patterns (method symbol + member-call target) previously only accepted `property_identifier`. The added alternation `[(property_identifier) (private_property_identifier)]` and the private-method symbol pattern don't collide with the existing patterns (they match different node types), and since `tsx.scm` only layers JSX additions on top, `.tsx` files are covered too.
- Tests: ran `test_typescript.py` locally — all 45 pass, including the two new `TestTypeScriptPrivateMembers` cases (`#increment` shows up as a method symbol and as a call target). Ruff is clean on the test file.

One note: the repo CI for this PR is currently `action_required` (first-time contributor — workflows awaiting maintainer approval), so only the Repowise code-health check has run. Local verification is solid. Approved.

RaghavChamadiya (approved):
Checking the `tsx.scm` point is what makes this ready to merge as-is. The issue says that file carries the same two patterns and needs the same change; it does not. `tsx.scm` is a JSX-only supplement that `parser._load_compiled_query` appends to `typescript.scm` for the tsx grammar variant, so `.tsx` files pick this up from the one edit. You said that in the description and you were right.

Both halves are covered too, the symbol and the call target, which is what the issue meant by invisible at both ends.

PR #1725, opened 2026-08-19
fix(ingestion): probe index.tsx and index.jsx for relative TS imports

### LINKED ISSUE #1717: [Bug] Relative TypeScript imports never probe index.tsx or index.jsx
**Summary**

Resolving a relative import to a directory index tries four filenames and misses
the two React ones. The same resolver already accepts `.tsx` and `.jsx` as bare
extensions, and the tsconfig path resolver in the same package already lists
both index forms, so this is an inconsistency inside our own code rather than a
missing feature.

**Where**

`packages/core/src/repowise/core/ingestion/resolvers/typescript.py:28-39`

```python
        exts: tuple[str, ...] = (
            ".ts",
            ".tsx",
            ".mts",
            ".cts",
            ".js",
            ".jsx",
            "/index.ts",
            "/index.mts",
            "/index.cts",
            "/index.js",
        )
```

`.tsx` and `.jsx` are present as bare extensions and absent as index forms.

Compare `packages/core/src/repowise/core/ingestion/tsconfig_resolver.py:50-57`,
which does list both:

```python
_INDEX_FILES = (
    "index.ts",
    "index.tsx",
    "index.mts",
    "index.cts",
    "index.js",
    "index.jsx",
    ...
```

**Repro**

In any TypeScript project, create `src/components/Button/index.tsx` exporting a
component, and import it from a sibling file as `./components/Button`. Index the
repository. The import does not resolve, so no edge is drawn to the component.

Renaming the file to `index.ts` resolves it, which isolates the cause to the
filename list.

**Expected**

`./components/Button` resolves to `components/Button/index.tsx`.

**Impact**

A directory holding a React component behind an `index.tsx` is a common layout.
Every import of it fails to resolve, so the component has no incoming edges and
can be reported as unreachable.

The gain has not been measured on our benchmark corpora, since neither is
`.tsx`-heavy. It is listed as a correctness fix rather than as a scored one.

**Done looks like**

`"/index.tsx"` and `"/index.jsx"` are added to the tuple, next to the existing
index entries. A resolver test imports a directory containing `index.tsx` and
asserts the resolved path.

sloemo01:
Fixes #1717

## Problem
A relative import to a directory probed `/index.ts` and `/index.js` but missed the two React index forms (`/index.tsx`, `/index.jsx`) that the bare-extension list and the tsconfig path resolver both accept. A directory-index React component therefore resolved to nothing and read as an unreachable external dep.

## Fix
Add `/index.tsx` and `/index.jsx` to the relative-import probe list, keeping the existing ordering (`/index.ts` before `/index.tsx`, matching the bare-extension order).

## Test
Parametrized test covering every index extension plus a ts-before-tsx ordering guard. `tests/unit/ingestion/test_typescript_resolver.py`: 70 passed.

### DIFF
--- a/packages/core/src/repowise/core/ingestion/resolvers/typescript.py
+++ b/packages/core/src/repowise/core/ingestion/resolvers/typescript.py
@@ -34,9 +34,11 @@ def resolve_ts_js_import(module_path: str, importer_path: str, ctx: ResolverCont
             ".js",
             ".jsx",
             "/index.ts",
+            "/index.tsx",
             "/index.mts",
             "/index.cts",
             "/index.js",
+            "/index.jsx",
         )
         if ctx.has_sfc_files:
             # Both the bare file (``./Foo`` -> ``Foo.vue``) and the directory
--- a/tests/unit/ingestion/test_typescript_resolver.py
+++ b/tests/unit/ingestion/test_typescript_resolver.py
@@ -79,6 +79,46 @@ class TestExplicitRelativeExtensions:
         assert result == "data/example.ts"
 
 
+class TestDirectoryIndexProbing:
+    """A relative import to a directory must probe its index files.
+
+    The resolver already probed ``/index.ts`` and ``/index.js`` but missed the
+    two React index forms (``/index.tsx``, ``/index.jsx``) that the bare
+    extension list and the tsconfig path resolver both accept — so a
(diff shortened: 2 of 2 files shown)


sloemo01:

swati510 (approved):

PR #1726, opened 2026-08-19
fix(ingestion): keep package qualifier on Go embeds

### LINKED ISSUE #1716: [Bug] A qualified Go embed drops its package and can bind an interface to itself
**Summary**

Go embedding is extracted by taking the bare type name, which strips the package
qualifier. `io.Reader` becomes `Reader`, so an embed of a standard library type
binds to whatever repo-local type happens to share the short name. When the
enclosing type has that same name, the type inherits from itself.

**Where**

`packages/core/src/repowise/core/ingestion/extractors/heritage/go.py`

Both branches call `bare_type_name` on the embed text before recording it, the
struct branch on a field with no name and the interface branch on a `type_elem`.

`bare_type_name` is shared and is documented as shape-only, so **it is not the
thing to change**. It is doing its job. The Go heritage extractor is the caller
that needs to keep the qualifier and resolve it.

**Repro**

Parse this file:

```go
package store

import "io"

type Reader interface {
	io.Reader
	Name() string
}

type Store struct {
	io.Closer
}
```

Observed:

```
HERITAGE:
   Reader --extends--> Reader   (line 5)
   Store --mixin--> Closer   (line 10)
```

The first line is a self-edge. `Reader` inherits from `Reader`, which is itself.
The second binds `io.Closer` to a bare `Closer`, which will match any type
called `Closer` the repository declares.

**Expected**

Neither relation is recorded. `io.Reader` and `io.Closer` are declared in the
standard library, not in this repository, so there is no local parent to point
at. A heritage relation should only be recorded when the qualifier resolves to a
file in the repository.

**Impact**

Two effects, and the second is the expensive one.

A self-edge is wrong on its face and can make a type appear to inherit its own
method set.

Binding a standard library embed to an unrelated local type inflates the
interface's method set. Downstream, structural interface satisfaction is
computed by checking whether a type's methods are a superset of the interface's
methods, so an inflated set makes real implementations stop qualifying and
suppresses correct `method_implements` edges.

**Done looks like**

A qualified embed keeps its qualifier and is resolved through the file's import
table. If the qualifier names an external package, no relation is recorded. An
unqualified embed keeps its current behaviour.

Tests: the file above produces no heritage relations. A second test where the
embedded type is declared in the same repository still produces the relation.

Do not change `bare_type_name` or its tests. Other callers depend on it
stripping qualifiers.

sloemo01:
Fixes #1716

## Problem
Go heritage extraction stripped the package qualifier via `bare_type_name`, so an embed of `io.Reader` became `Reader` and bound to whatever repo-local type shared the short name — and when the enclosing type had that same name, the type inherited from itself.

## Fix
Keep the qualifier (strip only type arguments) on both the struct and interface embed branches of `extractors/heritage/go.py`.

## Test
Pinned the new behaviour in the heritage type-name table (`ns.Qual` → `ns.Qual`, `ns.Both[Arg]` → `ns.Both`) and added a self-inheritance regression test asserting `io.Reader` stays `io.Reader` and never binds to a local `Reader`. `test_go_extractors.py` + `test_heritage_type_names.py`: 64 passed.

### DIFF
--- a/packages/core/src/repowise/core/ingestion/extractors/heritage/go.py
+++ b/packages/core/src/repowise/core/ingestion/extractors/heritage/go.py
@@ -5,7 +5,7 @@ from __future__ import annotations
 from tree_sitter import Node
 
 from ...models import HeritageRelation
-from ...type_names import bare_type_name
+from ...type_names import strip_type_arguments
 from ..helpers import node_text
 
 
@@ -33,12 +33,16 @@ def _extract_go_heritage(
             type_child = field_decl.child_by_field_name("type")
             if name_node is None and type_child is not None:
                 parent = node_text(type_child, src).strip().lstrip("*")
-                bare = bare_type_name(parent)
-                if bare:
+                # Keep the package qualifier: ``io.Reader`` must stay
+                # ``io.Reader``, not ``Reader``, or an embed of a stdlib type
+                # binds to whatever repo-local type shares the short name (and
+                # can inherit from itself). Type arguments are still stripped.
+                parent = strip_type_arguments(parent)
+                if parent:
                     out.append(
                         HeritageRelation(
                             child_name=name,
-                            parent_name=bare,
+                            parent_name=parent,
                             kind="mixin",
                             line=line,
                         )
@@ -53,12 +57,13 @@ def _extract_go_heritage(
             # is a generic bound, carrying no methods to inherit.
             if "|" in text or "~" in text:
                 continue
-            bare = bare_type_name(text)
-            if bare:
+            # Keep the package qualifier (see the struct branch above).
+            parent = strip_type_arguments(text)
+            if parent:
                 out.append(
                     HeritageRelation(
                         child_name=name,
-                        parent_name=bare,
+                        parent_name=parent,
                         kind="extends",
                         line=line,
                     )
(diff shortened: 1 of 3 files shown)


sloemo01:

RaghavChamadiya (approved):
This is the right fix and the self-inheritance test is the one that matters, since that is the case where the old behaviour was not just imprecise but circular.

Worth recording why the qualifier can be kept safely here. Every tier in `heritage_resolver._resolve_one` is an exact-string lookup keyed on the bare type name, so `io.Reader` now matches nothing and the relation is dropped rather than bound to the wrong type. Go is also absent from `_INHERITED_LANGUAGES`, so no ancestor walk was depending on those edges.

The codebase already carries the argument for this, in `call_resolver.py:66`:

```python
# C++ is absent because its heritage binds a qualified external
# parent to a same-named local type, which puts unrelated siblings in one
# hierarchy before this walk even runs.
```

That is the same bug you just fixed for Go, written down as a reason to switch a whole tier off for C++. Go was in the same trap without anyone noticing.

Two follow-ups this opens, both out of scope here and neither a blocker. A same-repo cross-package embed like `store.Cache` now drops instead of resolving, because no tier maps a qualifier to the imported package's files, and C++ still binds qualified parents to local types. I will file those separately.

PR #1727, opened 2026-08-19
fix(ingestion): capture Go method calls through a field

### LINKED ISSUE #1718: [Bug] Go method calls through a field produce no call site
**Summary**

A Go call written `a.b.Method()`, where the receiver is a field rather than a
plain variable, matches no call pattern, so no call site is recorded at all. The
call is missing from the graph rather than present and unresolved, which also
means it does not appear in any recall or coverage figure as a miss.

**Where**

`packages/core/src/repowise/core/ingestion/queries/go.scm:58-65`

```scheme
; Method call: obj.Method(args)
(call_expression
  function: (selector_expression
    operand: (identifier) @call.receiver
    field: (field_identifier) @call.target
  )
  arguments: (argument_list) @call.arguments
) @call.site
```

`operand:` requires an `identifier`. When the operand is itself a
`selector_expression`, as in `a.b.Method()`, this pattern does not match. The
chained pattern below it handles `operand: (call_expression)`, so
`f().Method()` is captured, but there is no pattern for a selector operand.

**Repro**

```go
package p

type inner struct{}

func (i inner) Do() {}

type outer struct{ in inner }

func (o outer) Run() {
	o.in.Do()
}
```

`o.in.Do()` produces no call site.

**Expected**

A call site is recorded with target `Do`. The receiver expression should be
captured so the resolver has something to type.

**Impact**

Measured against the Go compiler's own call graph on a large repository, method
calls through a field are missed 80% of the time, against 40% for a method call
on a plain local or parameter. That is 189 missed edges on one repository.

**Notes for whoever picks this up**

Capture and resolution should land separately, and the capture is the safe half.

Once a call site exists, the resolver has to decide what the receiver names. Do
not route the composed text to the bare-name global lookup: that tier resolves on
name alone and was hand-measured at about 29% precision, so feeding it a
composed receiver would add wrong edges faster than right ones. Route it to the
typed receiver path only, and drop the call site if that path declines to answer.

An equivalent change for Java was built and measured at zero net gain, and was
refused for that reason. Go was never attempted, so that result does not settle
this one, but it is a fair warning that capture alone may not move the number.

**Done looks like**

A new pattern accepts a `selector_expression` operand and records the receiver.
A parser test covers `o.in.Do()` and asserts one call site. A second commit,
measured separately, decides whether the resolver can type it.

sloemo01:
Fixes #1718

## Problem
A call written `a.b.Method()` — receiver is a field, not a plain variable — matched no call pattern, so no call site was recorded at all. The call was missing from the graph rather than present and unresolved, which also meant it did not appear in any recall or coverage figure as a miss.

## Fix
Add a pattern accepting a `selector_expression` operand and capture the whole receiver expression so the resolver has something to type. This is the **capture half only**, per the issue's guidance: resolution is a separate, measured change and must not route the composed receiver to the bare-name lookup.

## Test
Parser test asserting `o.in.Do()` yields one call site with receiver `o.in`, plus a guard that a plain `i.Do()` still captures. `test_go_extractors.py`: 10 passed; broader Go suite: 83 passed.

### DIFF
--- a/packages/core/src/repowise/core/ingestion/queries/go.scm
+++ b/packages/core/src/repowise/core/ingestion/queries/go.scm
@@ -64,6 +64,20 @@
   arguments: (argument_list) @call.arguments
 ) @call.site
 
+; Method call through a field: a.b.Method(args)
+; The operand is itself a selector_expression (``o.in`` in ``o.in.Do()``),
+; which the identifier-only pattern above does not match, so the call was
+; missing from the graph entirely. Capture the whole receiver expression so
+; the resolver has something to type; resolution is a separate, measured
+; change and must not route the composed text to the bare-name lookup.
+(call_expression
+  function: (selector_expression
+    operand: (selector_expression) @call.receiver
+    field: (field_identifier) @call.target
+  )
+  arguments: (argument_list) @call.arguments
+) @call.site
+
 ; Package-qualified call: pkg.Function(args)
 ; (same pattern as method call — receiver is the package alias)
 ; Captured by the selector_expression pattern above.
--- a/tests/unit/ingestion/test_go_extractors.py
+++ b/tests/unit/ingestion/test_go_extractors.py
@@ -80,3 +80,42 @@ class TestGoMethodReceiver:
         reset = [s for s in result.symbols if s.name == "reset"]
         assert reset
         assert reset[0].parent_name == "cache"
+
+
+class TestGoCallThroughField:
+    """A method call whose receiver is a field (``o.in.Do()``) must be captured.
+
+    The operand of the selector is itself a selector_expression, which the
+    identifier-only method-call pattern did not match — so the call was missing
+    from the graph entirely rather than present and unresolved. Capture the
+    whole receiver expression; resolution is a separate, measured change.
+    """
+
+    def test_field_receiver_call_site_is_captured(self, parser: ASTParser) -> None:
+        src = (
+            b"package p\n\n"
+            b"type inner struct{}\n\n"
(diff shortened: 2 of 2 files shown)


sloemo01:

RaghavChamadiya (approved):
Your CI red here was never yours. The three failures on the last run were `test_agent_matrix.py::test_a_published_tool_count_matches_the_registry` and its superseded-count sibling, all on `docs/business/COMMERCIAL.md`, and main fixed that in #1771 about an hour after your run started. Nothing in the Go matrix failed.

I checked rather than assuming. Your patch applied to current main gives:

```
tests/unit/ingestion/test_go_extractors.py + tests/unit/cli/test_agent_matrix.py
95 passed

tests/unit/ingestion/
2508 passed, 1 skipped
```

So I am merging on that instead of making you rebase for a signal you already earned.

On the change itself, holding the line at capture and refusing to touch resolution is the right instinct, and I traced what happens to `o.in` to confirm it costs nothing. A receiver-carrying site goes to `_resolve_member_call` and never reaches the bare-name free-call path, so the composed text cannot reach a global-unique lookup on `Do`. It then misses every member tier in turn, the Go package tier, the typed-receiver fallback, `_file_methods`, the alias and pair indexes, and resolves to `None`. The site is recorded and unresolved, which is exactly what the issue asked for and what makes the miss countable.

The second test earning its place is worth saying too. Widening a `.scm` pattern is the kind of change that quietly steals matches from the pattern beside it, and pinning that plain `i.Do()` still lands with one call site is what rules that out.

PR #1731, opened 2026-08-19
fix(ingestion): only treat a generated-file banner as generated

### LINKED ISSUE #1519: [Bug] _is_generated substring-matches "AUTO-GENERATED" in prose docblocks, silently dropping hand-written files from the index

sloemo01:
Fixes #1519

## Problem
`_is_generated` substring-matched the whole 512-byte header against the marker list, so a marker like `AUTO-GENERATED` that occurs naturally in prose silently dropped hand-written files from the index (19 endpoint files on one Payload CMS repo).

## Fix
A generated-file banner is a header convention: it sits on the first line or two, never buried in a paragraph. Require the marker within the first two lines, which fixes the general class of false positives for all six markers rather than just this one.

## Test
Adds tests for a first-line banner, a second-line banner, the exact prose false positive from the issue, and the generated-suffix path still winning. `test_traverser.py`: 101 passed.


RaghavChamadiya (approved):
Correcting myself: I said above that this was parked at `action_required`. That was true of the 09:37 run and I read the stale one. Your 11:31 rebase did trigger a fresh run and it is green across 3.11, 3.12, 3.13 and ruff, on this exact head. Nothing was ever blocked here. Merging.

PR #1732, opened 2026-08-19
fix(cli): reindex exits non-zero when every embed failed

### LINKED ISSUE #1495: reindex reports "Done!" and exits 0 when every embedding call failed
## Summary

`repowise reindex --embedder ollama` prints `Done!` and **exits 0** when *every* embedding call failed. Nothing in the exit status distinguishes a complete failure from a complete success, so an automated pipeline (or an agent) treats an empty vector index as a successful build.

## Repro

On a machine with ollama running but **without** the default embedding model pulled:

```
$ ollama pull nomic-embed-text     # any model that is NOT the default
$ repowise reindex . --embedder ollama
Using ollama embedder
Found 2891 wiki pages and 26 decision records to index.
  Warning: failed to embed symbol_spotlight:packages/shared/src/ids.ts::entityIdRefine:
    Client error '404 Not Found' for url 'http://localhost:11434/api/embed'
  ... (repeated per page)
  Indexing decisions... ━━━━━━ 2917/2917 0:00:13

Done! Indexed 0 items (2917 failed) -> /path/.repowise/lancedb
$ echo $?
0
```

`Indexed 0 items (2917 failed)` — and a zero exit code, under the word `Done!`.

## Why it matters

The failure is recoverable and obvious *if a human reads the last line*, but:

- exit 0 means CI, `&&` chains and agent tooling all proceed as if the index is built
- semantic search then silently degrades to keyword-only. Every hit comes back `"sources": ["fts"]`, and `mode: "concept"` returns byte-identical results to `mode: "auto"`. Retrieval quality drops measurably (on our repo, answer coverage went 0.80 → 0.67 across a fixed 15-question set) with no error anywhere
- the per-page warning is a raw `httpx` 404 string. A 404 from `/api/embed` almost always means "that model is not pulled", but nothing says so

## Suggested fix

1. Exit non-zero when `indexed == 0 && failed > 0` (arguably whenever `failed > 0`), and don't lead with `Done!`.
2. Fail fast: one probe embed before the loop, so 2,917 identical 404s become one clear error.
3. Make the 404 actionable. The default ollama model is `embeddinggemma` (`_DEFAULT_MODEL` in `core/providers/embedding/ollama.py`) and is not stated anywhere in `reindex --help`; a 404 could suggest `ollama pull embeddinggemma` or name the `OLLAMA_EMBEDDING_MODEL` / `REPOWISE_EMBEDDING_MODEL` override.

## Related

#1370 covers *degradation warnings* when a requested embedder is substituted — adjacent but distinct. Here the requested embedder is used, every call fails, and the command still reports success.

## Environment

repowise from `main` (post-#1454), Python 3.14, macOS 24.6, ollama 0.32.9.

sloemo01:
Fixes #1495

## Problem
`repowise reindex` printed `Done!` and exited 0 even when every embedding call failed, so an automated pipeline (or an agent) treated an empty vector index as a successful build.

## Fix
Exit non-zero (`click.Abort`) when nothing was indexed but items failed, so the failure is visible in the exit status rather than only in the printed count. A run that indexed nothing and failed nothing (empty wiki) still succeeds.

## Test
Adds a test asserting `click.Abort` is raised when every item fails to embed, and that an empty wiki (nothing indexed, nothing failed) does not abort. `test_reindex_cmd.py`: 2 passed.

### DIFF
--- a/packages/cli/src/repowise/cli/commands/reindex_cmd.py
+++ b/packages/cli/src/repowise/cli/commands/reindex_cmd.py
@@ -253,3 +253,10 @@ async def _reindex(repo_path, embedder_name: str, batch_size: int) -> None:
         + (f" ({below_floor} held back as too thin to index)" if below_floor else "")
         + f" -> {lance_dir}"
     )
+
+    # A reindex that indexed nothing but failed on every item is a failed
+    # build, not a success: an automated pipeline (or an agent) must not treat
+    # an empty vector index as a successful reindex. Exit non-zero so the
+    # failure is visible in the exit status, not just in the printed count.
+    if indexed == 0 and failed > 0:
+        raise click.Abort()
--- a/tests/unit/cli/test_reindex_cmd.py
+++ b/tests/unit/cli/test_reindex_cmd.py
@@ -5,6 +5,9 @@ from __future__ import annotations
 from pathlib import Path
 from typing import Any
 
+import click
+import pytest
+
 from repowise.cli.commands import reindex_cmd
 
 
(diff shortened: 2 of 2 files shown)


sloemo01:

RaghavChamadiya (approved):
Correcting the CI note in my comment above: I read the superseded 09:37 run. Your 11:31 rebase ran and is green across 3.11, 3.12, 3.13 and ruff on this head. Merging.

PR #1733, opened 2026-08-19
fix(ingestion): resolve module.symbol() calls through submodule bindings

### LINKED ISSUE #1193: dead-code: unused_export false-positive on module.symbol(...) attribute calls (only direct 'from x import symbol' resolves)
## Summary

`repowise dead-code --safe-only` flags a function as `unused_export` ("no importers", `safe_to_delete: true`) when it's only ever called through an attribute-style call on the imported module (`module.symbol(...)`), even though a real caller exists. Only a direct `from x import symbol` + bare `symbol()` call site is credited as usage — the import's scope (module-level vs. function-local) doesn't matter, and neither does the confidence/safe-only threshold.

This makes `unused_export`/`unused_internal` findings unsafe to auto-delete on any codebase that uses `import module` / `from pkg import module` + `module.func()` — a very common Python idiom (e.g. for lazy/dispatch-table style calling). We hit this running the automated dead-code cleanup job unattended: every candidate across 3 separate weekly runs (15 symbols total) was a false positive of exactly this shape, and a fresh scan of a ~1,800-file / ~36k-symbol repo just now found 209 `unused_export` findings with the same pattern, 0 `unreachable_file`/`zombie_package` findings (which don't share this resolution path).

## repowise version

0.13.0

## Minimal reproduction

```
.
├── sensors/
│   ├── __init__.py   (empty)
│   └── foo.py
├── caller.py
└── main.py
```

`sensors/foo.py`:
```python
def check():
    return "ok"
```

`caller.py` (module-level import, attribute-style call):
```python
from sensors import foo


def run():
    return foo.check()
```

`main.py` (so `run` itself isn't also flagged, isolating the bug to `check`):
```python
import caller

if __name__ == "__main__":
    print(caller.run())
```

```
$ repowise dead-code . --format json --safe-only
[
  {
    "kind": "unused_export",
    "file_path": "sensors/foo.py",
    "symbol_name": "check",
    "confidence": 0.7,
    "reason": "Public symbol 'check' has no importers",
    "safe_to_delete": true,
    "lines": 1,
    "primary_owner": "..."
  }
]
```

`check` has a real caller (`caller.py:run`, which is itself called from `main.py`), but is reported as safe-to-delete at confidence 0.7.

## Control — the only shape that resolves correctly

Changing `caller.py` to a direct symbol import + direct call makes the finding disappear entirely:

```python
from sensors.foo import check


def run():
    return check()
```

```
$ repowise dead-code . --format json --safe-only
[]
```

So the resolver correctly traces `from x import symbol; symbol()`, but not `from x import module; module.symbol()` — regardless of whether the import itself is at module scope or inside a function body (I tested both; same result).

## Impact

`dead_code/dynamic_markers.py`'s Python marker list (`importlib.import_module`, `__import__(`, `importlib.reload`, `pkgutil.iter_modules`) only lowers confidence for genuinely dynamic imports — it doesn't cover this case, because the import itself IS fully static; it's the attribute-call resolution that's missing. This isn't an edge case: `import module` + `module.func()` (vs. `from module import func`) is an extremely common pattern for namespacing, avoiding name collisions, or dispatch-table-style calling (e.g. iterating a list of modules and calling a conventionally-named function on each).

Given the false-positive rate this produces in practice (100% across every sample we've hit), I'd suggest either:
- Resolving attribute calls on an imported module/object back to the target symbol's definition in the "no importers" analysis, or
- Excluding `unused_export`/`unused_internal` from `--safe-only` until that resolution exists (a currently-safe-only finding of this kind can delete code with real callers).

Happy to share the exact repo pattern that surfaced this (a sensor-dispatch module that lazy-imports and calls `module.check()` for each of ~15 sensor modules) if useful, but the repro above reproduces it in isolation independent of that.

sloemo01:
Fixes #1193

## Problem
`from pkg import submodule` bound `submodule` to the package `__init__.py` (the fan-out's first target), so a later `submodule.symbol()` call resolved against a file that declares nothing and the call was missed. Dead-code then reported the symbol as safe-to-delete despite a real caller.

## Fix
Two-part:
- `resolve_python_import_all` now points each submodule binding's `source_file` at the submodule file it resolves to, not the package init.
- `build_import_name_maps` only back-fills a binding's `source_file` when it is unset, so the submodule pin survives.

Verified end-to-end: `from sensors import foo; foo.check()` now resolves to `sensors/foo.py::check` (0.88).

## Test
Adds regression tests for the submodule-binding pin and the non-submodule (package re-export) case. `test_python_resolver_fanout.py`: 9 passed; full `tests/unit/ingestion` + `tests/unit/dead_code`: 2855 passed (1 pre-existing env failure unrelated to this change).

### DIFF
--- a/packages/core/src/repowise/core/ingestion/import_index.py
+++ b/packages/core/src/repowise/core/ingestion/import_index.py
@@ -52,11 +52,17 @@ def build_import_name_maps(parsed_files: dict[str, ParsedFile]) -> ImportNameMap
                 for binding in imp.bindings:
                     if binding.local_name == "*":
                         continue
-                    binding.source_file = resolved
-                    maps.import_names[path][binding.local_name] = resolved
+                    # A binding whose source_file was already pinned to a
+                    # submodule file (``from pkg import submodule`` fan-out,
+                    # #1193) must keep it: ``imp.resolved_file`` is the package
+                    # ``__init__.py``, which declares nothing the submodule
+                    # name resolves to. Only back-fill when unset.
+                    if binding.source_file is None:
+                        binding.source_file = resolved
+                    maps.import_names[path][binding.local_name] = binding.source_file
                     maps.import_bindings[path][binding.local_name] = binding
                     if binding.is_module_alias:
-                        maps.module_aliases[path][binding.local_name] = resolved
+                        maps.module_aliases[path][binding.local_name] = binding.source_file
             else:
                 # Fallback for imports without binding data
                 for name in imp.imported_names:
--- a/packages/core/src/repowise/core/ingestion/resolvers/python.py
+++ b/packages/core/src/repowise/core/ingestion/resolvers/python.py
@@ -114,4 +114,14 @@ def resolve_python_import_all(
                         break
             if hit and hit != base:
                 targets.append(hit)
+                # Point the binding for this name at the submodule file, not
+                # the package ``__init__.py``. ``from pkg import submodule``
+                # binds ``submodule`` to ``pkg/submodule.py``, so a later
+                # ``submodule.symbol()`` call must resolve against that file
+                # (#1193). Without this the binding's source_file stays the
+                # package init, which declares nothing, and the call is missed.
+                for binding in imp.bindings:
+                    if binding.local_name == name:
+                        binding.source_file = hit
+                        break
     return tuple(dict.fromkeys(targets))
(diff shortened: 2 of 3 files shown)


RaghavChamadiya (approved):
Same correction as on #1731 and #1732: the run I read was the superseded one. Your 11:31 rebase is green across the full matrix on this head. Merging.

PR #1734, opened 2026-08-19
fix(cpp): use byte-safe AST node text

### LINKED ISSUE #1699: [Bug] C++ visibility helpers slice decoded text with byte offsets, hiding dllexport markers
**Summary**

`src` is a decoded string, but four helpers index into it using tree-sitter's
byte offsets. Any character that takes more than one byte earlier in the file
shifts every offset after it, so the slice returns the wrong text. For the export
check that means a real `__declspec(dllexport)` is missed and the symbol is
recorded as not exported. The same class of mistake has been fixed elsewhere in
the parser already, so there is a pattern to follow.

**Where**

Four sites, all C and C++ facing, each doing
`src[node.start_byte : node.end_byte]`:

* `packages/core/src/repowise/core/ingestion/extractors/visibility.py:415` and
  `:423`, both inside `_has_export_marker`
* `packages/core/src/repowise/core/ingestion/extractors/visibility.py:433`,
  inside `_has_file_scope_static`
* `packages/core/src/repowise/core/ingestion/parser_helpers.py:116`, inside
  `_qualified_cpp_parent`, which decides the parent class of every
  `void Foo::method()`

The fix already exists in the codebase. `node_text` in
`packages/core/src/repowise/core/ingestion/extractors/helpers.py:8-14` reads
`node.text` and decodes it properly, only falling back to slicing. These four
call sites bypass it.

**Repro**

A C++ header with a non-English character above an exported function, an author
name in a comment being the realistic case. The export marker is not found and
the symbol is recorded as not exported.

**Expected**

The symbol is detected as exported regardless of what appears earlier in the
file.

**Impact**

Real world C++ headers with author names or non-English comments. Low frequency,
but it silently mislabels visibility, which then feeds dead code detection.

**Done looks like**

All four sites use `node_text`, and a test with a fixture containing a
multi byte character above an exported C++ symbol asserts the symbol is detected
as exported. That test must fail before the change.

**Important caveat, please read before starting**

**This cannot be demonstrated on our own test repositories.** leveldb, fmt and
swift-nio contain zero non-English bytes across 242 C and C++ files, so a before
and after comparison on the corpus is byte for byte identical. The proof has to
be a made up fixture plus a guard test. If you go looking for a corpus diff you
will conclude the bug is not real.

QinXi-ai:
## Summary

- Replace four C/C++ helpers' decoded-string byte-offset slices with the existing `node_text()` helper, so earlier multibyte UTF-8 text cannot shift AST node extraction.
- Add end-to-end regressions for both affected outcomes: preserving `__declspec(dllexport)` metadata and attaching a qualified `Foo::method()` definition to `Foo`.

The root cause was using tree-sitter byte offsets to slice a decoded Python `str`. When a non-ASCII character appeared earlier in the file, those byte offsets selected the wrong characters, silently dropping either the export flag or the method-parent graph edge.

## Related Issues

Fixes #1699

## Test Plan

- [x] Confirmed both new multibyte regression tests fail on unmodified `main`.
- [x] C++ ingestion and parser tests pass (`118 passed`).
- [x] Full Python lint passes (`ruff check .`).
- [x] Web build is not applicable; this change only touches Python ingestion code and tests.

## Checklist

- [x] My code follows the project's code style.
- [x] I have added tests for the bug fix.
- [x] Relevant existing tests pass.
- [x] Documentation changes are not needed for this internal parser correction.

### DIFF
--- a/packages/core/src/repowise/core/ingestion/extractors/visibility.py
+++ b/packages/core/src/repowise/core/ingestion/extractors/visibility.py
@@ -16,6 +16,8 @@ import re
 from collections.abc import Callable
 from typing import TYPE_CHECKING
 
+from .helpers import node_text
+
 if TYPE_CHECKING:
     from tree_sitter import Node
 
@@ -412,7 +414,7 @@ def _has_export_marker(def_node: Node, src: str) -> bool:
     sibling = def_node.prev_sibling
     seen = 0
     while sibling is not None and seen < 4:
-        text = src[sibling.start_byte : sibling.end_byte]
+        text = node_text(sibling, src)
         if any(marker in text for marker in _CPP_EXPORT_MARKERS):
             return True
         sibling = sibling.prev_sibling
@@ -420,7 +422,7 @@ def _has_export_marker(def_node: Node, src: str) -> bool:
     # Also check the def_node's own leading children — some grammars
     # nest the declspec inside the function_definition.
     for child in def_node.children[:3]:
-        text = src[child.start_byte : child.end_byte]
+        text = node_text(child, src)
         if any(marker in text for marker in _CPP_EXPORT_MARKERS):
             return True
     return False
@@ -430,7 +432,7 @@ def _has_file_scope_static(def_node: Node, src: str) -> bool:
     """Return True if a ``static`` storage-class specifier appears in the leading declarators."""
     for child in def_node.children[:4]:
         if child.type == "storage_class_specifier":
-            text = src[child.start_byte : child.end_byte]
+            text = node_text(child, src)
             if "static" in text:
                 return True
     return False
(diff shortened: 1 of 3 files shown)


sloemo01 (approved):
Verified this against main — the fix is correct and the tests genuinely exercise it.

**Root cause confirmed:** tree-sitter `start_byte`/`end_byte` are *byte* offsets, but `visibility.py` was slicing a Python `str` (which indexes by code point) with them. Any multibyte character before a definition shifted the slice and could misread export markers / static storage / qualified scope.

**The fix uses the right primitive:** `node_text()` falls back to `node.text.decode()` when available, which handles byte offsets correctly. Good call reusing the existing helper rather than re-inlining.

**Tests are well-targeted:** the multibyte comment `// Author: 李` is placed directly before the definition, which is exactly where the byte-index drift manifests (comment bytes preceding the `__declspec`/qualified-name node).

Verified locally:
- `test_cpp_extractors.py` — 16 passed (incl. both new multibyte tests)
- `test_c_type_use.py` + `test_cpp_reference_edges.py` — 19 passed, no regression

One note for the author (not blocking): I checked the other `src[start_byte:end_byte]` slices in `go_interface_satisfaction.py` — those operate on a `bytes` buffer (`src: bytes`), so slicing by byte offset + `.decode()` is already correct there. This PR has the only `str`-slice instance. Thanks for catching it.

swati510 (approved):
LGTM

PR #1737, opened 2026-08-19
fix(mcp): widen transport_security to match --host

### LINKED ISSUE #1736: [Bug] --host makes the MCP server unreachable — 421 Misdirected Request
## Describe the Bug

`repowise mcp --transport streamable-http --host <non-loopback>` (also reproduces on
`--transport sse`) binds successfully and logs the correct address, but every client request is
rejected with `421 Misdirected Request` before reaching any tool. The server logs `Invalid Host
header: ...`. This happens for any non-default `--host` value (a LAN IP, a hostname, or
`0.0.0.0`) — the connection cannot be used at all, regardless of what `--host` is actually set to.

## Steps to Reproduce

1. `repowise mcp --transport streamable-http --host 172.21.12.48 --port 7338 --all`
2. From another host on the same LAN (or any client sending a matching `Host` header), point an
   MCP HTTP client at `http://172.21.12.48:7338/mcp`.
3. Every request returns `421 Misdirected Request`; the server logs `Invalid Host header:
   172.21.12.48:7338`.

## Expected Behavior

A request whose `Host` header matches the address `--host`/`--port` were bound to should be
served normally, the same as it is against the default `127.0.0.1`.

## Actual Behavior

Every request is rejected by the `mcp` SDK's built-in DNS-rebinding protection
(`mcp.server.transport_security.TransportSecurityMiddleware`), regardless of `--host`:

```
WARNING  Invalid Host header: 172.21.12.48:7338   transport_security.py:64
INFO:    172.21.129.8:58368 - "POST /mcp HTTP/1.1" 421 Misdirected Request
```

Root cause: `packages/server/src/repowise/server/mcp_server/_server.py` constructs the
module-level `FastMCP(...)` instance with no `host=` argument, at import time — before any CLI
`--host` is known. The `mcp` SDK's `FastMCP.__init__` sees the default loopback host and bakes a
DNS-rebinding allowlist scoped to loopback only (`127.0.0.1:*`/`localhost:*`/`[::1]:*`).
`run_mcp()` later rebinds the socket to the CLI's `--host` value via `mcp.settings.host`, but
never updates `mcp.settings.transport_security` to match — so the allowlist stays frozen on
loopback forever, and `TransportSecurityMiddleware._validate_host` rejects every non-loopback
`Host` header no matter what `--host` was actually given.

## Environment

- OS: Not OS-specific; reproduced on Linux (WSL2)
- Python version: 3.11+ (not version-specific)
- Repowise version: 0.41.0 (also confirmed present through the latest release, v0.44.0, and
  current `main`, by code inspection as of 2026-08-19)
- Installation method: pip

## Additional Context

I have a fix ready (`_configure_transport_security()`, widens the allowlist to the actual bind
host, or disables the check for a wildcard `0.0.0.0`/`::` bind) with tests, and will open a PR
shortly.

nathanpride:
FastMCP builds its module-level instance with no host= argument, at import time, before any CLI --host is known, so it auto-bakes a DNS-rebinding Host allowlist scoped to loopback only. run_mcp() later rebinds the socket via mcp.settings.host, but never updated the allowlist to match, so every request to a non-loopback --host failed Host validation with 421 Misdirected Request no matter what --host was given.

_configure_transport_security() rebuilds the allowlist around the actual bind host (or disables the now-unsatisfiable Host check for a wildcard bind, where nothing else currently gates access).

## Summary

- Widen the MCP HTTP/SSE transport's DNS-rebinding `Host` allowlist to match `--host`, instead of leaving it frozen on loopback (fixes `421 Misdirected Request` on any non-default `--host`)

## Related Issues

Fixes #1736
Related to #1400 (same function — `run_mcp` in `_server.py` — but a distinct concern: #1400 is about missing authentication on a wide-open bind, which this PR does not address)

## Test Plan

- [x] Tests pass (`pytest`) — `uv run pytest tests/unit/server/test_mcp_transport_security.py tests/unit/cli/test_mcp_transport.py`, 28 passed
- [ ] Web build passes (`npm run build`) *(if frontend changes)* — n/a, no frontend changes
- [x] Manual: `repowise mcp --transport streamable-http --host <non-loopback IP>`, then a request with a matching `Host` header — `200 OK` / clean `initialize` response, no `421`/`Invalid Host header` (verified against both a wildcard `0.0.0.0` bind and a concrete interface IP)

## Checklist

### DIFF
--- a/packages/server/src/repowise/server/mcp_server/_server.py
+++ b/packages/server/src/repowise/server/mcp_server/_server.py
@@ -9,6 +9,7 @@ from contextlib import asynccontextmanager
 from typing import Any
 
 from mcp.server.fastmcp import FastMCP
+from mcp.server.transport_security import TransportSecuritySettings
 from sqlalchemy.ext.asyncio import AsyncSession, async_sessionmaker
 
 from repowise.core.persistence.database import (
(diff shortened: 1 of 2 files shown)


RaghavChamadiya (approved):
Thanks @nathanpride. The diagnosis is the valuable part: `FastMCP` builds `mcp.settings.transport_security` at import time, before any CLI `--host` exists, so it bakes in a loopback-only allowlist. `run_mcp` then rebinds the socket through `mcp.settings.host` and nothing ever revisits the allowlist. That is why `--host` produced a 421 for every value, which reads like a networking problem and is not one.

`_bracket_if_ipv6` is the detail that shows you read the SDK rather than guessed. `TransportSecurityMiddleware` matches on `host.startswith(base + ":")`, so a bare `2001:db8::1` can never match the bracketed `[2001:db8::1]:7338` a client actually sends, and the "any colon means IPv6, because hostnames and IPv4 never contain one" reasoning is right and worth having written down.

On the wildcard case, I want to say plainly that I read it as a deliberate trade and I agree with it. Disabling DNS-rebinding protection sounds like a downgrade, but no single `Host` value can match a `0.0.0.0` bind, so the previous behaviour was not protection, it was a permanent 421. Nothing was reachable to protect. Turning it off makes the bind do what the operator asked and leaves the startup security warning as the gate until #1400 lands. Your docstring says exactly that, which is why it does not need a reviewer to reconstruct it.

Keeping loopback in the allowlist for a non-loopback bind is the right default too. An SSH tunnel or a client on the same box arrives as `127.0.0.1` and would otherwise break the moment someone set `--host`.

Full matrix green. Merging. This closes #1736.

PR #1750, opened 2026-08-19
fix(mcp): share paired-test filename heuristic

### LINKED ISSUE #1740: [Bug] Two different test-file heuristics give opposite answers for the same file
**Summary**

"Does this file have a test?" is answered by two separate implementations that
use different matching rules, so `get_health` and `get_risk` can disagree about
the same file in the same index.

`get_health` matches an exact basename set. `get_risk` matches a substring. A
file whose tests are named after behaviour rather than after the file, such as a
source file `foo.py` tested by `test_foo_strategies.py` and `test_foo_edge.py`
with no plain `test_foo.py`, is reported as untested by one tool and tested by
the other.

**Where**

`packages/core/src/repowise/core/analysis/health/engine.py:265-288`
builds a candidate set (`test_<stem>.py`, `<stem>_test.py`, `<stem>.test.ts`,
and so on) and tests set membership, so matching is exact.

`packages/server/src/repowise/server/mcp_server/tool_risk/assessment.py:210-233`
builds `LIKE` patterns wrapped in `%`, so `%test_foo%` also matches
`test_foo_strategies.py`.

**Repro**

In any indexed repo, find a source file whose only tests carry a suffix after
the stem. Then compare:

- `repowise health --target <file>` (reads the exact-set heuristic)
- `repowise risk --target <file>` (reads the substring heuristic)

`test_gap` and `has_test_file` will contradict each other.

**Expected**

One shared helper with one matching rule, called by both. Which rule is correct
is a separate question from the fact that there are currently two.

**Impact**

The two answers feed different user-visible claims, and `has_test_file` gates
the `untested_hotspot` biomarker, which carries a high severity weight. So the
stricter of the two heuristics can penalise a file that the other tool reports
as tested, in the same index, from the same question.

**Done looks like**

A single helper in core, called from both sites. A test that pins the agreed
behaviour for the suffixed-test-name case, in both directions.

Note for whoever picks this up: the longer-term direction is to answer this from
the graph rather than from filenames, since the graph already records which test
files reach a given source file. This issue is only about the two
implementations disagreeing, and a shared helper is the right scope here.

nightcityblade:
## Summary

- Reuse health analysis's exact paired-test filename heuristic in the risk tool.
- Add a regression proving health and risk agree on suffixed test filenames.

## Related Issues

Fixes #1740

## Test Plan

- [x] Tests pass (`pytest`): 17 focused health/risk tests passed
- [x] Lint passes (`ruff check .`): changed files pass Ruff and format checks
- [x] Web build passes (`npm run build`) *(if frontend changes)* — N/A (no frontend changes)

## Checklist

- [x] I have updated documentation if needed — N/A (no documentation change needed)

### DIFF
--- a/packages/core/src/repowise/core/analysis/health/engine.py
+++ b/packages/core/src/repowise/core/analysis/health/engine.py
@@ -281,9 +281,10 @@ def _has_paired_test_file(rel_path: str, path_basenames: set[str]) -> bool:
     """
     p = Path(rel_path)
     stem = p.stem
+    test_suffix = ".exs" if p.suffix == ".ex" else p.suffix
     candidates = {
         f"test_{stem}.py",
-        f"{stem}_test.py",
+        f"{stem}_test{test_suffix}",
         f"{stem}.test.ts",
         f"{stem}.test.tsx",
         f"{stem}.test.js",
@@ -293,7 +294,6 @@ def _has_paired_test_file(rel_path: str, path_basenames: set[str]) -> bool:
         f"{stem}.spec.js",
         f"{stem}.spec.mts",
         f"{stem}.spec.cts",
-        f"{stem}_test.go",
     }
     return not candidates.isdisjoint(path_basenames)
 
--- a/packages/server/src/repowise/server/mcp_server/tool_risk/assessment.py
+++ b/packages/server/src/repowise/server/mcp_server/tool_risk/assessment.py
@@ -10,12 +10,12 @@ from typing import Any
 from sqlalchemy import select, text
 from sqlalchemy.ext.asyncio import AsyncSession
 
+from repowise.core.analysis.health.engine import _has_paired_test_file, _path_basenames
 from repowise.core.persistence.models import (
     GitMetadata,
     GraphNode,
     Repository,
 )
-from repowise.core.persistence.sql import LIKE_ESCAPE, escape_like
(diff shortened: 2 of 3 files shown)


RaghavChamadiya (approved):
This is the fix, and you took it further than I asked in a way that is right.

I said `f"{stem}_test{p.suffix}"`. That would have produced `user_test.ex` for an Elixir source and missed `test/user_test.exs`, which was one of the exact pairs I put in the review as a regression. You caught that my own suggestion did not cover a case I had used to argue for it, and normalising `.ex` to `.exs` is the correct reading: Elixir test files are scripts, so the convention genuinely changes extension across the pair. Dropping the `_test.go` line once the general form covers it is the right cleanup too.

Taking the `_path_basenames` note means health and risk now share both rules, the basename rule and the match rule, which is the whole argument the issue makes one level up.

Full matrix is green on 1a320ac, all three Python versions plus ruff.

One thing I owe you rather than the other way round. My review named the `<stem>_test` half and missed the symmetric one: `test_{stem}.py` is still spelled `.py` only, so risk loses Ruby's minitest `test/test_user.rb`, which the old extension-blind `%test_{base}%` pattern did catch. That is the same regression shape I blocked you on, and I should have seen both halves at once. I am not going to make you do a third round for it; I will fold it into #1768, which is already open for Ruby's `_spec.rb` and touches the same candidate set.

PR #1751, opened 2026-08-19
fix(cli): make `coverage add` exit non-zero when it stores nothing

### LINKED ISSUE #1745: [Bug] coverage add exits 0 when no coverage is ingested
## Summary

`repowise coverage add` exits with status 0 when it ingests nothing, including when it reports `No indexed files found`. Automation that follows the normal Unix contract and checks only `$?` therefore records a successful refresh while the coverage store remains unchanged.

## Version check

- Rechecked on 2026-08-19: the installed CLI is 0.43.0 and its `coverage add` help still documents ingestion without describing a partial/no-op success exit contract.
- The latest upstream release is 0.44.0; release notes from 0.36.0 through 0.44.0 contain no corresponding fix.

## Reproduction

1. Initialize or select an index whose repository/file mapping cannot accept the supplied report (the observed case had no indexed files available for the report).
2. Run:

```bash
repowise coverage add coverage.xml
printf 'exit=%s\n' "$?"
```

3. Observe an ingestion refusal/no-op such as `No indexed files found` and `exit=0`.
4. Query `repowise coverage status`; no new per-file coverage was stored.

The same zero exit was observed for a severely partial ingest. In the measured case only 1 of 249 report files and 26 of approximately 29,317 map records were accepted.

## Expected behavior

Return non-zero when zero records are ingested. For partial ingestion, either return non-zero or provide a documented strict flag suitable for CI. Machine-readable output should expose requested, mapped, skipped, and stored counts.

## Impact

A coverage refresh is commonly scripted as `repowise coverage add ... || exit 1`. Today that guard cannot distinguish a complete ingest from a no-op, leaving stale or absent coverage behind while CI reports success.

## Related documentation ambiguity

A coverage.py `.coverage` file carrying contexts populated the per-test map but did not populate per-file line coverage in the measured workflow; generating Cobertura XML and running a second `coverage add` was required. The help says the command stores per-file coverage and additionally a per-test map, so the map-only state looks complete. Clarifying that input-specific behavior, or reporting the two stored record classes explicitly, would make the no-op/partial contract much easier to validate.

NgoQuocViet2001:
Fixes #1745.

## The problem

Every no-op path through `coverage add` printed a message and returned, so the process exited 0:

| path | before |
|---|---|
| auto-discovery found no report | `0` |
| no index yet (`repowise init` not run) | `0` |
| no indexed files | `0` |
| report parsed but nothing mapped to the repo tree | `0` |

A refresh scripted as `repowise coverage add ... \|\| exit 1` therefore could not tell a complete ingest from one that stored nothing, and CI recorded success over a stale or empty coverage store.

## The change

`_do()` now answers whether anything was stored, and the command turns that answer into the exit status. The discovery miss raises directly, since it happens before the ingest starts and has no `_do()` to report through.

## Partial ingests

Kept at 0 by default and put behind `--strict`, which is the second option the issue offers. A report that maps most of its files is still a real refresh, and failing it by default would break every setup with a build-prefix mismatch it has already learned to live with — whereas a `--strict` run is opt-in and says what it found:

```
--strict: 248 report file(s) did not map to the repo tree.
```

Happy to flip that to fail-by-default if you would rather; it is a one-line change and I asked on the issue before picking a side.

Not addressed here: the machine-readable requested/mapped/skipped/stored counts, and the `.coverage`-contexts documentation ambiguity. Both are worth doing and neither belongs in an exit-status fix.

## Test plan

- Ran: `uv run pytest tests/unit/cli/test_coverage_cmd.py -q` — 7 passed (3 pre-existing + 4 new).
- Ran: `uv run pytest tests/unit/cli/ tests/unit/health/test_coverage_discovery.py -q` — 2192 passed, 2 skipped, 2 xfailed.
- Ran: `uv run ruff check` and `uv run ruff format --check` on both changed files — clean, and the tree was already format-clean before the change.
- Checked: with only `coverage_cmd.py` reverted, 3 of the new assertions fail — so they are pinned to the behaviour, not to the test setup.

New coverage: the discovery miss exits 1 with its message intact, and a parametrised case pins the wiring in both directions (stored → 0, nothing stored → 1), which is the branch every one of the four no-op paths funnels through. Plus `--help` documenting `--strict`.

### DIFF
--- a/packages/cli/src/repowise/cli/commands/coverage_cmd.py
+++ b/packages/cli/src/repowise/cli/commands/coverage_cmd.py
@@ -68,6 +68,12 @@ def coverage_group() -> None:
     help="Force a parser instead of auto-detecting from content.",
 )
 @click.option(
+    "--strict",
+    is_flag=True,
+    default=False,
+    help="Also fail when some report files did not map to the repo tree.",
+)
+@click.option(
     "--verbose",
     "-v",
     is_flag=True,
@@ -78,6 +84,7 @@ def coverage_add(
     paths: tuple[str, ...],
     repo: str | None,
     coverage_format: str | None,
+    strict: bool,
     verbose: bool,
 ) -> None:
     """Ingest coverage (auto-discovers reports when none are given).
@@ -94,6 +101,10 @@ def coverage_add(
         repowise coverage add coverage/lcov.info
         repowise coverage add .coverage            # per-test map from coverage.py
         repowise coverage add web.lcov api.lcov    # merged hit-wins
+
+    Exits non-zero when nothing was stored, so `coverage add ... || exit 1`
+    tells a complete ingest from a no-op. With --strict it also exits
+    non-zero when some report files did not map to the repo tree.
     """
     configure_cli_logging(verbose=verbose)
 
@@ -125,7 +136,7 @@ def coverage_add(
                 "then re-run, or pass a path:\n"
                 "  [cyan]repowise coverage add path/to/report[/cyan]"
             )
-            return
+            raise click.exceptions.Exit(1)
         console.print(
             f"Discovered {len(report_paths)} report(s): "
             + ", ".join(p.name for p in report_paths[:5])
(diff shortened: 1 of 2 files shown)


RaghavChamadiya (approved):
Verified against current main: applies clean, targeted tests green locally, CI green. Merging.

PR #1752, opened 2026-08-19
refactor(perf): share the sink-reachability walk between crossfn and gated

### LINKED ISSUE #1741: [Bug] crossfn and gated duplicate a 61-line reachability walk
**Summary**

`collect_crossfn_io_in_loop` and `collect_blocking_io_under_lock` are the same
function. Both run the same pre-checks, build the same `CallGraphIndex`, collect
the same sink set, run the same reverse BFS, and walk the same output loop. The
only differences are which `PerfFnFacts` field supplies the entry set and which
kind string the resulting hit carries.

Their per-function helpers, `_hits_for_function` and `_lock_hits_for_function`,
are duplicated the same way.

The repository's own duplication detector reports this pair, so it is visible in
`repowise health` output on this repo.

**Where**

`packages/core/src/repowise/core/analysis/health/perf/crossfn.py`
`packages/core/src/repowise/core/analysis/health/perf/gated.py`

`gated.py`'s docstring already states the relationship: "Mirrors
`crossfn.collect_crossfn_io_in_loop` exactly: the only difference is the entry
set is `PerfFnFacts.lock_call_targets` ... instead of loop-nested callees."

**Expected**

One parameterised implementation taking the entry-set accessor and the hit kind,
with two thin callers preserving the current public names.

**Impact**

Maintenance only, but real: the cross-function N+1 detector and the
lock-held-I/O detector are two of the engine's differentiating signals, and a fix
to the shared walk currently has to be made twice or it silently applies to only
one of them.

**Done looks like**

A single shared collector. Both existing entry points keep their names and
signatures. The existing tests for both detectors pass unchanged, which is the
main correctness gate here since the behaviour must not move.

NgoQuocViet2001:
Closes #1741.

## What changed

`collect_crossfn_io_in_loop` and `collect_blocking_io_under_lock` are the same function twice — the same pre-checks, the same index, the same bare-sink set, the same multi-source reverse BFS, and the same per-function first-hop resolution, down to `_hits_for_function` / `_lock_hits_for_function`.

The walk now lives once in `perf/sink_reach.py`, parameterised by the entry-set accessor and the hit kind, next to the other perf primitives. Both entry points keep their names and signatures and become thin callers:

```python
return collect_sink_reaching_hits(
    walked, graph,
    entries=lambda fact: fact.loop_call_targets,   # lock_call_targets in gated
    kind=CROSSFN_KIND,                             # LOCK_IO_KIND in gated
    index=index, max_depth=max_depth,
)
```

## Rebuilt on current `main`, not rebased

#1832 moved first-hop matching from callee-name to call-line accurate, via `index.resolve_call_targets(...)`, stamping each hit with `resolution_basis`. @sloemo01 [pointed out](https://github.com/repowise-dev/repowise/pull/1752#issuecomment-5380278075) that resolving the textual conflict would have quietly reinstated the old name-based loop and dropped that.

So this branch was reset onto `main` and the de-duplication re-derived on top, rather than merged. The shared walk resolves the same way `main` does and passes the basis through, so collapsing the two copies costs none of the exact-edge accuracy:

```python
targets, basis = index.resolve_call_targets(a_sid, call_line, target_name)
```

`callees_by_name` is gone from the branch entirely.

## New coverage

Nothing asserted `resolution_basis` in the perf suite before — the only assertions on it live under `tests/unit/server/`. Two cases added, one per pass, where a single owner calls two same-named callees on different lines so name matching cannot pick the right one:

```python
graph.add_edge("owner.py::run", "pure.py::load", edge_type="calls", call_lines=[9])
graph.add_edge("owner.py::run", "db.py::load",   edge_type="calls", call_lines=[4])
...
assert hit.path == ("owner.py::run", "db.py::load")
assert hit.resolution_basis == "call-site"
```

Simulating the naive resolve (swapping `resolve_call_targets` for a name-based loop) fails 4 tests: these two plus the existing `test_unresolved_loop_call_does_not_reuse_same_name_outside_loop`. These also answer the bot's "no test file imports the changed files" note.

## One difference preserved rather than unified

The lock pass sets `PerfHit.func_start` and the loop pass does not. That is carried through as `carry_func_start` rather than made consistent, because making them agree would move behaviour and this is a de-duplication. The new lock test asserts it, so the asymmetry is at least written down. Happy to unify in a follow-up.

## Test plan

- Ran: `uv run python -m pytest tests/unit/health/ -q` — **1215 passed**.
- Ran: `uv run python -m pytest tests/unit/health/test_perf_crossfn.py tests/unit/health/test_perf_phase7b.py -q` — **47 passed**.
- Ran: `uv run ruff check` and `uv run ruff format --check` on all four changed files — clean.

### DIFF
--- a/packages/core/src/repowise/core/analysis/health/perf/crossfn.py
+++ b/packages/core/src/repowise/core/analysis/health/perf/crossfn.py
@@ -48,13 +48,10 @@ from collections.abc import Iterable
 from typing import TYPE_CHECKING, Any
 
 from .callgraph import CallGraphIndex
-from .reachability import path_to_sink, reachable_to_sink
+from .sink_reach import collect_sink_reaching_hits
 
 if TYPE_CHECKING:
-    # Imported lazily at runtime (see ``_hits_for_function``) to avoid a cycle:
-    # the walker pulls in ``perf.io_boundaries`` while the ``complexity``
-    # package is still initialising, so ``crossfn`` cannot import it eagerly.
-    from ..complexity import FileComplexity, PerfFnFacts, PerfHit
+    from ..complexity import FileComplexity, PerfHit
 
 # Cross-function hits carry this so the biomarker can phrase them distinctly
 # while still scoring under the same ``io_in_loop`` / ``performance`` budget.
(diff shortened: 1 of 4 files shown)


RaghavChamadiya (approved):
Thanks @NgoQuocViet2001, and @sloemo01 for the review that made this land safely. Flagging that the conflict *hid* a semantic change rather than just being textual is the thing that would have cost us the call-line accuracy #1832 added, and it is not something a three-way merge would have told anyone.

Resetting onto main and re-deriving rather than resolving the conflict was the right instinct and I want to name why: on a rebase the old name-based `callees_by_name` loop sits on the "ours" side of every hunk, so a conflict resolution that looks clean can quietly reinstate it. Rebuilding makes that impossible instead of merely unlikely. I checked, and `callees_by_name` appears nowhere in the branch.

Verified the rest against your head:

`_hits_for_function` in `sink_reach.py` is a faithful merge of both originals, not a rewrite of either. Same `seen` dedupe on target name, same `path_to_sink` chain, same `break` after the first reaching callee, and `index.resolve_call_targets(a_sid, call_line, target_name)` with `resolution_basis=basis` on every hit, matching what both passes do on main today. `carry_func_start` preserving the disagreement between the two callers rather than unifying it is exactly right for a de-duplication, and saying so in the docstring means the next person knows it is a fact about the callers and not an oversight here. `max_depth` is forwarded from both entry points, so the depth bound still belongs to the caller.

Correcting @sloemo01 on the test coverage was worth more than agreeing would have been. "The perf suite was not guarding the field you were worried about, here is the one test that would have caught a full revert, and here is why a partial one would have slipped through" is a better answer than either "you are right" or "it is covered". Adding a case per pass where one owner calls two same-named callees on different lines pins the thing that actually distinguishes call-line resolution from name matching, and `test_lock_io_first_hop_also_carries_the_resolution_basis` closes the gap on the lock side that had no assertion at all.

Green on all three Pythons. Merging.

PR #1753, opened 2026-08-19
fix(perf): only count a prisma verb as a database sink on a member call

### LINKED ISSUE #1501: Perf analyzer treats any function named 'aggregate' as a database call, producing false N+1 findings on pure code
### Summary

The performance analyzer's `io_in_loop` / N+1 detector classifies **any call to a function named `aggregate` as a database boundary**, including locally defined pure functions. In a client-side TypeScript codebase with no database access at all, this produced 6 of our 20 performance findings, all `boundary_kind: "db"`, both direct and cross-function.

### Environment

- repowise 0.41.0+f44f8a2d (pipx install), Windows 11
- Indexed repo: React/TypeScript frontend + Supabase backend; the flagged files are pure browser-side utilities and their vitest tests

### Reproduction

```ts
// math.ts — a pure local function, no I/O anywhere
export function aggregate(values: readonly Ohlc[]): Ohlc | null {
  if (values.length === 0) return null;
  return values.reduce(roll);
}

export function reconstructs(parent: Ohlc, children: readonly Ohlc[]): boolean {
  const rolled = aggregate(children);   // pure call
  ...
}

// view.ts
for (let start = 0; start < sorted.length; start += ratio) {
  const chunk = sorted.slice(start, start + ratio);
  const rolled = aggregate(chunk.map(toWorldOhlc));   // flagged
}
```

Run the health analysis and the loop call site is flagged:

> `io_in_loop` (medium) — "a database call runs once per loop iteration (N+1 / IO-in-loop)" — `details: {"boundary_kind": "db", "cross_function": false}`

Test files that loop over fixtures and call `reconstructs` get the cross-function variant:

> "a database call is reached once per loop iteration through `__module__` -> `reconstructs` (cross-function N+1)"

### Expected

A call should only count as a database boundary when the callee resolves to a known client/ORM surface (e.g. a member call on a query-builder or driver object), not on the bare name `aggregate`. Here the callee is a same-repo import of a pure function; the import binding is statically resolvable.

### Actual

`boundary_kind: "db"` findings on pure arithmetic, in files (and their tests) that cannot reach any database. Name-matching also seems to be the only trigger: identical loops calling the same logic under a different name are not flagged.

NgoQuocViet2001:
Fixes #1501.

## The problem

`ts_js.sink_kind` matched `PRISMA_METHODS` on the bare method name, with no member-call requirement:

```python
if method in PRISMA_METHODS:  # distinctive prisma-client verbs
    return "db"
```

`aggregate` is in that set, so a locally defined pure function called it a database round trip:

```ts
export function aggregate(values: readonly Ohlc[]): Ohlc | null { ... }   // no I/O anywhere

for (const chunk of sorted) {
  const rolled = aggregate(chunk);        // io_in_loop, boundary_kind: "db"
}
```

The reporter saw 6 of 20 performance findings come from this in a browser-side codebase with no database, direct and cross-function, and the tests that loop over fixtures inherited them. `groupBy`, `upsert` and `findMany` are the same shape.

## The change

A prisma query is always reached through the client — `prisma.user.aggregate(...)` — so that arm is now gated on `is_attribute`, which the walker already computes and passes.

`prisma.user.aggregate(...)` still reports. A bare `aggregate(...)` no longer does.

This is the rule the module's own header already states — "Only DISTINCTIVE method names are trusted without import resolution". These names are distinctive as *client methods*; as free functions they are ordinary vocabulary, which is exactly the collision the header excludes `create` / `update` / `count` / `exec` for.

`TS_FS_METHODS` is deliberately left matching bare, since `import { readFileSync } from "node:fs"` is a real destructured binding with no root to resolve.

## Test plan

- Ran: `uv run pytest tests/unit/health/test_perf_dialects.py -q` — 131 passed (130 pre-existing + 1 new).
- Ran: `uv run pytest tests/unit/health/ -q` — 1163 passed.
- Ran: `uv run ruff check` and `uv run ruff format --check` on both changed files — clean.
- Checked: with `ts_js.py` reverted the new test fails, so it is pinned to the behaviour.

The new case reproduces the issue's snippet (a local `aggregate` called in a loop → no hits), asserts `prisma.user.aggregate` in a loop still reports `("io_in_loop", "db")`, and sweeps `groupBy` / `upsert` / `findMany` as bare local functions.

### DIFF
--- a/packages/core/src/repowise/core/analysis/health/perf/dialects/ts_js.py
+++ b/packages/core/src/repowise/core/analysis/health/perf/dialects/ts_js.py
@@ -163,7 +163,12 @@ class TsJsPerfDialect(BasePerfDialect):
             if method in TS_SINK_METHODS or awaited:
                 return root_kind
             return None
-        if method in PRISMA_METHODS:  # distinctive prisma-client verbs
+        # Distinctive prisma-client verbs, but only as a MEMBER call. A prisma
+        # query is always reached through the client (``prisma.user.aggregate``),
+        # so a bare ``aggregate(xs)`` / ``groupBy(xs)`` / ``upsert(x)`` is an
+        # ordinary local function and was being reported as a database round
+        # trip -- an N+1 finding on pure arithmetic, in code with no database.
+        if is_attribute and method in PRISMA_METHODS:
             return "db"
         if method in TS_FS_METHODS:  # distinctive sync-fs verbs
             return "filesystem"
--- a/tests/unit/health/test_perf_dialects.py
+++ b/tests/unit/health/test_perf_dialects.py
@@ -442,6 +442,33 @@ def test_ts_nested_io_requires_collection_outer_loop():
     assert ("nested_loop_with_io", "db") in _hits("typescript", nested)
 
 
+def test_ts_prisma_verb_is_a_sink_only_as_a_member_call():
+    """A bare ``aggregate(xs)`` is a local function, not a database round trip.
+
+    The prisma lexicon is reached through the client (``prisma.user.aggregate``),
+    so matching the bare name flagged pure arithmetic as an N+1 in a codebase
+    with no database at all.
+    """
+    pure = (
+        "function aggregate(xs){ return xs.reduce(roll); }"
+        " function view(sorted, ratio){ for (const chunk of sorted) {"
+        " const rolled = aggregate(chunk); } }"
+    )
+    assert _hits("typescript", pure) == []
+
+    # The real thing still reports, through the client object.
+    real = (
+        "async function f(prisma, xs){ for (const x of xs) {"
+        " await prisma.user.aggregate({ where: x }); } }"
+    )
+    assert ("io_in_loop", "db") in _hits("typescript", real)
+
+    # And the other bare names the lexicon carries are equally not sinks.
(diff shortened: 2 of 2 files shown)


RaghavChamadiya (approved):
Thanks @NgoQuocViet2001 for gating on the receiver instead of trimming the lexicon. `aggregate`, `groupBy` and `upsert` are real prisma verbs, so dropping them from `PRISMA_METHODS` would have traded one wrong answer for another.

The gate reads correctly against the surrounding code: `is_attribute` is already a keyword-only parameter on `sink_kind`, and the branch directly above it applies the same receiver-first logic to imported I/O packages. So the prisma branch was the outlier rather than the rule, and this brings it in line.

The test earns its place by asserting both directions, that a bare `aggregate(xs)` in a loop is silent and that `prisma.user.aggregate` in a loop still reports `io_in_loop`. Looping over the other three bare names is what makes it a regression test rather than an example.

Non-blocking: the diff also rewraps four lines in `test_perf_dialects.py` that this change does not touch. CI only runs `ruff check`, not `ruff format`, so nothing needs undoing, just worth knowing for next time since it makes the real change harder to find in the diff.

Green on 3.11, 3.12, 3.13 and ruff. Merging.

PR #1754, opened 2026-08-19
fix(pages): scope and rank a module page's decision records

### LINKED ISSUE #1587: [Bug] Module pages are handed the whole repo's decision list, then cut it to five
## Describe the Bug

Every module page is handed the entire repository's decision list, not the decisions that relate to that module. The template then shows the first five of it.

`page_generator/levels.py:288` passes `run.decisions_all` straight through:

```python
decision_records=run.decisions_all,
```

and `module_page.j2:48` renders `{% for dr in ctx.decision_records[:5] %}`.

There's no filter to the module's own files anywhere in between, and no ranking either, so the five that get shown are just the first five in the list.

The scoped data already exists. `build_decision_maps` (`helpers.py:244-265`) returns two things, `decisions_by_file` and `decisions_all`, and file pages use the scoped one 114 lines earlier in the same file:

```python
# levels.py:174, file pages
decision_records=run.decisions_by_file.get(p.file_info.path),
# levels.py:288, module pages
decision_records=run.decisions_all,
```

So this looks like the wrong map being passed rather than a missing capability.

## Steps to Reproduce

1. Index a repo that has more decision records than a handful, across different areas.
2. Open any module page in the wiki.
3. The "decisions" section shows five decisions, and for most modules they'll be about unrelated parts of the codebase.

## Expected Behavior

A module page shows decisions that govern that module's files, ranked so the five it has room for are the five worth reading.

## Actual Behavior

It shows five decisions from a repo-wide pool, chosen by list position. On a repo with decisions concentrated in one area, the same five appear on every module page regardless of subject.

## Additional Context

Two things make this worth more than it looks:

`module_page` is one of the LLM-written page types, so this list isn't just rendered, it goes into the prompt (`page_generator/pertype.py:331`). The model is being given unrelated decisions as context for the module it's documenting.

`DecisionRecord` already carries `confidence` and `staleness_score`, and `editor_files/fetcher.py:265` orders by `staleness_score` for a different surface, so the ranking signal exists and is used elsewhere.

The same shape appears at `page_generator/levels.py:443` for `repo_overview` (`decisions_all[:10]`, and `repo_overview.j2:70` cuts to 8). There, repo-wide scope is correct and only the ranking is missing, so it's the smaller half of this.

A fix probably wants a guard test in the style of `tests/unit/test_no_test_path_copies.py` so a page type can't go back to taking an unfiltered slice.

NgoQuocViet2001:
Fixes #1587.

## The problem

`levels.py` handed every module page the whole repository's list and the template showed the first five of it:

```python
# levels.py:174, file pages
decision_records=run.decisions_by_file.get(p.file_info.path),
# levels.py:288, module pages
decision_records=run.decisions_all,
```

The scoped map already existed, 114 lines earlier in the same file — the wrong one was being passed. On a repo whose decisions cluster in one area that is the same five on every module page, about parts of the codebase the page does not document. And since `module_page` is LLM-written, that list is prompt context, not just rendered output.

## The change

`decisions_for_files(decisions_by_file, paths)` collects the decisions governing the module's own files. Two details:

- **De-duplicated** on `(title, source, evidence_file)` — the `decision_records` uniqueness constraint — so a decision affecting several of a module's files appears once, while two genuinely distinct records sharing a title are kept apart.
- **Ranked** by `confidence`, descending, with a stable sort so ties keep discovery order. A missing or unparseable confidence sorts last rather than raising.

`repo_overview` keeps its repo-wide scope, which is correct there; only the choice of ten was list position, so it takes `rank_decisions(...)[:10]`.

## Guard test

The issue asks for one in the style of `test_no_test_path_copies`, so `test_no_page_type_slices_the_repo_wide_decision_list_unranked` reads `levels.py` and fails on any `decisions_all[...]` slice that does not go through `rank_decisions`. Repo-wide scope stays available; taking its first N by position does not.

Its ceiling is the same as its model's: it matches on the name `decisions_all`, so a future copy under another name is invisible to it. It catches the shape that actually occurred here.

## Test plan

- Ran: `uv run pytest tests/unit/generation/test_decision_scoping.py -q` — 5 passed (all new).
- Ran: `uv run pytest tests/unit/generation/ -q` — 1448 passed.
- Ran: `uv run ruff check` on `page_generator/` and the new test, and `ruff format --check` — clean, and both changed files were format-clean before the change.
- Checked: with `levels.py` reverted the guard test fails, so it is pinned to the call site rather than to the helper.

Behaviour cases cover scoping (another module's decisions do not leak in, an unindexed path yields nothing), the ordering, the de-duplication, and that a repeated title from a different evidence file is not collapsed.

## Not changed

`staleness_score` is not part of the ranking: `build_decision_maps` does not carry it into the payload, and threading it through is a wider change than this. `confidence` alone already replaces "whichever was written first" with a real signal — happy to fold staleness in as a follow-up if you want both.

### DIFF
--- a/packages/core/src/repowise/core/analysis/health/perf/dialects/ts_js.py
+++ b/packages/core/src/repowise/core/analysis/health/perf/dialects/ts_js.py
@@ -163,7 +163,12 @@ class TsJsPerfDialect(BasePerfDialect):
             if method in TS_SINK_METHODS or awaited:
                 return root_kind
             return None
-        if method in PRISMA_METHODS:  # distinctive prisma-client verbs
+        # Distinctive prisma-client verbs, but only as a MEMBER call. A prisma
+        # query is always reached through the client (``prisma.user.aggregate``),
+        # so a bare ``aggregate(xs)`` / ``groupBy(xs)`` / ``upsert(x)`` is an
+        # ordinary local function and was being reported as a database round
+        # trip -- an N+1 finding on pure arithmetic, in code with no database.
+        if is_attribute and method in PRISMA_METHODS:
             return "db"
         if method in TS_FS_METHODS:  # distinctive sync-fs verbs
             return "filesystem"
--- a/packages/core/src/repowise/core/generation/page_generator/helpers.py
+++ b/packages/core/src/repowise/core/generation/page_generator/helpers.py
@@ -10,6 +10,7 @@ from __future__ import annotations
 
 import re
 from collections import defaultdict
+from collections.abc import Iterable
 from datetime import UTC, datetime
 from pathlib import Path
 from typing import Any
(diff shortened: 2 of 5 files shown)


RaghavChamadiya (approved):
Thanks @NgoQuocViet2001, the part that makes this more than a scoping fix is `test_no_page_type_slices_the_repo_wide_decision_list_unranked`. Greping `levels.py` for a bare `decisions_all[` turns "we fixed the module page" into "no page type can quietly reintroduce this", which is the difference between a fix and a fix that holds.

The reasoning in the comment is the real argument for it: these records are not only rendered, they go into the prompt as context for the page being written. So on a repo where decisions cluster in one area, every module page was being written against the same five records regardless of what the module does. Keeping repo-wide scope on the overview and only ranking the ten there is the right split.

The dedupe key deserves a mention too. Matching `(title, source, evidence_file)` to the `decision_records` uniqueness constraint is what stops two genuinely different records with a shared title collapsing into one, and you tested that case rather than assuming it.

This also carries the `is_attribute` gate from your #1753, which merged a moment ago. The content is identical so it resolves to nothing, but that is why GitHub may show this branch oddly for a minute.

PR #1767, opened 2026-08-20
fix(parser): avoid duplicate Python decorators

### LINKED ISSUE #1698: [Bug] Every Python decorator is stored twice on the symbol
**Summary**

Python decorators are collected in two places. The query file captures them, and
then the parser walks the parent node and appends the same ones again. So
`@pytest.fixture` is stored as `['@pytest.fixture', '@pytest.fixture']` on every
decorated symbol.

**Where**

`packages/core/src/repowise/core/ingestion/queries/python.scm:41-51` captures
`(decorator) @symbol.modifiers` on both the function and the class pattern.

`packages/core/src/repowise/core/ingestion/parser.py:642-645` then does it again:

```python
if def_node.parent and def_node.parent.type == "decorated_definition":
    for sibling in def_node.parent.children:
        if sibling.type == "decorator":
            modifier_texts.append(_node_text(sibling, src))
```

**Repro**

Index any Python file with a decorated function and read the symbol's modifiers.
Every decorator appears twice.

**Expected**

One entry per decorator.

**Impact**

Stated honestly, this is mild. Anything that just checks whether a decorator is
present is unaffected, and our fixture pass stops at the first match so it does
not care. It doubles the stored list on the language that uses decorators most,
and anything that counts decorators is wrong today.

Python only. `decorated_definition` is a Python node type, and the Rust, C# and
C++ attribute paths have no second collector.

**Done looks like**

Remove whichever collector is redundant. Prefer deleting the parser side append,
since the query file capture is the declared mechanism. Add a test asserting a
decorated function has exactly one entry and a doubly decorated function has
exactly two.

**One practical note**

`parser.py` is a busy file with active work landing on it. Branch from
`origin/main` and expect to rebase.

cipher-v:
## Summary

- Remove the redundant parser-side Python decorator collection.
- Capture one or more decorators through the Python Tree-sitter query and add regression tests for single and multiple decorators.

## Related Issues

Closes #1698

## Test Plan

- [x] Python parser tests pass (`31 passed`)
- [x] Complete parser test suite passes (`223 passed`)
- [x] `git diff --check` passes
- [x] Web build not applicable because this change does not affect the frontend
- [x] Related generation integration tests pass (`52 passed`)

## Checklist

- [x] I have added tests for the bug fix
- [x] Documentation changes are not required for this internal parser fix

### DIFF
--- a/packages/core/src/repowise/core/ingestion/parser.py
+++ b/packages/core/src/repowise/core/ingestion/parser.py
@@ -688,10 +688,6 @@ class ASTParser:
 
             # Visibility
             modifier_texts = [_node_text(m, src) for m in modifier_nodes]
-            if def_node.parent and def_node.parent.type == "decorated_definition":
-                for sibling in def_node.parent.children:
-                    if sibling.type == "decorator":
-                        modifier_texts.append(_node_text(sibling, src))
 
             # Rust: outer attributes (#[...]) are preceding siblings of the item
             rust_attrs: list[str] = []
--- a/packages/core/src/repowise/core/ingestion/queries/python.scm
+++ b/packages/core/src/repowise/core/ingestion/queries/python.scm
@@ -40,7 +40,7 @@
 
 ; Decorated function or class — captures the decorator as a modifier
 (decorated_definition
-  (decorator) @symbol.modifiers
+  (decorator)+ @symbol.modifiers
   (function_definition
     name: (identifier) @symbol.name
     parameters: (parameters) @symbol.params
@@ -48,7 +48,7 @@
 )
 
 (decorated_definition
-  (decorator) @symbol.modifiers
+  (decorator)+ @symbol.modifiers
   (class_definition
     name: (identifier) @symbol.name
   ) @symbol.def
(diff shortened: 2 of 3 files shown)


RaghavChamadiya (approved):
Thanks @cipher-v. This is the version of the fix I wanted, because it removes the cause rather than compensating for it.

The duplicate was never in the parser loop. `python.scm` captured a singular `(decorator)`, and a singular node in a tree-sitter pattern matches once, so a definition carrying two decorators produced two matches of the same pattern with one decorator each. `(decorator)+` collects them into a single match, and the payoff is the four lines that then delete themselves:

```python
-            if def_node.parent and def_node.parent.type == "decorated_definition":
-                for sibling in def_node.parent.children:
-                    if sibling.type == "decorator":
-                        modifier_texts.append(_node_text(sibling, src))
```

That block was a Python-shaped special case sitting in a function that serves every language. Getting rid of it is worth more than the bug it was papering over.

Asserting source order in `test_function_multiple_decorators_recorded_once_in_source_order` is the right call too. `@cache` before `@app.post(...)` is not the same program as the reverse, so "recorded once" alone would have been an under-specified guarantee.

Another PR (#1743) arrived a day earlier for this, de-duplicating in `_extract_symbols` with an extra pass over every match. I took yours because a net deletion in the ingestion hot path beats an added pass, and I have explained that on their thread so it reads as a technical call rather than a race.

Full matrix green. Merging. This closes #1698. #1886 is assigned to you.

PR #1772, opened 2026-08-20
fix(pipeline): report dynamic hint extraction failures on update path…

### LINKED ISSUE #1697: [Bug] Dynamic-hint failures on the update path are swallowed silently
**Summary**

The dynamic hint step in the incremental pipeline is wrapped in a bare
catch-everything block that does nothing. If it fails, the run still reports
success and the graph quietly loses every dynamic edge. The step right above it
already reports failure properly, and this one was simply never included in that
fix.

**Where**

`packages/core/src/repowise/core/pipeline/incremental.py:213`

```python
except Exception:
    pass  # dynamic hints are best-effort, same as the init phase
```

The pattern to copy is in the same function at lines 191 to 196, which logs a
structured warning and prints a line telling the user what the consequence is.

**Repro**

Make `HintRegistry().extract_all` raise, for example by adding a temporary
`raise` in the registry, then run `repowise update`. The run completes, reports
success, and prints nothing. The resulting graph has no dynamic edges.

**Expected**

The step stays best effort and must not start raising, but the failure is
reported rather than hidden.

**Impact**

Dynamic edges are a live rescue for dead code detection, worth 38 findings on
leveldb alone. Losing them silently means an updated index reports dead code
that a freshly built index does not, with nothing in the output to explain why.

**Done looks like**

The handler logs a structured warning and prints a user facing line naming the
consequence, matching the neighbouring handler above it. A test asserts the
warning fires when extraction raises. Around six lines, copied from the sibling.

Shivang9983:
… (#1697)

## Summary

- Replaced silent `except Exception: pass` block in `repowise.core.pipeline.incremental` with structured `structlog` warning and a    
   user-facing consequence log.
- Preserved best-effort non-failing behavior during incremental updates when dynamic hint extraction fails.
- Added regression test `test_dynamic_hint_failure_logs_warning_and_does_not_raise` to assert proper warning emission without            
   crashing graph construction.

-

## Related Issues

Fixes #1697

## Test Plan

Ran unit test suite locally:
```bash
pytest tests/unit/pipeline/test_update_graph_convergence.py

- [ x ] Tests pass (`pytest`)

## Checklist

- [ x ] My code follows the project's code style
- [ x ] I have added tests for new functionality
- [ x ] All existing tests still pass

### DIFF
--- a/packages/core/src/repowise/core/pipeline/incremental.py
+++ b/packages/core/src/repowise/core/pipeline/incremental.py
@@ -210,11 +210,13 @@ def build_repo_graph(
         graph_builder.add_dynamic_edges(dynamic_edges)
         if dynamic_edges:
             log(f"Dynamic hint edges added: [cyan]{len(dynamic_edges)}[/cyan]")
-            
+
     except Exception as exc:
-        logger.warning("dynamic_hints_extraction_failed", error=str(exc), exc_info=True)
-        log(f"[yellow]Dynamic hint extraction failed: {exc}. Dynamic edges will be missing from this update.[/yellow]")
-        
+        logger.warning("dynamic_hints_extraction_failed", error=str(exc))
+        log(
+            f"[yellow]Dynamic hint extraction failed: {exc}. Dynamic edges"
+            "will be missing from this update.[/yellow]"
+        )
 
     return parsed_files, source_map, graph_builder, repo_structure, len(file_infos)
 
--- a/packages/core/src/repowise/core/pipeline/incremental.py
+++ b/packages/core/src/repowise/core/pipeline/incremental.py
@@ -210,8 +210,11 @@ def build_repo_graph(
         graph_builder.add_dynamic_edges(dynamic_edges)
         if dynamic_edges:
             log(f"Dynamic hint edges added: [cyan]{len(dynamic_edges)}[/cyan]")
-    except Exception:
-        pass  # dynamic hints are best-effort, same as the init phase
+            
+    except Exception as exc:
+        logger.warning("dynamic_hints_extraction_failed", error=str(exc), exc_info=True)
+        log(f"[yellow]Dynamic hint extraction failed: {exc}. Dynamic edges will be missing from this update.[/yellow]")
+        
 
     return parsed_files, source_map, graph_builder, repo_structure, len(file_infos)
 
--- a/tests/unit/pipeline/test_update_graph_convergence.py
+++ b/tests/unit/pipeline/test_update_graph_convergence.py
@@ -11,17 +11,17 @@ These tests build both graph shapes over the same small git repo and assert
 node/edge convergence plus equality of the centrality-cache signatures (the
 exact keys the cache hits on).
 """
-
 from __future__ import annotations
 
 from pathlib import Path
 from types import SimpleNamespace
+from unittest.mock import patch
 
 from repowise.core.ingestion import ASTParser, FileTraverser, GraphBuilder
 from repowise.core.ingestion.git_indexer import GitIndexer
 from repowise.core.ingestion.git_indexer.tiers import GitIndexTier
 from repowise.core.ingestion.graph._centrality_cache import subgraph_signature
-from repowise.core.pipeline.incremental import rebuild_graph_and_git
+from repowise.core.pipeline.incremental import build_repo_graph, rebuild_graph_and_git
 
 
 def _build_git_repo(tmp_path: Path) -> None:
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Review by Ayush7614 — CHANGES_REQUESTED** [review]
### Request Changes

Intent is correct — dynamic hints are best-effort and should warn instead of silent `pass`. No fake code: `packages/core/src/repowise/core/pipeline/incremental.py:29` `logger = structlog.get_logger(__name__)` exists, `logger.warning("dynamic_hints_extraction_failed", error=str(exc), exc_info=True)` at `incremental.py:193` is valid, and `build_repo_graph(...):192-194` `log("[yellow]Dynamic hint extraction failed...")` matches `LogFn` usage elsewhere.

**Blocking — CI lint FAILURE** (`Python lint (ruff)` `FAILURE` on `runs/32370307094/job/96690208258`, `mergeStateStatus: BLOCKED`):

- `packages/core/src/repowise/core/pipeline/incremental.py:191` trailing whitespace + `195:1` blank line with spaces — `ruff format --check` fails:
  ```
  191: '            '
  195: '        '
  ```
- `tests/unit/pipeline/test_update_graph_convergence.py:14` missing blank line after module docstring, `107` line-length, `128` `RUF059` 4 unused unpacks (`parsed_files`, `source_map`, `repo_structure`, `count`), `132-134` extra blank lines + missing wrapping — `ruff check` reports 4 `RUF059` + `ruff format` diff. File also missing final newline (`\ No newline at end of file`).

**Fix:**
```bash
ruff format packages/core/src/repowise/core/pipeline/incremental.py tests/unit/pipeline/test_update_graph_convergence.py
# then fix RUF059 — prefix unused with _:
_parsed_files, _source_map, _graph_builder, _repo_structure, _count = build_repo_graph(...)
# or use _ = build_repo_graph(...) and assert returned graph_builder
# + ensure EOF newline
```

**Non-blocking suggestions:**
- Test at `test_update_graph_convergence.py:119-138` only asserts `capture_log` contains `"Dynamic hint extraction failed"` / `"Dynamic edges will be missing"` — it never asserts structured `logger.warning` was called with `exc_info=True`. Consider adding `patch("repowise.core.pipeline.incremental.logger.warning")` assertion or renaming test to reflect console-only check.
- Patch target `repowise.core.ingestion.dynamic_hints.HintRegistry.extract_all` works because `HintRegistry` is imported inside `build_repo_graph:181` after patch starts — fine, but document that coupling; if refactored to top-level import, target must change.

Please run `ruff check && ruff format --check` locally and push — happy to approve once lint is green.

Ayush7614 (approved):
### Approved — lint fixed

Re-review after `21ec002 style(pipeline): resolve ruff lint errors`.

**Verification (pr-1772-new `21ec002`):**
- `packages/core/src/repowise/core/pipeline/incremental.py:214-219` — trailing whitespace / blank lines fixed, now `logger.warning("dynamic_hints_extraction_failed", error=str(exc))` + `log("[yellow]Dynamic hint extraction failed: ... Dynamic edges will be missing...")` split over 2 lines matching sibling `framework_edges_failed:192` format. `ruff check` ✅ `ruff format --check` ✅ (both files).
- `tests/unit/pipeline/test_update_graph_convergence.py:119-135` — `RUF059` fixed (`_parsed_files, _source_map, _repo_structure, _count`), line-length collapse at `subgraph_signature` line, blank-line + EOF newline fixed; `ruff` ✅.
- `statusCheckRollup` now shows `Repowise / code health SUCCESS`; full `CI` matrix not visible for this fork push (previously 8 jobs) but local `ruff` is the former blocker and is now clean. No fake APIs — `logger = structlog.get_logger(__name__)` at `incremental.py:29` exists, `HintRegistry.extract_all` patch target correct (import inside `build_repo_graph:203` so patch before call works).

Previous `CHANGES_REQUESTED` was solely for lint (`Python lint (ruff) FAILURE`, `mergeStateStatus: BLOCKED`). That is resolved — `exc_info=True` removal to match sibling is intentional per commit message and fine for structured logging.

LGTM — ready to merge (dismissing previous request).

RaghavChamadiya (approved):
Thanks @Shivang9983, and sorry for the wait after you turned this around so quickly. You did everything I asked: the four RUF059 unpacks are underscore-prefixed, the handler now matches its sibling exactly (no `exc_info`, wrapped strings), and the whitespace and trailing-newline points are closed. Full matrix green on 3.11, 3.12, 3.13 and ruff.

The reason this is worth having is the one from the review: the framework-edges handler two blocks above already said an update path has to report what the init path reports, or an index degrades differently depending on how it was built and nothing says which. The dynamic-hint block was the gap in that, not a disagreement with it.

One thing I am merging anyway rather than sending you round again for a single character. The adjacent string literals join without a space:

```python
f"[yellow]Dynamic hint extraction failed: {exc}. Dynamic edges"
"will be missing from this update.[/yellow]"
```

renders as `Dynamic edgeswill be missing from this update.` You have waited long enough on this one, so I will take that space in a follow-up rather than hold the fix for it. Flagging it because it is a genuinely easy trap: implicit concatenation across a line break swallows the space at the seam every time, and the sibling handler you copied avoids it only because its break happens to fall after a space.

Merging. This closes #1697.

PR #1782, opened 2026-08-21
fix(ingestion): constrain chained C++ calls by return type

RaghavChamadiya:
## Summary

- Carry AST-derived inner-call evidence through Java, C++, C#, and TypeScript call sites.
- Normalize declared return types and decline mixed-return overload families.
- Enable only the audited C++ future<T>.get() path, preventing unrelated methods named get from becoming call edges; the other language lanes remain behavior-preserving.
- Keep call-site evidence payloads composable with parallel contributor work.

## Results

- Removes 285 incorrect site-level C++ fallback resolutions, corresponding to 16 collapsed graph call edges.
- Source audit: 30/30 sampled removals were correct.
- Graph and export output are deterministic across repeated treatment runs.
- Dead-code findings are unchanged.
- Median graph-build cost on Seastar: 8.076 to 8.658 ms/file (+7.21%); median peak memory +3.89 MB (~12.7%).

## Testing

- Focused parser, signature, resolver, cache, and tsconfig tests: 53 passed.
- Broader graph, re-export, dead-code/reachability, persistence, export, and server-consumer selection: 342 passed.
- Ruff check passed.

Two existing execution-flow provenance assertions remain failing with fixture and endpoint code unchanged from the base revision; this change does not touch that behavior.


sloemo01 (commented):
Nice one — return-type chaining is a clean add, and I dig that it ships C++-only behind `PRODUCTION_RETURN_TYPE_CHAIN_LANGUAGES` so nothing changes for everyone else.

One thing I'd tighten while you're here: `receiver_call` is being tacked on as a dynamic attr (`calls[-1].receiver_call = ...`) instead of just being a real field on `CallSite` — even though `CallReceiver` already exists. It works on a plain `@dataclass`, but it trips up `dataclasses.replace`, which you can see for yourself in the module-call branch: the `replace(...)` silently drops `receiver_call` and you have to glue it back by hand. That's the kind of thing that bites later.

Since `CallReceiver` is already there, maybe just declare `receiver_call: CallReceiver | None = None` on `CallSite`. Then `replace()` carries it for free and those five `# type: ignore[attr-defined]` lines disappear. And you can keep the default `None` so extraction stays "only set when we actually captured a chained call."

Also nice that the chained captures went into the `.scm` files rather than scanning text — the field-name reads make it formatting-proof.

swati510 (approved):

PR #1789, opened 2026-08-21
fix(context): used_by from call edges, not header/impl pairing

### LINKED ISSUE #1655: [Bug] used_by is populated from the header/impl pairing, not from resolved call edges — names a file that never references the symbol, omits the one that does
## Summary

`used_by` appears to be populated from the **header/implementation pairing**, not from resolved `calls` / `references` edges. The result is wrong in both directions at once:

- it **names a file that never mentions the symbol**, and
- it **omits the file that actually calls it** — even when the call edge resolved successfully.

For a consumer this is worse than a missing edge. `repowise context <sym>` is one of the first things an agent reaches for, and a fabricated `used_by` invites reasoning about a relationship that does not exist in the source.

This was originally observed alongside #1600/#1601/#1602 but left out of that batch, because at the time I could not reproduce it outside a large tree and did not want to file something unproven. It reproduces minimally now.

## Repro

Three files, no build system needed.

`src/thing.h`
```cpp
#pragma once

namespace demo
{
    class Thing
    {
    public:
        int Value() const;
    };

    int FreeValue();
}
```

`src/thing.cpp`
```cpp
#include "thing.h"

namespace demo
{
    int Thing::Value() const
    {
        return 1;
    }

    int FreeValue()
    {
        return 2;
    }
}
```

`src/main.cpp`
```cpp
#include "thing.h"

int main()
{
    demo::Thing t;
    int a = t.Value();          // cross-file MEMBER call
    int b = demo::FreeValue();  // cross-file FREE call
    return a + b;
}
```

```
git init && git add -A && git commit -m repro
repowise init . --mode fast --no-prose --no-editor-setup --no-claude-md -y
repowise context Value
repowise context FreeValue
```

## Actual

```
used_by: src/thing.h      (for Value)
used_by: src/thing.h      (for FreeValue)
```

`src/thing.h` contains **zero** references to either symbol — it holds only their declarations. `src/main.cpp`, which calls both, is absent from `used_by` in both cases.

Note the second one especially: the run reports

```
[info] Call edges resolved  total=1
[info] Graph built          edge_types={'defines': 10, 'imports': 3, 'calls': 1}
```

so a `calls` edge for `FreeValue` **was** resolved from `main.cpp` — and `used_by` still does not mention `main.cpp`. That is what makes me think `used_by` is not reading the call/reference edges at all.

The same run logs

```
[info] header_pair_edges  added=1
```

which matches the single bogus `used_by` entry, and the count tracks the number of header/impl pairs in every tree I have tried (1 here, 3 in a 6-file variant).

## Expected

Either of these would be fine:

- `used_by` lists the files that actually reference the symbol (here: `src/main.cpp`), or
- `used_by` is empty when nothing references it.

What should not happen is a file appearing there purely because it is the paired header. If the header→impl link is useful to surface, it seems like it wants to be its own relation (`declared_in`, say) rather than being folded into `used_by`, since "declares" and "uses" mean different things to anything consuming the graph.

## Impact

`dead-code` seems unaffected — a symbol with a fabricated `used_by` is still correctly reported (verified separately: a genuinely-unused free function carries `used_by: <paired header>` and is still flagged). So this looks confined to the graph-surface consumers: `context`, `get_context`, `get_symbol`, and anything else answering "who uses this".

On a real ~2,500-file C++23 tree, **every** `used_by` edge I sampled was of this kind — 9 for 9, none correct.

## Environment

- repowise 0.43.0 installed from `main` (`pip install git+https://github.com/repowise-dev/repowise@main`), at/after `fe173c62d4fd`
- Windows 11, Python 3.13.7
- C++ (also visible with `--mode fast`, no LLM provider configured)

Installed from `main` rather than a release because #1600/#1601/#1602 merged after `v0.43.0` was tagged, so no released version carries them yet — happy to re-check against `v0.44.0` when it lands.

sloemo01:
## What

Fixes #1655. `get_context`'s `used_by` for a symbol was populated from the C++ header/impl pairing, not from resolved call/reference edges — it named the paired header (which never references the symbol) and omitted the file that actually calls it.

## Root cause

The `used_by` query matched `GraphEdge.target_node_id == sym.file_path` (the bare path). Symbol nodes are keyed `<path>::<name>`, and call/reference edges point *at* that symbol node — so the query never matched a single call edge. It only ever caught file→file edges, which for a C++ symbol is the synthetic header/source `imports` pair (minted with `imported_names = []`).

## Fix

`targets.py` now matches **either** kind of evidence:

1. **symbol→symbol `calls` / `references` edges** targeting the symbol node (`<path>::<name>`) — the real cross-file C++ callers.
2. **file→file `imports` edges whose `imported_names` actually names the symbol** — gated on the name so the empty-imports header pair is not served as a fabricated user.

Distinct sources afterwards, so a file that both imports and calls a symbol is one user. The Python file-import path (`imported_names`-based) is preserved — existing ranking tests pass unchanged.

## Tests

- `test_used_by_ranking.py` — 6 passed (4 existing + 2 new):
  - `test_call_edge_to_symbol_node_is_a_user` — a cross-file C++ call edge now surfaces the caller in `used_by`.
  - `test_empty_header_pair_edge_is_not_a_user` — the synthetic header pair (empty `imported_names`) is excluded.
- Targeted context/used_by suites: **42 passed**, no regression.

This is confined to the graph-surface consumers (`context` / `get_context` / `get_symbol`); `dead-code` was already correct per the issue.

### DIFF
--- a/packages/server/src/repowise/server/mcp_server/tool_context/targets.py
+++ b/packages/server/src/repowise/server/mcp_server/tool_context/targets.py
@@ -804,6 +804,25 @@ async def _resolve_one_target(
             # files joined by both an import and a call are one user of this
             # symbol — that is a guard, not a fix: the same corpus holds zero
             # duplicate rows, so it costs nothing and prevents nothing today.
+            #
+            # A symbol's users live at two node depths, so two kinds of edge have
+            # to be matched, not one:
+            #
+            #   1. symbol -> symbol ``calls`` / ``references`` edges, which point
+            #      at the symbol node (``<path>::<name>``). For a C++ free
+            #      function the only evidence of a use is such an edge, so
+            #      querying by the bare ``sym.file_path`` never sees it.
+            #   2. file -> file ``imports`` edges, where the source file really
+            #      does import the symbol. These must be gated on the edge's
+            #      ``imported_names`` actually naming the symbol, otherwise the
+            #      synthetic C++ header/source pair (``imported_names = []``)
+            #      is served up as a "user" even though the paired header never
+            #      references the symbol.
+            #
+            # Match either: symbol-node call edges, or import edges whose
+            # ``imported_names`` mentions the symbol. Distinct sources
+            # afterwards, so a file that both imports and calls is one user.
+            sym_node_id = f"{sym.file_path}::{sym.name}"
             res = await session.execute(
                 select(GraphEdge.source_node_id, GraphNode.pagerank)
                 .outerjoin(
@@ -813,8 +832,11 @@ async def _resolve_one_target(
                 )
                 .where(
                     GraphEdge.repository_id == repo_id,
-                    GraphEdge.target_node_id == sym.file_path,
                     GraphEdge.edge_type.notin_(NON_DEPENDENCY_EDGE_TYPES),
+                    or_(
+                        GraphEdge.target_node_id == sym_node_id,
+                        GraphEdge.imported_names_json.contains(sym.name),
+                    ),
                 )
             )
             best_rank: dict[str, float] = {}
--- a/tests/unit/server/mcp/test_used_by_ranking.py
+++ b/tests/unit/server/mcp/test_used_by_ranking.py
@@ -162,3 +162,109 @@ async def test_one_user_joined_by_two_edge_types_is_listed_once(
     used_by = (await _card(session, repository, "AuthService"))["docs"]["used_by"]
 
     assert used_by.count("src/auth/middleware.py") == 1
+
+
+async def test_call_edge_to_symbol_node_is_a_user(session, repository, populated_db) -> None:
+    """#1655: a C++ ``used_by`` must come from resolved call edges.
+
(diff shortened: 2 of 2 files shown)


sloemo01:
Main already carries this fix, so I am closing rather than rebasing a four-hundred-commit-stale branch.

`tool_context/targets.py` on main keys `used_by` on the symbol's own node id (`getattr(sym, "symbol_id", None) or getattr(sym, "node_id", None)`), which is the change this branch made, and its comment gives the same reason: a `calls` edge is symbol-to-symbol and targets a `path::Name`, so keying on the file path could only ever answer "who imports this symbol's file".

Main's `tests/unit/server/mcp/test_used_by_ranking.py` asserts the same two contracts in its docstring, the source keying and the ranking of which twenty survive the cut.

Main's version also goes further than this one did: it matches both node depths, symbol-to-symbol call and reference edges plus file-to-file imports gated on `imported_names` actually naming the symbol, and it cites the corpus numbers behind the ranking.

The merge conflict was the useful signal, since it was the same lines changed on both sides.

PR #1793, opened 2026-08-21
fix(server): make CORS browser-compliant and configurable via env

Ayush7614:
## What

Every `repowise serve --host` deployment had **broken CORS**:

- `packages/server/src/repowise/server/app.py:445-452` set `CORSMiddleware(allow_origins=["*"], allow_credentials=True, ...)`
- Browsers **reject** `Access-Control-Allow-Origin: *` with `Allow-Credentials: true` (preflight fails) — legitimate browser UI on a team host broke, while `curl`/`fetch(mode:no-cors)` ignored CORS and got cross-site credentialed access from any origin anyway.

Auth via `deps.verify_api_key` already exists per-router (most routers have `dependencies=[Depends(verify_api_key)]`), but CORS misconfig made the browser boundary wrong.

## Fix — `app.py:445-465`

- Default (no env): `allow_origins=["*"]`, `allow_credentials=False` — **browser-spec valid**, safe for `localhost` dev without config.
- When `REPOWISE_CORS_ORIGINS` env is set (comma-separated, e.g. `https://example.com,https://app.example.com`), use explicit origins with `allow_credentials=True` (required for bearer auth from browsers).
- Warns if `REPOWISE_CORS_ALLOW_CREDENTIALS` is set with wildcard (explicit misconfig that browsers reject).

## Verification

```
REPOWISE_CORS_ORIGINS="" -> ["*"] False  # local dev, valid
REPOWISE_CORS_ORIGINS="https://example.com, https://app.example.com" -> explicit origins True
```
- `ruff check/format` clean
- No open issue tracks CORS (searched `CORS`/`allow_credentials` — 0 hits); related #1736 is about host-header 421, not CORS. Gap found by reading `app.py` CORS block and comparing to browser spec.

## Risk

- No behavior change for localhost dev (still `*` but now spec-compliant without credentials — previously credentials with * was rejected anyway).
- Team deployments set `REPOWISE_CORS_ORIGINS` to their actual origins once — no break for those already on loopback.

### DIFF
--- a/packages/server/src/repowise/server/app.py
+++ b/packages/server/src/repowise/server/app.py
@@ -458,11 +458,31 @@ def create_app() -> FastAPI:
         lifespan=lifespan,
     )
 
-    # CORS — allow all origins for local development
+    # CORS — configurable; default allows local dev but is browser-spec compliant.
+    # Browsers reject `Access-Control-Allow-Origin: *` with `Allow-Credentials: true`
+    # (preflight fails). When REPOWISE_CORS_ORIGINS is unset we allow any origin
+    # without credentials (safe for local dev); when credentials are needed the
+    # operator must set explicit origins via REPOWISE_CORS_ORIGINS.
+    cors_origins_env = os.environ.get("REPOWISE_CORS_ORIGINS", "").strip()
+    if cors_origins_env:
+        cors_origins = [o.strip() for o in cors_origins_env.split(",") if o.strip()]
+        cors_allow_credentials = True
+    else:
+        cors_origins = ["*"]
+        cors_allow_credentials = False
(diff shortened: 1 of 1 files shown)


RaghavChamadiya (approved):
Thanks @Ayush7614. The detail that makes this easy to approve is that `allow_origins=["*"]` together with `allow_credentials=True` is already rejected by every spec-compliant browser, so the combination on `app.py:531-532` has never worked for the case it looks like it supports. Moving to a permissive default without credentials, and letting an operator name real origins through `REPOWISE_CORS_ORIGINS` when they need cookies, is the only shape that actually functions.

Worth saying for anyone reading later: this is a browser header contract only, so `verify_api_key` is unaffected either way, and a local or curl user sees no change.

PR #1795, opened 2026-08-21
fix(health): count Python match arms toward cyclomatic complexity

### LINKED ISSUE #1774: [Bug] Python match arms are not counted as branches, so max_ccn undercounts match-heavy functions
## Describe the Bug

Python `match` statement arms do not appear to be counted as branches by the complexity walker, so `max_ccn` is reported lower than independent tools measure — sometimes far lower.

This matters more than a cosmetic metric gap: we used Repowise's health output to *generate* the acceptance criteria for a refactoring epic, so the understated baselines propagated into the work items themselves.

## Evidence

Three functions, measured with `lizard` and `ruff --select C901 --config lint.mccabe.max-complexity=N` (bisected) against Repowise's `max_ccn`:

| function (Python 3.14) | Repowise `max_ccn` | lizard CCN | notes |
|---|---|---|---|
| a FastAPI router built from two `match` statements | **3** | **8** | biggest gap; the function is essentially *only* `match` arms |
| `recover_orphaned_work` (task module) | **19** | **29** | 10-point gap |
| `stage_cloud_window` (task module) | **23** | **24** | ~1 point; ruff reads 18 here |

The first case is the clearest: a router whose two handlers are built almost entirely from `match` arms reported `max_ccn = 3`, while lizard measured 8 for the same functions. Removing the `match` arms from consideration is the only hypothesis we found that explains a 3-vs-8 reading on that shape.

The third row is included deliberately as a **counter-example** — there Repowise sat *between* ruff (18) and lizard (24), which is ordinary metric-definition drift rather than a defect. So this is not "Repowise is always low"; it looks specific to `match`.

## Steps to Reproduce

1. Index a Python 3.10+ repo containing a function whose branching is predominantly `match` / `case` arms.
2. Compare `max_ccn` from `get_health` / `repowise health` against `lizard` for the same function.
3. The larger the proportion of branching expressed as `match` arms, the larger the gap.

## Expected

Each `case` arm contributes to cyclomatic complexity, the way `elif` branches do — so a `match` with N arms reads comparably to an equivalent `if/elif` chain.

## Actual

`match`-heavy functions report a CCN close to their non-`match` control flow alone, so `complex_method` findings are missed on functions that other tools flag.

## Why it matters

`match` is idiomatic in modern Python for exactly the dispatch-heavy code that most warrants a complexity signal — routers, protocol handlers, state machines. Those are the functions most likely to be under-reported.

## Related

- #451 (top-level arrow/lambda functions invisible to the complexity walker) — same class of problem, different language and node kind. This report is the Python `match` analogue.

## Environment

- Python 3.14 codebase, ~304 source files
- Comparison tools: `lizard`, `ruff` mccabe (`C901`, bisected via `max-complexity`)

sloemo01:
## What

Fixes #1774. Python `match`/`case` arms are not counted as branches, so `max_ccn` undercounted match-heavy functions — sometimes far lower than independent tools measure.

## Root cause

`_is_flat_match` in `cyclomatic.py` classified Python's `match_statement` as a "flat" switch, and flat-match arms are deliberately counted **once for the dispatch, not per arm** (design note at `languages.py:37-38`). But Python `match`/`case` is *pattern matching* — each `case` is a distinct branch point, structurally the same decision as an `if`/`elif` chain. The suppression that's correct for C `switch`/Rust `match` dispatch was being over-applied to Python.

Measured on `main` before the fix:
```
# match, 3 arms                     # equivalent if/elif
match x:                                if x == 1: return "a"
    case 1: return "a"                  elif x == 2: return "b"
    case 2: return "b"                  elif x == 3: return "c"
    case 3: return "c"                  else: return "z"
    case _: return "z"
match 3 arms   -> ccn = 2        (1 entry + 1 match; all arms dropped)
if/elif 3      -> ccn = 4        (1 + 3 branches, each counted)
```

## Fix

`_is_flat_match` now returns `False` for Python's `match_statement`, so each `case_clause` counts a branch like an `elif`. Rust `match_expression` and other dispatch switches keep the flat behaviour — the existing `flat_match.rs` fixture is unchanged.

## Behavioral note on the large-method detector

This surfaces a deliberate design conflict: the `LargeMethodDetector` had a test asserting a 31-case Python dispatch table sits at CCN 2 and is "layout, not a size smell" (thus dodging the `_CCN_FLOOR = 3` gate). With arms now counted, that same 65-line dispatcher reports CCN 33 and is correctly flagged as a large method. I updated that test to match the corrected behavior — a 31-case dispatch is real branching and should be flagged.

## Tests

- New fixture `python/match.py` + `test_python_match_arms_count_toward_ccn` (4-arm match → CCN 5)
- Updated `test_large_method_fires_on_match_dispatch` (was `..._skips_flat_match_dispatch`)
- Full `tests/unit/health/`: **1187 passed**, no regressions
- Ruff clean

This is a behavior change worth a maintainer's eye — it makes Python match-heavy functions report higher (and more accurate) CCN, which will surface new large-method/brain-method findings on real codebases.

### DIFF
--- a/packages/core/src/repowise/core/analysis/health/complexity/cyclomatic.py
+++ b/packages/core/src/repowise/core/analysis/health/complexity/cyclomatic.py
@@ -197,7 +197,17 @@ def _is_flat_match(node: Node, lmap: LanguageNodeMap) -> bool:
     ``loop``, ``if_let``, ``while_let`` expressions, and no ``block``
     with multiple statements).  Flat matches contribute 1 CCN point for
     the match keyword itself but do NOT count each arm individually.
+
+    Python's ``match_statement`` is excluded: ``match``/``case`` in Python
+    is *pattern matching* — each ``case`` is a distinct branch point,
+    structurally the same decision as ``if``/``elif``. Suppressing its arms
+    like a C-style flat switch would report a ``match`` with N arms as ~N
+    points too low (measured: a 3-arm ``match`` gave CCN 2 while the
+    equivalent ``if``/``elif`` gave 4). Rust ``match_expression`` and other
+    languages' dispatch switches keep the flat behaviour.
     """
+    if node.type == "match_statement":
+        return False
     complex_types = lmap.branch_kinds | lmap.loop_kinds | lmap.switch_kinds
     cases = _collect_case_children(node, lmap)
     if not cases:
--- a/tests/fixtures/lang_samples/python/match.py
+++ b/tests/fixtures/lang_samples/python/match.py
@@ -1,0 +1,27 @@
+# Python match/case is pattern matching — each case is a distinct branch
+# point, structurally the same decision as if/elif. Unlike a flat Rust
+# match (or C-style switch), the arms must count toward CCN.
+#
+# python_match 3 arms -> base 1 + 3 cases = 4
+def python_match(x):
+    match x:
+        case 1:
+            return "one"
+        case 2:
+            return "two"
+        case 3:
+            return "three"
+        case _:
+            return "other"
+
+
+# Equivalent if/elif — must report the same CCN as python_match above.
+def python_if_elif(x):
+    if x == 1:
+        return "one"
+    elif x == 2:
+        return "two"
+    elif x == 3:
+        return "three"
+    else:
+        return "other"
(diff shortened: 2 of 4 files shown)


RaghavChamadiya (approved):
Verified against current main: applies cleanly, matches lizard on all three shapes, health suite green locally. Fixture comment fix and the HEALTH_ANALYZER_VERSION bump are ours to carry, per the comment above.

PR #1796, opened 2026-08-21
fix(health): don't count comments in a parameter list as parameters

### LINKED ISSUE #1775: [Bug] Annotated[..., Form(...)] signatures inflate the reported parameter count (5 params reported as 15)
## Describe the Bug

A FastAPI route handler with 5 parameters is reported by Repowise as taking **15 parameters**. The signature uses `Annotated[...]` with `Form(...)`/`Query(...)` metadata and carries an explanatory comment block, and the commas inside those brackets — and apparently inside comments — appear to be counted as top-level parameter separators.

The result is a spurious "too many parameters" finding on a function that has a perfectly ordinary arity.

## Evidence

The function (Python 3.14, FastAPI):

```python
async def trigger_scan(
    request: Request,
    session: Annotated[AsyncSession, Depends(get_session)],
    # <9-line comment block explaining the pattern below, containing several commas>
    agent_id: Annotated[str, Form(pattern=r"^[a-z0-9]+(-[a-z0-9]+)*$", max_length=128)],
    scan_root: Annotated[str, Form()],
    subpath: Annotated[str, Form()] = "",
) -> HTMLResponse:
```

Ground truth via `ast`:

```
count: 5
names: ['request', 'session', 'agent_id', 'scan_root', 'subpath']
```

Repowise's finding list: `trigger_scan takes 15 parameters`.

A naive comma-split of the parameter-list source region reproduces an inflated count: `Annotated[AsyncSession, Depends(...)]` contributes one extra, `Form(pattern=..., max_length=128)` another, and the comment block's prose commas contribute the rest.

## Steps to Reproduce

1. Index a Python repo with FastAPI-style handlers using `Annotated[T, Form(...)]` / `Annotated[T, Query(...)]` with multiple keyword arguments inside the metadata call.
2. Add an interleaved comment block containing commas inside the parameter list (idiomatic where a validation pattern needs explaining).
3. Inspect the health findings for that function.
4. The reported parameter count exceeds the real arity, scaling with bracket-internal and comment commas.

## Expected

Parameter count comes from the parsed signature — e.g. `len(posonlyargs + args + kwonlyargs)` — so bracket-internal and comment commas cannot contribute.

## Actual

Counts appear to be derived from splitting the signature text on commas, producing 15 for a 5-parameter function and a spurious high-arity finding.

## Why it matters

`Annotated[...]` metadata is the standard way to express validation in FastAPI and increasingly common in Pydantic-based code generally, so the false positive lands precisely on idiomatic modern Python. It also erodes trust in the findings list as a whole: we caught this only because a developer checked the signature by hand, and thereafter treated every findings-list claim as needing verification.

## Note on scope

The *score* metrics (`nloc`, nesting) looked sane on this file — this is specific to the parameter-count finding.

## Environment

- Python 3.14, FastAPI, ~304 source files
- Ground truth from `ast.parse` on the same file Repowise indexed

sloemo01:
## What

Fixes #1775. A comment inside a Python parameter list was counted as a parameter, inflating the reported arity of `Annotated[T, Form(...)]` / `Annotated[T, Query(...)]` FastAPI handlers and tripping a spurious high-arity finding.

## Root cause

`_count_parameters` in `ast_utils.py` counts every named child of the parameter-list node that isn't in its skip list. A `comment` node inside the signature is named but carries no arity — so an explanatory comment block between parameters (idiomatic when a validation pattern needs explaining) was counted as a parameter. The `Annotated[...]` / `Form(...)` commas are already contained within their own `typed_parameter` nodes, so they don't leak; the comment node was the only gap.

Measured before the fix: a 5-parameter handler with one interleaved comment reported `param_count=6`; on the issue's real signature (more comment commas) it scaled to 15.

## Fix

Add `"comment"` to the skip list in `_count_parameters`, alongside the splat tokens, `keyword_separator`/`positional_separator`, and punctuation. A comment node is named but never an arity-bearing parameter.

## Tests

- Added `test_python_comment_in_parameter_list_not_counted` — a 5-parameter `Annotated`/`Form` handler with an interleaved comma-laden comment now reports `param_count=5`.
- Full `tests/unit/health/`: **1136 passed**, no regressions.
- Ruff clean.

Note: this is scoped to the comment leak (#1775). The separate question of whether a bare `self`/`cls` node is skipped for *methods* is a distinct pre-existing behaviour, untouched here.

### DIFF
--- a/packages/core/src/repowise/core/analysis/health/complexity/ast_utils.py
+++ b/packages/core/src/repowise/core/analysis/health/complexity/ast_utils.py
@@ -240,7 +240,10 @@ def _count_parameters(fn_node: Node) -> int:
         # A bare ``*`` (keyword-only marker) and a bare ``/`` (positional-only
         # marker) parse as named ``keyword_separator`` / ``positional_separator``
         # nodes but carry no arity, so they must be skipped alongside the
-        # ``*``/``**`` splat tokens and the punctuation.
+        # ``*``/``**`` splat tokens and the punctuation. A ``comment`` node is
+        # also named but carries no arity — an explanatory comment block inside
+        # the parameter list (idiomatic for a validation pattern) must not
+        # count as a parameter (#1775).
         if child.type in (
             "(",
             ")",
@@ -252,6 +255,7 @@ def _count_parameters(fn_node: Node) -> int:
             "**",
             "keyword_separator",
             "positional_separator",
+            "comment",
         ):
             continue
         if child.type == "declArg":
--- a/tests/unit/health/test_complexity_walker.py
+++ b/tests/unit/health/test_complexity_walker.py
@@ -95,6 +95,31 @@ def test_python_complex_method_ccn():
     assert many.ccn >= 9, f"expected CCN ≥ 9, got {many.ccn}"
 
 
+def test_python_comment_in_parameter_list_not_counted():
+    """#1775: a comment inside a parameter list must not inflate param_count.
+
+    ``Annotated[T, Form(...)]`` metadata and an explanatory comment block are
+    idiomatic in FastAPI/Pydantic signatures. The comment node is named but
+    carries no arity; before the fix it was counted, so a 5-parameter handler
+    reported 6+ (and worse on longer signatures) and tripped a spurious
+    high-arity finding.
+    """
+    src = """async def trigger_scan(
+    request: Request,
+    session: Annotated[AsyncSession, Depends(get_session)],
+    # comment with, several, commas explaining the pattern
+    agent_id: Annotated[str, Form(pattern=r\"^[a-z0-9]+(-[a-z0-9]+)*$\", max_length=128)],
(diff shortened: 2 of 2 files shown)


RaghavChamadiya (approved):
Thanks @sloemo01 for putting this next to the existing skips rather than filtering comments somewhere upstream. `keyword_separator` and `positional_separator` are already there for exactly the same reason, a named node that carries no arity, so `comment` belongs in that tuple and nowhere else.

The test case is well chosen. A comment containing commas inside an `Annotated[...]` signature is the shape that actually shows up in FastAPI handlers, and it is also the shape where a naive comma-count would go wrong in a different way, so the test would catch a regression that took either route.

Worth noting how this lands beyond the reported case: high-arity findings feed the maintainability score, so an inflated `param_count` was not just a wrong number on one page, it was quietly pushing files toward a finding they did not earn. Any language whose parameter list can hold a comment gets that back, not only Python.

PR #1797, opened 2026-08-21
fix(risk): emit is_hotspot from the backend, not client re-derivation

### LINKED ISSUE #1788: get_risk target rows omit is_hotspot, so every renderer re-derives it without the absolute floors
## Summary

`get_risk`'s per-target rows carry `hotspot_score` but no `is_hotspot`, so every renderer has to re-derive the badge by thresholding the score. The score alone is not the backend's definition, so each renderer that does this gets it subtly wrong in the same way.

## Mechanism

`git_indexer/enrich.py:218` gates the flag on two things:

```python
        if churn_pct >= 0.75 and meets_hotspot_floors(meta):
            meta["is_hotspot"] = True
```

and the floors exist precisely because the score on its own is not enough:

```python
def meets_hotspot_floors(meta: dict) -> bool:
    """Absolute activity floors for hotspot classification (issue #361).

    The churn percentile is repo-relative, so on a quiet repo the top
    quartile degenerates to "any file touched recently".
    """
```

A client sees only `churn_percentile` / `hotspot_score`, so it can reproduce the `>= 0.75` half and cannot reproduce the floors. On a small or quiet repo it badges files the backend does not consider hotspots, which is the degeneracy #361 was filed to fix, reintroduced one layer out.

## Evidence this is a real trap and not a hypothetical

It has now cost two rounds on one PR. #1480 first rendered `0.91` as `1th pct` and gated its badge on `score >= 80`, which can never fire for a 0-1 value; its tests passed because the fixtures invented `hotspot_score: 91`. The corrected version derives the badge from `score >= 0.75`, matching `enrich.py`, and is as close as a client can get. It still cannot apply the floors.

The scale is a second trap in the same field: `hotspot_score` is a 0-1 fraction across the MCP boundary, while the same-named `churn_percentile` on `ChurnComplexity` in `analysis/health/churn_complexity.py` is 0-100.

## Proposed fix

Emit `is_hotspot` on the per-target row in `tool_risk/assessment.py`, read from the stored `GitMetadata` flag rather than recomputed, so there is exactly one definition of hotspot and it lives where the floors are. Clients then read the boolean and use the score only for display.

Worth deciding at the same time whether `global_hotspots` entries should carry it too. They are selected as hotspots already, so it may be redundant there, and a redundant-but-consistent field is arguably better than a field that appears on one row shape and not another.

Once it ships, the derived threshold in the UI should be removed rather than left as a fallback, otherwise the wrong answer stays reachable.

## Not in scope

Changing the floors or the 0.75 cut. This is about which side of the wire applies them.

sloemo01:
## What

Fixes #1788. `get_risk` per-target rows carried `hotspot_score` but not `is_hotspot`, so every renderer re-derived the badge by thresholding the score — which can't reproduce the backend's definition and reintroduces the quiet-repo degeneracy #361 was filed to fix.

## Root cause

`is_hotspot` is computed once in `git_indexer/enrich.py` and gated on **two** things: `churn_pct >= 0.75` **and** `meets_hotspot_floors(meta)`. The floors exist because the churn percentile is repo-relative — on a quiet repo the top quartile degenerates to "any file touched recently". A client sees only the score, so it can reproduce the `>= 0.75` half but not the floors, and badged files the backend does not consider hotspots.

## Fix

- **`assessment.py`**: emit `is_hotspot` on the per-target row (and the no-git-metadata fallback) read from the stored `GitMetadata` flag — the single source of truth where the floors already live.
- **`get_risk.py`**: `global_hotspots` entries are hotspots by construction (selected by `is_hotspot == True`), but carry the field for a consistent row shape across both target types.
- **`artifacts.tsx`**: trust the backend flag and **remove** the `score >= 0.75` fallback, so the wrong answer is not left reachable. Dropped the now-unused `HOTSPOT_SCORE_THRESHOLD`.
- Updated the two UI artifacts tests to send `is_hotspot` rather than rely on the derived threshold.

## Tests

- Backend: `tests/unit/server/mcp/test_risk.py` + `test_hotspot_health_surfaces.py` — **21 passed**, no regressions.
- Ruff clean on the backend files.
- UI: changed `artifacts.test.tsx` fixtures to the new contract; type-check runs in CI (Web UI lint + type check).

### DIFF
--- a/packages/server/src/repowise/server/mcp_server/tool_risk/assessment.py
+++ b/packages/server/src/repowise/server/mcp_server/tool_risk/assessment.py
@@ -413,6 +413,7 @@ async def _assess_one_target(
 
     if meta is None:
         result_data["hotspot_score"] = 0.0
+        result_data["is_hotspot"] = False
         result_data["dependents_count"] = dep_count
         result_data["co_change_partners"] = []
         result_data["primary_owner"] = None
@@ -453,6 +454,12 @@ async def _assess_one_target(
     bus_factor = getattr(meta, "bus_factor", 0) or 0
 
     result_data["hotspot_score"] = hotspot_score
+    # Emit the backend's hotspot classification, not a client re-derivation.
+    # ``is_hotspot`` lives on the stored GitMetadata row and already applies
+    # the absolute activity floors (issue #361); a client that thresholds the
+    # score alone can reproduce the churn half but not the floors, so on a
+    # quiet repo it would badge files the backend does not consider hotspots.
+    result_data["is_hotspot"] = bool(getattr(meta, "is_hotspot", False))
     result_data["dependents_count"] = dep_count
     result_data["co_change_partners"] = co_changes
     result_data["primary_owner"] = owner
--- a/packages/server/src/repowise/server/mcp_server/tool_risk/get_risk.py
+++ b/packages/server/src/repowise/server/mcp_server/tool_risk/get_risk.py
@@ -161,6 +161,7 @@ async def get_risk(
             entry = {
                 "file_path": h.file_path,
                 "hotspot_score": h.churn_percentile,
+                "is_hotspot": True,
                 "primary_owner": h.primary_owner_name,
             }
             # Silent on files with no counted fixes, so a repo without fix
--- a/packages/ui/__tests__/chat/artifacts.test.tsx
+++ b/packages/ui/__tests__/chat/artifacts.test.tsx
@@ -59,7 +59,7 @@ describe("chat artifact renderers", () => {
         data={{
           targets: [
             // Wire scores are 0–1 fractions (rank / total), not 0–100.
-            { file_path: "src/hot.ts", churn_percentile: 0.99 },
+            { file_path: "src/hot.ts", churn_percentile: 0.99, is_hotspot: true },
           ],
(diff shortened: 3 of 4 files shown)


RaghavChamadiya (approved):
Thanks @sloemo01, and thanks in particular for deleting `HOTSPOT_SCORE_THRESHOLD` instead of leaving it as a fallback. Keeping it as an `||` branch would have looked safer and been worse: the derived answer would still win whenever the backend said false, which is exactly the case the floors exist to handle.

The reasoning holds up against the source. `is_hotspot` on the `GitMetadata` row applies the absolute activity floors from #361, and a client thresholding `hotspot_score >= 0.75` can reproduce the churn percentile but has no way to see the floors, because the floors are not a function of the percentile. On a quiet repo the top quartile is still the top quartile, so the UI was badging files the backend deliberately does not call hotspots.

Setting `is_hotspot = False` on the `meta is None` branch matters too. Without it that path emits a row missing the key, and every consumer has to decide what absent means, which is how the re-derivation got written in the first place.

`"is_hotspot": True` on the `global_hotspots` rows in `get_risk.py` is correct rather than lazy: those rows come from the hotspot query, so membership is the classification.

Green on the full matrix. Merging. This closes #1788.

PR #1798, opened 2026-08-21
fix(security): store provider_config 0600 and redact keys in validate response

Ayush7614:
## What

`provider_config.json` (`~/.repowise/provider_config.json`) holds plaintext API keys (openai/anthropic/etc.) but was written **0644** — any user on a shared host can `cat` it. The `/providers/{id}/validate` endpoint also returned raw `str(exc)` which SDKs embed as `Bearer sk-...` on 401, leaking keys to the UI response.

## Fix

- **`provider_config.py:179-196`** `_save_config`: `chmod 0600` on tmp and final path via `contextlib.suppress(OSError)` (best-effort on Windows). Adds `_redact_key` helper (`re.sub(r'sk-[A-Za-z0-9\-_]{4,}', 'sk-***')`)
- **`routers/providers.py:11,132,145`** — import `_redact_key` at top, redact both instantiation and generation errors in `validate_provider`.

## Verification

```python
os.environ['REPOWISE_CONFIG_DIR']=tmp; set_api_key('openai','sk-test-12345'); assert oct(_config_path().stat().st_mode)[-3:] == '600'
_redact_key('Bearer sk-proj-abc123') == 'Bearer sk-***'
```
- `ruff check/format` clean
- `pytest tests/unit/server/test_provider_config.py` — 16 passed
- No open issue tracks perms/scrubbing (searched `chmod`/`scrub` 0 hits)

### DIFF
--- a/packages/server/src/repowise/server/provider_config.py
+++ b/packages/server/src/repowise/server/provider_config.py
@@ -22,6 +22,7 @@ later CLI run in the repo picks it up.
 
 from __future__ import annotations
 
+import contextlib
 import json
 import logging
 import os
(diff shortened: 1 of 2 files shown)


RaghavChamadiya (approved):
Thanks @Ayush7614. Two independent leaks in one pass, and both fixes are the modest version of themselves, which is why this is easy to take. Writing `provider_config.json` at 0600 through `contextlib.suppress(OSError)` keeps Windows working where `chmod` is a no-op, and redacting in the two `except` paths at `providers.py:137` and `:142` stops a provider exception echoing a key back through the validate response.

One non-blocking note for whoever picks it up next, not a reason to hold this. `_redact_key` matches `sk-` prefixed keys, so OpenAI, Anthropic and OpenRouter material is covered, but a Gemini key resolved through `GEMINI_API_KEY` or `GOOGLE_API_KEY` in this same file has a different shape and would still pass through if its SDK embeds it in an error string. Worth a follow-up that widens the pattern per provider. The 0600 write and the partial redaction are both strict improvements over what is on main today, so this goes in as is.

PR #1800, opened 2026-08-21
fix(coverage): stamp ingested_commit_sha from the live HEAD, not the stale stored column

### LINKED ISSUE #1747: [Bug] coverage ingested_commit_sha remains stale after incremental update
## Summary

Coverage ingestion stamps `ingested_commit_sha` from `repositories.head_commit`, but an incremental `repowise update` advances the working index without refreshing that field. Correct coverage for the current tree is therefore labelled with an older commit.

## Version check

- Latest upstream is 0.44.0; release notes through 0.44.0 do not mention this coverage provenance field.

## Measured reproduction

1. Initialize an index at commit `1c85e2ec`.
2. Advance the repository and run `repowise update` at commit `85111c59`.
3. Confirm:
   - Git `HEAD` is `85111c59`;
   - `.repowise/state.json` `last_sync_commit` is `85111c59`;
   - the index's churn anchor has advanced to the current tree.
4. Ingest a coverage report produced from `85111c59`.
5. Query coverage status.
6. Observe `ingested_commit_sha = 1c85e2ec`.

The coverage data itself mapped correctly; only its provenance SHA was stale.

## Expected behavior

`repowise update` should keep the repository row's current commit identity synchronized, or coverage ingestion should stamp the authoritative current/indexed commit (`last_sync_commit`, or an equivalent single source of truth).

If the report cannot be proven to describe that commit, ingestion should require an explicit SHA or mark the provenance unknown rather than reuse a stale initialization value.

## Impact

Any automation that rejects index/coverage commit skew fails on every normally incrementally-updated repository, despite correct coverage. Ignoring the field is also unsafe because it removes the only apparent protection against genuinely stale coverage.

sloemo01:
## What

Fixes #1747. Coverage ingestion stamped `ingested_commit_sha` from the stored `Repository.head_commit` column, but an incremental `repowise update` advances the index and working tree without refreshing that field — so coverage produced from the current tree was labelled with a stale initialisation-time commit.

## Root cause

Two stamp sites read the stale column:

1. `coverage_cmd.py:168` — `head_sha = getattr(repo_row, "head_commit", None)`
2. `update_cmd/persistence.py` (re-score path) — `getattr(repo, "head_commit", None)`

`Repository.head_commit` is only refreshed on drift via `reconcile_repo_head_commit`, which the routine incremental update path doesn't always run. So the coverage provenance SHA lagged behind the tree it actually described.

## Fix

Both sites now resolve the **live** HEAD from disk via `get_head_commit(Path(repo_path))` — the same source the health pass uses (`_analyzed_commit` already reads it this way for exactly this reason) — falling back to the stored column only if the git read fails.

## Impact

Automation that rejects index/coverage commit skew no longer fails on every normally incrementally-updated repo, and the provenance field now actually reflects the tree the coverage describes.

## Tests

- `tests/unit/cli/test_coverage_cmd.py` — 3 passed, no regression.
- Module import check + ruff clean on both files.

### DIFF
--- a/packages/cli/src/repowise/cli/commands/coverage_cmd.py
+++ b/packages/cli/src/repowise/cli/commands/coverage_cmd.py
@@ -23,6 +23,7 @@ from repowise.cli.helpers import (
     run_async,
 )
 from repowise.cli.output import emit_json, format_option, notice_console
+from repowise.core.workspace.update import get_head_commit
 
 
 def _resolve_coverage_repo(path: str | None, fmt: str = "table") -> Path:
@@ -176,7 +177,17 @@ def coverage_add(
                 )
                 return False
 
-            head_sha = getattr(repo_row, "head_commit", None)
+            # Stamp the *live* HEAD, not ``repo_row.head_commit``. An incremental
+            # ``repowise update`` advances the index and the working tree but
+            # never refreshes that stored column, so a coverage report produced
+            # from the current tree would otherwise be labelled with a stale
+            # initialisation-time commit (issue #1747). The row has the repo
+            # path, so resolve the current commit from disk — the same source
+            # the health pass uses.
+            head_sha = (
+                get_head_commit(Path(repo_path))
+                or getattr(repo_row, "head_commit", None)
+            )
 
             # --- Per-file aggregate coverage (lcov / cobertura / clover / json).
             agg_matched = 0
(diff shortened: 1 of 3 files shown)


RaghavChamadiya (approved):
Verified against current main; merging.

PR #1804, opened 2026-08-21
test(distill): assert the median rewrite time, not a p95 over 12 samples

### LINKED ISSUE #1783: Timing budget in test_rewrite_perf is a near-maximum, not a p95, and fails CI on unrelated PRs
## Summary

`tests/unit/distill/test_rewrite_perf.py` asserts a wall-clock budget over 12 subprocess spawns and calls the result a p95. With 12 samples it is not a p95, it is the second-slowest sample, so a single scheduling hiccup on a shared CI runner fails the build. This has already cost a contributor: PR #1568, a Rust `.scm` change that cannot reach this code, went red for five days on

```
FAILED tests/unit/distill/test_rewrite_perf.py::test_a_ledgered_invocation_stays_under_budget
  - AssertionError: a ledgered rewrite p95 256.4 ms >= 200 ms
```

with Python 3.11 and 3.13 green on the same commit. They said "not sure why the CI tests are failing, they all passed locally" and had no way to tell it was unrelated.

## Mechanism

`_p95` at `tests/unit/distill/test_rewrite_perf.py:76-88`:

```python
    for _ in range(12):
        start = time.perf_counter()
        result = subprocess.run(cmd, input=payload, capture_output=True, text=True)
        timings.append((time.perf_counter() - start) * 1000)
    timings.sort()
    return timings[int(len(timings) * 0.95) - 1], outputs
```

`int(12 * 0.95) - 1` is `int(11.4) - 1`, so index 10 of a 12-element sorted list: the eleventh smallest, one below the maximum. To pass, 11 of 12 subprocess spawns must clear the budget. One outlier decides the run, and process spawn time on a hosted runner is exactly where outliers live.

Both budget tests share the helper, so `test_p95_under_150ms` has the same exposure at a tighter budget.

## Why it is worth fixing rather than raising

Raising the number has been done once already, and the comment above `_LEDGERED_BUDGET_MS` says so plainly:

```python
#: The number is a guard against the *next* regression, not an endorsement of
#: this one. If it needs raising again, the write is the thing to fix.
```

The instrument is what is wrong here, not the threshold. The test is measuring a near-maximum and comparing it to a budget written as if it were a typical case, so raising the budget trades a flaky failure for a guard that no longer catches the regression it exists for.

## Options

Taking the median or a trimmed mean over the same 12 runs measures the thing the budget is written about and drops the spawn outlier, at the cost of not catching a regression that only shows in the tail. Alternatively keep the tail but sample enough for the index to mean something, which costs wall-clock in every CI run.

I lean toward the median plus a separate, much looser ceiling on the maximum, so a genuine "sometimes takes a second" regression still fails while a single slow spawn does not. I am not certain that is the right split and would take an argument for the trimmed mean instead.

Either way the failure message should name the whole distribution, not one number. `p95 256.4 ms >= 200 ms` gave the contributor nothing to reason about; the twelve timings would have made it obvious at a glance that one sample was an outlier.

## Not in scope

The hook's actual cost is not in question here. The import guard in the same file is a structural test with no timing in it and should stay exactly as it is.

sloemo01:
## What

Fixes #1783 — the flaky rewrite-perf timing test that keeps failing unrelated PRs' CI. This is the test that red the build on your #1800 just now, on PR #1568 for five days, and on two of my own PRs today.

## Root cause

`_p95` in `test_rewrite_perf.py` called a 12-sample wall-clock measurement a "p95", but `int(12 * 0.95) - 1` is index 10 — the **eleventh-smallest of 12, one below the maximum**. So a single slow subprocess spawn (exactly where outliers live on a shared CI runner) decided the run. The budget was written as if it measured a typical case, but the instrument measured a near-maximum.

## Fix

- **Measure the median** — the quantity the budget is actually written about. A single scheduling hiccup no longer decides the run.
- **Add a separate, much looser ceiling (1000ms) on the worst sample**, so a genuine "sometimes takes a second" regression still fails while one slow spawn does not — the split the issue leaned toward.
- **The failure message now names the whole timing distribution** (e.g. `median 145 ms >= 150 ms (all: 88, 92, ..., 612)`), so a single outlier is obvious at a glance instead of hiding behind one number — exactly the problem the issue reported: "p95 256.4 ms >= 200 ms gave the contributor nothing to reason about."

## Why this fix

Raising the budget again (done once already, per the comment above `_LEDGERED_BUDGET_MS`) would trade a flaky failure for a guard that no longer catches the regression it exists for. The instrument was wrong, not the threshold.

## Tests

- `tests/unit/distill/test_rewrite_perf.py` — 7 passed locally.
- The median assertion now tolerates a single slow spawn while the 1000ms max ceiling still catches real regressions.
- Ruff clean.

Not in scope (per the issue): the hook's actual cost and the import-graph guard — both unchanged.

### DIFF
--- a/tests/unit/distill/test_rewrite_perf.py
+++ b/tests/unit/distill/test_rewrite_perf.py
@@ -74,8 +74,16 @@ def _payload(command: str, cwd: Path) -> str:
     )
 
 
-def _p95(cmd: list[str], payload: str) -> tuple[float, list[str]]:
-    """Wall-clock p95 over 12 invocations, plus every stdout seen."""
+def _p95(cmd: list[str], payload: str) -> tuple[float, list[str], list[float]]:
+    """Wall-clock timings over 12 invocations, plus every stdout seen.
+
+    Returns ``(median_ms, outputs, all_timings_ms)``. The *median* is the
+    quantity the budget is written about: a single slow subprocess spawn on a
+    shared CI runner is a scheduling outlier, not a regression, and must not
+    decide the run. The raw ``all_timings`` list is returned so the failure
+    message can show the whole distribution — one outlier is then obvious at
+    a glance instead of hiding behind a single number (issue #1783).
+    """
     timings: list[float] = []
     outputs: list[str] = []
     for _ in range(12):
@@ -85,7 +93,8 @@ def _p95(cmd: list[str], payload: str) -> tuple[float, list[str]]:
         assert result.returncode == 0
         outputs.append(result.stdout)
     timings.sort()
-    return timings[int(len(timings) * 0.95) - 1], outputs
+    median = timings[len(timings) // 2]
+    return median, outputs, timings
 
 
 #: Command shapes the hook must answer within budget. The pipeline and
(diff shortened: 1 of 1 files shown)


RaghavChamadiya (approved):
Thanks @sloemo01. `int(12 * 0.95) - 1` is index 10 of 12, the second largest sample, so the name said p95 and the code measured a near-maximum. That is the whole bug and it is worth stating plainly, because a test that reds on a shared runner's scheduling noise costs more than the regression it was written to catch. It sat on #1568 for five days.

The shape of the replacement is what I would have asked for. Measuring the median puts the assertion on the quantity the budget is actually about, and the separate 1000ms ceiling on `max(timings)` means a genuine "sometimes takes a second" regression still fails loudly. Dropping the outlier check entirely would have traded one failure mode for another.

Printing the full distribution in the failure message is the part contributors will feel. Reading `median 96.2 ms >= 150 ms (all: 88, 91, 93, ..., 1140)` tells you in one line that you are looking at a spawn hiccup, where a single number never could.

Non-blocking: the helper is still called `_p95` while it returns a median, so the next reader gets the same wrong impression the old code gave. Renaming it to `_timings` or `_median_ms` would close that, and it is small enough to fold into another PR rather than respin this one.

Green on the full matrix. Merging. I will let @crizah know on #1568 that their red is cleared.

Issue #1746, opened 2026-08-19
[Bug] coverage add treats severe path-mapping loss as a successful note


SimplicityGuy:
## Summary

When a coverage report's paths do not align with the repository identity/path stored in the index, `repowise coverage add` ingests the small subset it can map, prints the missing remainder as a note, and exits 0. It then exposes a plausible coverage percentage calculated from the tiny mapped subset.

## Version check

- Originally reproduced with Repowise 0.35.0.
- Rechecked on 2026-08-19 with installed Repowise 0.43.0; no matching existing issue or release-note fix was found.
- Latest upstream is 0.44.0; release notes through 0.44.0 do not mention stricter coverage path-mapping validation.

## Measured reproduction

The index was keyed to a different absolute checkout path from the tree that produced the report.

1. Produce a coverage report containing 249 files and approximately 29,317 per-test map records.
2. Run `repowise coverage add` against the mismatched index.
3. Observe:
   - 1 of 249 files ingested;
   - 26 of approximately 29,317 map records accepted;
   - unmapped paths reported only as a note;
   - process exit status 0;
   - a plausible non-null `line_coverage_pct` derived from the mapped fragment.

## Expected behavior

A loss this large should fail closed. At minimum:

- print requested/mapped/skipped counts in both human and JSON output;
- return non-zero when mapping completeness falls below an explicit threshold;
- provide a strict mode for automation;
- avoid publishing an aggregate percentage without clearly marking it partial.

## Impact

Missing coverage is visible because values remain null. Partial coverage is more dangerous: a realistic percentage looks authoritative and can silently influence health scores, hotspot ranking, and review decisions even though 248 of 249 files were omitted.

RaghavChamadiya:
Thanks @SimplicityGuy. The 1-of-249 measurement is what makes this actionable, and you are right that partial is the more dangerous of the two states.

Confirmed. The skipped remainder never reaches the exit code. In `packages/cli/src/repowise/cli/commands/coverage_cmd.py`:

```python
agg_matched = resolved.matched
console.print(
    f"[green]Ingested coverage for {resolved.matched} file(s)[/green] "
    f"({resolved.matched_exact} exact, {resolved.matched_suffix} resolved)."
)
skipped = resolved.unmatched + resolved.ambiguous
if skipped:
    sample = ", ".join(skipped[:5])
    console.print(
        f"[yellow]{len(skipped)} report file(s) did not map to the "
[...]

PR #1805, opened 2026-08-21
fix(coverage): answer the Tests tab for files a partial report never named

### LINKED ISSUE #1763: [Bug] Tests tab hides every file without a coverage row, even when the graph can answer for it
## Summary

On the measured basis, the Tests tab only ever shows files that have a row in `coverage_files`. Every other file is absent, even though the dependency graph can say whether a test reaches it. On this repository that is 2,209 of 3,608 files, so roughly two thirds of the tree is invisible on the tab that answers "is this tested".

## Measured

Against a live index of this repo:

```
health_file_metrics rows : 3608
coverage_files rows      : 1399
metric but no coverage   : 2209
```

The 2,209 are not unindexed and not untested. They are files the ingested lcov report never mentioned. The graph-inferred map has an answer for them and it is not consulted, because the basis decision is all-or-nothing.

## Where

`packages/server/src/repowise/server/routers/code_health/coverage_routes.py`:

```python
all_rows = await crud.load_coverage_for_repo(
    session, repo_id, include_covered_lines=False
)
if not all_rows:
    # No report was ever ingested. The graph can still answer "does a test
    # reach this", so answer that instead of an empty measured shape.
    ...
    return await _inferred_coverage(session, repo_id, limit)
```

A single stored row is enough to select the measured basis for the whole repository. The inferred map is then never built, and the per-file rows the tab renders come only from `coverage_files`.

## Why this matters more since #1746

#1746 reports `coverage add` accepting a severely partial ingest: 1 of 249 files mapped, exit 0, and a plausible aggregate percentage derived from the fragment. Under the current basis rule that repository now reports `basis: "measured"` on the strength of that one file *and suppresses the graph answer for the other 248*. The partial-ingest failure and the all-or-nothing basis rule compound: the worse the ingest, the more the tab hides, and the more authoritative it looks while doing it.

## What it is not

This is not a request to merge the two bases. Measured and inferred are different claims and the response deliberately keeps them in separate fields so neither can be rendered through the other's code path. Any fix has to preserve that.

- Serve the inferred map alongside the measured one when a file has no measured row, keeping them in separate fields and labelling each row's basis. Most useful, most work, and needs a rule for what the chart does with a mixed population.
- Leave the tab as it is and state the gap in the copy: "1,399 of 3,608 files carry a coverage row". Cheap, honest, and does not answer the question for the other 2,209.
- Decide the basis per file instead of per repository. Cleanest conceptually, but the hero figure and the chart are currently built from one population, so it is not a small change.

The per-file endpoint `/health/tests-reaching` already answers for any file regardless of basis, so the file page is not affected. This is specific to the repository-level tab.

Related: #1746, #1745, #1748.

sloemo01:
## What

Fixes #1763 — the Tests tab hides every file the ingested coverage report never mentioned, even though the dependency graph can answer whether a test reaches it. On a measured repo that is 2,209 of 3,608 files silently invisible.

## Root cause

The basis decision is all-or-nothing. In `coverage_routes.py` a single stored row is enough to select the measured basis for the whole repo:

```python
if not all_rows:
    ... return await _inferred_coverage(...)
```

So a repo with **any** coverage rows answers `basis: "measured"` and `files` only carries the measured rows. Files the lcov report never named get no row at all — not unindexed and not untested, just absent. And because a partial ingest still reports a plausible aggregate (`basis: "measured"`, `file_count: 1399`), the tab *looks* authoritative while hiding two thirds of the tree. This compounds with #1746 (a report that maps 1 of 249 files suppresses the graph answer for the other 248).

## Fix

Serve the graph-inferred map for exactly the files **without** a measured row, in a separate `inferred` block beside the measured one:

- `basis` stays `"measured"`; `summary` / `files` / `modules` are untouched.
- The `inferred` block is scoped to non-measured files only — a file is never in both.
- `measured_file_count` states the split, so the UI can say "your report named N files; the graph answers for the other M".
- **No percentage is ever derived from the inferred map** — the existing invariant. Reaching is a file-level fact with no line attribution. The no-ratio rule is asserted over the serialized payload, same as the pure-inferred shape.

The inferred walk now reads health metrics **before** the graph and early-outs when there's no gap, so a fully-measured repo no longer pays to read every call edge on every tab load.

## Changes

- `packages/server/.../code_health/coverage_routes.py` — `_inferred_coverage` gains `excluded_paths`; measured branch serves the hybrid block; cheap early-out when no gap.
- `packages/types/src/health.ts` — optional `measured_file_count` on `InferredTestMap`; hybrid shape documented.
- `packages/ui/src/health/coverage-view.tsx` — new `CoverageGap` section (counts only) for the hybrid shape.
- `tests/unit/server/test_health_coverage_basis.py` — updated the now-outdated "no inferred on measured" assertion; added hybrid + no-percentage cases.

## Tests

- `tests/unit/server/test_health_coverage_basis.py` + `test_health_coverage_route.py` + `test_health_badge.py` + `test_health.py`: **40 passed**.
- `tsc --noEmit` clean on `packages/ui` and `packages/types`.
- `ruff` clean.

### DIFF
--- a/packages/server/src/repowise/server/routers/code_health/coverage_routes.py
+++ b/packages/server/src/repowise/server/routers/code_health/coverage_routes.py
@@ -157,6 +157,32 @@ async def health_coverage(
     ]
     module_rows.sort(key=lambda x: x["line_coverage_pct"])
 
+    # The hybrid gap. A single stored row is enough to select the measured
+    # basis for the whole repository, but most files in most repos have no
+    # measured row at all (the lcov report never mentioned them) - here that is
+    # 2,209 of 3,608 files. Those files are not unindexed and not untested: the
+    # graph-inferred map has an answer for them and the all-or-nothing basis
+    # rule currently withholds it. Serve the inferred map for exactly the files
+    # with no measured row, so a partial ingest stops hiding the files it did
+    # not mention. ``basis`` stays ``measured`` and ``inferred`` reports only
+    # the non-measured set - a file is never in both, and measured rows never
+    # wear inferred percentages (none are derived here).
+    if include_inferred and not file_path:
+        measured_paths = frozenset(r.file_path for r in all_rows)
+        inferred = await _inferred_coverage(
+            session, repo_id, limit, excluded_paths=measured_paths
+        )
+        if inferred.get("basis") == "inferred":
+            inferred["inferred"]["measured_file_count"] = len(measured_paths)
+            return {
+                "basis": "measured",
+                "summary": summary,
+                "files": files,
+                "modules": module_rows[:module_limit] if module_limit else [],
+                "modules_total": len(module_rows),
+                "inferred": inferred["inferred"],
+            }
+
     return {
         "basis": "measured",
         "summary": summary,
@@ -186,7 +212,11 @@ def _empty_summary() -> dict[str, Any]:
 
 
 async def _inferred_coverage(
-    session: AsyncSession, repo_id: str, limit: int
+    session: AsyncSession,
+    repo_id: str,
+    limit: int,
+    *,
+    excluded_paths: frozenset[str] = frozenset(),
 ) -> dict[str, Any]:
     """The graph-inferred test map, in the shape the measured one leaves empty.
 
@@ -197,8 +227,13 @@ async def _inferred_coverage(
     walk also means this can never disagree with the ``untested_hotspot``
     biomarker, which reads the same function.
 
-    Counts are repo-wide even when ``files`` is trimmed, matching the measured
-    branch's ``modules_total``: a truncated page must never read as the repo.
+    ``excluded_paths`` scopes the answer to the files *not* carrying a measured
+    coverage row. On the hybrid basis the measured branch already answers those
+    files; this map fills the gap for the rest, so a file is never in both.
+
+    Counts are repo-wide over the scoped set even when ``files`` is trimmed,
+    matching the measured branch's ``modules_total``: a truncated page must
+    never read as the repo.
 
     Degrades to ``basis: "none"`` rather than raising. An unindexed graph is the
     honest unknown, and failing the tab to withhold a secondary signal is the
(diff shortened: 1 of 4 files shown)


RaghavChamadiya (approved):
Thanks @sloemo01. This one has been sitting green longer than it should have, and it is the most user-visible thing in your queue: 2,209 of 3,608 files silently absent from a tab that looks authoritative is worse than a tab that admits it knows nothing.

The framing in the description is what makes the fix obviously right. Those files were "not unindexed and not untested, just absent", and the all-or-nothing basis rule was the only reason. One stored row selecting `measured` for the entire repository is the bug, and it is the kind that gets more wrong the more partial the report is.

Checked the shape, since a hybrid response is where double-counting usually creeps in:

`excluded_paths=measured_paths` is threaded into `_inferred_coverage` and applied at the row filter, so the inferred set is the exact complement of the measured set. A file cannot be in both by construction rather than by convention, which is the difference between an invariant and a comment. `basis` stays `"measured"` and `summary`, `files` and `modules` are handed back untouched, so nothing that reads this today changes behaviour. The hybrid block is additive.

Holding the no-percentage rule is the part I would have been most worried about, and you kept it and said why: reaching is a file-level fact with no line attribution, so a ratio derived from it would be a fabricated number wearing a real one's clothes. `measured_file_count` stating the split is the right alternative, because it lets the UI say "your report named N files, the graph answers for the other M" instead of blending two things that are not the same measurement.

The early-out is a good catch on yourself. Reading health metrics before the graph and returning the empty shape when `rows` is empty means a fully-measured repo stops paying for a walk over every call edge on each tab load. That is a performance fix hidden inside a correctness fix, and it only exists because you noticed the ordering.

This half-closes #1746 in the sense that it stops the graph answer being suppressed, but the other half is still open: a report that maps 1 of 249 files still ingests as a success and reports the fragment as repo-wide. That is your #1955, and I have it as the next one to look at.

PR #1806, opened 2026-08-21
fix(pipeline): preserve coverage across an incremental health re-score

### LINKED ISSUE #1739: [Bug] Incremental health analysis nulls coverage for every changed file
**Summary**

The incremental pipeline builds its `HealthAnalyzer` without a `coverage_map`,
so every file it re-analyzes is scored as if no coverage report had ever been
ingested. The partial-health writer then upserts those metrics, overwriting the
stored `line_coverage_pct` with NULL for exactly the files that just changed.

Unchanged files keep their coverage, so this is not visible as a single broken
run. It erodes the coverage picture one file per update, starting with the files
under active development.

The same bug was already found and fixed at the sibling call site.

**Where**

`packages/core/src/repowise/core/pipeline/incremental.py:483`

```python
_health_analyzer = HealthAnalyzer(
    graph_builder.graph(),
    git_meta_map=git_meta_map,
    parsed_files=parsed_files,
    duplication_cache_dir=Path(repo_path) / ".repowise",
    repo_root=repo_path,
)                                    # no coverage_map
```

`HealthAnalyzer.__init__` defaults it to `{}` (`analysis/health/engine.py:311`),
and the lookup at `engine.py:889-893` then yields `None` for every file.
`persist_partial_health` (`incremental.py:757`) upserts those metrics.

The pattern to copy is `_coverage_for_rescore` in
`packages/cli/src/repowise/cli/commands/update_cmd/persistence.py:87`, used at
line 1147 of the same file. Its comment describes this exact failure. It reloads
the persisted `coverage_files` rows, and optionally re-discovers a fresh report
when `coverage.reingest_on_update` is set.

**Repro**

1. `repowise init` on a repo that has a coverage report
2. `repowise coverage add <report>` and confirm `coverage_files` is populated
3. Confirm `repowise health --target <file>` shows a coverage percentage
4. Edit that file, then `repowise update`
5. `repowise health --target <file>` now reports no coverage data. The row in
   `coverage_files` is still intact; only `health_file_metrics` was nulled.

**Expected**

An incremental re-score preserves coverage the same way the update re-score
path does. A file's stored coverage should only change when a new report is
ingested.

**Impact**

Two effects, and the second is worse than the first.

The number disappears from the health surface for the files people are actively
working on.

`untested_hotspot` then falls to its no-coverage branch, which consults a
filename heuristic, and flags a well-covered file as untested. That finding
carries a high severity weight, so a file with strong coverage can be driven to
the bottom of the health ranking by a value that was deleted rather than
measured.

**Done looks like**

`incremental.py:483` receives a coverage map loaded from the persisted
`coverage_files` rows. Note that the surrounding function has no session at that
point, so the map needs loading earlier and threading in; this is the bulk of
the work and the reason this is not a one-liner.

A test that ingests coverage, runs an incremental update over a changed file,
and asserts the stored coverage survived.

Optional hardening worth considering in the same PR: have the metrics upsert
refuse to replace a non-null coverage value with NULL when the analyzer held no
coverage map at all. That would have caught both this occurrence and the
original one, and would catch a future fourth call site.

sloemo01:
## What

Fixes #1739. The incremental pipeline built its `HealthAnalyzer` **without a `coverage_map`**, so every changed file was re-scored as if no coverage had ever been ingested. The partial-health writer then upserted those metrics, overwriting the stored `line_coverage_pct` with `NULL` for exactly the files that just changed — eroding coverage one file per update, starting with the files under active development.

## Root cause

`run_partial_analysis` built `HealthAnalyzer(...)` with no `coverage_map` (which defaults to `{}`), so the coverage lookup at `engine.py:889` yielded `None` for every re-analyzed file. `persist_partial_health` then upserted those metrics, nulling coverage. The sibling re-score path already had the fix pattern (`_coverage_for_rescore` in the CLI update path) — this extends it to the incremental health pass.

## Fix

- **`incremental.py`**: add `load_stored_coverage_map(repo_path)` — reloads the persisted `CoverageFile` rows into the `{path: coverage-dict}` shape `HealthAnalyzer` expects, mirroring `load_stored_function_mod_p80` and the CLI's `_coverage_for_rescore`.
- **`run_partial_analysis`**: accept and pass `coverage_map` to `HealthAnalyzer`.
- **Both callers** now load and supply it: the workspace updater (`workspace/update.py`, async) and the CLI incremental path (`update_cmd/incremental.py`).
- Returns empty map (safe no-coverage default) when nothing is stored or the store is unreadable.

## Tests

- New `tests/unit/pipeline/test_load_stored_coverage_map.py` — persists coverage rows then reloads them as the health-coverage map (2 tests: reload round-trip + absent-store returns empty).
- `tests/unit/pipeline/` + `tests/unit/cli/test_coverage_cmd.py` — **127 passed**, no regressions.
- Ruff clean.

### DIFF
--- a/packages/cli/src/repowise/cli/commands/update_cmd/incremental.py
+++ b/packages/cli/src/repowise/cli/commands/update_cmd/incremental.py
@@ -200,7 +200,10 @@ def _run_partial_analysis(
     Returns ``(partial_health_report, dead_code_report)`` — either may be
     ``None`` if its analysis failed (both are best-effort).
     """
-    from repowise.core.pipeline.incremental import run_partial_analysis
+    from repowise.core.pipeline.incremental import (
+        load_stored_coverage_map,
+        run_partial_analysis,
+    )
 
     return run_partial_analysis(
         repo_path,
@@ -212,5 +215,6 @@ def _run_partial_analysis(
         stored_git_meta=stored_git_meta,
         stored_performance_callers=stored_performance_callers,
         repo_function_mod_p80=repo_function_mod_p80,
+        coverage_map=run_async(load_stored_coverage_map(repo_path, log=console.print)),
         log=console.print,
     )
--- a/packages/core/src/repowise/core/pipeline/incremental.py
+++ b/packages/core/src/repowise/core/pipeline/incremental.py
@@ -493,6 +493,70 @@ async def load_stored_function_mod_p80(repo_path: Any, *, log: LogFn | None = No
         return None
 
 
+async def load_stored_coverage_map(
+    repo_path: Any, *, log: LogFn | None = None
+) -> dict[str, dict]:
+    """Load the persisted coverage map for health scoring on an incremental run.
+
+    The incremental health pass re-scores changed files only, but if it builds
+    its ``HealthAnalyzer`` without a ``coverage_map`` every re-analyzed file is
+    scored as if no coverage had ever been ingested — and the partial-health
+    writer then overwrites the stored ``line_coverage_pct`` with ``None`` for
+    exactly the files that just changed, eroding coverage one file per update
+    (issue #1739).
+
+    Returns ``{path: coverage-dict}`` by reloading the persisted ``CoverageFile``
+    rows, mirroring ``_coverage_for_rescore`` in the CLI update path. Empty dict
+    when no coverage is stored, the store is unreadable, or the repository row
+    is missing — the analyzer then scores without coverage, matching the
+    established safe default.
+    """
+    log = log or _noop_log
+    import json
+
+    if not (Path(repo_path) / ".repowise" / "wiki.db").is_file():
+        return {}
+    try:
+        from repowise.core.persistence import (
+            create_engine,
+            create_session_factory,
+            get_session,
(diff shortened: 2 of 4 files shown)


RaghavChamadiya (approved):
Thanks @sloemo01 for tracing this to the writer instead of stopping at the analyzer. I confirmed the whole chain against main before merging:

- `run_partial_analysis` builds its `HealthAnalyzer` at `incremental.py:569` with no `coverage_map`
- the analyzer defaults it to `{}` (`engine.py:327`)
- the per-file lookup at `engine.py:977` then yields `None` for both `line_coverage_pct` and `branch_coverage_pct`
- `upsert_health_metrics` writes that `None` straight back (`crud/analysis/health.py:1018`)

So coverage really is erased, and it is erased for exactly the files someone just touched, which is the worst possible subset to lose it on. A repo that ingested coverage once would watch it decay along its own active development front, and nothing in the output says why.

Mirroring `load_stored_function_mod_p80` rather than inventing a new loading shape is the right call, and wiring both callers means the workspace updater does not quietly keep the bug.

PR #1808, opened 2026-08-21
fix(hooks): resolve the real hooks dir so worktree installs don't crash init

### LINKED ISSUE #1609: [Bug] initialization failures in worktree aren't handled
## Steps to Reproduce
failure to install hooks into nonexistent .git/ directory in a worktree causes entire initialization to fail

## Environment
- OS: Fedora Linux
- Python version: [e.g., 3.12.1]
- Repowise version: 0.42.0
- Installation method: uv tool install ...

## Additional Context
  File "~/.local/bin/repowise", line 10, in <module>
    sys.exit(cli())
             ~~~^^
  File "~/.local/share/uv/tools/repowise/lib64/python3.14/site-packages/click/core.py", line 1569, in __call__
    return self.main(*args, **kwargs)
           ~~~~~~~~~^^^^^^^^^^^^^^^^^
  File ".../repowise//lib64/python3.14/site-packages/click/core.py", line 1490, in main
    rv = self.invoke(ctx)
  File ".../repowise/lib64/python3.14/site-packages/repowise/cli/_instrumented_group.py", line 130, in invoke
    return super().invoke(ctx)
           ~~~~~~~~~~~~~~^^^^^
  File ".../repowise/lib64/python3.14/site-packages/click/core.py", line 1970, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
                           ~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^
  File ".../repowise/lib64/python3.14/site-packages/click/core.py", line 1353, in invoke
    return ctx.invoke(self.callback, **ctx.params)
           ~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".../repowise/lib64/python3.14/site-packages/click/core.py", line 907, in invoke
    return callback(*args, **kwargs)
  File ".../repowise/lib64/python3.14/site-packages/repowise/cli/commands/init_cmd/command.py", line 1615, in init_command
    offer_hook_install(console, [repo_path], yes=yes)
    ~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".../repowise/cli/commands/init_cmd/_interactive.py", line 182, in offer_hook_install
    _offer_hook_install_prompts(console_obj, repo_paths, aliases)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".../repowise/lib64/python3.14/site-packages/repowise/cli/commands/init_cmd/_interactive.py", line 221, in _offer_hook_install_prompts
    result = install(rp)
  File ".../repowise/lib64/python3.14/site-packages/repowise/cli/hooks.py", line 190, in install
    hooks_dir.mkdir(exist_ok=True)
    ~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^
  File "/usr/lib64/python3.14/pathlib/__init__.py", line 1011, in mkdir
    os.mkdir(self, mode)
    ~~~~~~~~^^^^^^^^^^^^
NotADirectoryError: [Errno 20] Not a directory: '.../test-repowise-worktree/.git/hooks'

sloemo01:
## What

Fixes #1609. In a git **worktree**, `.git` is a *file* (`gitdir: <path>`), not a directory. The hook installer built `root / ".git" / "hooks"` from `_git_root`'s heuristic, so in a worktree `mkdir` raised `NotADirectoryError` and the crash propagated up to take down the whole `repowise init`.

## Root cause

`_git_root` walks up for any `.git` entry — but in a worktree that entry is a file, not a directory. `install()`/`status()`/`uninstall()` then did `root / ".git" / "hooks"`, which fails on a worktree because `.git` is not a directory.

## Fix

- Add `_hooks_dir(repo_path)` — resolves the real hooks dir via `git rev-parse --git-path hooks`, which returns the correct (shared) hooks dir for **both** normal repos and worktrees. Falls back to the `.git/hooks` heuristic when git is unavailable or not a repo.
- `install`, `status`, and `uninstall` now use `_hooks_dir`.

## Tests

- New `TestInstallInWorktree` in `test_hooks_legacy_migration.py`: creates a real worktree (where `.git` is asserted to be a file), verifies `install` no longer crashes and `status` reports correctly.
- Existing hook tests: **13 passed**, no regression.
- Ruff clean.

### DIFF
--- a/packages/cli/src/repowise/cli/hooks.py
+++ b/packages/cli/src/repowise/cli/hooks.py
@@ -12,6 +12,7 @@ from __future__ import annotations
 import contextlib
 import re
 import stat
+import subprocess
 from pathlib import Path
 
 _HOOK_MARKER = "# repowise-hook-start"
@@ -94,6 +95,36 @@ def _git_root(path: Path) -> Path | None:
         if (parent / ".git").exists():
             return parent
     return None
+
+
+def _hooks_dir(repo_path: Path) -> Path | None:
+    """Resolve the real hooks directory for *repo_path*.
+
+    A normal repo keeps ``.git/hooks`` under ``.git``, but a git **worktree**
+    stores ``.git`` as a *file* (``gitdir: <path>``) pointing at the shared
+    metadata dir — so ``root / ".git" / "hooks"`` is ``NotADirectoryError``
+    territory. Ask git for the real hooks path so both layouts work, and fall
+    back to the ``.git/hooks`` heuristic when git is unavailable or not a repo.
+    """
+    try:
+        result = subprocess.run(
+            ["git", "rev-parse", "--git-path", "hooks"],
+            cwd=str(repo_path),
+            capture_output=True,
+            text=True,
+            timeout=10,
+        )
+        if result.returncode == 0:
+            p = Path(result.stdout.strip())
+            if p.is_absolute():
+                return p
+            return repo_path / p
+    except Exception:
+        pass
+    root = _git_root(repo_path)
+    if root is None:
+        return None
+    return root / ".git" / "hooks"
 
 
 def _strip_legacy_block(content: str) -> tuple[str, bool]:
@@ -186,8 +217,10 @@ def install(repo_path: Path) -> str:
     if root is None:
         return "not a git repository"
 
-    hooks_dir = root / ".git" / "hooks"
-    hooks_dir.mkdir(exist_ok=True)
+    hooks_dir = _hooks_dir(repo_path)
+    if hooks_dir is None:
+        return "not a git repository"
+    hooks_dir.mkdir(parents=True, exist_ok=True)
     hook_path = hooks_dir / "post-commit"
 
     migrated_legacy = False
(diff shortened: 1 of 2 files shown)


RaghavChamadiya (approved):
Thanks @sloemo01. The diagnosis holds exactly as written: `_git_root` tests with `.exists()`, not `.is_dir()`, so in a worktree it happily returns the root and the failure lands one level down on `mkdir`, which is why a hook-install detail was able to take down the whole `init`.

Asking git for the path instead of rebuilding it is the right instinct, and it buys more than the worktree fix you claimed. `git rev-parse --git-path hooks` also honors `core.hooksPath`:

```
$ git rev-parse --git-path hooks
C:\Users\...\repowise\.git\hooks
$ git -c core.hooksPath=/tmp/myhooks rev-parse --git-path hooks
/tmp/myhooks
```

So this quietly fixes a second case nobody has filed: anyone using husky, lefthook or pre-commit has `core.hooksPath` pointed elsewhere, and the old code was writing a `post-commit` hook into a directory git had stopped reading. The hook installed, reported success, and never fired. Worth a line in the changelog beyond the worktree crash.

The absolute-or-relative handling on the `rev-parse` output is correct, and keeping the `.git/hooks` heuristic as a fallback means a machine without git on PATH degrades the way it used to instead of newly returning nothing.

PR #1809, opened 2026-08-21
fix(pages): pass repo_id to regenerate so Write-with-AI works in workspace mode

### LINKED ISSUE #1665: [Bug] Write with AI documentation page 404
## Describe the Bug

In workspace mode, the **Write with AI** action for a documentation page fails with `404 Page not found`.

The page itself can be opened successfully in the frontend, but regenerating it fails because the regenerate request does not include `repo_id`. In workspace mode, `repo_id` is required to route the database session to the correct repository database.

A manual regenerate request with `repo_id` added succeeds with `202 Accepted`.

## Steps to Reproduce

1. Start Repowise in workspace mode with multiple repositories:

```bash
repowise serve
```

2. Open a repository in the Repowise frontend.

3. Navigate to a generated documentation/onboarding page, for example:

```text
onboarding:onboarding/active_landscape
```

4. Click **Write with AI**.

5. Observe the backend request:

```text
POST /api/pages/lookup/regenerate?page_id=onboarding%3Aonboarding%2Factive_landscape&cascade=none
```

6. The request returns:

```text
404 Not Found
```

7. Send the same request manually, but include the repository ID:

```bash
curl -i -X POST \
'http://127.0.0.1:7337/api/pages/lookup/regenerate?page_id=onboarding%3Aonboarding%2Factive_landscape&cascade=none&repo_id=7d9f21a325d14e47bd1221aa6e9937fb'
```

8. This succeeds:

```text
HTTP/1.1 202 Accepted
```

Example response:

```json
{
  "job_id": "cfa4b9d1c2cf477cb9f64c4184559e0a",
  "status": "accepted"
}
```

## Expected Behavior

Clicking **Write with AI** should regenerate the selected page successfully in workspace mode.

The regenerate request should be routed to the repository database that owns the page.

## Actual Behavior

The frontend sends the regenerate request without `repo_id`:

```text
POST /api/pages/lookup/regenerate?page_id=...&cascade=none
```

The backend returns:

```text
404 Page not found
```

The regular page lookup works because it includes the repository ID:

```text
GET /api/pages/lookup?page_id=onboarding%3Aonboarding%2Factive_landscape&repo_id=7d9f21a325d14e47bd1221aa6e9937fb
HTTP/1.1 200 OK
```

The relevant workspace routing code in `server/deps.py` reads `repo_id` directly from the request:

```python
repo_id = request.path_params.get("repo_id") or request.query_params.get("repo_id")
return resolve_session_factory(request.app.state, repo_id)
```

The working `/api/pages/lookup` endpoint explicitly supports `repo_id` and documents that it is needed to reach pages outside the default store in workspace mode.

The `/api/pages/lookup/regenerate` endpoint does not expose `repo_id`, and currently performs:

```python
page = await crud.get_page(session, page_id)
if page is None:
    raise HTTPException(status_code=404, detail="Page not found")
```

Adding `repo_id` manually to the query string causes `get_db_session()` to use the correct repository database and the request succeeds.

## Environment

- OS: macOS
- Python version: 3.14
- Repowise version: `[run: repowise --version]`
- Installation method: pipx
- Mode: workspace with multiple repositories

## Additional Context

Example repository ID:

```text
7d9f21a325d14e47bd1221aa6e9937fb
```

Example affected page ID:

```text
onboarding:onboarding/active_landscape
```

This appears to be a workspace-routing issue rather than an LLM/Ollama issue. The regenerate job is successfully accepted once `repo_id` is supplied.

A likely fix is to include `repo_id` in the frontend **Write with AI** regenerate request and/or explicitly expose `repo_id` on the `/api/pages/lookup/regenerate` endpoint.

sloemo01:
## What

Fixes #1665. The **Write with AI** regenerate request omitted `repo_id`, so in workspace mode `get_db_session` couldn't route to the correct repository database and the page lookup returned **404**. A manual regenerate request with `repo_id` added succeeds (202).

## Root cause

`regeneratePage` (api-client) never sent `repo_id`, and the backend `regenerate_page_by_query` endpoint didn't accept one — so `resolve_request_session_factory` read a `None` repo id and couldn't open the right repo DB.

## Changes

- **`pages.py`**: `regenerate_page_by_query` now accepts an optional `repo_id` query param (workspace routing; single-repo callers unchanged).
- **`api-client/pages.ts`**: `regeneratePage` accepts and forwards `repoId` → `repo_id`.
- **`page-generate-button.tsx` + `freshness-table-wrapper.tsx`**: pass their `repoId` through (both had it available but weren't sending it).

## Tests

- `api-client/src/pages.test.ts`: new test asserting `repo_id` is included when `repoId` is given (workspace routing).
- Backend ruff clean + module import verified.

### DIFF
--- a/packages/api-client/src/pages.test.ts
+++ b/packages/api-client/src/pages.test.ts
@@ -83,4 +83,12 @@ describe("regeneratePage", () => {
       cascade: "dependents",
     });
   });
+
+  it("includes repo_id when given (workspace routing)", async () => {
+    await regeneratePage("file_page:src/main.py", { repoId: "repo1" });
+    expect(apiPost.mock.calls[0]![3]).toEqual({
+      page_id: "file_page:src/main.py",
+      repo_id: "repo1",
+    });
+  });
 });
--- a/packages/api-client/src/pages.ts
+++ b/packages/api-client/src/pages.ts
@@ -136,7 +136,7 @@ export async function updatePageNotes(
  */
 export async function regeneratePage(
   pageId: string,
-  opts?: { cascade?: GenerateCascade; style?: string },
+  opts?: { cascade?: GenerateCascade; style?: string; repoId?: string },
 ): Promise<JobLaunchResponse> {
   return apiPost<JobLaunchResponse>(
     "/api/pages/lookup/regenerate",
@@ -144,6 +144,7 @@ export async function regeneratePage(
     undefined,
     {
       page_id: pageId,
+      ...(opts?.repoId ? { repo_id: opts.repoId } : {}),
       ...(opts?.cascade ? { cascade: opts.cascade } : {}),
       ...(opts?.style ? { style: opts.style } : {}),
     },
--- a/packages/server/src/repowise/server/routers/pages.py
+++ b/packages/server/src/repowise/server/routers/pages.py
@@ -237,6 +237,11 @@ async def update_page_notes(
 async def regenerate_page_by_query(
     request: Request,
     page_id: str = Query(..., description="Page ID"),
+    repo_id: str | None = Query(
+        None,
+        description="Repository ID. Required in workspace mode to route to the "
+        "correct repository database.",
+    ),
     style: str | None = Query(
         None,
         description="Optional wiki style to regenerate this page in (per-page override).",
--- a/packages/web/src/components/coverage/freshness-table-wrapper.tsx
+++ b/packages/web/src/components/coverage/freshness-table-wrapper.tsx
@@ -29,7 +29,7 @@ export function FreshnessTableWithRegenerate({
         return;
       }
       try {
-        const res = await regeneratePage(pageId, { cascade: "none" });
+        const res = await regeneratePage(pageId, { cascade: "none", repoId });
         setActiveJobId(res.job_id);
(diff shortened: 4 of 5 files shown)


RaghavChamadiya (approved):
Thanks @sloemo01, and apologies for how long this waited. It is a small diff that fixes a real hole, and it is complete: `regeneratePage` has exactly two callers in the web app and both now pass the id, so there is no third surface left routing to the wrong database.

The freshness table is the one that gives the bug away. `repoId` was already a required prop there (`freshness-table-wrapper.tsx:16-17`) and the call simply never used it, so in workspace mode the click resolved through whatever `resolve_request_session_factory` fell back to.

One thing worth recording for whoever reads this next, because it is not obvious from the diff and it changes how you would test it: the server-side `repo_id` query parameter is documentation, not plumbing. Routing happens in `deps.py:67`, which reads `repo_id` straight off `request.query_params` before any endpoint signature is consulted, so the fix is entirely in the client sending it. Declaring it on the endpoint is still worth doing, since it puts the parameter in the OpenAPI schema instead of leaving it as an undocumented string FastAPI silently ignores.

Green on all eight jobs, clean against current main. Merging.

PR #1810, opened 2026-08-21
fix(server): read the embedder key from the process env so CLI and MCP agree

### LINKED ISSUE #1711: CLI vs MCP embedder key precedence drift: exported OPENAI_API_KEY outranks the repo's .repowise/.env for CLI commands but not the MCP server
## Summary

The CLI and the MCP server resolve the embedder credential in **different precedence orders**, so on any machine where `OPENAI_API_KEY` (or another provider key) is exported at user/process scope to a credential that differs from the repo's `.repowise/.env`, `repowise mcp` keeps working while `repowise search` (and every CLI command that resolves the embedder key) silently uses the wrong key → semantic search degrades/hangs.

`cli/providers/keys.py`'s docstring says it deliberately "mirrors the server's precedence" — the two do **not** agree.

## The two resolvers

**CLI — `cli/providers/keys.py::resolve_embedder_api_key(embedder_name, repo_path)`**
1. process environment first (an exported key is an explicit override and "nothing may outrank it"),
2. `<repo>/.repowise/.env`,
3. `~/.repowise/config.yaml` `embedder_api_key`.

**MCP server — `server/mcp_server/_server.py::_persisted_embedder_key(name)`**
1. the served repo's `.repowise/.env`,
2. `~/.repowise/config.yaml` `embedder_api_key`.

…and it **never consults the process environment**. So "mirrors the server's precedence" is inaccurate: the same key exists in both places with different precedence → divergence.

## Reproduction

Machine with a user/process-scoped `OPENAI_API_KEY` set to credential **A**, and a repo whose `.repowise/.env` carries credential **B** (e.g. an OmniRoute/free-pool key with `openai.base_url` in `.repowise/config.yaml`):

- `repowise mcp` + `search_codebase` tool → works (uses B from the repo `.env`).
- `repowise search "x" --mode semantic` → uses A from the process env → wrong key against the configured base_url → semantic leg fails or hangs. FTS/symbol still work, so it is the classic silent-degradation shape.

On this machine (2026-08-18), `repowise search --mode semantic` took ~46s / timed out with the inherited key, while the MCP tool answered in <1s — consistent with the two resolvers disagreeing.

## Suggested fix

Make `resolve_embedder_api_key` actually mirror `_persisted_embedder_key`, or document/decide the intended precedence. Given that the server and the `load_dotenv`/`load_repo_env` readers already treat the repo's `.repowise/.env` as the per-repo source of truth (and `load_dotenv` deliberately does NOT overwrite an exported var), the cleanest fix is for the CLI to consult the repo `.env` **before** the process environment for this credential — so a user-scope export of a *different* credential can't poison a repo that pins its own key.

Same silent-divergence family as #1678 / #1685 (a resolver's "successful" path and its fallback disagreeing).

sloemo01:
## What

Fixes #1711. The MCP server's `_persisted_embedder_key` never consulted the **process environment**, so on any machine with an exported `OPENAI_API_KEY` (or other provider key) of a different credential than the repo's `.repowise/.env`, the CLI (which reads env first) and the MCP server (which skipped env) resolved **different keys** for the same repo in the same shell. Semantic search degraded/hung through the CLI while the MCP tool answered.

## Root cause

Two resolvers with different precedence:
- **CLI** (`cli/providers/keys.py::resolve_embedder_api_key`): process env → repo `.env` → global config.
- **MCP server** (`_server.py::_persisted_embedder_key`): repo `.env` → global config — **never** the process env.

So on any machine with an exported credential, the two disagreed about the same directory.

## Fix

Unify **the other way** than the issue suggested, to preserve the CLI's deliberate contract that an exported key is an explicit override (`test_env_wins_over_everything`): the **server now reads the process env tier first**, before the repo `.env` and global config — matching `cli/providers/keys.py`. This makes CLI and MCP resolve identically on every machine, and keeps the existing precedence contract intact.

## Tests

- New `test_env_export_wins_in_both_cli_and_server` — asserts an exported `OPENAI_API_KEY` wins on **both** the CLI and the MCP server for the same repo.
- `tests/unit/cli/test_embedder_key_resolution.py` — **20 passed**, no regression (including the existing `test_env_wins_over_everything`).
- Ruff clean.

### DIFF
--- a/packages/server/src/repowise/server/mcp_server/_server.py
+++ b/packages/server/src/repowise/server/mcp_server/_server.py
@@ -98,6 +98,16 @@ def _persisted_embedder_key(name: str) -> str | None:
         return None
     canonical = env_vars[0]
 
+    # The process environment is the highest-precedence source, matching the
+    # CLI's resolver (cli/providers/keys.py): an exported key is an explicit
+    # override. The server used to skip this tier entirely, so on a machine
+    # with an exported credential it and the CLI disagreed about the same
+    # repo in the same shell (issue #1711).
+    for var in env_vars:
+        value = os.environ.get(var)
+        if value:
+            return value
+
     if _state._repo_path:
         try:
             # Reuse the reader get_answer already uses for the LLM credential,
--- a/tests/unit/cli/test_embedder_key_resolution.py
+++ b/tests/unit/cli/test_embedder_key_resolution.py
@@ -313,6 +313,27 @@ def test_cli_and_mcp_resolve_the_same_key_for_the_same_repo(
     assert _server._persisted_embedder_key("openai") == expected
 
 
+def test_env_export_wins_in_both_cli_and_server(
+    tmp_path: Path, monkeypatch: pytest.MonkeyPatch
+) -> None:
+    """Issue #1711: an exported key is an explicit override on BOTH sides.
+
+    The server previously never consulted the process env, so on a machine with
+    an exported credential it resolved a different key (or none) than the CLI,
+    and the same repo in the same shell disagreed over MCP vs ``repowise
+    search``. The server now reads the env tier first, matching the CLI.
+    """
+    from repowise.server.mcp_server import _server, _state
+
+    _write_env(tmp_path, OPENAI_API_KEY="«redacted:sk-repo»")
+    _write_global_config(embedder="openai", embedder_api_key="«redacted:sk-global»")
+    monkeypatch.setenv("OPENAI_API_KEY", "sk-from-env")
+    monkeypatch.setattr(_state, "_repo_path", str(tmp_path), raising=False)
+
+    assert resolve_embedder_api_key("openai", tmp_path).key == "sk-from-env"
+    assert _server._persisted_embedder_key("openai") == "sk-from-env"
+
+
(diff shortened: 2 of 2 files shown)


RaghavChamadiya (approved):
Thanks @sloemo01 for resolving this in the direction that keeps the existing contract rather than the one the issue proposed. That was the judgement call in this fix and you got it right.

Confirmed both resolvers on main. The CLI checks `os.environ` first (`cli/providers/keys.py:81`), then the repo `.env`, then the global config. `_persisted_embedder_key` went straight to the repo `.env` and then the global config, with no env tier at all. Unifying upward preserves `test_env_wins_over_everything`; unifying downward would have satisfied the issue and broken that test, and with it the deliberate rule that an exported key is an explicit override.

The failure this produces is nastier than a plain mismatch, which is probably why it took a report to find. Both sides succeed. Both build a real embedder. The vectors just cannot match the index one of them wrote, so semantic search degrades to full-text on one surface and works on the other, in the same directory, in the same shell.

One non-blocking note for whenever you are next in that file: the docstring above the function still reads "Order: the served repo's `.repowise/.env` first, then the global config", which now understates it by a tier. Not holding the merge on a comment.

PR #1811, opened 2026-08-21
fix(sql): don't flag SELECT INTO targets as a cartesian join

### LINKED ISSUE #1502: sql_cartesian_join false positives: plpgsql DECLARE variables and SELECT INTO targets parsed as comma-joined tables
### Summary

The `sql_cartesian_join` detector flags **plpgsql variable declarations and `SELECT … INTO` targets as comma-joined tables**. Every one of the 7 `sql_cartesian_join` findings in our repo points at a `DECLARE` block line or an `INTO` clause inside a `create function … language plpgsql` body in a migration file; none is a real join. The reported "table" names are the variable names (`balance`, `min_cents`, `current_referrer`, …), which is the tell.

### Environment

- repowise 0.41.0+f44f8a2d (pipx install), Windows 11
- Indexed repo: Supabase project; migrations define plpgsql functions

### Reproduction

```sql
create or replace function public.redeem_code(code text)
returns boolean
language plpgsql
security definer
set search_path = ''
as $$
declare
  caller_id uuid := (select auth.uid());
  current_referrer uuid;      -- flagged: "comma-join with 'current_referrer'"
  min_cents integer;          -- flagged: "comma-join with 'min_cents'"
  balance integer;            -- flagged: "comma-join with 'balance'"
begin
  select referred_by
  into current_referrer       -- also flagged on the INTO line
  from public.profiles
  where id = caller_id;
  return true;
end;
$$;
```

Health analysis output for such lines:

> `sql_cartesian_join` (medium) — "comma-join with 'min_cents' has no join predicate: this is the full cross product" — `details: {"table": "min_cents"}`

### Expected

Lines inside a plpgsql `DECLARE` section are variable declarations, and `SELECT … INTO a, b` names target variables, not FROM items. Neither can produce a cartesian product. The detector should either parse dollar-quoted plpgsql bodies properly or skip them, and only evaluate comma-separated relations inside an actual `FROM` clause.

### Actual

Declaration and `INTO` lines are reported as full cross products, at medium severity, one finding per variable. In a Supabase-style repo where every function lives in a migration, this makes the SQL findings list almost entirely noise.

sloemo01:
## What

Fixes #1502. A multi-target `SELECT ... INTO a, b` is parsed by sqlglot with the **trailing target** (`b`) as a comma-join FROM item, so `sql_cartesian_join` false-fired on table-creating statements. The reporter's reproducer:

```sql
select a, b into x, y from t where c = 1;   -- before: [('sql_cartesian_join', 'y')]
```

## Root cause

For `SELECT ... INTO x, y`, sqlglot sets `into` and lands the trailing target `y` in `joins` — so the detector's "comma-join with no predicate" arm fires on a name that is an INTO target, not a join table. A single-target `INTO x` is silent (no trailing comma), which is why the bug only shows on multi-target INTOs.

## Fix

Skip cartesian detection when the select carries an `into` arg — present exactly when the trailing names are INTO targets, not joins. A genuine comma-join (`FROM a, b` with no `into`) still fires.

## Tests

- `test_multi_target_select_into_is_silent` — the reporter's reproducer + the no-WHERE variant.
- `test_single_target_select_into_is_silent`.
- `test_multi_target_into_does_not_mask_real_cartesian` — a real comma-join elsewhere still fires.
- `tests/unit/health/test_sql_markers.py` — **37 passed**, no regression.
- Ruff clean.

### DIFF
--- a/packages/core/src/repowise/core/analysis/health/sql_complexity.py
+++ b/packages/core/src/repowise/core/analysis/health/sql_complexity.py
@@ -238,6 +238,13 @@ def _statement_smells(text: str, dialect: str | None) -> list[PerfHit]:
         for sel in stmt.find_all(exp.Select):
             if sel.args.get("where"):
                 continue
+            # ``SELECT ... INTO a, b`` — a multi-target INTO — is parsed with
+            # the trailing target (``b``) as a comma-join FROM item, so a
+            # table-creating statement looks like a cartesian join (issue
+            # #1502). ``into`` is present exactly when the trailing names are
+            # INTO targets, not joins; skip those.
+            if sel.args.get("into"):
+                continue
             for join in sel.args.get("joins") or ():
                 if (
                     not join.args.get("kind")  # CROSS/OUTER etc. are explicit intent
--- a/tests/unit/health/test_sql_markers.py
+++ b/tests/unit/health/test_sql_markers.py
@@ -186,6 +186,21 @@ class TestCartesianJoin:
         assert _smell_kinds("SELECT a.x FROM a JOIN b ON a.id = b.id") == []
         assert _smell_kinds("SELECT a.x FROM a JOIN b USING (id)") == []
 
+    def test_multi_target_select_into_is_silent(self) -> None:
+        """Issue #1502: ``SELECT ... INTO a, b`` parses the trailing target as
+        a comma-join, so it must not be flagged as a cartesian join."""
+        assert _smell_kinds("select a, b into x, y from t where c = 1;") == []
+        assert _smell_kinds("select a, b into x, y from t;") == []
+
+    def test_single_target_select_into_is_silent(self) -> None:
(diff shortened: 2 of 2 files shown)


RaghavChamadiya (approved):
Thanks @sloemo01. The part that makes this the right fix is that `into` is not a heuristic, it is the parser telling you what those trailing names are. sqlglot puts the extra INTO targets in the FROM item list because that is how the tokens sit, but it also records that the statement had an INTO, so gating on `sel.args.get("into")` asks the question directly instead of pattern-matching the symptom.

`test_multi_target_into_does_not_mask_real_cartesian` is the test I was going to ask for. A `continue` inside a per-select loop is easy to write one indentation level too high, and that test is what proves the escape is scoped to the statement that earned it and not the file.

Covering the single-target form too is worth it even though it was never broken, since it pins that this stays silent if someone later narrows the check to multi-target only.

PR #1812, opened 2026-08-21
fix(status): surface a broken repository/path mapping as invalid

### LINKED ISSUE #1748: [Bug] status reports a healthy indexed repository when ingest path mapping is broken
## Summary

`repowise status` reports an index as healthy and current even when its repository/path identity is broken enough that downstream coverage ingestion can map only a handful of files. The reported file count describes the working tree, not the repository row/path mapping used by ingestion.

## Version check

- Latest upstream is 0.44.0; release notes through 0.44.0 do not mention an index-to-working-tree mapping preflight.

## Measured reproduction

In the broken state:

- only 9 files were mappable through the repository identity used by ingestion;
- `repowise status` reported `indexed: true`;
- `last_sync_commit` was correct;
- `health.file_count` reported 2,251;
- `repowise health` output was indistinguishable from the working state;
- `repowise context <known-file>` was indistinguishable from the working state;
- coverage ingestion still exited 0 despite the mapping loss.

A coverage report then mapped only 1 of 249 files and 26 of approximately 29,317 per-test records.

## Expected behavior

Provide a cheap validity/preflight check that verifies the active repository row and indexed absolute-path identity can resolve the current checkout. `repowise status` should either include that result or avoid presenting the index as healthy when this invariant fails.

Useful machine-readable fields would include:

- active repository identity;
- indexed root versus current root;
- resolvable indexed-file count;
- working-tree file count;
- mapping-valid boolean and diagnostic reason.

## Impact

There is currently no inexpensive command an automation can run before a costly coverage job to distinguish a working index from this broken state. All obvious health checks are green, and the eventual ingest also exits 0, so a plausible partial dataset can silently replace valid coverage.

sloemo01:
## What

Fixes #1748. `repowise status` treated `.repowise/` existence alone as "indexed", so a store whose `Repository` row no longer matches the checkout's `local_path` was presented as **healthy — with 0 files** — while downstream coverage ingestion silently lost most of the tree.

## Root cause

`_query_repo_counts` matches `Repository.local_path == str(repo_path)`. When the stored identity diverges from the current checkout it returns `(0, 0)`, but `indexed` was still `true` (`.repowise/` exists), so a broken mapping read as a healthy index with no signal that anything was wrong.

## Fix

Add `_path_mapping_valid(repo_path)`: checks that a `Repository` row resolves the current `repo_path` (the identity ingestion actually uses), and carry it as **`mapping_valid`** on every workspace status row — so a diverged index is surfaced instead of presented as healthy.

## Tests

- `test_path_mapping_valid_with_no_db_is_true` — an unindexed dir is not a broken mapping.
- `test_path_mapping_valid_false_when_repo_row_is_missing` — `.repowise/` exists but no row matches → invalid.
- `tests/unit/cli/test_status_storage.py` + `test_format_json_rollout.py` + `test_workspace_autodetect_e2e.py` — **51 passed**, no regression.
- Ruff clean.

### DIFF
--- a/packages/cli/src/repowise/cli/commands/status_cmd.py
+++ b/packages/cli/src/repowise/cli/commands/status_cmd.py
@@ -104,6 +104,49 @@ def _query_repo_counts(repo_path: Path) -> tuple[int, int]:
         return 0, 0
 
 
+def _path_mapping_valid(repo_path: Path) -> bool:
+    """True when a ``Repository`` row matches *repo_path*'s ``local_path``.
+
+    ``repowise status`` used to treat ``.repowise/`` existence alone as
+    "indexed", so a store whose repository/path identity no longer resolves
+    the current checkout was presented as healthy — with 0 files — and the
+    downstream coverage ingestion silently lost most of the tree (issue
+    #1748). The repository row is the identity ingestion actually uses; when
+    it does not match, the index and the checkout have diverged.
+    """
+
+    async def _check() -> bool:
+        from sqlalchemy import select as sa_select
+
+        from repowise.core.persistence import (
+            create_engine,
+            create_session_factory,
+            get_session,
+        )
+        from repowise.core.persistence.models import Repository
+
+        url = get_db_url_for_repo(repo_path)
+        await reconcile_schema_best_effort(url)
+        engine = create_engine(url)
+        sf = create_session_factory(engine)
+        try:
+            async with get_session(sf) as session:
+                res = await session.execute(
+                    sa_select(Repository.id).where(Repository.local_path == str(repo_path))
+                )
+                return res.scalar_one_or_none() is not None
+        finally:
+            await engine.dispose()
+
+    db_path = get_repowise_dir(repo_path) / "wiki.db"
+    if not db_path.exists() and not db_configured():
+        return True  # nothing to diverge from
+    try:
+        return run_async(_check())
+    except Exception:
+        return False
+
+
 def _query_page_count(repo_path: Path) -> int:
     """Return the number of generated wiki pages for a repo, or 0."""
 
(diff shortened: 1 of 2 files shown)


sloemo01:
Reworked, and you were right about the predicate.

I read #1748's repro again with your point in mind and it holds up: for `status` to have printed `indexed: true` with a correct `last_sync_commit`, the `Repository` row had already resolved by `local_path`, since that is the same lookup `_query_page_count` goes through. So in the state they measured, the row-exists check returns `True` and the row still reads healthy. The divergence is one level down, as you said.

**The predicate is now the file-count comparison you named.** `_mapping_report` returns both counts and compares them, counting the working tree with the same traverser ingestion uses so the two numbers describe the same file set rather than two notions of "a file":

```python
elif tree and indexed * 4 < tree:
```

The ratio is wide on purpose. Their repro is three orders of magnitude apart, so anything near parity is left alone; an index one edit behind is the ordinary state of a working repository and a tight threshold would flag every mid-edit tree. `indexed == 0` against a non-empty tree is reported separately, since that is a different statement from "partially resolved".

**Both surfaces, including the single-repo one you flagged as untouched.** `mapping_valid` and the full report ride the single-repo `--format json`, and the reason prints before the state table. In the workspace table the status column now reads the check *before* the count, because `files == 0` was producing exactly the `empty` reading you objected to.

**On the vacuous test.** Verified rather than taken on faith: a zero-byte `wiki.db` raises `no such table: repositories`, so the old test returned through its own `except` without running the query, and would have passed with the `select` wrong. Replaced with tests that build real stores through `upsert_repository` and count real `GraphNode` file rows. Dropping the comparison block now fails three of them, including the JSON preflight.

On the third point, raw string equality: not fixed here. `_query_page_count` and `_query_repo_counts` do the same comparison at `status_cmd.py:68` and `:127`, and the new report makes a third, so a normalization change belongs in one pass across all three rather than in this one. Say the word and I will follow up with it.

6 new tests, 2,178 pass in `tests/unit/cli/`, `ruff check` clean. The one failure in that suite (`test_agent_target_baseline::test_integration_writes_match_baseline`) is pre-existing on `main`, which I confirmed in a clean worktree.

PR #1813, opened 2026-08-21
feat(generation): REPOWISE_LANGUAGE env pins the generation language

### LINKED ISSUE #1756: [Feature] Add environment variable for natural language
## Problem

When deploying Repowise via Docker, the CLI already provides an option to specify the generation language during repository initialization. However, when a repository is initialized through the UI (Add Repository window), there is no way to set the language at that point – the documentation is generated in English by default. As a result, users must regenerate the documentation from the command line afterward, just to switch to their preferred language, which adds unnecessary friction.

To bridge this gap, I suggest introducing a dedicated environment variable to define the desired language globally. Since both the UI and CLI operations run in the same container environment, the backend could read this variable and apply it during UI‑based initialization as well. This would make the language setting consistent with the existing provider and model environment variables, and completely eliminate the need for an extra CLI regeneration step solely for language changes.

## Proposed Solution

Environment variable name can be REPOWISE_LANGUAGE or something like that.

sloemo01:
## What

Closes #1756. A repo initialized through the **UI / Docker deploy** had no way to set the generation language — it always fell back to English, forcing a CLI regeneration to switch. The CLI already has `--language`, but the UI path bypasses it.

## Fix

Add `_resolve_language(repo_path, config)` — generation language in precedence order:

1. repo-local `.repowise/config.yaml` `language` (CLI/init)
2. **`REPOWISE_LANGUAGE` env** (new — pins the language for a Docker/UI deploy without a CLI flag)
3. default `"en"`

Both CLI and UI/backend generation resolve through the pipeline's config step, so the env var applies to UI-based init in a container. Matches the existing `REPOWISE_EMBEDDER` / provider env-var pattern.

## Tests

- New `test_run_pipeline_language_from_repowise_language_env` in `test_pipeline_reasoning.py` — sets `REPOWISE_LANGUAGE=pt`, runs the pipeline, asserts the resolved language is `pt` (not the English default).
- `tests/unit/generation/test_pipeline_reasoning.py` — **2 passed**, no regression.
- Ruff clean.

### DIFF
--- a/packages/core/src/repowise/core/pipeline/orchestrator.py
+++ b/packages/core/src/repowise/core/pipeline/orchestrator.py
@@ -16,6 +16,7 @@ from __future__ import annotations
 
 import asyncio
 import contextlib
+import os
 import time
 from dataclasses import dataclass, field
 from pathlib import Path
@@ -44,6 +45,19 @@ from .phases.git import _run_git_indexing, drop_transient_git_signals
 from .phases.ingestion import _run_ingestion, reparse_for_resume
 from .resume import ResumePhase
 from .resume.controller import ResumeController
+
+
+def _resolve_language(repo_path: Path | str, config: Any | None = None) -> str:
+    """Generation language, in precedence order (issue #1756).
+
+    repo-local ``config.yaml`` > ``REPOWISE_LANGUAGE`` env (so a Docker/UI
+    deploy can pin the language without a CLI flag) > default ``"en"``.
+    """
+    if config is None:
+        from repowise.core.repo_config import load_repo_config
+
+        config = load_repo_config(repo_path)
+    return config.get("language") or os.environ.get("REPOWISE_LANGUAGE") or "en"
 
 logger = structlog.get_logger(__name__)
 
@@ -687,7 +701,7 @@ async def run_pipeline(
                 max_concurrency=concurrency,
                 reasoning=resolve_reasoning(config=_cfg),
                 wiki_style=_style,
-                language=_cfg.get("language", "en"),
+                language=_resolve_language(repo_path, config=_cfg),
             )
 
         # The mode decides how pages are rendered, not the caller's config.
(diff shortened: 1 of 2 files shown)


RaghavChamadiya (approved):

sloemo01:
Pinged as asked, with one thing you should know: this branch had a real conflict with `main` that I did not notice, and it was the reason no CI ever ran on it.

`gh api repos/repowise-dev/repowise/pulls/1813 --jq .mergeable_state` was `dirty`, and GitHub does not start workflows for a conflicted PR. Three pushes spawned nothing while my other branches ran CI normally, which I first read as a queue or permissions problem. It was neither. Merging `main` in and resolving the one conflict fixed it:

```
before: mergeable_state=dirty      no run, three pushes in a row
after:  mergeable_state=unstable   CI in_progress, 8 checks reporting
```

The conflict was trivia in `core/repo_config.py`: main added `contextlib`, `hashlib` and `json` to the import block while this branch added `os` for the resolver. All four are kept.

Worth flagging that the branch being conflicted while my others were clean is itself the signal I should have read first. If a push produces no run and a sibling branch produces one, check `mergeable_state` before anything else.

**On your three points:**

**The ten sites are wired.** `resolve_language` now lives beside `load_repo_config` in `core/repo_config.py`, and every site you listed calls it: `job_executor.py:993` and `:1251`, `generate_cmd/command.py:255`, `update_cmd/command.py:457` and `:1501`, `deterministic.py:159`, `upgrade_flow.py:457`, `restyle_cmd.py:246`, `workspace_cmd.py:773`, plus the orchestrator call it already had. The helper moved below the logger.

**The test asserts the wiring now, not the helper.** It spies on the `language=` argument `PageGenerator` is constructed with. Confirmed by sabotage: reverting that argument to `_cfg.get("language", "en")` fails the test, which the old version would have passed. That took two attempts to patch correctly — the phase imports `PageGenerator` inside the function body from `repowise.core.generation`, so patching the phase module raises `AttributeError`.

**Verification.** 1,588 tests pass in `tests/unit/generation/` on the merged base, and 0 checks failing. `ruff check` clean on all ten changed files.

Issue #1678, opened 2026-08-17
Semantic search silently returns zero on first query: cold vector-search path exceeds 8s _safe_vector timeout and the exception is swallowed


MrJ55:
﻿### Summary
`repowise search "<prose>" --mode semantic` returns **zero results** for the first query in a fresh process, even when the vector index is healthy (reindex OK, doctor drift 0.0%). The vector leg silently degrades to FTS/symbol-only. Warm follow-up queries return results instantly. This mirrors the failure mode described in #1495 (silent degradation, exit 0, no error) but the root cause here is a **cold-path timeout**, not the ollama build failure.

### Version
- repowise 0.43.0 (installed via `uv tool install repowise`)
- Windows 10
- Index: comet-mcp (and reproduces on repowise-dev/repowise's own repo, 4329-page wiki)
- Embedder: openai / text-embedding-3-small; 1536-dim, 104 rows; `repowise doctor`: `SQL â†” Vector Store: OK`, drift 0.0%

### Repro
```bash
repowise init --no-prose          # structural wiki
repowise reindex --embedder openai
repowise search "how are errors handled" --mode semantic          # => No results found
repowise search "how are errors handled" --mode semantic --full --format json
# => {"results": [], "_meta": {"embedder_degraded": false, ...}}
```

### Root cause (traced)
`_safe_vector` (packages/server/.../mcp_server/tool_search.py) runs:
```python
with contextlib.suppress(Exception):
    return await asyncio.wait_for(ctx.vector_store.search(query, limit=limit), timeout=8.0)
return []
```
The vector search has a **cold path**: LanceDB connect + first embed + first ANN query. Measured on this machine: store open 6.3s, **first search 13.4s**, warm search 0.19s. On every first query per process the 8.0s timeout fires, `contextlib.suppress` swallows the `TimeoutError`, and the vector leg returns `[]`. Same 8.0s timeout in `_non_decision_fallback`. `embedder_degraded` stays `false` because the embedder itself built fine â€” only the query timed out.

### Expected
The vector leg should either allow the cold path time (raise/make-configurable timeout), or a timeout should surface as `embedder_degraded`/a warning instead of silently returning `[]` (same class as #1495's silent-fallout concern).

### Workaround (local)
[...]

PR #1820, opened 2026-08-21
fix: make the edenai embedder selectable and honoured by serve

### LINKED ISSUE #1815: [Bug] The edenai embedder is documented but cannot be selected
## Describe the Bug

`edenai` is registered as an embedder and documented as a valid value, but it cannot actually be selected. Four separate lists were left on the older five-backend set, so the flag is rejected, neither interactive prompt offers the option, the advanced-mode default never notices the key, and on the HTTP server the value falls through to the keyless mock.

This came in with #705, which added the embedder and the docs but did not update those lists. That one is on me, so I am opening this to give the fix somewhere to point rather than sliding it in quietly.

The docs already promise it, which is what makes it a bug rather than a missing feature:

- `docs/reference/CLI_REFERENCE.md` lists `--embedder ... edenai ...` in two tables
- `docs/reference/CONFIG.md` lists `edenai` among the `REPOWISE_EMBEDDER` values

## Steps to Reproduce

1. `repowise init --embedder edenai`
2. `repowise init` in advanced mode, then look at the `Embedder for RAG` prompt
3. Same prompt with only `EDENAI_API_KEY` exported
4. `REPOWISE_EMBEDDER=edenai` with the HTTP server

The first and last reproduce without a real key:

```python
import click
from repowise.cli.commands.init_cmd.command import init_command
init_command.make_context("init", ["--embedder", "edenai"])
# click.exceptions.UsageError: Invalid value for '--embedder': 'edenai' is not one of
# 'gemini', 'openai', 'openrouter', 'ollama', 'mock'.
```

```
$ EDENAI_API_KEY=dummy REPOWISE_EMBEDDER=edenai python -c \
    "from repowise.server.app import _build_embedder; print(type(_build_embedder()).__name__)"
WARNING  embedder.mock_active: set REPOWISE_EMBEDDER=gemini, openai, openrouter, or ollama for real RAG
KeylessEmbedder
```

## Expected Behavior

`edenai` behaves like `openrouter` everywhere: accepted by the flag, offered in both prompts, detected from `EDENAI_API_KEY` as the prompt default, and resolved to `EdenAIEmbedder` by the server.

## Actual Behavior

- `--embedder edenai` exits with `Invalid value for '--embedder'`
- both embedder prompts offer only `gemini`, `openai`, `openrouter`, `ollama`, `mock`
- a machine whose only key is `EDENAI_API_KEY` defaults to `mock`, so semantic search runs on vectors that cannot match the index
- the server logs `embedder.mock_active` and returns `KeylessEmbedder`

The last one is the same failure mode as #306, and the same one #324 fixed for the MCP server by going through the shared registry. `mcp_server/_server.py::_resolve_embedder` is correct today. `server/app.py::_build_embedder` is the remaining hardcoded chain, and it stops at `openrouter`.

## Environment

- OS: Ubuntu (Linux)
- Python version: 3.11.15
- Repowise version: 0.44.0, and `main` at 9c16233
- Installation method: uv workspace from source

## Additional Context

The exact spots, all on current `main`:

- `packages/cli/src/repowise/cli/commands/init_cmd/command.py`, the `--embedder` `click.Choice` and its help string. The `--provider` help string also omits `edenai`, though that one is only undocumented rather than rejected, since it has no `Choice`.
- `packages/cli/src/repowise/cli/ui/mode_selection.py`, `embedder_choices` in advanced config, the `click.Choice` in the index-only Search prompt, and `_resolve_embedder_from_env`, which is a third copy of the detection chain and does not read `EDENAI_API_KEY`. The shared resolver in `cli/providers/embedders.py` does.
- `packages/server/src/repowise/server/app.py`, `_build_embedder` and its docstring and warning.

One thing I deliberately left out of the fix: the numbered embedder menu in `serve_cmd.py`, plus `_get_or_prompt_api_key` and `_set_api_key_env` next to it. Adding a fourth entry there changes a menu you laid out by hand, and those two helpers duplicate `_EMBEDDER_KEY_ENV` from `cli/providers/keys.py`, which already knows about `edenai`. That felt like a separate change and your call rather than mine. Happy to do it either way.

I have a PR ready for the rest, with tests.

MVS-source:
## Summary

- Makes the `edenai` embedder selectable: adds it to the `--embedder` `click.Choice` and its help, to both interactive embedder prompts, and to the env detection that supplies the advanced-mode default.
- Fixes the HTTP server falling through to the keyless mock on `REPOWISE_EMBEDDER=edenai`, which made the server report healthy while semantic search ran on vectors that cannot match the index.
- Adds tests for the flag, both prompts, the detection precedence, and the server resolution, and fills the `edenai` gap in the existing `resolve_embedder` parametrize.

## Related Issues

Fixes #1815

Follow-up to #705, which added the embedder and the docs but left four older lists behind. My omission, so this closes it out.

## Test Plan

- [x] Tests pass (`pytest`)

No frontend change, so the web build is untouched.

What I ran:

- `pytest tests/unit/cli/test_edenai_reachability.py tests/unit/server/test_app_embedder_backends.py` gives 13 passed, and `tests/unit/cli/test_shared_helpers.py` 25 passed with the two new parametrize cases.
- `pytest tests/unit/cli tests/unit/server` for regressions, since the change touches prompt construction that other UI tests drive. This branch gives 6 failed, 4500 passed. I then ran the same command on `main` at 9c16233 with nothing of mine applied, and it gives **9 failed, 4482 passed**, so this branch fails strictly fewer tests than the base and every failure on it also fails without the change. They are all in `test_update_e2e.py`, `test_update_persist_reliability.py`, and `test_watch_cmd.py`, none of which this diff touches, and the set varies between runs, which is what flaky looks like.

  One of them I can now explain, and it corrects something I said on #705. The two `test_watch_cmd::TestSingleRepoTrigger` failures are `OSError: [Errno 28] inotify watch limit reached` on my machine, not anything in your code. Sorry for the earlier noise on that.
- `ruff check .` clean across the repo. For formatting I checked against ruff 0.6.9, the version your `.pre-commit-config.yaml` pins, rather than the 0.15.6 the venv resolves under `>=0.6,<1`. Every line this PR adds is already formatted. Two of the files carry one pre-existing line each that 0.6.9 would rewrap, in `mode_selection.py` and `command.py`, both untouched here and both already like that on `main`, so I ran `format --diff` rather than `format --write` and left them alone.

Both failures reproduce before the change and pass after:

```
--embedder edenai      REFUSE -> ACCEPTE
REPOWISE_EMBEDDER=edenai -> KeylessEmbedder -> EdenAIEmbedder
```

## Checklist

Docs needed no change here: `CLI_REFERENCE.md` and `CONFIG.md` already list `edenai`, which is what made this a bug rather than a feature.

## Notes on two judgement calls

`_resolve_embedder_from_env` in `mode_selection.py` reads `EDENAI_API_KEY` **last**, immediately before the mock, mirroring the shared resolver in `cli/providers/embedders.py`. Same reasoning you gave on #705: an unrelated key in the environment must not outrank a backend the user was already resolving to. There is a parametrized test pinning that, so it cannot drift back.

For `server/app.py` I added a branch next to the others rather than routing through `get_embedder`. Your #324 moved the MCP server onto the shared registry, and `_build_embedder` is the last hardcoded copy, so converging it would fix this class of bug for every registered embedder instead of one. I did not do it here because it changes behaviour for the four existing backends and this PR should stay small, but say the word and I will send that separately.

I left the numbered embedder menu in `serve_cmd.py` alone for the same reason, details in the issue.

I do work at Eden AI.

### DIFF
--- a/packages/cli/src/repowise/cli/commands/init_cmd/command.py
+++ b/packages/cli/src/repowise/cli/commands/init_cmd/command.py
@@ -411,7 +411,7 @@ def _run_generation_phase(
     default=None,
     help=(
         "LLM provider name (anthropic, openai, openrouter, gemini, "
-        "deepseek, kimi, ollama, litellm, codex_cli, opencode, mock)."
+        "deepseek, kimi, ollama, litellm, codex_cli, opencode, edenai, mock)."
     ),
 )
 @click.option("--model", default=None, help="Model identifier override.")
@@ -419,8 +419,11 @@ def _run_generation_phase(
     "--embedder",
     "embedder_name",
     default=None,
-    type=click.Choice(["gemini", "openai", "openrouter", "ollama", "mock"]),
-    help="Embedder for RAG: gemini | openai | openrouter | ollama | mock (default: auto-detect).",
+    type=click.Choice(["gemini", "openai", "openrouter", "ollama", "edenai", "mock"]),
+    help=(
+        "Embedder for RAG: gemini | openai | openrouter | ollama | edenai | mock "
+        "(default: auto-detect)."
+    ),
 )
 @click.option("--skip-tests", is_flag=True, default=False, help="Skip test files.")
 @click.option("--skip-infra", is_flag=True, default=False, help="Skip infrastructure files.")
--- a/packages/cli/src/repowise/cli/commands/reindex_cmd.py
+++ b/packages/cli/src/repowise/cli/commands/reindex_cmd.py
@@ -19,7 +19,7 @@ from repowise.cli.ui import BRAND_STYLE, OWL_SPINNER
 @click.argument("path", required=False, default=None)
 @click.option(
     "--embedder",
-    type=click.Choice(["gemini", "openai", "openrouter", "ollama", "mock", "auto"]),
+    type=click.Choice(["gemini", "openai", "openrouter", "ollama", "edenai", "mock", "auto"]),
     default="auto",
     help="Embedder to use. 'auto' detects from env vars / config.",
 )
--- a/packages/cli/src/repowise/cli/ui/mode_selection.py
+++ b/packages/cli/src/repowise/cli/ui/mode_selection.py
@@ -450,7 +450,7 @@ def _prompt_generation(
 
     # Embedder selection
     detected_embedder = _resolve_embedder_from_env()
-    embedder_choices = ["gemini", "openai", "openrouter", "ollama", "mock"]
+    embedder_choices = ["gemini", "openai", "openrouter", "ollama", "edenai", "mock"]
     result["embedder"] = click.prompt(
         "  Embedder for RAG",
         default=detected_embedder,
@@ -644,7 +644,7 @@ def _prompt_index_only_search(console: Console, result: dict[str, Any]) -> None:
     result["embedder"] = click.prompt(
         "  Embedder for semantic search (mock = full-text only)",
         default="mock",
-        type=click.Choice(["mock", "ollama", "gemini", "openai", "openrouter"]),
+        type=click.Choice(["mock", "ollama", "gemini", "openai", "openrouter", "edenai"]),
     )
 
 
@@ -658,6 +658,11 @@ def _resolve_embedder_from_env() -> str:
         return "openrouter"
     if os.environ.get("OLLAMA_EMBEDDING_MODEL"):
         return "ollama"
+    # Last, like the shared resolver in cli/providers/embedders.py: an unrelated
+    # EDENAI_API_KEY in the environment must not outrank a provider the user was
+    # already resolving to.
+    if os.environ.get("EDENAI_API_KEY"):
+        return "edenai"
     return "mock"
 
 
--- a/packages/server/src/repowise/server/app.py
+++ b/packages/server/src/repowise/server/app.py
@@ -103,6 +103,7 @@ def _build_embedder():
         gemini     — GeminiEmbedder via GEMINI_API_KEY / GOOGLE_API_KEY env var
         openai     — OpenAIEmbedder via OPENAI_API_KEY env var
         openrouter — OpenRouterEmbedder via OPENROUTER_API_KEY env var
+        edenai     — EdenAIEmbedder via EDENAI_API_KEY env var
     """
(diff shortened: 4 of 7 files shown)


RaghavChamadiya (approved):
Thanks @MVS-source. Filing the bug against your own merge before anyone noticed, and saying plainly that #705 was your omission, is the part worth naming here. That is a harder thing to do than the fix.

I checked the two judgement calls rather than taking the description for them.

The `_resolve_embedder_from_env` ordering is right. It mirrors the shared resolver at `cli/providers/embedders.py:74-84` exactly, gemini then openai then openrouter then ollama then edenai then mock, so `EDENAI_API_KEY` only decides when nothing else has. That matters more than usual given you work at Eden AI, and putting your own backend last in the precedence chain is the correct answer rather than the convenient one. The parametrized test pinning it means it cannot quietly drift later either.

The default model is also right: `amazon/amazon.titan-embed-text-v2:0` matches `EdenAIEmbedder.__init__`, not the value the registry docstring claims.

On the four symptoms in #1815, this covers all four. The two that make it worth taking before a release rather than after are the ones that are quiet:

`_prompt_index_only_search` is a second prompt that is easy to miss when you are reading for the obvious one, and `_resolve_embedder_from_env` is the one where nothing errors at all. A machine whose only key is `EDENAI_API_KEY` gets the mock, the server logs healthy, and semantic search runs on vectors that cannot match the index. That is the same shape as #1678, which shipped a fix for this morning, and it is not a failure anyone diagnoses from the outside.

Two follow-ups, neither for this PR.

Yes to converging `_build_embedder` on `get_embedder`, and you are right that it is the last hardcoded copy after #324 moved the MCP server across. Send it separately after the release, since it changes behaviour for four existing backends and that wants its own diff and its own week.

`registry.py:9` says edenai defaults to `openai/text-embedding-3-small`, which is wrong against the class default your PR correctly uses. Pre-existing, unrelated to this change, worth a line whenever you are next in that file.

The flaky `test_watch_cmd` runs are an inotify limit on your machine and not something you needed to apologise for. Running the same command on `main` to show your branch fails strictly fewer tests than the base was the right way to settle it, and more rigour than most people bring to a "tests are flaky" claim.

Crediting @sloemo01 in the reindex commit for the one your own sweep missed was a good call.

PR #1833, opened 2026-08-22
perf(cli): cut the per-call startup floor

### LINKED ISSUE #1712: CLI per-call ~6s fixed startup floor makes CLI one-shots uneconomical for automation/hooks (serve/mcp path is <1s)
## Summary

Every `repowise` CLI invocation spawns a fresh Python process whose fixed startup cost dominates short commands, making the CLI one-shot pattern uneconomical for tool automation / hooks (e.g. extensions, editors, CI).

## Observations (Windows, repowise 0.43.0, 2026-08-18)

Timed a batch of CLI calls from the same repo / cwd:

| command | wall time |
|---|---|
| `repowise status` | ~6.0 – 7.1 s |
| `repowise decision list` | ~6.0 – 6.9 s |
| `repowise search --mode symbol` | ~6.2 s |
| `repowise search --mode concept` | ~46 s (see #1678/#1685 for the vector/LLM legs) |

The ~6 s floor is present even for cheap, purely-local metadata commands (`status`), i.e. it is process startup (Python import graph + DB/index open), not work. It is roughly constant across commands of very different cost.

## Impact

An integration that shells out to `repowise <cmd>` per call (as a Pi extension or editor hook does) pays ~6 s before doing anything, and can't afford to run per-event. This is exactly why CLI-backed hooks fall back to no-op: the latency exceeds any sane per-call budget. The MCP server path (`repowise mcp` / `search_codebase`) answers in <1 s by contrast, because it keeps the process alive.

## Suggested direction (not blocking)

- A long-lived mode that reuses the loaded index (e.g. documenting/widening `repowise serve --no-ui`, or a `--once`-style batch entrypoint, or making `repowise mcp` usable as a stdio client for one-off CLI-style calls) so callers hit <1 s.
- Or a `repowise status --json` fast path that defers expensive imports where possible.

Not a correctness bug — a performance/enablement gap for CLI automation.

sloemo01:
Fixes #1712

## Problem
Every `repowise` CLI invocation pays a ~1.1s fixed import floor on top of actual work, making one-shot calls (`status`, `coverage status`, hooks, editor integrations) uneconomical.

Root cause: `repowise.cli.ui/__init__.py` is an **eager facade**. Importing *any* submodule — even the cheap `ui.brand.format_bytes` that `status` needs — runs the package `__init__`, which eagerly imports every interactive UI submodule (`mode_selection` → `repo_scanner` → `core.ingestion` → `networkx` + `tree-sitter`). Measured with `python -X importtime`: importing `status_cmd` pulled `networkx` (0.045s) + `tree_sitter` (0.19s) + `core.ingestion` (~1.1s cumulative).

## Fix
Convert `ui/__init__.py` to a **lazy facade** (PEP 562 `__getattr__`). Each public name maps to its owning submodule and is imported only on first access. Cheap commands no longer pay for the interactive UI's heavy graph deps.

## Measured
Cold `repowise status`:
- Before: **~1.9s**
- After: **~0.5s** (~4x)

Import profile of `status_cmd`: `core.ingestion`/`networkx`/`tree_sitter` → **0** modules loaded.

## Tests
- Added regression test asserting a fresh interpreter importing `ui.brand` loads no `core.ingestion`/`networkx`/`tree_sitter`/`lancedb`/`scipy`.
- Existing facade-parity test (all 44 public names resolve) and interactive-UI tests all pass.
- `ruff` clean.

### DIFF
--- a/packages/cli/src/repowise/cli/ui/__init__.py
+++ b/packages/cli/src/repowise/cli/ui/__init__.py
@@ -3,109 +3,88 @@
 This package was split out of the former single ``ui.py`` module. The submodules
 group the helpers by concern; this façade re-exports every previously-public name
 so existing ``from repowise.cli.ui import ...`` call sites are unchanged.
+
+The façade is lazy: each public name is resolved on first access via
+:func:`__getattr__` rather than imported eagerly at package import time. That
+matters for CLI startup — importing *any* submodule (e.g. ``ui.brand``, the one
+cheap ``status``/``coverage`` commands need) used to run this package's
+``__init__``, which eagerly pulled in the heavy interactive modules
+(``mode_selection`` -> ``repo_scanner`` -> ``core.ingestion`` -> networkx +
+tree-sitter, ~1.1s on a cold process). Issue #1712. ``import repowise.cli.ui``
+still resolves ``__all__`` cheaply because the name -> submodule mapping lives in
+this module; only the bodies are imported when actually accessed.
 """
 
 from __future__ import annotations
 
-from repowise.cli.ui.brand import (
-    BRAND,
-    BRAND_STYLE,
-    DIM,
-    ERR,
-    OK,
-    VALUE,
-    WARN,
-    format_bytes,
-    format_elapsed,
-    key_value_table,
-    print_banner,
-    print_phase_header,
-    print_section,
-)
-from repowise.cli.ui.env_persistence import load_dotenv
-from repowise.cli.ui.mascot import (
-    OWL_SPINNER,
-    THINKING_FRAMES,
-    banner_text,
-    mini,
-)
-from repowise.cli.ui.mode_selection import (
-    LARGE_REPO_FILE_THRESHOLD,
-    interactive_advanced_config,
-    interactive_customize_offer,
-    interactive_fast_mode_offer,
-    interactive_generate_docs_toggle,
-    interactive_mode_select,
-    print_index_only_intro,
-    prompt_file_page_volume,
-    prompt_wiki_style,
-    should_offer_fast_mode,
-)
-from repowise.cli.ui.progress import MaybeCountColumn, RichProgressCallback
-from repowise.cli.ui.provider_selection import (
-    ProviderSelection,
-    interactive_provider_config_select,
-    interactive_provider_select,
-)
-from repowise.cli.ui.repo_scanner import (
-    RepoScanInfo,
-    print_scan_summary,
-    quick_repo_scan,
-)
-from repowise.cli.ui.result_panels import (
-    build_completion_panel,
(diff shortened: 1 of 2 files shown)


Ayush7614 (approved):
Verified: `packages/cli/src/repowise/cli/ui/__init__.py:1` converts eager facade to lazy PEP 562 __getattr__, so importing ui.brand no longer pulls core.ingestion/networkx/tree-sitter. Measured 1.9s→0.5s cold status. Regression test asserts no heavy deps on fresh import. Correct perf fix — LGTM.

RaghavChamadiya (approved):
Thanks @sloemo01. I measured this rather than taking the number on trust, and it holds up better than the description claims.

Importing the one cheap thing `status` needs, on my machine, three cold processes each:

```
main          858.5 ms / 644.1 ms / 582.8 ms    networkx=True   tree_sitter=True
this branch   115.2 ms / 102.6 ms /  98.8 ms    networkx=False  tree_sitter=False
```

The second half of that is the part I care about more than the milliseconds. On main, `from repowise.cli.ui.brand import format_bytes` has `networkx` and `tree_sitter` resident by the time it returns, because touching any submodule runs the package `__init__` and that eagerly walks `mode_selection` to `repo_scanner` to `core.ingestion`. On this branch neither is ever loaded. That is the difference between a facade that is cheap and one that merely looks cheap.

Diagnosing it as an eager facade rather than reaching for a lazy import inside each call site is the right call. There are a lot of `from repowise.cli.ui import ...` sites and fixing them individually would have been a much larger diff that had to be maintained forever; fixing the package that makes them expensive is one file.

Two things I checked because a lazy facade is exactly where a rename or a typo goes unnoticed until runtime:

The exported surface is unchanged. `__all__` on main is 43 names and `sorted(_LAZY)` here is the same 43, byte-identical as a set, so nothing is gained or lost and no existing `from repowise.cli.ui import X` can start failing. Deriving `__all__` from `_LAZY` rather than maintaining a second list is what keeps that true as things get added.

The `_loaded` cache matters and the docstring says why. The MCP and server paths are long-lived processes, and `__getattr__` fires on every attribute access, not just the first, so without the cache each access would re-enter `import_module`. That is cheap after the first, since `sys.modules` catches it, but it is not free, and caching the module object skips the lookup entirely.

One consequence worth knowing rather than acting on: PEP 562 `__getattr__` returns `Any`, so a type checker can no longer see through this facade to the real signatures of the 43 names. CI lints with `ruff check .` and does not type-check Python, so nothing breaks today, and the trade is clearly worth it for a 6x startup cut. If a type-checking gate ever lands, the answer is an `if TYPE_CHECKING:` block re-declaring the eager imports, which costs nothing at runtime.

PR #1834, opened 2026-08-22
fix(workspace): single-flight workspace updates

### LINKED ISSUE #1831: [Bug] Workspace updates are not single-flighted, so one rebase spawns a full workspace update per commit
## Summary

The single-flight lock that stops two `repowise update` runs from racing is held per member repo, not per workspace. In workspace mode a second invocation therefore does not defer. It walks the member list, wins the lock on whichever repos the first one has not reached yet, and does real ingestion on those. With N invocations you get N concurrent parse pools rather than one run and N-1 quick exits.

The post-commit hook fires one `repowise update` per commit with no coalescing of its own, and `git rebase` replays commits one at a time. So rebasing a five commit branch inside a workspace primary starts five workspace updates at the same second.

This is the workspace half of #95. That issue was about two concurrent updates on the same workspace and it is closed, but the guard that closed it only covers a single repo.

## What actually happened

Rebasing a five commit branch in a workspace primary. From `git reflog`:

```
3282e555 HEAD@{08-22 14:35}: rebase (finish): returning to refs/heads/fix/...
3282e555 HEAD@{08-22 14:35}: rebase (pick): ...
6108e04d HEAD@{08-22 14:35}: rebase (pick): ...
ec4ef7f3 HEAD@{08-22 14:35}: rebase (pick): ...
a80656f3 HEAD@{08-22 14:35}: rebase (pick): ...
31977c38 HEAD@{08-22 14:35}: rebase (pick): ...
363a476a HEAD@{08-22 14:35}: rebase (start): checkout origin/main
```

Five picks, and fifty minutes later five updates were still running, each parented by the hook:

```
  PID   parent   mem     cpu     started
77156    90076   578 MB  2842s   14:35:42
78392    94752  1859 MB  3089s   14:35:42
17324    13772   656 MB  2864s   14:35:42
86036    19912  1032 MB  2185s   14:35:43
27344    10588   771 MB  3061s   14:35:43

parent 90076: repowise.exe update
  grandparent: sh.exe :: sh ".git/hooks/post-commit"
```

Together 4.9 GB, five cores pinned, and the machine at 1.4 GB free of 31.2 GB. Killing them returned 6.4 GB.

They also broke each other. `.repowise/.update.log` is a wall of:

```
Future exception was never retrieved
future: <Future finished exception=BrokenProcessPool('A process in the process pool was terminated abruptly while the future was running or pending.')>
```

interleaved with normal progress lines from different members, which is what five parse pools sharing one machine looks like.

Nothing in the output says another update is in flight. The deferral message in `update_cmd/command.py:932` never printed, because in workspace mode that code path is not the one taken.

## Where

`packages/core/src/repowise/core/workspace/update.py:774-794`, inside the per member loop:

```python
# Per-repo single-flight lock. The post-commit hook fires a
# new ``repowise update`` for every commit; without this guard,
# rapid-fire commits race on save_state, each pass starts from
# the same stale base, and the wiki never converges to HEAD.
# Check + acquire are one atomic exclusive create.
existing = _try_acquire_lock(path, new_head)
if existing is not None:
    ...
    return RepoUpdateResult(alias=alias, updated=False, ...)
```

The comment names the exact scenario. The lock is correct for what it guards: two runs never ingest the same member at once. What it cannot express is "a workspace update is already in flight", because its key is the member path.

`update_cmd/command.py:923` is the single repo equivalent and does bail the whole run:

```python
existing_lock = try_acquire_update_lock(repo_path, head)
if existing_lock is not None:
    ...
    return UpdateOutcome.DEFERRED
```

So the behaviour differs by mode. Outside a workspace, five hook fires give you one update and four immediate deferrals. Inside a workspace, five hook fires give you five updates.

The hook itself (`packages/cli/src/repowise/cli/hooks.py:60`) has no guard and is not really the place for one:

```sh
(
  cd "$ROOT" || exit 1
  if command -v repowise >/dev/null 2>&1; then
    repowise update >> "$LOG" 2>&1
  fi
) &
```

It writes `.update.queued` first, but that marker is only read by `augment_cmd` to suppress a staleness warning (`helpers.py:355-380` documents this). It is not a mutual exclusion primitive and was never meant to be one.

## Why the members deferring individually does not save you

Two invocations racing over the same member list do not serialise, they interleave. Each wins the members the other has not locked yet, so both keep finding work and both keep going. The more members a workspace has, the more work each concurrent run finds before it runs out, so a large workspace makes it worse rather than better.

It is worth saying that the workspace file here had drifted to 151 entries, 75 of them indexed, which is my own mess and not a product bug. But the shape does not depend on that. Three members and two concurrent runs reproduce it, just with less to go around.

## Repro

```bash
# in a workspace root with two or more indexed members
repowise update &
repowise update &
wait
```

Both do real ingestion. Expected is one run and one message naming the pid holding the lock, which is what the same two commands produce in a non-workspace repo.

Or the way it actually shows up: install the post-commit hook in a workspace primary and rebase a branch with several commits.

## Environment

- OS: Windows 11
- Python 3.13.5
- Repowise 0.45.0
- Installed with uv, and separately into a conda env, which is the one the hook resolved

- Take the single-flight lock at the workspace root as well as per member, so a second workspace update defers as a whole and writes `.update.pending` for the workspace. Matches what single repo mode already does and reuses the existing lock primitive.
- Have the hook check for a live lock before spawning at all. Cheaper for the common case, since the expensive part is Python startup plus DB open on a process that is about to bail anyway, and five of those on a rebase is still five interpreters. Does not fix a directly invoked `repowise update`, so probably not on its own.
- Coalesce in the hook by target commit: if `.update.queued` is fresh, overwrite the target and skip spawning. The marker already carries `target_commit` and the running update already knows how to roll forward through `.update.pending`, so most of the machinery exists.
- Bound the parse pool globally rather than per process, so concurrent runs degrade instead of oversubscribing. Worth considering regardless, since the `BrokenProcessPool` failures above are a second symptom and probably related to #1394.

The first is the smallest change that makes workspace mode behave like single repo mode, and the others are useful on top rather than instead.

## Related

- #95, the single repo version of this, closed by the lock that this issue is about the limits of
- #1394, parse pool memory, which is what turns "several concurrent runs" into an unresponsive machine

sloemo01:
## What

Fixes #1831 — workspace updates are now single-flighted at the **workspace** level, so a rebase that produces N commits no longer spawns N full workspace update passes.

### Root cause

The single-flight update lock (`update_lock.py`) lives at `<repo>/.repowise/.update.lock` — **per member repo**. It correctly stops two updates from racing on the *same* repo's index, but it does nothing to stop the post-commit hook from firing a fresh `repowise update` per rebase commit, each of which runs a full `update_workspace()` pass over every stale member. A rebase that rewrites N commits therefore triggers N redundant whole-workspace re-indexes.

### The fix

- **`packages/core/src/repowise/core/update_lock.py`** — generalize the lock helpers behind an explicit lock-path primitive (`_try_acquire_lock_at`, `_release_lock_at`, `_read_lock_at`) and add a **workspace** lock at `<workspace>/.repowise-workspace/.update.lock` (`update_workspace_lock` / `release_workspace_lock`). The workspace guard inherits the exact same crash/liveness/PID-reuse semantics as the per-repo lock.
- **`packages/core/src/repowise/core/workspace/update.py`** — `update_workspace()` now acquires the workspace lock before the parallel pass. If another workspace update already holds it, the whole run defers (`skipped_reason="in_flight"`) and writes a `.update.pending` marker per stale member so the in-flight pass rolls forward to the latest HEAD (mirroring the existing per-repo deferral). The lock is released in a `finally` after the pass.

A rebase's post-commit hooks now coalesce: the first invocation wins the workspace lock and updates everything; the rest record pending heads and exit instead of re-indexing the whole workspace.

### Tests

- `tests/unit/cli/test_update_lock.py` — workspace lock path isolation, single-flight, and stale-lock recovery.
- `tests/unit/workspace/test_update.py` — a concurrent `update_workspace` defers to an in-flight workspace update (writes pending markers, updates nothing); the lock is released after a successful pass.

### Verification

- `uv run ruff check .` — all checks passed
- `uv run pytest tests/unit/cli/test_update_lock.py tests/unit/workspace/test_update.py` — 50 passed

### DIFF
--- a/packages/core/src/repowise/core/update_lock.py
+++ b/packages/core/src/repowise/core/update_lock.py
@@ -42,6 +42,87 @@ def update_lock_path(repo_path: Path) -> Path:
     return Path(repo_path) / ".repowise" / UPDATE_LOCK_FILENAME
 
 
+def workspace_update_lock_path(workspace_root: Path) -> Path:
+    """Path of the workspace-level single-flight update lock.
+
+    Unlike :func:`update_lock_path` (per-repo), this guards the whole
+    workspace so a rebase's N post-commit hooks coalesce into one full
+    ``update_workspace`` pass instead of N redundant ones. Lives under the
+    workspace data dir (``.repowise-workspace/``) so it is shared by every
+    member repo's update.
+    """
+    return Path(workspace_root) / ".repowise-workspace" / UPDATE_LOCK_FILENAME
+
+
+def _try_acquire_lock_at(
+    lock_path: Path,
+    target_commit: str | None,
+) -> dict[str, Any] | None:
+    """Core exclusive-create acquire against an explicit lock path.
+
+    Shared by the per-repo (:func:`try_acquire_update_lock`) and workspace
+    (:func:`update_workspace_lock`) single-flight guards — one
+    implementation for both, so the workspace guard inherits the same
+    crash/liveness/coalescing semantics as the per-repo one.
+    """
+    from repowise.core.procutils import process_create_token
+
+    payload = {
+        "pid": os.getpid(),
+        "pid_create_token": process_create_token(os.getpid()),
+        "target_commit": target_commit,
+        "started_at": time.time(),
+    }
+    data = json.dumps(payload)
+    tmp_path = lock_path.with_name(
+        f"{UPDATE_LOCK_FILENAME}.{os.getpid()}.{threading.get_ident()}.tmp"
+    )
+
+    def _read_existing() -> dict[str, Any] | None:
+        return _read_lock_at(lock_path)
+
+    for _ in range(2):
+        try:
+            lock_path.parent.mkdir(parents=True, exist_ok=True)
+            tmp_path.write_text(data, encoding="utf-8")
+            # Atomic create-with-content: the lock either doesn't exist or
+            # holds a complete payload — contenders can never read a
+            # half-written file.
+            os.link(tmp_path, lock_path)
+        except FileExistsError:
+            existing = _read_existing()
+            if existing is not None:
+                return existing
+            # Stale lock: clear it and retry the exclusive create.
+            with contextlib.suppress(OSError):
+                lock_path.unlink(missing_ok=True)
+            continue
+        except OSError:
+            return None
+        finally:
+            with contextlib.suppress(OSError):
+                tmp_path.unlink(missing_ok=True)
+        return None
+    # Both create attempts lost a race against a stale lock that was then
+    # unlinked. Rather than falling through to a bare read — which can return
+    # ``None`` ("acquired") with *no lock file on disk*, letting a caller
+    # proceed without holding the lock — make one final exclusive create.
+    # Winner: return ``None`` (owned). Loser: report the fresh winner.
+    # Still-unreadable degrades to acquired, as everywhere.
+    try:
+        lock_path.parent.mkdir(parents=True, exist_ok=True)
+        tmp_path.write_text(data, encoding="utf-8")
+        os.link(tmp_path, lock_path)
+    except FileExistsError:
+        return _read_existing()
+    except OSError:
+        return None
+    finally:
+        with contextlib.suppress(OSError):
+            tmp_path.unlink(missing_ok=True)
+    return None
+
+
 def try_acquire_update_lock(repo_path: Path, target_commit: str | None) -> dict[str, Any] | None:
     """Atomically acquire the update lock. ``None`` means acquired.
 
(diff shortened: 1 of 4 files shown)


Ayush7614 (approved):
Verified: `packages/core/src/repowise/core/update_lock.py:42` generalizes to _try_acquire_lock_at and adds workspace lock at .repowise-workspace/.update.lock; `packages/core/src/repowise/core/workspace/update.py:1` acquires before parallel pass and defers with pending marker if in-flight (#1831). Coalesces N rebase hooks into one pass. Tests test_update_lock.py + test_update.py 50 passed. LGTM.

RaghavChamadiya (approved):
Thanks @sloemo01. Extracting `_try_acquire_lock_at` and giving the workspace guard the same primitive is the part that makes this safe to take. A second lock implementation would have been the obvious way to write this and it would have drifted: the hard-link-over-tempfile acquire, the two-attempt stale retry, the final exclusive create instead of a bare read, and the PID-creation-token identity check are all subtle and all now have exactly one copy. I read the refactor as a pure move and could not find a semantic change in it.

I traced the lock lifetime specifically, since a leaked workspace lock would wedge every member repo rather than one. The acquire sits immediately above the `try`, with nothing between them that can return or raise, and `_release_workspace_lock(workspace_root)` is in the `finally` that covers the whole parallel pass through `return results`. The two earlier returns both happen with no lock held: the no-stale-repos case is above the acquire, and the deferral case only runs when the acquire lost. So there is no path that exits holding it.

Putting the lock under `.repowise-workspace/` rather than reusing a member's `.repowise/` is right, and `test_workspace_lock_lives_under_workspace_data_dir` is the test that stops someone collapsing the two paths later. Making the deferral write a `.update.pending` per stale member, and `mkdir` the `.repowise` dir first, means a first-time member that has never been indexed still records its head instead of throwing.

One non-blocking observation. The workspace lock is taken after the staleness scan, so N coalesced hook invocations each still walk every member's git state before deferring. That is a very large improvement over N full re-indexes and I am not asking you to change it, but on a big workspace the scan is not free, and hoisting the acquire above the scan is the next step if #1831 turns out not to be fully quiet.

Full matrix green on f3953284.

PR #1835, opened 2026-08-22
fix(web): render API timestamps as UTC

### LINKED ISSUE #1821: Web UI dates display 8 hours behind the real UTC time on non-UTC clients
## Summary

The web UI renders every API-supplied timestamp **offset by the browser's local timezone** instead of as UTC. On a machine in `Asia/Shanghai` (UTC+8), every timestamp the UI shows is **8 hours earlier** than the actual UTC value stored in the DB and returned by the API.

## Verification

I reproduced this end-to-end on `repowise v0.44.0` against a freshly-initialized repo (CST timezone):

| Source | Value |
|---|---|
| Server clock (UTC) | `2026-08-21T13:43:26Z` |
| `git log -1 --format=%ai` (real local time of last commit) | `2026-08-21 21:42:53 +0800` |
| API: `GET /api/repos/<id>/overview-summary` → `top_hotspots[0].last_commit_at` | `"2026-08-21T13:23:33"` *(no `Z`; value is UTC)* |
| UI: Commits table → `Committed` column for the same commit | `Aug 21, 2026, 1:23 PM` *(8 h late)* |

I confirmed with Node that `new Date("2026-08-21T13:23:33").toLocaleString()` (no tz suffix in input) parses as local time in a UTC+8 browser and renders `1:23 PM`, while `new Date("2026-08-21T13:23:33Z").toLocaleString()` renders `9:23 PM` — matching the real commit time.

## Steps to reproduce

1. Install repowise 0.44.0 on a machine whose OS timezone is *not* UTC (e.g. `Asia/Shanghai`, `America/New_York`).
2. `repowise init` any repo, `repowise serve`, open the UI.
3. Visit **Commits** or **Code Health**.
4. Compare the `Committed` / `Last touched` column to `git log --format=%ai` for the same commit.

**Expected:** UI shows the same wall-clock value as `git log` (or at minimum, the UTC value that the API intends).
**Actual:** UI shows the wall-clock value **minus the browser's UTC offset**.

## Root cause

The backend serializes `DateTime(timezone=True)` columns with `datetime.isoformat()`. On SQLite (which the default `wiki.db` uses), the tz info is dropped on read, so the API returns naive strings like `"2026-08-21T13:23:33"` — *timezone-naive but the value is UTC*.

The frontend's date helpers in the published web bundle look like this (minified, from the v0.44.0 tarball):

```js
// relative-time helper
function l(e) {
  let t = "string" == typeof e ? new Date(e) : e;
  let r = Math.floor((Date.now() - t.getTime()) / 1e3);
  // ... returns "5m ago", etc.
}

// date helper (e.g. "Aug 21, 2026")
function c(e) {
  let t = "string" == typeof e ? new Date(e) : e;
  return new Intl.DateTimeFormat("en-US", {
    month: "short", day: "numeric", year: "numeric"
  }).format(t);          // ❌ no timeZone option
}

// date+time helper (e.g. "Aug 21, 2026, 1:23 PM")
function i(e) {
  let t = "string" == typeof e ? new Date(e) : e;
  return new Intl.DateTimeFormat("en-US", {
    month: "short", day: "numeric", year: "numeric",
    hour: "numeric", minute: "2-digit"
  }).format(t);          // ❌ no timeZone option
}
```

Per ECMAScript, `new Date("2026-08-21T13:23:33")` (no offset) is parsed as **local time**. On a UTC+8 browser, that's interpreted as 13:23 CST, then `.format(t)` displays in local CST → `1:23 PM`. The actual UTC value is 13:23 (CST 21:23), so the displayed time is 8 hours behind reality.

`fw` and `Tk` (relative-time helpers) also base their "X ago" math on `new Date(e)` without tz, so their drift is similar — the threshold minutes may be off by the local offset, but it's harder to spot at a glance.

For reference, the one place in the UI that *does* display dates correctly is in `app/repos/[id]/stats/page.tsx`, where the "Root commit" string is rendered with an explicit `timeZone: "UTC"` option:

```ts
new Date(t.first_commit_at).toLocaleDateString(undefined, {
  day: "numeric", month: "short", year: "numeric",
  timeZone: "UTC"   // ← here
})
```

That call site is correct; the helpers are not.

## Suggested fix

Add `timeZone: "UTC"` to the two `Intl.DateTimeFormat` calls in the date helpers module:

```diff
 function c(e) {
   let t = "string" == typeof e ? new Date(e) : e;
-  return new Intl.DateTimeFormat("en-US", { month: "short", day: "numeric", year: "numeric" }).format(t);
+  return new Intl.DateTimeFormat("en-US", { month: "short", day: "numeric", year: "numeric", timeZone: "UTC" }).format(t);
 }
 function i(e) {
   let t = "string" == typeof e ? new Date(e) : e;
-  return new Intl.DateTimeFormat("en-US", { month: "short", day: "numeric", year: "numeric", hour: "numeric", minute: "2-digit" }).format(t);
+  return new Intl.DateTimeFormat("en-US", { month: "short", day: "numeric", year: "numeric", hour: "numeric", minute: "2-digit", timeZone: "UTC" }).format(t);
 }
```

For the relative-time helpers, the same problem applies to the `Date.now() - t.getTime()` baseline. Two reasonable approaches:

1. Add `timeZone: "UTC"` only to the formatter calls — the relative-time math will be slightly off near the boundary but acceptable.
2. Or fix it server-side: emit `Z` (or `+00:00`) on every datetime from the JSON encoder so naive ISO stops being ambiguous. The frontend patch is one file; the backend patch is two (encoder + the SQLite-read fix). Both are worth doing, but the frontend patch alone unblocks the user-visible UI bug.

## Workaround

Until the fix ships:

1. Open the UI in a browser whose TZ is set to UTC — the UI then renders the same numbers as the API. Practical only for personal viewing.
2. Use the API directly (`curl /api/repos/<id>/overview-summary`) and interpret the timestamps yourself as UTC.

## Environment

- repowise 0.44.0 (pip-installed via `uv tool`)
- Browser: any modern Chromium/Firefox
- Server timezone: any non-UTC timezone reproduces; UTC hides the bug

sloemo01:
## Summary

Fixes #1821. The repowise API returns UTC wall-clock strings with no trailing `Z` (e.g. `"2026-08-21T13:23:33"`). The web UI parsed these with `new Date()` — which interprets a timezone-less ISO string as *local* time — and then formatted them in the browser's local timezone. On a non-UTC client (e.g. `Asia/Shanghai`, UTC+8) every API-supplied timestamp rendered **8 hours behind** the real UTC value.

### Root cause
```js
new Date("2026-08-21T13:23:33").toLocaleString()  // 1:23 PM in UTC+8 — 8h late
new Date("2026-08-21T13:23:33Z").toLocaleString() // 9:23 PM — correct
```

### Fix
- **`packages/ui/src/lib/format.ts`** — added a `parseDate` helper that treats a bare ISO string (no timezone suffix) as UTC and appends `Z`. `formatDate`, `formatDateTime`, `formatRelativeTime`, and `formatRelativeTimeOrNull` now parse via `parseDate` and render with `timeZone: 'UTC'`, so the displayed wall-clock always matches the value the API returned.
- Replaced all raw `new Date(...).toLocaleString()/toLocaleDateString()` calls across the UI and web surface with the UTC-aware formatters — including the **Committed** column (security findings) and **Last touched** column (dead-code findings) called out in the issue, plus coupling, coverage, health, overview, code-health, commits timeline, and chart bucket labels.
- **Tests** — added unit tests in `packages/ui/__tests__/lib/format.test.ts` covering `parseDate` (bare ISO as UTC, explicit `Z`/offset preserved, `Date` pass-through) and `formatDate`/`formatDateTime` rendering in UTC.

### Test plan
- `packages/ui`: `npx vitest run __tests__/lib/format.test.ts` — **16 passed**.
- `packages/ui` & `packages/web`: `tsc --noEmit` — **0 errors**.

Closes #1821

### DIFF
--- a/packages/ui/__tests__/lib/format.test.ts
+++ b/packages/ui/__tests__/lib/format.test.ts
@@ -3,7 +3,10 @@ import {
   formatAgeDays,
   formatBytes,
   formatCompact,
+  formatDate,
+  formatDateTime,
   formatPercent,
+  parseDate,
 } from "../../src/lib/format";
 
 describe("formatCompact", () => {
@@ -73,5 +76,44 @@ describe("formatBytes", () => {
   it("returns an em dash for invalid input", () => {
     expect(formatBytes(-1)).toBe("—");
     expect(formatBytes(Number.NaN)).toBe("—");
+  });
+});
+
+describe("parseDate", () => {
+  it("treats a bare ISO string (no tz suffix) as UTC", () => {
+    // "2026-08-21T13:23:33" must mean 13:23 UTC, not local time.
+    expect(parseDate("2026-08-21T13:23:33").toISOString()).toBe(
+      "2026-08-21T13:23:33.000Z",
+    );
+  });
+
+  it("keeps an explicit Z or offset as-is", () => {
+    expect(parseDate("2026-08-21T13:23:33Z").toISOString()).toBe(
+      "2026-08-21T13:23:33.000Z",
+    );
+    expect(parseDate("2026-08-21T21:23:33+08:00").toISOString()).toBe(
+      "2026-08-21T13:23:33.000Z",
+    );
+  });
+
+  it("passes Date instances through unchanged", () => {
+    const d = new Date("2026-08-21T13:23:33Z");
+    expect(parseDate(d)).toBe(d);
+  });
+});
+
+describe("formatDate / formatDateTime render UTC, not the local zone", () => {
+  it("renders the bare API string as the same wall-clock in UTC", () => {
+    // With a plain `new Date("2026-08-21T13:23:33")`, a UTC+8 browser would
+    // render 21:23 on the 21st as the local reading of 13:23 UTC. Parsing as
+    // UTC and pinning timeZone: "UTC" keeps the shown wall-clock fixed at the
+    // value the API returned.
+    expect(formatDate("2026-08-21T13:23:33")).toBe("Aug 21, 2026");
+    expect(formatDateTime("2026-08-21T13:23:33")).toBe("Aug 21, 2026, 1:23 PM");
+  });
+
+  it("renders an explicit Z timestamp at the same UTC wall-clock", () => {
+    expect(formatDate("2026-08-21T13:23:33Z")).toBe("Aug 21, 2026");
+    expect(formatDateTime("2026-08-21T13:23:33Z")).toBe("Aug 21, 2026, 1:23 PM");
   });
 });
--- a/packages/ui/src/commits/code-evolution-chart.tsx
+++ b/packages/ui/src/commits/code-evolution-chart.tsx
@@ -22,6 +22,7 @@ import {
   TooltipTrigger,
 } from "../ui/tooltip";
 import { cn } from "../lib/cn";
+import { parseDate } from "../lib/format";
 
 /**
  * The repo's development "story arc" — a stacked-area timeline of how the
@@ -81,13 +82,13 @@ const STACK_ORDER: CommitCategory[] = [
 type Mode = "share" | "volume";
 
 function formatBucketLabel(startIso: string, granularity: string): string {
-  const d = new Date(startIso);
+  const d = parseDate(startIso);
   if (Number.isNaN(d.getTime())) return startIso;
-  const month = d.toLocaleString("en-US", { month: "short" });
+  const month = d.toLocaleString("en-US", { month: "short", timeZone: "UTC" });
   if (granularity === "week") {
-    return `${month} ${d.getDate()}`;
+    return `${month} ${d.getUTCDate()}`;
   }
-  return `${month} ${String(d.getFullYear()).slice(2)}`;
+  return `${month} ${String(d.getUTCFullYear()).slice(2)}`;
 }
 
 /** Dominant category across a slice of buckets (by summed count). */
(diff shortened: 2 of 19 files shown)


Ayush7614 (approved):
Verified: `packages/ui/src/lib/format.ts:1` parseDate treats bare ISO as UTC (append Z) and formatDate/formatDateTime render with timeZone UTC, fixing 8h shift for Asia/Shanghai clients (#1821). All raw new Date().toLocaleString sites replaced. Tests __tests__/lib/format.test.ts 16 passed, tsc clean. LGTM.

RaghavChamadiya (approved):
Thanks @sloemo01. The diagnosis is precise: a bare ISO string with no zone suffix is parsed as *local* time by `new Date`, and the API returns exactly that shape, so every timestamp shifted by the viewer's offset. A UTC+8 reader saw 21:23 for a 13:23 UTC event and had no way to know.

Doing both halves is what makes it correct. Parsing as UTC alone would still let `toLocaleString` render in the viewer's zone, and pinning `timeZone: "UTC"` alone would not fix the parse. The tests pin both, and the `+08:00` case proves an explicit offset is still honoured rather than being clobbered by the UTC assumption, which is the regression a blunter fix would introduce.

Routing `code-evolution-chart.tsx` through `parseDate` rather than leaving its own `new Date(startIso)` is the right instinct too. A bucket label off by a day at a month boundary is the kind of wrong that never gets reported, it just quietly makes the chart untrustworthy.

Green on the full matrix, web build included. Merging.

PR #1836, opened 2026-08-22
fix(agent-targets): stop re-rendering .mcp.json when the merge refuses

### LINKED ISSUE #1603: [Bug] init writes project-local .mcp.json / .vscode files with no opt-out, and reformats unrelated entries in a tracked file
## Describe the Bug

`repowise init` writes project-local editor files — including **`.mcp.json`, which is a tracked file in many repos** — and there is no flag that opts out of it. `--no-editor-setup` and `REPOWISE_SKIP_EDITOR_SETUP=1` both cover only machine-wide registration; `--no-claude-md` and `--no-codex` cover their own files. Nothing covers `.mcp.json` or `.vscode/`.

Two distinct problems:

1. **Unrequested writes to a tracked file.** `.mcp.json` is modified and `.vscode/mcp.json` + `.vscode/extensions.json` are created, with no flag available to prevent it.
2. **Unrelated content is reformatted.** The whole of `.mcp.json` is re-serialised, so pre-existing, unrelated server entries come back with different formatting. In my case an existing SonarQube entry had its `"-e", "VAR"` argument pairs expanded onto separate lines, producing diff noise in a file repowise had no reason to touch beyond appending its own entry.

## Steps to Reproduce

```bash
cd some-repo-with-a-tracked-mcp-json
git status --short          # clean
export REPOWISE_SKIP_EDITOR_SETUP=1
repowise init --no-prose --no-claude-md --no-agents --no-codex \
               --no-distill-hook --no-editor-setup --no-onboarding -y
git status --short
```

## Expected Behavior

With `--no-editor-setup` *and* `REPOWISE_SKIP_EDITOR_SETUP=1` set, I expected no editor/agent integration files to be written at all — or, failing that, a documented flag to decline the project-local ones, consistent with `--no-claude-md` and `--no-codex`.

If `.mcp.json` is written, I'd expect a minimal edit that appends the `repowise` server and leaves unrelated entries byte-identical.

## Actual Behavior

```
 M .mcp.json
?? .vscode/extensions.json
?? .vscode/mcp.json
```

`.mcp.json` diff contains both the new `repowise` server (expected, had I asked for it) and reformatting of an unrelated pre-existing entry:

```diff
-        "-e", "SONARQUBE_TOKEN",
-        "-e", "SONARQUBE_ORG",
-        "-e", "SONARQUBE_URL",
+        "-e",
+        "SONARQUBE_TOKEN",
+        "-e",
+        "SONARQUBE_ORG",
+        "-e",
+        "SONARQUBE_URL",
```

## Environment

- OS: Windows 11 Pro (10.0.22631)
- Python version: 3.13.7
- Repowise version: 0.42.0
- Installation method: pip (venv)

## Additional Context

The `--editor-setup` help text does say "Project-local files are unaffected (see `--no-claude-md`, `--no-codex`)", so the machine-wide/project-local split is intentional and documented. The gap is that the referenced escape hatches don't cover `.mcp.json` or `.vscode/` — so a user who has explicitly declined every integration they know how to decline still gets a modified tracked file.

Suggestions, in rough order of value:

1. A `--no-mcp-config` / `--no-vscode` flag (or fold both into `--no-editor-setup`, which reads as the natural home for them).
2. Make the `.mcp.json` edit minimal — parse, insert the one key, preserve existing formatting — so it doesn't produce diff noise in unrelated entries.
3. Print the project-local files that will be written before writing them, since some are tracked and will show up in the user's next commit.

sloemo01:
Part of #1603

## What

When `repowise init` wrote the repo-shared, tracked `.mcp.json`, it re-serialised the whole document from a dict. Pre-existing, unrelated server entries (e.g. a SonarQube entry with compact `["-e", "SONARQUBE_TOKEN"]` args) came back reformatted, one arg per line, in a file repowise had no business re-rendering end to end.

This makes the `.mcp.json` write a minimal edit: only the `mcpServers.repowise` value and its separator are touched, so every other byte of the file survives. This is the no-reformat half of #1603 only. It does not add an opt-out, and the issue stays open for that.

## The refusal that was not honoured

`merge_json_object_member` returns `KEPT` when the document's shape is one it cannot edit safely: `mcpServers` present but not an object, or an insertion point it cannot resolve. `KEPT` means "leave this file alone."

The caller treated it as a prompt to re-render instead:

```python
action = merge_json_object_member(config_path, "mcpServers", "repowise", new_entry)
if action is FileAction.KEPT:
    action = write_json_config(config_path, {"mcpServers": {"repowise": new_entry}})
```

That writes the whole document from a dict, so a `.mcp.json` holding `"mcpServers": null` or `[]` (not exotic for a hand-edited or partially-migrated file) lost every other server in it and every top-level sibling key. The old code did not have this hole.

`KEPT` is now returned to the caller with a reason, alongside the `written.reason` the target was already discarding when recording the project-scope result.

## Test Plan
- `test_project_install_keeps_a_non_object_mcp_servers_entry` pins the refusal: the file is byte-identical afterwards and the row carries a reason.
- `test_project_install_does_not_drop_other_servers` pins the destructive case: a sibling key survives.
- Both fail with the re-render fallback restored, and the first also fails when `written.reason` is dropped again.
- 178 tests pass in `test_agent_targets.py`; `ruff check` clean.

### DIFF
--- a/packages/cli/src/repowise/cli/agent_targets/formats/json_merge.py
+++ b/packages/cli/src/repowise/cli/agent_targets/formats/json_merge.py
@@ -221,6 +221,270 @@ def write_json_config(path: Path, data: dict) -> FileAction:
     return action
 
 
+# ---------------------------------------------------------------------------
+# Surgical (minimal-edit) JSON merge
+#
+# ``write_json_config`` renders the whole document from a dict, which is the
+# right call for config we own. It is the wrong call for a tracked, repo-shared
+# file such as the root ``.mcp.json`` that other tools also write into: the
+# file is parsed to a dict and re-serialised end to end, so pre-existing,
+# unrelated server entries come back reformatted (issue #1603). The function
+# below performs a *minimal edit*: only the bytes of the entry we own are
+# touched, everything else in the file stays byte-identical.
+# ---------------------------------------------------------------------------
+
+
+def _skip_ws(text: str, i: int, end: int) -> int:
+    while i < end and text[i] in " \t\r\n":
+        i += 1
+    return i
+
+
+def _find_object_member(
+    text: str, obj_start: int, obj_end: int, key_target: str
+) -> tuple[int, int, int, int] | None:
+    """Locate a ``"key": value`` member inside the object spanning *text*.
+
+    *obj_start* points at the ``{`` and *obj_end* just past its ``}``. Returns
+    ``(key_start, key_end, value_start, value_end)`` for the member whose key
+    equals *key_target*, or ``None`` when the key is not present. Strict JSON
+    only — this is only ever called after :func:`load_json_object` has already
+    validated the document, so the scan is free to be lax about errors.
+    """
+    decoder = json.JSONDecoder()
+    i = obj_start + 1
+    while True:
+        i = _skip_ws(text, i, obj_end)
+        if i >= obj_end or text[i] == "}":
+            return None
+        if text[i] == ",":
+            i += 1
+            continue
+        key_start = i
+        try:
+            key, kend = decoder.raw_decode(text, i)
+        except json.JSONDecodeError:
+            return None
+        i = _skip_ws(text, kend, obj_end)
+        if i >= obj_end or text[i] != ":":
+            return None
+        i = _skip_ws(text, i + 1, obj_end)
+        try:
+            _, vend = decoder.raw_decode(text, i)
+        except json.JSONDecodeError:
+            return None
+        if key == key_target:
+            return key_start, kend, i, vend
+        i = _skip_ws(text, vend, obj_end)
+
+
+def _container_indent(text: str, obj_start: int, obj_end: int) -> str:
+    """Return the whitespace prefix indenting the object's members.
+
+    Derived from the whitespace preceding the matching closing brace, so a
+    member inserted into the object is indented to sit with its siblings no
+    matter the file's formatting. Falls back to ``"  "`` for one-line objects.
+    """
+    i = obj_end - 1
+    j = i
+    while j > obj_start and text[j] in " \t\r\n":
+        j -= 1
+    if j != i:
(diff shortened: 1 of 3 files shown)


Ayush7614 (approved):
Verified: `packages/cli/src/repowise/cli/agent_targets/formats/json_merge.py:221` adds surgical merge_json_object_member that upserts only mcpServers.repowise preserving unrelated entries byte-identical, with env preservation and idempotency, fixing #1603 reformatting bug. claude_code.py now uses it. Tests cover byte-identical, idempotent, env. Solid minimal-edit fix — LGTM.

sloemo01:
Fallback fixed, and the title no longer claims the opt-out.

You were right on both counts. There is no opt-out in the diff, so the title and body now describe what it actually does: the no-reformat half of #1603, `Part of #1603` rather than `Fixes`.

On the destructive fallback, your reading of it is exact. `KEPT` is the sentinel for "I cannot edit this safely", and the caller answered it by writing the whole document from a dict:

```python
if action is FileAction.KEPT:
    action = write_json_config(config_path, {"mcpServers": {"repowise": new_entry}})
```

For `"mcpServers": null` or `[]` that drops every other server and every top-level sibling key, which is the file `KEPT` exists to protect. The refusal is now returned to the caller with a reason instead.

One more instance of the same shape, which I only found because of your note: the target's `install` was calling `result.record(written.path, written.action)` for project scope and dropping `written.reason`. `codex.py` passes all three. So even a correctly-returned refusal reached the log as a bare `kept` with no explanation, which is the thing `FileWrite`'s docstring says a `KEPT` must never be.

Two tests, both failing with the fallback restored:

- `test_project_install_keeps_a_non_object_mcp_servers_entry` — the file is byte-identical afterwards, and the row carries a reason. Also fails if `written.reason` is dropped again.
- `test_project_install_does_not_drop_other_servers` — a sibling top-level key survives.

178 tests pass in `test_agent_targets.py`, `ruff check` clean.

PR #1837, opened 2026-08-22
fix(ingestion): make the per-file git-index timeout generous and configurable

### LINKED ISSUE #1781: [Feature] prevent premature timeout during long-running distill/cmake builds
## Problem

When executing long-running compilation commands (such as running cmake via distill on large repositories for the first time), the agent incorrectly assumes the process is stuck. As a result, it prematurely terminates the execution before the build can finish.

## Proposed Solution
To prevent the agent from prematurely terminating long-running compilation processes, we propose introducing a tracking mechanism via a unique reference ID and an Model Context Protocol (MCP) command.

### On-Demand Progress Checks:
Implement an MCP command that uses a unique Reference ID to poll the running process and return the last 5 to 10 lines of the build log. This allows the agent to actively verify that compilation is still active.

### Automated Heartbeat Logging:
Configure the build process or the agent to automatically stream a small buffer (e.g., the last 4 to 5 lines of output) every 30 seconds. This continuous feedback loop will signal to the agent that the task is progressing normally and prevent an artificial timeout. 

## Alternatives Considered
I updated agents.md to instruct the agent to increase its internal timeout or explicitly wait longer before assuming a process has frozen. However, this is a passive workaround and does not provide the active feedback needed for exceptionally large repositories.

sloemo01:
## Summary

The per-file git-indexing timeout was a hardcoded 45s, so a file whose log/blame/rename subprocess was legitimately still working got recorded as partial data. That budget is thin for a first-time full index of a large repository.

## Changes
- Raise the per-file git-index timeout default 45s → 120s.
- Make it configurable via `REPOWISE_GIT_INDEX_TIMEOUT_S` (positive float seconds), matching the existing `REPOWISE_*_TIMEOUT_S` resolver pattern; invalid values fall back to the default.
- Regression tests for the default, the env override, and malformed values.

## Scope
This changes `_FILE_INDEX_TIMEOUT_SECS` in the git indexer. It does not touch `distill`, which has no timeout of its own; #1781 stays open as the design question for what `distill` should emit while a long build runs.

## Test Plan
- `ruff check` passes on changed files.
- New tests pass; related git-indexer tests pass (no regression).

### DIFF
--- a/packages/core/src/repowise/core/ingestion/git_indexer/_constants.py
+++ b/packages/core/src/repowise/core/ingestion/git_indexer/_constants.py
@@ -7,6 +7,8 @@ the commit-depth defaults, decay half-lives, and skip heuristics.
 from __future__ import annotations
 
 import contextlib
+import math
+import os
 import re
 from typing import Any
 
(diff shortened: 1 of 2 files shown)


Ayush7614 (approved):
Verified: `packages/core/src/repowise/core/ingestion/git_indexer/_constants.py:385` raises per-file git index timeout 45s→120s and makes it configurable via REPOWISE_GIT_INDEX_TIMEOUT_S with finite/positive validation (#1781). Prevents premature partial data on large first-time indexes. Tests for default/env/bad values pass. LGTM.

sloemo01:
Done, all three:

- Retitled to `fix(ingestion): make the per-file git-index timeout generous and configurable`.
- Dropped `Closes #1781`. The body now says explicitly that this does not touch `distill`, and that #1781 stays open as the design question for what `distill` should emit while a long build runs. Removed the cmake/distill sentence from the summary and from the test module's docstring.
- On the justification: I do not have a case where the 45s budget actually fired on a real repo. You are right that the argument was "45s could be too short", which is a guess. I've left that claim out rather than dress it up.

The subsystem point landed: `_FILE_INDEX_TIMEOUT_SECS` in `git_indexer` and `distill` are unrelated, and I had them conflated because the issue title mentioned a long cmake build.

PR #1841, opened 2026-08-22
fix(health): recognize Ruby <stem>_spec.rb test pairing

### LINKED ISSUE #1768: [Bug] Ruby's <stem>_spec.rb is missed by both test-pairing heuristics, so every RSpec repo reads as untested
**Summary**

Ruby's standard test-file layout is `<stem>_spec.rb` under `spec/`, and neither
of the two "does this file have a test?" heuristics matches it. So an RSpec repo
with full spec coverage reports `test_gap: true` from `get_risk` and
`has_test_file: false` from `get_health`, for every source file.

**Where**

`packages/core/src/repowise/core/analysis/health/engine.py:276-299`, the
candidate set in `_has_paired_test_file`, has `.spec.ts`, `.spec.js`,
`.spec.mts` and `.spec.cts`, all with a dot before `spec`. Ruby uses an
underscore, so `user_spec.rb` is not a candidate.

`packages/server/src/repowise/server/mcp_server/tool_risk/assessment.py:229-233`
builds `%{base}.spec.{ext}%` from the target extension, which for `user.rb`
gives `%user.spec.rb%`. That is not a Ruby convention either, so the pattern
that looks like it should cover Ruby covers nothing.

**Repro**

```
lib/user.rb
spec/user_spec.rb
```

Both heuristics answer "no test". Checked against both rules directly:

```
src/user.rb   spec/user_spec.rb   health=False  risk=False
```

**Expected**

`<stem>_spec.<ext>` alongside the existing `<stem>_test.<ext>` spelling. Crystal
uses the same `_spec.cr` convention and has a language spec in the tree, so it
comes along for free.

**Impact**

`has_test_file` gates the `untested_hotspot` biomarker, which carries a high
severity weight, so a Ruby or Crystal repo has every hotspot penalised for
missing tests it actually has. The graph-reachability tier in `_check_test_gap`
can rescue some files on the risk side, but health has no such tier for this
flag, and neither has one when call resolution does not reach.

**Done looks like**

`f"{stem}_spec{p.suffix}"` in the shared candidate set, and a test pinning
`lib/user.rb` + `spec/user_spec.rb` as paired.

Note: this is deliberately scoped as a sibling of the `<stem>_test.<ext>`
generalisation being discussed on #1750, and should land after it so the two
do not conflict in the same candidate set. #1740 has the background on why
there were two rules at all.

sloemo01:
Fixes #1768

## Problem
Ruby's standard test layout is `<stem>_spec.rb` under `spec/` (underscore before `spec`, not a dot). Neither test-pairing heuristic recognized it:
- `_has_paired_test_file` in `engine.py` only had `.spec.ts/.spec.js/.spec.mts/.spec.cts` (dot form), so every RSpec repo read as `has_test_file: false` from `get_health`.
- The risk tool (via `_check_test_gap`) delegates to the same shared heuristic, so it also reported `test_gap: true`.

## Fix
Added `{stem}_spec{ext}` (underscore form) alongside the existing `{stem}_test{ext}` and dot-form `.spec.*` candidates. This matches Ruby `user_spec.rb` and, because Crystal shares the `_spec.cr` convention, comes along for free.

Note: the `%{base}.spec.{ext}%` builder previously in the risk tool was already refactored to delegate to the shared `_has_paired_test_file`, so the single engine.py change covers both `get_health` and `get_risk`.

## Tests
- `test_paired_test_basenames.py`: added `test_ruby_and_crystal_underscore_spec_pairing`; oracle kept in lockstep.
- `test_risk.py`: extended the suffix-convention parametrize with `lib/user.rb -> spec/user_spec.rb` and `src/user.cr -> spec/user_spec.cr`.

`uv run ruff check` passes; the affected tests plus the health/persistence suites pass.

### DIFF
--- a/packages/core/src/repowise/core/analysis/health/engine.py
+++ b/packages/core/src/repowise/core/analysis/health/engine.py
@@ -254,6 +254,7 @@ def _has_paired_test_file(rel_path: str, path_basenames: set[str]) -> bool:
     candidates = {
         f"test_{stem}.py",
         f"{stem}_test{test_suffix}",
+        f"{stem}_spec{test_suffix}",
         f"{stem}.test.ts",
         f"{stem}.test.tsx",
         f"{stem}.test.js",
--- a/packages/core/src/repowise/core/analysis/health/engine.py
+++ b/packages/core/src/repowise/core/analysis/health/engine.py
@@ -252,7 +252,7 @@ def _has_paired_test_file(rel_path: str, path_basenames: set[str]) -> bool:
     stem = p.stem
     test_suffix = ".exs" if p.suffix == ".ex" else p.suffix
     candidates = {
-        f"test_{stem}.py",
+        f"test_{stem}{test_suffix}",
         f"{stem}_test{test_suffix}",
         f"{stem}_spec{test_suffix}",
         f"{stem}.test.ts",
--- a/tests/unit/health/test_paired_test_basenames.py
+++ b/tests/unit/health/test_paired_test_basenames.py
@@ -25,6 +25,7 @@ def _reference_has_paired_test_file(rel_path: str, all_paths: set[str]) -> bool:
     candidates = {
         f"test_{stem}.py",
         f"{stem}_test.py",
+        f"{stem}_spec.rb",
         f"{stem}.test.ts",
         f"{stem}.test.tsx",
         f"{stem}.test.js",
@@ -104,7 +105,6 @@ class TestPairedTestBasenames:
         assert _path_basenames(set()) == set()
         assert not _has_paired_test_file("anything.py", set())
 
-
 class TestPascalPairing:
     """Delphi/FPC's ``u``-prefixed unit pairs with a standalone console test
     program named ``Test<Stem>.dpr`` (the ``u`` dropped) -- confirmed against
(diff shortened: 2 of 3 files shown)


Ayush7614 (approved):
Verified: `packages/core/src/repowise/core/analysis/health/engine.py:289` adds `{stem}_spec{ext}` candidate alongside dot-form, fixing Ruby <stem>_spec.rb (#1768) and Crystal _spec.cr pairing for _has_paired_test_file. Covers both get_health and get_risk via shared helper. Tests in test_paired_test_basenames + test_risk parametrize cover it. LGTM.

RaghavChamadiya (approved):
Thanks @sloemo01 for folding in the other half. `test_{stem}{test_suffix}` plus the `{stem}_spec{test_suffix}` candidate is exactly what was needed, and the parametrized cases in `test_risk.py` pin all three layouts (`user_spec.rb`, `user_spec.cr`, `test_user.rb`) so the Ruby minitest form is covered too.

One non-blocking note for later. `_reference_has_paired_test_file` in `tests/unit/health/test_paired_test_basenames.py` picked up `{stem}_spec.rb` but not `test_{stem}{test_suffix}`, so the oracle and `engine.py` now disagree on a Ruby `test_user.rb`. Nothing fails today because `CORPUS` has no `.rb` paths, but whoever adds one will get a confusing red. Worth squaring up next time that file is open.

Issue #1497, opened 2026-08-13
e2e specs are ingested but get no wiki page, so get_answer cannot answer test-infrastructure questions


alexander-mitelman:
## Summary

End-to-end test specs are ingested into the graph but never get a `file_page`, and `get_answer` searches pages — so any question about test infrastructure is unanswerable, with no signal that a whole area is missing.

## Numbers from our repo

TypeScript monorepo, 2,741 files ingested (`repowise init . --no-prose`):

```
wiki_pages where page_type='file_page'   2106      (of 2741 files ingested)
apps/*/e2e/** tracked in git               59
  ... of those, with a wiki page            0      <-- none
```

Ordinary `*.test.ts` files fare better but are also thinned: 135 of the first 200 have pages.

## What it looks like in use

Our Playwright setup is the most heavily documented convention we have — a React hydration gate that every console e2e spec depends on. Asking for it:

> "How do console e2e tests avoid clicking server-rendered markup before React has attached?"

`get_answer` returns `confidence: "low"` and five `fallback_targets`, none of them related (`apps/console/src/features/harvests/server/harvest.ts`, `apps/portal/src/shared/session-epoch.ts`, …). The two files that answer it — `apps/console/e2e/helpers/hydration.ts` and `apps/console/e2e/fixtures.ts` — have no page, so the retrieval cannot reach them. `git grep -il hydrat` finds them in one attempt.

The information is in the graph: `search_codebase` in `symbol` mode *does* return symbols from these files. Only the page layer is missing.

## Why this matters more than a coverage gap

Test files are where a codebase's hard-won conventions live — the gotcha comments, the "we tried X and it broke" notes, the fixtures everyone must use. That is exactly the material an agent most needs and least able to infer. And because the omission is silent, the answer that comes back isn't "I don't have that", it's five confident, wrong files.

## Ask

- If skipping e2e specs is deliberate, could it be surfaced (a note in `get_overview` / the init summary, or a `--include-tests` flag)?
- If it isn't, e2e specs seem worth a page — they're small, high-signal, and already ingested.

I didn't pass `--skip-tests`; the init summary reported only `2 by .gitignore, 1 by filename pattern, 18 oversized, 11 binary, 4 generated, 79 unknown type` as exclusions, which doesn't account for the ~635 file/page gap.

## Environment

repowise from `main` (post-#1454), `repowise init . --no-prose --yes`, pnpm workspace, 20 packages.

sloemo01:
Taking this

PR #1842, opened 2026-08-22
fix(generation): give e2e spec files a wiki page

sloemo01:
## What

End-to-end test spec files (`apps/*/e2e/**`, `*.spec.ts` under an `e2e/` tree, etc.) are **silently dropped from ingestion entirely**, so they never get a `file_page`. Because `get_answer` searches wiki pages, any question about test infrastructure comes back as five confident, unrelated files — with no signal a whole area is missing.

## Root cause

The `e2e` directory is in `_BLOCKED_DIRS` in `packages/core/src/repowise/core/ingestion/traverser.py`. It was grouped with `fixtures/` and `conftest/` as "browser-driven test rigs whose content rarely answers code questions", but unlike those it silently excludes **every file** under an `e2e/` directory before parsing — before a `file_page` could ever be selected. This is why the init summary reported only `2 by .gitignore, 1 by filename pattern, 18 oversized, 11 binary, 4 generated, 79 unknown type` as exclusions: the `e2e/` drop happened in the walker's blocklist, which the summary doesn't count.

## Fix

Remove `e2e` from `_BLOCKED_DIRS` so e2e specs are traversed and parsed like any other test file. They are stamped `is_test=True` via `is_test_related_path()` (the `e2e` token was already in `_TEST_DIR_TOKENS`), so:

- they get a `file_page` and are retrievable by `get_answer`, and
- every existing downstream filter (`skip_tests`, scoring's test penalty, test-dilution handling) still works exactly as before.

This matches how `tests/`, `spec/` and `__tests__/` are already treated — the traverser comment already documented that those directories are indexed and tagged `is_test=True`.

## Tests

- Added `test_traverses_e2e_specs` to `tests/unit/ingestion/test_traverser.py`: an `apps/console/e2e/**` tree is traversed and every file is tagged `is_test=True`.
- `uv run ruff check .` — all checks passed.
- `uv run pytest tests/unit/ingestion/ tests/unit/generation/ tests/unit/test_test_paths.py tests/unit/test_no_test_path_copies.py -q` — all pass (one unrelated pre-existing git-rebase sandbox failure excluded).

## Notes

The issue offered two acceptable resolutions: surface the skip (`--include-tests` flag / overview note) or just give e2e specs a page. Since these files are small, high-signal, and already meant to be indexed as tests (the `#`-comment above `_BLOCKED_DIRS` explicitly says test dirs are not blocked and are tagged `is_test=True`), making them actually indexed is the cleaner fix — no flag needed, and nothing is silently hidden.

### DIFF
--- a/packages/core/src/repowise/core/ingestion/traverser.py
+++ b/packages/core/src/repowise/core/ingestion/traverser.py
@@ -223,10 +223,20 @@ _BLOCKED_DIRS: frozenset[str] = frozenset(
         # is_test_related_path() so downstream consumers can filter them
         # when appropriate.
         #
+        # e2e/ used to sit here because Playwright/Cypress rigs were
+        # treated as browser-driven test scaffolding whose content rarely
+        # answers code questions. But end-to-end specs are exactly where a
+        # codebase's hard-won conventions live -- the hydration gate every
+        # console e2e test depends on, the fixtures everyone must use -- and
+        # silently dropping them left the wiki unable to answer any
+        # test-infrastructure question with no signal a whole area was
+        # missing. e2e files are now indexed like other tests and tagged
+        # is_test=True via is_test_related_path(), so they get a file_page
+        # and stay filterable when a consumer wants to exclude them.
+        #
         # The following ARE still blocked because they typically hold
-        # binary fixtures, generated artifacts, or browser-driven test
-        # rigs whose content rarely answers code questions:
-        "e2e",
+        # binary fixtures, generated artifacts, or scaffolding whose content
+        # rarely answers code questions:
         "fixtures",
         "conftest",
     }
--- a/tests/unit/ingestion/test_traverser.py
+++ b/tests/unit/ingestion/test_traverser.py
@@ -154,6 +154,35 @@ class TestFileTraverser:
         assert not any(p.startswith("Library/") for p in paths)
         assert not any(p.startswith("Temp/") for p in paths)
 
+    def test_traverses_e2e_specs(self, tmp_path: Path) -> None:
+        """e2e spec files are indexed so they can get a file_page (#1497).
+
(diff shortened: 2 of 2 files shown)


Ayush7614 (approved):
Verified: `packages/core/src/repowise/core/ingestion/traverser.py:223` removes e2e from _BLOCKED_DIRS so e2e specs are traversed and tagged is_test=True via is_test_related_path — fixes silent drop where file_page never created for test-infra questions. TestTraversesE2eSpecs covers apps/console/e2e tree. Matching handling to tests/spec. LGTM.

RaghavChamadiya (approved):
Thanks @sloemo01, and sorry this waited so long for a look.

This is the necessary half of #1497 and the reporter's own mechanism was one level off, which is worth recording. They wrote that e2e specs "are ingested but never get a file_page", and inferred the gap was in the page layer. It is not: `_BLOCKED_DIRS` is consulted during the directory walk itself (`traverser.py:622`), so an `e2e/` directory is never descended into and no page could ever exist for anything inside it. Taking `"e2e"` out of that set is the only change that can fix it, and no amount of work in selection would have.

The safety properties that make this comfortable to take are worth naming, since the obvious worry is a large Playwright suite flooding the wiki. These files still come back `is_test=True` via `is_test_related_path()`, so `--skip-tests` and every downstream filter keep working, `_PENALTY_TEST` still down-ranks them in scoring so production code wins under a page cap, and `max_file_pages` still bounds the total. They become eligible, not privileged.

Rewriting the comment rather than deleting the line is the right instinct too. That block is a record of why each entry is there, and a silent deletion would have left the next reader assuming e2e was an oversight.

Note that #2018 also targets #1497, from a different angle (the importance floor and the test penalty in selection, which is the `tests/`-directory half rather than the `e2e/` half). They are complementary rather than competing; I will say so on that thread.

PR #1843, opened 2026-08-22
refactor(cli): add shared warn() helper, unify warning stream

### LINKED ISSUE #1368: cli: add a shared warn() helper and unify console vs err_console warning stream
Residual from the #852 config-warning fix (PR `fix/852-config-warnings`). The `[yellow]Warning:[/yellow]` markup is now spelled at 5+ sites; the canonical site (`helpers.py:773`) prints to `err_console` while the new sites print to `console`. A shared `warn(text)` helper would stop the drift and unify the stream.

sloemo01:
Fixes #1368

Adds a shared `warn()` helper to `repowise.cli.helpers` that prints the yellow `Warning:` prefix to `err_console` (stderr), and routes every CLI warning site through it so warnings unify on the stderr stream instead of being split across `console`/stdout and `err_console`.

Sites updated:
- `helpers.py` — `_persist_provider_key` env-save failure, `resolve_provider` config validation warnings
- `worktree.py` — `--include-submodules` state conflict
- `reindex_cmd.py` — embed/skip failures
- `init_cmd/command.py` — local-provider concurrency warning
- `workspace_cmd.py` — indexing failure
- `mcp_cmd.py` — missing `.repowise` directory

Behavior is unchanged (still a yellow `Warning:` prefix), only the destination stream is normalized to stderr. Adds `TestWarn` unit tests asserting warnings land on stderr with the prefix and nothing leaks to stdout.

### DIFF
--- a/packages/cli/src/repowise/cli/commands/init_cmd/command.py
+++ b/packages/cli/src/repowise/cli/commands/init_cmd/command.py
@@ -46,6 +46,7 @@ from repowise.cli.helpers import (
     run_async,
     save_config_partial,
     save_state,
+    warn,
 )
 from repowise.cli.platform import telemetry
 from repowise.cli.providers import resolve_embedder
@@ -338,8 +339,8 @@ def _run_generation_phase(
     # Warn when a local provider runs with default concurrency
     local_providers = ("ollama", "codex_cli", "claude_cli", "opencode")
     if provider.provider_name in local_providers and concurrency > 4:
-        console.print(
-            f"  [{WARN}]Warning:[/] {provider.provider_name} is a local provider "
+        warn(
+            f"  {provider.provider_name} is a local provider "
             f"running with concurrency={concurrency}. "
             f"If you see timeout errors, try [bold]--concurrency 1[/bold]."
         )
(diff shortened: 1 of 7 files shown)


Ayush7614 (approved):
Verified: `packages/cli/src/repowise/cli/helpers.py:1` adds shared `warn()` → err_console (stderr) with yellow Warning prefix, and all sites routed through it (#1368). Unifies stream split (was console vs err_console). TestWarn asserts stderr + prefix, nothing on stdout. No behavior change, no extra code — LGTM.

RaghavChamadiya (approved):
Thanks @sloemo01, and thanks @Ayush7614 for the review with line citations back in August. Sorry this sat after that.

This closes #1368, and the part worth naming is the stream split rather than the duplication. `workspace_cmd.py:627` was printing its indexing-failure warning to `console` while the equivalents in `helpers.py` went to `err_console`, so one warning landed on stdout and the rest on stderr. Anyone piping workspace output into a file or another process was getting that one warning interleaved into their data, silently. Routing everything through a single `warn()` fixes a correctness problem, not just a consistency one.

Verified on current main that no `warn()` existed anywhere in `packages/cli` and that the same yellow `Warning:` markup was spelled out independently across seven modules.

PR #1849, opened 2026-08-22
fix(risk): normalize the git_metadata target path in get_risk

### LINKED ISSUE #1279: v0.37.0 get_risk returns hotspot_score=0, primary_owner=null despite data in wiki.db — column name mismatch
Version: 0.37.0
Repo type: Azure DevOps, .NET/C# monorepo

Issue: get_risk MCP tool returns hotspot_score=0, primary_owner=null, co_change_partners=[], trend="unknown", and risk_summary="no git metadata available" for files with rich behavioral data in wiki.db.

Root cause (confirmed via schema inspection): The MCP layer queries legacy column names that no longer exist in the v0.37.0 git_metadata table:

MCP queries	Actual column name
hotspot_score	temporal_hotspot_score
primary_owner	primary_owner_name
(not surfaced)	co_change_partners_json
Path resolution is not the issue — exact path match confirmed with COUNT(*) = 1. The row exists; the columns being queried do not.

Direct DB evidence for TimeSeriesContext.cs:

temporal_hotspot_score = 8.95
primary_owner_name     = 'Nikola.Glisic'
bus_factor             = 10
churn_percentile       = 0.9994
is_hotspot             = 1
co_change_partners_json: 15 entries (non-empty array)
MCP returns: hotspot_score=0, primary_owner=null, co_change_partners=[]

Workaround: Query wiki.db SQLite directly until fixed.

Also affected: bus_factor, churn_percentile, prior_defect_count are not exposed in the MCP response schema at all — these fields exist in git_metadata but have no corresponding keys in the get_risk return object.

sloemo01:
## Summary
Fixes #1279: `get_risk` returned `hotspot_score=0`, `primary_owner=null`, `co_change_partners=[]`, and a `"no git metadata available"` summary for files that do have rich `git_metadata` rows.

## Root cause
`_assess_one_target` looked up the `GitMetadata` row by **exact string equality** on `file_path` against the caller's raw target. The index stores `file_path` POSIX-relative (`src/auth/service.py`), but callers reach `get_risk` through git tools, shell completion, or editors that hand over:
- backslash form (`src\auth\service.py`, common on Windows),
- a leading `./`,
- a trailing separator,
- or a repo-absolute path (`/abs/repo/src/auth/service.py`).

Any of these made the lookup miss, so `meta` was `None` and `_assess_one_target` emitted the indistinguishable "no git metadata available" card — the exact payload the reporter saw. (The v0.37 column names were already correct; the lookup simply never found the row.)

## Change
- Add `normalize_target_path(target, repo_root)` in `assessment.py`: converts backslashes to `/`, strips a leading `./` and any leading slash, collapses duplicate/trailing separators, and makes a repo-absolute path relative to `repository.local_path`.
- In `_assess_one_target`, normalize the target once and key every file-path lookup (git_metadata, graph edges/nodes, test gap, security signals) on the normalized form, while keeping the response keyed by the caller's exact string.

## Tests
Added regression tests in `tests/unit/server/mcp/test_risk.py`:
- `test_get_risk_normalizes_target_path` — backslash, `./`, trailing-slash forms now populate `hotspot_score`/`primary_owner`/`co_change_partners`.
- `test_get_risk_repo_absolute_target_path` — repo-absolute form resolves.
- `test_normalize_target_path` — parametrized unit coverage of the helper.

Ruff + all `test_risk.py` / `test_risk_dependency_edges.py` tests pass.

### DIFF
--- a/packages/server/src/repowise/server/mcp_server/tool_risk/assessment.py
+++ b/packages/server/src/repowise/server/mcp_server/tool_risk/assessment.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 import contextlib
 import json
 from datetime import UTC, datetime
+from pathlib import Path
 from typing import Any
 
 from sqlalchemy import select, text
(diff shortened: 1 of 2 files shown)


Ayush7614 (approved):
Verified: `packages/server/src/repowise/server/mcp_server/tool_risk/assessment.py:1` adds `normalize_target_path` handling backslash, ./, trailing slash, repo-absolute → POSIX-relative, fixing #1279 where exact-string lookup returned hotspot_score=0. All lookups now use normalized path while response keeps raw target. Parametrized tests in `test_risk.py` cover forms. Correct fix — LGTM.

RaghavChamadiya (approved):
Thanks @sloemo01, the prefix strip is right and the comment above it explains why, so the next person will not reach for `lstrip` again. The parametrized cases carry the regression directly: `.github/workflows/ci.yml` and `.claude/TRIAGE.md` both come back with their leading dot intact.

Two other things worth calling out as good. Resolving against `repository.local_path` instead of the process cwd matters, since the MCP server does not run from the repo. And keeping the response keyed by the caller's exact string while only the lookup is normalized means nobody's client breaks on the key changing shape.

PR #1850, opened 2026-08-22
fix(cli): don't promise retry for config errors that cannot self-heal

### LINKED ISSUE #1372: cli: update degraded panel promises retry for config errors that cannot self-heal
Residual from the #852 config-warning fix (PR `fix/852-config-warnings`). `render_degraded` says "(will retry on the next update)", but an embedder degradation caused by a bad env var will not heal on retry — every future run re-warns. Reword the panel or split config errors into a separate channel.

sloemo01:
Closes #1372

## Summary
The degraded panel promised '(will retry on the next update)' for every best-effort step that failed during `repowise update`. That promise is only honest for transient failures (lock contention, a recovered server, a network blip) that genuinely re-run and heal on the next update. A config-driven step — the canonical case is the embedder with a bad `REPOWISE_EMBEDDER` value, a missing/bad API key, or an unreachable configured endpoint — reads the same config on the next run and fails identically, so promising a retry is a promise nothing will keep.

## Change

`render_degraded` in `update_cmd/reporting.py` now splits degraded entries into two channels:

- **Retryable** (transient) steps keep the honest '(will retry on the next update)' message.
- **Config-cannot-self-heal** steps (currently the embedder) surface with the actual remedy instead: *'Fix the config, then run `repowise reindex`'*.

A helper (`_split_degraded`) and a step-name allowlist (`_CONFIG_CANNOT_SELF_HEAL_STEPS`) keep the classification explicit and easy to extend.

## Test Plan

- `tests/unit/cli/test_update_reporting.py` — new `TestDegradedPanelRetryPromise` covering: retryable step still promises retry, a config embedder error says fix-config/reindex (and does *not* say 'will retry'), mixed entries split across both channels, and empty/None render nothing.
- `uv run pytest tests/unit/cli/test_update*.py` — 146 passed.
- `ruff check` + `ruff format --check` — clean.

### DIFF
--- a/packages/cli/src/repowise/cli/commands/update_cmd/reporting.py
+++ b/packages/cli/src/repowise/cli/commands/update_cmd/reporting.py
@@ -163,19 +163,67 @@ def render_degraded(degraded: list[str] | None) -> None:
 
     These used to be swallowed (``except Exception: pass``), so the update
     claimed clean success while, say, git metadata or graph nodes silently
-    stayed at the previous commit. The run still exits 0 — every listed step
-    is retried by the next update — but the panel must not say "complete"
-    without this block when something was skipped.
+    stayed at the previous commit. The run still exits 0, but the panel must
+    not say "complete" without this block when something was skipped.
+
+    Not every degraded step heals on the next update. A transient failure
+    (lock contention, a server that recovered, a network blip) re-runs and
+    fixes itself, so promising a retry is honest. A step that failed because
+    the *config* is broken — the canonical case is the embedder: a bad
+    ``REPOWISE_EMBEDDER`` value, a missing/bad API key, or an unreachable
+    endpoint the user configured — reads the same config on the next run and
+    fails identically, so telling the user "will retry on the next update" is
+    a promise nothing will keep. Those steps are surfaced as a config error
+    with the way to fix them instead.
     """
     if not degraded:
         return
-    console.print()
-    console.print(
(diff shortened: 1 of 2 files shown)


Ayush7614 (approved):
Verified: `packages/cli/src/repowise/cli/commands/update_cmd/reporting.py:1` splits degraded panel into retryable vs config-cannot-self-heal (`_CONFIG_CANNOT_SELF_HEAL_STEPS`) so bad REPOWISE_EMBEDDER/key no longer promises 'will retry on next update' — fix for #1372. Tests in `test_update_reporting.py` cover retryable/config/mixed/empty. Honest UX improvement, no extra code — LGTM.

RaghavChamadiya (approved):
Thanks @sloemo01. "Will retry on the next update" is the kind of line that costs trust slowly: a user who reads it and re-runs, and gets the identical panel, learns to distrust the whole report rather than that one sentence. Splitting on whether a retry can actually change the outcome is the right axis to split on.

The embedder is the right case to draw the line at, and the docstring says why in the terms a reader needs: a bad `REPOWISE_EMBEDDER`, a missing key, an unreachable endpoint. All three are read again identically on the next run. Naming `repowise reindex` in the config branch is what makes the message actionable, since a plain update would not re-embed even after the config is fixed.

Deriving the class from the step name before the first colon keeps `_split_degraded` a pure function over the strings that already exist, so nothing upstream has to start tagging its failures. That is the version of this that does not spread.

PR #1851, opened 2026-08-22
fix(workspace): persist generation-phase degradation so state.json is not clean

### LINKED ISSUE #1369: cli: init generation-phase embedder degradation can still go silent after a clean header probe
Residual from the #852 config-warning fix (PR `fix/852-config-warnings`). The init header probes `build_embedder` and warns on degradation, but the generation phase rebuilds the embedder; if that second build fails (e.g. local Ollama goes down mid-run), the run silently writes mock vectors with no warning. Thread the built instance or move the check into the shared generation entry (`init_cmd/generation.py`).

sloemo01:
## Summary

A workspace init that silently falls back to keyless vectors wrote a clean `state.json`. Semantic search was off with nothing in the record saying so, and `structlog` is pinned to ERROR unless `-v`, so a scripted run had no way to notice. The single-repo flow reports this through `run_warnings`, persisted as `degraded` (#2108); the workspace flow builds its own generation progress bar after the pipeline callback is gone, so nothing collected its warnings.

This is the workspace half only, reduced from an earlier three-hunk version per review. `command.py` and `generation.py` in that version were made redundant by #2108, which reports generation-phase degradation through the progress callback's warning channel, so both hunks are dropped here.

## Changes
- Thread a warnings list through `_run_workspace_generation` and `_run_workspace_deterministic_generation` into `run_repo_generation`, which #2108 already extends with the generation callback's warnings.
- Fold that list into `state.json` as `degraded` via `_fold_generation_warnings`, using `setdefault(...).extend(...)` so an entry an earlier phase recorded is not dropped.
- Tests for the fold, the extend-not-assign rule, and that `warnings` stays optional for callers that omit it.

## Scope
The workspace path only. Nothing here changes the embedder predicate: the earlier version's `isinstance(embedder_impl, KeylessEmbedder) and (requested_real or name == "ollama")` reduces to `name != "mock"`, and since #2108 owns that logic, this PR does not duplicate it.

## Test Plan
- 6 new tests; sabotage-verified (dropping either helper's `warnings=` fails its test; assigning instead of extending fails the fold test).
- 68 tests pass across the new file and its neighbours (`test_workspace_deterministic`, `test_workspace_cmd`, `test_workspace_init_provider`).
- `ruff check` passes on changed files.

### DIFF
--- a/packages/cli/src/repowise/cli/commands/init_cmd/command.py
+++ b/packages/cli/src/repowise/cli/commands/init_cmd/command.py
@@ -1453,6 +1453,13 @@ def init_command(
     # this run, so persistence/state below treat it as index-only. So does
     # `no_provider`: there is no model to do the work either way.
     effective_index_only = index_only or cost_declined or no_provider
+    # The generation phase rebuilds the embedder, and that second build can
+    # degrade even when the header probe was clean (issue #1369). The phase
+    # records it on ``result``; fold it into the run record so state.json's
+    # ``degraded`` list reports it just like a header-probe degradation would.
+    gen_embedder_degraded = getattr(result, "embedder_degraded", None)
+    if gen_embedder_degraded:
+        run_warnings.append(gen_embedder_degraded)
     print_phase_header(
         console,
         total_phases,
--- a/packages/cli/src/repowise/cli/commands/init_cmd/generation.py
+++ b/packages/cli/src/repowise/cli/commands/init_cmd/generation.py
@@ -50,6 +50,7 @@ from repowise.core.generation.models import (
     count_stub_fallbacks,
     is_stub_fallback,
 )
+from repowise.core.providers.embedding.base import KeylessEmbedder
 
 __all__ = [
     "COST_GATE_USD",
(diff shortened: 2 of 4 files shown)


Ayush7614 (approved):
Verified diff: `packages/cli/src/repowise/cli/commands/init_cmd/generation.py:1` records build-time degradation to `result.embedder_degraded` and folds into `state.json degraded` for both single-repo (`command.py`) and workspace (`workspace.py`) flows, as described for #1369. `mock` stays silent correctly. Dedicated regression tests in `test_init_failure_reporting.py` verify degraded vs healthy vs mock. Low risk, honest fix — LGTM, no fake content.

sloemo01:
Reduced it, keeping the workspace fix.

Dropped the `command.py` and `generation.py` hunks as redundant, and the PR is now workspace-only:

- Threads a warnings list through `_run_workspace_generation` and `_run_workspace_deterministic_generation` into `run_repo_generation`, which #2108 already extends with the generation callback's warnings, so the workspace fold rides the channel you described rather than a new attribute.

Both of your code notes were right and are fixed:

- **Assign vs append.** The old line was `state["degraded"] = [gen_embedder_degraded]`. It is now `state.setdefault("degraded", []).extend(gen_warnings)`, pulled into `_fold_generation_warnings` so the rule is directly testable. There is a test that fails if it goes back to assigning, with an entry an earlier phase recorded already in `state`.
- **The predicate.** Deleted with the hunk that held it, so the reduction to `embedder_name_resolved != "mock"` is moot here. #2108 owns that logic; this PR does not restate it.

The subsystem point I got wrong the first time: I read "long build" in #1781 and assumed the git-indexer timeout was the same thing. They are not related, and I should have checked before writing the summary around it.

Verified: 6 new tests, sabotage-checked. Dropping either helper's `warnings=` fails its test, and assigning instead of extending fails the fold test. 68 tests pass across the new file and `test_workspace_deterministic`, `test_workspace_cmd`, `test_workspace_init_provider`. `ruff check` clean.

sloemo01:
Checked both asks from the bot's contract notice, on the head of this PR (`c97026db9`). The notice is advisory and nobody had run them, so here is the evidence rather than an assumption.

**The 4 callers of the changed signature:** all pass.

```
tests/unit/cli/test_workspace_deterministic.py .....          (5 passed)
tests/unit/cli/test_workspace_degraded_reporting.py ......    (6 passed)
                                                        11 passed
```

`_run_workspace_deterministic_generation` is keyword-only, and the callers named in the notice pass `repo_path=`, `result=`, `embedder_name_resolved=`, `embedder_was_requested=` and the remainder by keyword, so the parameters this PR adds do not reach them.

**The files it flagged as importing the changed code:** all pass.

```
tests/integration/test_cli.py tests/unit/cli/test_augment_cmd.py
                                                        59 passed
```

One process note: my first attempt at this ran in a worktree sitting two commits behind the PR head, which is not the artefact the bot indexed. The numbers above are from `c97026db9`.

Issue #1370, opened 2026-08-09
cli: update deterministic page path and other embedder builds still degrade silently (R3 coverage)


SomSamantray:
Residual from the #852 config-warning fix (PR `fix/852-config-warnings`). R3 (warn on every command that requests an embedder and receives a degraded one) is implemented for init header, update's decision store, and search; `update_cmd/deterministic.py:171-175`, `reindex_cmd.py`, `generate_cmd/engine.py`, `upgrade_flow.py` and doctor's repo checks still build embedders without a degradation warning.

PR #1852, opened 2026-08-22
fix(cli): warn on silent embedder degradation in update paths

sloemo01:
## What

Closes the residual R3 gap from #852/#1370: embedder-degradation warnings were reported once **per call site** instead of once **per invocation**.

`repowise update` builds the embedder for its decision semantic-dedup store (`incremental.py`) *and* again for the deterministic page path (`deterministic.py`); a workspace update builds one per repo. When an embedder is unavailable, `build_embedder` printed the same full "could not be built" warning for **every** build, stacking the identical sentence multiple times in one run and burying the signal.

`build_embedder`'s degradation warning now fires **at most once per process** — matching the once-per-invocation gate the init header probe uses — so the first build of a run reports the degradation and later builds in the same invocation stay quiet.

## Why now

- `reindex`, `generate`, `upgrade_flow`, doctor's repo repair, and the deterministic page path all route through `build_embedder`, so this single gate covers all the R3 sites named in #1370 at once, including the `update` deterministic path.

## Change

- `packages/cli/src/repowise/cli/providers/embedders.py` — once-per-invocation gate (`_degradation_warned`) around the degradation warning, plus a `reset_degradation_warning()` hook so tests / embedded runners can re-arm it.
- `tests/unit/cli/test_embedder_resolution.py` — tests that three degraded builds in one invocation warn exactly once, and that the warning returns after a reset.
- `tests/unit/cli/conftest.py` — autouse fixture re-arms the gate between tests.

## QA

- `ruff check` and `ruff format --check` clean on changed files.
- CLI unit suite: 2191 passed (the 1 failure is a pre-existing platform-path baseline, unrelated).

### DIFF
--- a/packages/cli/src/repowise/cli/providers/embedders.py
+++ b/packages/cli/src/repowise/cli/providers/embedders.py
@@ -8,6 +8,27 @@ from typing import Any
 
 from repowise.cli.providers.keys import embedder_key_env_vars, resolve_embedder_api_key
 
+# Once-per-invocation gate for the degradation warning below. A single command
+# trips the fallback in several sites: ``repowise update`` builds the embedder
+# for the decision semantic-dedup store and again for the deterministic page
+# path, and a workspace run builds one per repo. Printing the full warning each
+# time stacks the same sentence over and over, which buries the signal. The
+# first fallback in a process reports it; later ones are just as degraded but
+# stay quiet, exactly as the init header probe does.
+_degradation_warned = False
+
+
+def reset_degradation_warning() -> None:
+    """Clear the once-per-invocation gate.
+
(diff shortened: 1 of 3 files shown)


Ayush7614 (approved):
Verified via diff: adds once-per-invocation gate `_degradation_warned` in `packages/cli/src/repowise/cli/providers/embedders.py:1` with `reset_degradation_warning()` for tests, and autouse re-arm in `conftest.py`. Fixes R3 gap #852/#1370 where `repowise update` built embedder twice (incremental + deterministic) and spammed identical warnings. Tests cover 3 builds warn once + reset re-arms. Clean, no fake/unneeded code — LGTM.

PR #1853, opened 2026-08-22
fix(workspace): reconcile listed repos with the configured (postgres) DB

### LINKED ISSUE #1034: [Bug] GET /api/workspace lists repos from the current directory, but doesn't check if they've already been indexed in the database when using postgresql
## Describe the Bug

When repowise is used with a postgres DB, the workspace will list repositories by walking the root directory, and ignore anything in the database

## Steps to Reproduce

1. Run `repowise init` with `REPOWISE_DB_URL` set to a postgres DB
2. Run `repowise serve` with the database
3. See that the dashboard tells you that there are repositories that have not been indexed, even though they have been and are in the database

## Expected Behavior

I expected repowise to take into account repositories indexed in the postgres DB, not just repos indexed with SQLite

## Actual Behavior

Only repos indexed with SQLite are returned

## Environment

- OS: Ubuntu 24.04
- Python version: 3.12.3
- Repowise version: 0.34.0
- Installation method: uv

## Additional Context

Here's an image showcasing how the repo list on the sidebar (loaded from `/api/repos`) shows two repositories, but the workspace says they're not indexed:

Image

Image

sloemo01:
## Summary

Fixes #1034.

When the server is configured with a PostgreSQL database (`REPOWISE_DB_URL` /
`REPOWISE_DATABASE_URL`), `GET /api/workspace` and `GET /api/workspace/graph`
listed repos by walking the root directory and reading a repo-local
`.repowise/wiki.db` SQLite file. Repos indexed into the configured (postgres)
DB keyed by `local_path` have no local `wiki.db`, so they were reported as
`needs_index`/`missing_dir` with zero stats — even though they were already
indexed.

## Change

`_query_repo_stats` and `_query_top_language` in
`packages/server/src/repowise/server/routers/workspace.py` now:

- prefer the repo-local SQLite `wiki.db` when present (unchanged behaviour),
- otherwise fall back to querying the configured session factory
  (`app.state.session_factory`) keyed by the repo's absolute `local_path`,
  aggregating the same tables (`repositories`, `graph_nodes`, `wiki_pages`,
  `git_metadata`, `health_file_metrics`) the SQLite path reads.

The helpers became `async`; both callers (`/api/workspace`, `/api/workspace/graph`)
were updated to `await` them and pass the session factory.

## Tests

- Added `test_configured_db_repo_reported_as_indexed` covering a repo indexed
  in the configured DB (no local `wiki.db`) being reported as `indexed` with
  its `repo_id`.
- Updated the existing `TestQueryRepoStats` unit tests for the now-async
  helpers.

### DIFF
--- a/packages/server/src/repowise/server/routers/workspace.py
+++ b/packages/server/src/repowise/server/routers/workspace.py
@@ -12,8 +12,12 @@ import sqlite3
 from pathlib import Path
 
 from fastapi import APIRouter, Depends, HTTPException, Query, Request
+from sqlalchemy import func, select
 
 from repowise.core.docs_mode import resolve_docs_mode
+from repowise.core.persistence import crud
+from repowise.core.persistence.database import get_session
+from repowise.core.persistence.models import GitMetadata, GraphNode, HealthFileMetric, Page
 from repowise.server.deps import (
     get_cross_repo_enricher,
     get_workspace_config,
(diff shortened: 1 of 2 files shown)


Ayush7614 (approved):
Verified: `pytest tests/unit/server/test_workspace_router.py -q` 45/45 passed including new `test_configured_db_repo_reported_as_indexed` (Postgres fallback without local `wiki.db`). Fixes #1034 — `_query_repo_stats` and `_query_top_language` now prefer repo-local SQLite then fall back to `session_factory` keyed by `local_path`, and `sync_workspace` checks the DB row before re-indexing (addressing @akshatmalik-bruh's point).\n\nThank you @akshatmalik-bruh for the thorough review — both observations were spot-on (session-factory consolidation vs dual path, and the `POST /api/workspace/sync` db_path.exists bug). @sloemo01 addressed the second point with a fix that keeps 45/45 green while leaving the larger consolidation as a clean follow-up as you both agreed. Great collaboration.\n\nLGTM — approved.

sloemo01:
Green now, so this is ready for the merge whenever you get to it.

All eight checks pass on `e2fe39711` (3.11 / 3.12 / 3.13 tests, ruff, Structurizr, VS Code extension, Web UI, tree-sitter floor; Integration tests skip on this path as usual). Still `BLOCKED` only because it needs your approval, not because anything is outstanding on my side.

I also owed you a reply here: the two API tests you asked about now pin the SQLite path per repository rather than the whole table, which is what caught the `sync_workspace` half in the first place.

sloemo01:
Rebased onto `main` and pushed; the branch was in `dirty` because of a real conflict.

One conflict, in `packages/server/src/repowise/server/routers/workspace.py`, and it was only the import block: `main` now imports the workspace read helpers from `repowise.core.workspace.reads` (#2530) where this branch adds the persistence imports it needs for the database fallback. Both sides are independent additions, so both are kept.

The rest of the file needed no hand-resolution. #2530 deleted the local `_contract_entry` / `_contract_link` projections in favour of `reads.contract_row` / `reads.link_row`, and the rebase takes that deletion; this branch never touched those helpers. `get_workspace_config` and every dependent route are intact.

Checked after the rebase:
- `tests/unit/server/test_workspace_router.py` -> 74 passed
- `tests/unit/server` + `tests/unit/workspace` -> 1689 passed
- `ruff check` on both changed files -> clean

The PR's own change is unchanged: same two files, `+241/-32`, and the database fallback path is byte-identical to before the rebase.

PR #1855, opened 2026-08-22
fix(health): exclude Rust inline test code from perf findings

NullSense:
## Summary

- The Phase-7b centrality gate (`perf.gated.collect_centrality_gated`) flags Rust test helpers as production perf findings. It decides a `hot_path_sync_io` / `nested_loop_quadratic` hit from only the walker's per-function facts and whether the file is hot/churny, and neither of those knows a function is test-only.
- Adds `FileComplexity.rust_test_line_ranges`, computed once from the tree `walk_file` already parses, and checks it at the one place both gated markers are produced.
- Measured on a real corpus: `hot_path_sync_io` findings went from 122 to 56. All 66 removed findings were test code; all 56 remaining findings were production code, unchanged.

## Related Issues

None.

## Test Plan

- [x] `uv run pytest tests/unit/health/test_perf_rust_test_filter.py -q` — 9 passed
- [x] `uv run pytest tests/unit/health/ -q` — 1213 passed
- [x] `uv run ruff check .`
  ```
  Change risk for main..HEAD: touches files that have broken before · 0th percentile of this repo's fix-bearing files
    fix(health): exclude Rust inline test code from perf findings
    +277 / -2 lines · 4 files · 3 dirs · 2 subsystems · entropy 1.36 · author exp 0  (fix)

  File                                                                    Lines   Prior fixes
  packages/core/src/repowise/core/analysis/health/complexity/walker.py      61            2.6
  packages/core/src/repowise/core/analysis/health/complexity/models.py      11            0.9

  Diff shape: Typical · 53rd percentile of recent commits by size and spread
  ```

## Checklist

## Out of scope

`perf.gated.collect_blocking_io_under_lock` (the cross-function lock-to-I/O pass) reads the same `PerfFnFacts.lock_call_targets` / `bare_sink_kind` and could plausibly hit the identical false-positive shape for a test helper that locks and calls into a sink. I didn't extend the filter there. No measurement backs it, and it's a separate cross-function reachability pass, not the same choke point. Worth a follow-up if it turns out to matter in practice.

---

### DIFF
--- a/packages/core/src/repowise/core/analysis/health/complexity/models.py
+++ b/packages/core/src/repowise/core/analysis/health/complexity/models.py
@@ -302,3 +302,14 @@ class FileComplexity:
     # flips a file from "untested" to "tested", so it can silence a finding
     # but never invent one.
     has_inline_tests: bool = False
+    # 1-indexed ``(start_line, end_line)`` spans of Rust test-only code: a
+    # ``#[cfg(test)]``-gated ``mod``/``impl`` (whole span, including any
+    # undecorated helper fns nested inside it) or a directly ``#[test]`` /
+    # ``#[tokio::test]`` / ``#[rstest]``-marked ``fn``. Computed once from the
+    # SAME parsed tree ``walk_file`` already builds — no extra parse. Rust-only
+    # (empty for every other language); the Phase-7b centrality gate
+    # (``perf.gated.collect_centrality_gated``) uses it to keep a
+    # ``PerfFnFacts.func_start`` line from ever emitting a ``hot_path_sync_io``
+    # / ``nested_loop_quadratic`` hit for inline test code the file-level
+    # ``is_test`` heuristic can't see.
+    rust_test_line_ranges: tuple[tuple[int, int], ...] = field(default_factory=tuple)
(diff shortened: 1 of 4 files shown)


Ayush7614 (approved):
Verified: `pytest tests/unit/health/test_perf_rust_test_filter.py -q` 9 passed, full `tests/unit/health/ -q` 1213 passed. Correctly adds `rust_test_line_ranges` via same-tree walk and gates `collect_centrality_gated` at the single choke point — measured 122→56 findings with only test-code hits removed. Shared attr helper `_eh_rust_attr_is_test` keeps perf and error-handling in sync, and span suppression avoids double-counting. LGTM.

RaghavChamadiya (approved):
Gating at `collect_centrality_gated` is the right place: it is the single point both `hot_path_sync_io` and `nested_loop_quadratic` pass through, so this is one fix instead of two that drift apart later. Computing the ranges from the tree the walker already parsed means it costs nothing extra.

The test I would have asked for is the one that is already here: `test_rust_test_line_ranges_cover_the_whole_mod_span` proves an undecorated helper nested inside a `#[cfg(test)] mod` is covered too. That is the actual bug, since a file-level flag could already silence a per-file marker while the per-function gate had no equivalent.

Returning early for any non-Rust language keeps this from touching other languages at all, which the parametrized test pins.

The earlier red on this branch was a flaky timing budget in `distill/`, nothing to do with your change. It is filed as #2372. Sorry for the delay, and for the four weeks before that.

PR #1856, opened 2026-08-22
fix(mcp): make local-endpoint budgets configurable

NullSense:
## Summary

- `tool_answer/config.py`: `_SYNTHESIS_MAX_TOKENS` (1024, hardcoded) now reads `REPOWISE_SYNTHESIS_MAX_TOKENS`. A reasoning model spends its whole budget on hidden thinking before emitting an answer token, so the call comes back empty at the old cap; measured against nanbeige, 1024 tokens returned nothing, 8192 produced a correct, cited answer.
- `_answer_pipeline.py` / `_helpers.py`: `_EMBED_TIMEOUT_S` (8.0, hardcoded) now reads `REPOWISE_EMBED_TIMEOUT_S` via a new `embed_timeout_s()`, mirroring the sibling `vector_search_timeout_s()`. A cold-loading local embedding model blows the old 8s budget; `get_answer` still exits 0 with the semantic leg silently dropped, visible only at WARNING.
- One PR because it's one root cause: constants tuned for a warm remote service, with no override, that degrade silently on a local one. Both follow the existing pattern for `REPOWISE_EMBEDDING_TIMEOUT` / `REPOWISE_VECTOR_SEARCH_TIMEOUT_S`: an unparseable or non-positive override warns and keeps the default.

## Related Issues

None.

## Test Plan

- [x] `uv run pytest tests/unit/ -q` — 14242 passed, 12 skipped, 2 xfailed. 2 failures, both pre-existing on unmodified `main` and unrelated to this change: a git-rebase test that depends on local git config, and a timezone-sensitive episode-dating test.
- [x] `uv run ruff check .`
  ```
  Change risk for main..HEAD: touches files that have broken before · 90th percentile of this repo's fix-bearing files
    fix(mcp): make local-endpoint budgets configurable
    +269 / -12 lines · 7 files · 3 dirs · 2 subsystems · entropy 2.19 · author exp 0  (fix)

  File                                                                       Lines   Prior fixes
  packages/server/src/repowise/server/mcp_server/tool_answer/config.py         36           11.7
  packages/server/src/repowise/server/mcp_server/_helpers.py                   36            6.4
  tests/unit/server/mcp/test_answer_synthesis_timeout.py                       47            5.8
  packages/server/src/repowise/server/mcp_server/_answer_pipeline.py           19            4.9
  packages/server/src/repowise/server/mcp_server/tool_answer/synthesis.py       6            2.8

  Diff shape: Typical · 57th percentile of recent commits by size and spread
  ```
  This flags real history: these are exactly the answer-pipeline files this fix targets, and they carry a documented run of prior fixes (`config.py` alone shows a recency-weighted 11.7). That matches the PR's own premise — this is a bug-prone area — rather than indicating a new problem in this diff; the full-suite result above is the direct check on that.

## Checklist

---

### DIFF
--- a/packages/server/src/repowise/server/mcp_server/_answer_pipeline.py
+++ b/packages/server/src/repowise/server/mcp_server/_answer_pipeline.py
@@ -64,7 +64,9 @@ from repowise.server.mcp_server._graph_files import (
     node_to_file,
 )
 from repowise.server.mcp_server._helpers import (
+    _EMBED_TIMEOUT_ENV,
     _VECTOR_TIMEOUT_ENV,
+    embed_timeout_s,
     vector_search_timeout_s,
 )
 from repowise.server.mcp_server._prose_symbols import symbol_backed_pages
@@ -132,10 +134,11 @@ _PAGERANK_BIAS_MAX = 0.3
 # rank-3/4 hit (~3.0-3.5).
 _GRAPH_EXPAND_DAMPING = 0.7
 
-# Budget for embedding the question. The searches that used to embed inline were
-# bounded at 8s including the embed, so the round-trip keeps that ceiling now
-# that it happens on its own.
-_EMBED_TIMEOUT_S = 8.0
+# Budget for embedding the question is resolved live by embed_timeout_s()
+# (see _helpers.py) — REPOWISE_EMBED_TIMEOUT_S, falling back to the 8s a warm
+# hosted endpoint needs. The searches that used to embed inline were bounded
+# at that same 8s including the embed, so the round-trip keeps that ceiling
+# now that it happens on its own.
 
 # Which retrieval legs actually ran for the current question (finding A18).
 #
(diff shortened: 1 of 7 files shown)


Ayush7614 (approved):
Verified: `pytest tests/unit/server/mcp/test_answer_synthesis_timeout.py tests/unit/server/mcp/test_embed_timeout.py -q` 76 passed. Makes two hardcoded budgets configurable (`REPOWISE_SYNTHESIS_MAX_TOKENS` for reasoning models spending budget on hidden thinking, `REPOWISE_EMBED_TIMEOUT_S` via `embed_timeout_s()` mirroring `vector_search_timeout_s()`) with validate-and-warn fallback, correct 120s cap, and updated degraded message advertising the env var. Hotspot history (config.py 11.7 fixes) matches the bug-prone area being fixed — not a new risk. LGTM.

RaghavChamadiya (approved):
Thanks @NullSense, and sorry this waited since 22 August without a reply.

Mirroring the existing `REPOWISE_VECTOR_SEARCH_TIMEOUT_S` shape is what makes this easy to take: same resolver pattern, same upper cap, same fallback to the old constant on a blank or unparseable value, so nothing moves for anyone who does not set the variables. The parametrized bad-value tests cover that directly.

The part I appreciated is that `test_embed_timeout.py` drives the real `question_vector()` call site with a slow embedder instead of testing the resolver on its own, which is the version that would actually catch a regression.

PR #1857, opened 2026-08-22
feat(embedding): declare a model's width without sending dimensions

NullSense:
## Summary

- `OpenAIEmbedder` can only override an embedding model's width by also sending that width to the API as the `dimensions` parameter. That's correct for a Matryoshka model, but it breaks a model of non-default width that isn't Matryoshka-capable: the API rejects the request outright (e.g. nvidia/Nemotron-3-Embed-1B, 2048-wide, served OpenAI-compatible, errors with "does not support Matryoshka embeddings; dimensions must be unset").
- Adds `declared_dimensions` / `REPOWISE_EMBEDDING_DECLARED_DIMS`, a second override that sets the declared width without ever sending it. Precedence: `dimensions=` > `REPOWISE_EMBEDDING_DIMS` > `declared_dimensions=` > `REPOWISE_EMBEDDING_DECLARED_DIMS` > the `_DIMS` table > 1536. Only the first two are ever sent on the wire.
- The width-mismatch error now names whichever of the three sources actually chose the number, instead of always pointing at `_DIMS` or `REPOWISE_EMBEDDING_DIMS`.

## Related Issues

None.

## Test Plan

- [x] `uv run pytest tests/unit/test_persistence/test_openai_embedder.py -q` — 52 passed
- [x] `uv run pytest tests/unit/test_persistence/ tests/unit/test_providers/ tests/unit/server/mcp/test_embedder_resolution.py tests/unit/cli/test_embedder_key_resolution.py -q` — 416 passed
- [x] `uv run ruff check .` (the actual CI lint gate; `ruff format --check` isn't run in CI, and this PR's files are clean under it anyway)
  ```
  Change risk for main..HEAD: touches files that have broken before · 52nd percentile of this repo's fix-bearing files
    feat(embedding): declare a model's width without sending dimensions
    +212 / -36 lines · 2 files · 2 dirs · 2 subsystems · entropy 0.98 · author exp 0

  File                                                              Lines   Prior fixes
  packages/core/src/repowise/core/providers/embedding/openai.py       142           1.0
  tests/unit/test_persistence/test_openai_embedder.py                 106           1.0

  Diff shape: Typical · 48th percentile of recent commits by size and spread
  ```

## Why this shape, not a `_DIMS` table entry

The straightforward fix is hardcoding `"nemotron-embed-1b": 2048` in `_DIMS`. That covers exactly one deployment; every other non-Matryoshka model of non-default width hits the identical wall. The existing `dimensions=` design already treats width as something the operator declares for whatever model they're running, not something the code enumerates by name — `declared_dimensions` extends that same idea to the one case it didn't cover.

## Checklist

---

### DIFF
--- a/packages/core/src/repowise/core/providers/embedding/openai.py
+++ b/packages/core/src/repowise/core/providers/embedding/openai.py
@@ -22,7 +22,13 @@ Dimensions:
     text-embedding-ada-002  → 1536 dims
 
     REPOWISE_EMBEDDING_DIMS (or the ``dimensions`` arg) overrides the width and
-    is passed to the API so the returned vectors match. See ``OpenAIEmbedder``.
+    is passed to the API so the returned vectors match — for a model that
+    supports a variable (Matryoshka) width.
+
+    REPOWISE_EMBEDDING_DECLARED_DIMS (or the ``declared_dimensions`` arg)
+    overrides the width *without* passing it to the API, for a model whose
+    width is fixed and that rejects the ``dimensions`` parameter outright.
+    See ``OpenAIEmbedder``.
 """
 
 from __future__ import annotations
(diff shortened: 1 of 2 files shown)


Ayush7614 (approved):
Verified: `pytest tests/unit/test_persistence/test_openai_embedder.py -q` 52 passed. Correctly adds `declared_dimensions`/`REPOWISE_EMBEDDING_DECLARED_DIMS` as declare-only width (never sent as `dimensions` param) for non-Matryoshka models like Nemotron-3-Embed-1B that 400 on `dimensions`. Precedence `dimensions > REPOWISE_EMBEDDING_DIMS > declared_dimensions > REPOWISE_EMBEDDING_DECLARED_DIMS > _DIMS > 1536` matches spec, error messages now name the actual source, and the strict-endpoint regression test proves the param is never sent. LGTM.

RaghavChamadiya (approved):
Separating "this model's width is fixed and it rejects the parameter" from "reshape the output to this width" is the right distinction, and the docstring carries it well enough that the next person will not collapse the two again. The Nemotron case is a real one and there was no way to express it before.

What makes this safe to take is that it changes nothing unless someone sets it: the precedence falls through to the existing table and the 1536 default untouched, `dimensions` still wins when both are given, and the declared width feeds the same mismatch check that was already there instead of bypassing it. Folding the two near-identical parsing blocks into `_validate_dimensions` and `_parse_dimensions_env` leaves the file smaller than it was.

Sorry this sat since 22 August.

PR #1858, opened 2026-08-22
feat(embedding): apply asymmetric query/document prefixes to embed calls

NullSense:
## Summary

- The embedder protocol has one `embed()` method, called identically for indexing and for querying — nothing says which side a text is on, and no provider prefixes anything. Adds `kind: str = "query" | "document"` through `Embedder.embed()`, `VectorStore.embed_texts()`/`search()`/`search_many()`, and every store/provider that calls them, plus `resolve_embed_prefix()` reading `REPOWISE_EMBED_QUERY_PREFIX` / `REPOWISE_EMBED_DOC_PREFIX` (empty by default, so existing deployments are byte-identical).
- Measured locally (Nemotron-3-Embed-1B, ~1,200 wiki pages, 11 queries): hit@5 went from 3/11 (27.3%) bare to 7/11 (63.6%) with `query: ` / `passage: ` prefixes; hit@10 4/11 → 7/11; MRR 0.236 → 0.332.
- The three decision-dedup call sites (`find_duplicate_decision`, `find_related_decisions`, `find_related_decisions_many`) now pass `kind="document"` explicitly, so they keep embedding symmetrically once an asymmetric prefix is configured — left alone, they'd have silently picked up the new default `kind="query"` through the same `VectorStore.search()` real document search uses, breaking near-duplicate matching.

## Related Issues

None.

## Test Plan

- [x] `uv run pytest tests/unit/test_persistence/test_openai_embedder.py tests/unit/server/mcp/test_answer_question_embedding.py tests/unit/server/mcp/test_why_search_payload.py tests/unit/server/mcp/test_keyless_vector_leg.py tests/unit/server/mcp/test_retrieval_leg_visibility.py -q` — 92 passed (every file touched to fix a `kind`-signature break, plus the new prefix tests)
- [x] `uv run ruff check .`
- [ ] `uv run pytest tests/unit/ -q` (full suite) — **not verified**. A second commit on this branch (`test: accept kind= in every embed()/embed_texts() test double`) fixes `embed()`/`embed_texts()` test doubles in `tests/unit/cli`, `tests/unit/persistence`, and `tests/unit/server/mcp` whose signature didn't accept the new `kind` parameter, per that commit's own message; I have not independently re-run the full suite to confirm it's green afterward, so I'm not claiming that result here. No production code changes in that second commit.
  ```
  Change risk for main..HEAD: touches files that have broken before · 5th percentile of this repo's fix-bearing files
    test: accept kind= in every embed()/embed_texts() test double
    +490 / -54 lines · 25 files · 8 dirs · 2 subsystems · entropy 3.22 · author exp 0

  File                                                                   Lines   Prior fixes
  packages/server/src/repowise/server/mcp_server/tool_why.py                2            8.4
  packages/server/src/repowise/server/mcp_server/_answer_pipeline.py        4            4.9
  tests/unit/server/mcp/test_keyless_vector_leg.py                          3            2.9
  tests/unit/cli/test_embedder_resolution.py                                6            2.9
  tests/unit/server/mcp/test_embedder_resolution.py                         3            2.8

  Diff shape: Elevated · 68th percentile of recent commits by size and spread
    Riskier than most commits in this repo.
  ```
  `risk` flags this as elevated on diff shape (25 files, 8 dirs) — expected for a protocol-signature change that touches every caller, and consistent with the full-suite verification gap noted above rather than a new finding.

## The wrinkle: decision dedup

`find_duplicate_decision` / `find_related_decisions` / `find_related_decisions_many` compare decision text to other decision text, and their own docstring already states the invariant this depends on: "the stored vector and the query side always embed the same shape." Decision vectors are stored via `upsert_decision_vectors`, which calls `embed_texts()` at its `"document"` default, so the dedup lookup has to search at `"document"` too — but those three functions look up existing decisions through the same generic `VectorStore.search()` / `search_many()` real document search uses, whose default kind is `"query"`. All three call sites now pass `kind="document"` explicitly.

## Checklist

- [ ] All existing tests still pass

---

### DIFF
--- a/packages/core/src/repowise/core/analysis/decisions/semantic_match.py
+++ b/packages/core/src/repowise/core/analysis/decisions/semantic_match.py
@@ -357,7 +357,13 @@ async def find_duplicate_decision(
     if not query:
         return None
     try:
-        results = await store.search(query, limit=SEARCH_FETCH)
+        # kind="document": this is decision text compared against other
+        # decision text (near-duplicate matching), not a natural-language
+        # question against stored documents. Stored decision vectors always
+        # embed at the document prefix (see upsert_decision_vectors); an
+        # asymmetric-prefix embedder that got the default "query" here would
+        # search with the wrong framing and silently miss real duplicates.
+        results = await store.search(query, limit=SEARCH_FETCH, kind="document")
     except Exception:
         return None
 
@@ -397,7 +403,8 @@ async def find_related_decisions(
     if not query:
         return []
     try:
-        results = await store.search(query, limit=SEARCH_FETCH)
+        # kind="document" — same reasoning as find_duplicate_decision above.
+        results = await store.search(query, limit=SEARCH_FETCH, kind="document")
     except Exception:
         return []
     return _related_from_results(results, lo=lo, hi=hi, exclude_ids=exclude_ids, limit=limit)
@@ -452,7 +459,10 @@ async def find_related_decisions_many(
     try:
         # Empty query texts still occupy their slot (keeps results aligned)
         # but must not reach the embedder — some providers reject "".
-        all_results = await store.search_many([q or " " for q in queries], limit=SEARCH_FETCH)
+        # kind="document" — same reasoning as find_duplicate_decision above.
+        all_results = await store.search_many(
+            [q or " " for q in queries], limit=SEARCH_FETCH, kind="document"
+        )
     except Exception:
         return [[] for _ in items]
     if len(all_results) != len(items):
--- a/packages/core/src/repowise/core/persistence/vector_store/_base.py
+++ b/packages/core/src/repowise/core/persistence/vector_store/_base.py
@@ -195,15 +195,27 @@ class VectorStore(ABC):
         for page_id, text, metadata in items:
             await self.embed_and_upsert(page_id, text, metadata)
 
-    async def embed_texts(self, texts: list[str]) -> list[list[float]] | None:
+    async def embed_texts(
+        self, texts: list[str], *, kind: str = "document"
+    ) -> list[list[float]] | None:
         """Embed *texts* in batched embedder requests, without upserting.
 
         Lets a caller that needs the raw vectors (e.g. decision dedup, which
-        searches *and* upserts the same text) pay for one batched embedding
-        instead of one round-trip per item. Returns ``None`` when the backend
-        holds no embedder — callers must fall back to the per-item text APIs.
-        Chunked so a large input can't blow the embedder's per-request token
-        cap; each text is capped at :data:`EMBED_TEXT_MAX_CHARS`.
+        searches *and* upserts the same text — pass ``kind="document"``, the
+        default, so both directions embed identically) pay for one batched
+        embedding instead of one round-trip per item. Returns ``None`` when
+        the backend holds no embedder — callers must fall back to the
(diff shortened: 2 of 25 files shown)


Ayush7614 (approved):
Verified subset: 107 passed (`test_openai_embedder`, `test_answer_question_embedding`, `test_why_search_payload`, `test_keyless_vector_leg`, `test_retrieval_leg_visibility`, `test_embed_kind_prefix`).\n\nChange is valid and working: adds `kind='query'|'document'` through `Embedder.embed()` → `VectorStore.embed_texts/search/search_many` → all providers, with `resolve_embed_prefix()` reading `REPOWISE_EMBED_QUERY_PREFIX`/`REPOWISE_EMBED_DOC_PREFIX` (empty by default → byte-identical). Decision dedup call sites correctly pass `kind='document'` to keep symmetric embedding. Blast-radius flagged (25 files, 8 dirs) is expected for a protocol-signature change — no new issues spotted, handling of empty text and invalid kind looks correct. LGTM. Note: full `tests/unit` not re-verified in CI yet due to prior test-double fixes, recommend confirming full suite green before merge.

RaghavChamadiya:
Thanks @NullSense, and sorry all four of these sat since 22 August without a word from me. #1855, #1856 and #1857 are dealt with; #1856 and #1857 are merged and #1855 is waiting on a CI re-run for a flaky budget that is not yours (#2372).

Merging #1857 will have knocked this one to conflicting, since both touch `__init__`, `_resolve_dimensions` and `embed()` in `openai.py`. That is my sequencing and not a problem with your work. `git fetch origin && git rebase origin/main` should be mechanical, the two changes do different things to the same region.

On this one I want a second pass before it goes in, for one reason.

The defaults are right and I want to say so first: `kind="document"` on the embed path and `kind="query"` on the search path match what every existing caller actually does, so with the prefixes empty nothing moves. And catching the decision-dedup sites yourself, where decision text is compared to decision text and both sides must be `document`, is the kind of thing that would have been a silent recall bug six months from now if you had not.

The gap is what happens when someone turns it on. `resolve_embed_prefix()` is empty by default, so the feature is opt-in, but the moment an operator sets `REPOWISE_EMBED_QUERY_PREFIX` on a deployment that already has a populated vector store, every stored document vector was embedded without the document prefix while every new query carries the query prefix. The two sides stop sharing a framing and recall degrades quietly, with no error. Your own hit@5 numbers are the argument for how large that effect is: a measurement that moves 27% to 64% in the intended direction moves comparably in the wrong one.

Nothing in the PR tells that operator their index needs rebuilding first. I looked for an existing guard in `providers/embedding/` that would catch an embedding-config change and found none, so this is not a regression you are introducing, it is a new way to reach a failure the codebase has never had to detect.

What I would like before merging, and the first is the one that blocks:

1. Some way for an operator to find out. A note in the PR is not enough since nobody reads it at the point of failure. A warning when a prefix is set and the store already holds vectors embedded without one, or failing that a line in the doctor output, or at minimum documentation next to the env vars saying "set these, then reindex".
2. The full unit suite run once. Your test plan says plainly that you did not re-run it after the commit that threaded `kind=` through the test doubles, and I would not merge a signature change touching 25 files across 8 directories without that. I appreciate you saying so instead of implying it was green.

The design is right and I want this. Ping me when the rebase and those two are done and I will take another look.

PR #1859, opened 2026-08-22
fix(workspace): scope walk-budget counter to fixture paths

### LINKED ISSUE #1854: [Bug] Read counter in test_extraction_walk_budget is process-global, not scoped to the tree under test, and fails CI on unrelated PRs
## Summary

`tests/unit/workspace/test_extraction_walk_budget.py::test_each_file_is_read_once_per_repo` asserts an aggregate cap on file reads, but the counter it reads from patches `Path.read_text` and `Path.read_bytes` on the class, so it counts every read anywhere in the process for the duration of the test, not the reads contract extraction performed on the fixture repos.

The budget has two reads of headroom. Any three extra reads from anywhere fail the build.

This went red on PR #1846 with

```
FAILED tests/unit/workspace/test_extraction_walk_budget.py::test_each_file_is_read_once_per_repo
  - assert 19 <= (2 * 6)
```

on Python 3.12, with 3.11 and 3.13 green on the same commit. Rerunning the identical commit with no change passed. The fixture contains a `pyproject.toml` and a `package.json`, and that PR was the one that started reading manifests, so it looked like a direct hit and cost real time to rule out. It was not: on that test's path `run_contract_extraction` is called with no `workspace_index`, and the new code returns before reading anything.

## Mechanism

The counter fixture, `test_extraction_walk_budget.py:70-94`:

```python
    original_read_text = Path.read_text
    original_read_bytes = Path.read_bytes

    def counted_read_text(self, *args, **kwargs):
        counts["reads"] += 1
        return original_read_text(self, *args, **kwargs)
    ...
    monkeypatch.setattr(Path, "read_text", counted_read_text)
    monkeypatch.setattr(Path, "read_bytes", counted_read_bytes)
```

There is no filter on `self`. A read of any path, by any code, on any thread alive during the test, increments the same counter the assertion at line 131 checks.

The measured baseline on the two fixture repos is 10 reads:

- 8 source reads, the 4 files per repo that are in the extractors' combined extension set, times 2 repos
- 2 reads of `pyproject.toml`, one per repo, from `traverser.py:1101 _collect_console_scripts` by way of `_console_script_tables`

That second pair is not contract extraction re-reading a source file. It is a different subsystem reading a manifest, and it already consumes a fifth of the budget.

The denominator makes it tighter than intended. `2 * len(REPO_SOURCE)` is 12 because `REPO_SOURCE` has six entries, two of which are the manifests. The comment directly above the assertion says the opposite:

```python
    # 4 source files per repo are in the extractors' combined extension set
    # (api.py, db.py, client.ts, queue.js); the manifests are not source.
    assert counters["reads"] <= 2 * len(REPO_SOURCE)
```

So the budget counts manifests as source in the denominator, while the comment excludes them, and counts manifest reads in the numerator that the budget was never written about.

I could not identify what produced the nine extra reads on 3.12 specifically. That run's log is full of leaked `aiosqlite` worker threads from earlier tests still raising `RuntimeError: Event loop is closed` during this window, up to `Thread-2250`, so a leaked thread is the obvious suspect, but I have not proven any of them calls `Path.read_text`. The point is that with a process-global counter I should not have to know. Nothing outside the fixture repos is this test's business.

## Why it is worth fixing rather than raising

Raising the cap trades a flaky failure for a weaker guard. The test exists because five extractors each used to walk and re-read the tree, and the number it protects is small by design. A cap loose enough to absorb unrelated process noise is also loose enough to miss a second read creeping back into one extractor, which is the regression it was written for.

The instrument is what is wrong here, not the threshold.

## Options

Scoping the counter to paths under the workspace root under test is the smallest change and removes the entire class of contamination:

```python
def counted_read_text(self, *args, **kwargs):
    if _under_test(self):
        counts["reads"] += 1
    return original_read_text(self, *args, **kwargs)
```

I lean further, to asserting the invariant the test name actually claims. Count reads per resolved path and assert no file is read more than twice, rather than capping a sum. That catches a single file being re-read even when the total stays under budget, which the aggregate cannot, and it stops the manifest reads and the source reads sharing one pool.

Either way the failure message should name the distribution. `assert 19 <= (2 * 6)` gave nothing to reason about; a per-path count would have shown at a glance that the extra reads were not on fixture files at all.

`test_one_tree_walk_per_repo` at line 118 shares the fixture and has the same global-patch exposure through `FileTraverser.traverse`. Its surface is much smaller, since only repowise code constructs a `FileTraverser`, so I would scope it for consistency rather than because it is failing.

## Not in scope

The budget itself is right and should stay. One walk and one boundary detection per repo is the invariant worth protecting, and `test_one_tree_walk_per_repo` is a good test. The leaked `aiosqlite` connections visible in the 3.12 log are a separate problem and probably deserve their own issue.

akshatmalik-bruh:
## Summary

- Scoped `Path.read_text`, `Path.read_bytes`, and `FileTraverser.traverse` counter patches in `test_extraction_walk_budget.py` strictly to paths under `tmp_path`, eliminating contamination from process-global noise (e.g. leaked async worker threads).
- Replaced the loose aggregate read count assertion (`assert reads <= 2 * len(REPO_SOURCE)`) with a strict per-path assertion (`MAX_READS_PER_FILE = 1`) that surfaces exact over-read paths and read counts on failure.

## Related Issues

fixes #1854

## Test Plan

- [x] Tests pass (`pytest tests/unit/workspace/test_extraction_walk_budget.py`) — 8/8 passed
- [x] Full workspace suite passes (`pytest tests/unit/workspace/`) — 682/682 passed
- [x] Regression detection verified — injected synthetic extra reads and confirmed failure message lists offending paths and exact call counts

## Checklist

### DIFF
--- a/tests/unit/workspace/test_extraction_walk_budget.py
+++ b/tests/unit/workspace/test_extraction_walk_budget.py
@@ -45,6 +45,15 @@ REPO_SOURCE = {
     "svc_b/queue.js": "channel.sendToQueue('jobs', payload);\n",
 }
 
+# Source files in the extractor's combined extension set.  The manifests
+# (pyproject.toml, package.json) are not in this set: .toml/.json are not
+# among the extensions any extractor claims, so iter_source_files never reads
+# them via read_bytes.  pyproject.toml is still read once via read_text by
+# _collect_console_scripts during FileTraverser.traverse.
+_SOURCE_FILES = frozenset(
+    rel for rel in REPO_SOURCE if not rel.endswith((".toml", ".json"))
+)
+
 
 def _make_repo(root: Path) -> None:
     for rel, content in REPO_SOURCE.items():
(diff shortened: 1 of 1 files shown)


Ayush7614 (approved):
Verified: scoping `Path.read_text`/`read_bytes` and `FileTraverser.traverse` counters to `tmp_path` eliminates process-global contamination (leaked async workers), and per-path `MAX_READS_PER_FILE=1` gives actionable failure messages. `pytest tests/unit/workspace/test_extraction_walk_budget.py` 8/8 passed, full `tests/unit/workspace/` 682/682 passed. Only test changes, health gate passed. LGTM.

RaghavChamadiya (approved):
Thanks @akshatmalik-bruh. The real problem here was not the flake, it was that the old assertion could not distinguish the two things it was measuring. `counters["reads"] <= 2 * len(REPO_SOURCE)` is a total with slack in it, so one file read eight times passed as long as three others were never read at all, and any unrelated thread touching the filesystem spent the same budget. It was a leaky global counter wearing the shape of an invariant.

Per-path counting with `MAX_READS_PER_FILE = 1` asserts the property the test is named after. "Each file is read once per repo" is now literally what is checked, and the failure message names the offending paths with their counts, so a regression points at the extractor that caused it instead of at a number nobody can decompose.

Scoping to `tmp_path` is the right filter and the reasoning in the docstring is worth keeping. Patching `Path.read_text`/`read_bytes` at the class level is the only interception point that catches every call site, and that is exactly why it needs a path predicate: a leaked aiosqlite worker has no business in this tally.

The comment block explaining why `pyproject.toml` is read once and `package.json` is not read at all is the part I would have had to reconstruct otherwise. Tracing it to `_SKIP_GENERATED_CHECK` rather than saying "manifests are not source" means the next person can tell whether a changed count is a bug or a legitimate shift.

Non-blocking: `_SOURCE_FILES` is computed at module level but nothing reads it now that the assertion is per-path, so it can go.

PR #1865, opened 2026-08-22
fix(mcp): prefer most-specific workspace repo

### LINKED ISSUE #1818: [Bug] repowise mcp <path> serves the workspace primary, not the repo you pointed it at
## Summary

`repowise mcp <path>` does not serve the repo at `<path>`. In any workspace whose primary repo sits at `path: .`, the detection loop matches that entry for every descendant path, so the explicit argument is discarded and the server comes up serving the workspace primary.

The code already intends the opposite. `_server.py:392-394` reads:

```python
# Override default repo to the one the path points at
if ws_repo_alias and ws_config.get_repo(ws_repo_alias):
    ws_config.default_repo = ws_repo_alias
```

That override is correct and never gets a chance to do anything, because `ws_repo_alias` has already been resolved to the primary.

## Repro

This repository is a workspace with three members:

```yaml
# .repowise-workspace.yaml
default_repo: repowise
repos:
- path: .
  alias: repowise
  is_primary: true
- path: backend
  alias: backend
- path: frontend
  alias: frontend
```

Point the MCP server at `backend`, which is a declared member, and ask it what it is serving:

```
$ repowise mcp ./backend
# tools/call list_repos ->
"workspace_root": "C:/Users/ragha/Desktop/repowise"
"default_repo": "repowise"
```

Expected `backend`. Every unqualified tool call in that session answers about `repowise` instead. The workaround exists (`repo='backend'` on tools that take it) but nothing tells the caller they need it, and an agent has no reason to suspect the server is not serving what it was pointed at.

The same thing happens to a repo that is *not* a workspace member. I indexed `test-repos/microdot`, which has its own complete `.repowise/`, then ran `repowise mcp` against it:

```
$ repowise mcp test-repos/microdot
# get_context(targets=["src/microdot/microdot.py"]) ->
{"error": "Target not found: 'src/microdot/microdot.py'"}
# get_overview() -> "Repository Overview: repowise"
# list_repos() -> repos: [repowise, backend, frontend]
```

microdot is not in the registry, so its index is unreachable through MCP entirely. The server reports success throughout and answers confidently about a different codebase.

## Where

`packages/server/src/repowise/server/mcp_server/_server.py:346-361`:

```python
# Determine which repo the given path belongs to
resolved = _Path(repo_path).resolve()
repo_alias = None
for entry in ws_config.repos:
    entry_abs = (ws_root / entry.path).resolve()
    try:
        resolved.relative_to(entry_abs)
        repo_alias = entry.alias
        break
    except ValueError:
        continue

if repo_alias is None:
    # Path is inside workspace but doesn't match a repo — use default
    primary = ws_config.get_primary()
    repo_alias = primary.alias if primary else ws_config.repos[0].alias
```

`relative_to` is a containment test, not an identity test, and `(ws_root / ".").resolve()` is `ws_root` itself. Every path under the workspace is contained by it, so the `.` entry matches on the first iteration and `break` ends the search before any more specific member is considered. First match wins, and the broadest possible entry is first.

The `repo_alias is None` fallback is a red herring here. For the microdot case the comment says "path is inside workspace but doesn't match a repo", but that branch is never reached: the path did match, incorrectly, against `.`.

## Why this is not just a nested-repo edge case

The `backend` case is the one that matters. `path: .` for the primary is what `repowise init` produces for the ordinary "monorepo with siblings" layout, so any workspace built the normal way has a first entry that shadows every other member. Pointing MCP at a specific member is exactly what someone does when they want to scope an agent to one repo, and it silently does not work.

## No way to override

The other commands that can land in a workspace take `--no-workspace`: `ask`, `context`, `costs`, `dead-code`, `claude-md` and `agents` all have it. `mcp` does not. There is no flag, and no env var, that makes `repowise mcp <path>` serve just that path.

The help text also promises the behaviour that does not happen, at `mcp_cmd.py:148`:

```
repowise mcp /path/to/repo       # stdio, specific repo
```

- Prefer the most specific match instead of the first. Collect every entry that contains the path and take the longest `entry_abs`, so `backend` beats `.`. Smallest change, fixes the member case, leaves the non-member case falling back to the primary.
- Treat a path that resolves to its own indexed repo as single-repo mode, so a nested repo with its own `.repowise/` is served directly rather than being absorbed. Fixes the microdot case, and needs a decision about whether being inside a workspace should ever override an explicit argument.
- Add `--no-workspace` to `mcp` for parity with the other commands. Useful on its own and worth doing whichever of the above is chosen, since it is the only escape hatch when detection gets it wrong.

Refusing outright when the path is inside a workspace but is not a member is also defensible, and would at least be loud. What is not defensible is answering as a different repository without saying so.

cipher-v:
## Summary

- Make MCP workspace detection consider all containing repo paths and prefer the deepest match, so a declared member such as `backend` is not shadowed by the workspace root (`.`).
- Add real temporary-workspace regression tests covering declared members, the workspace root, and the current non-member fallback to the primary repo.

## Related Issues

Fixes #1818

## Test Plan

- [x] `uv run pytest tests/unit/server/test_mcp_workspace_detection.py tests/unit/server/test_mcp_workspace.py -q` — 33 passed
- [x] `uv run pytest tests/unit/server/ -q` — 2,359 passed
- [x] `uv run ruff check packages/server/src/repowise/server/mcp_server/_server.py tests/unit/server/test_mcp_workspace_detection.py`
- [x] Web build not applicable — no frontend changes

## Checklist

- [x] Documentation is not required for this internal bug fix

### DIFF
--- a/packages/server/src/repowise/server/mcp_server/_server.py
+++ b/packages/server/src/repowise/server/mcp_server/_server.py
@@ -346,14 +346,18 @@ def _detect_workspace(repo_path: str | None):
         # Determine which repo the given path belongs to
         resolved = _Path(repo_path).resolve()
         repo_alias = None
+        best_match_depth = -1
         for entry in ws_config.repos:
             entry_abs = (ws_root / entry.path).resolve()
             try:
                 resolved.relative_to(entry_abs)
-                repo_alias = entry.alias
-                break
             except ValueError:
                 continue
+
+            match_depth = len(entry_abs.parts)
+            if match_depth > best_match_depth:
+                repo_alias = entry.alias
+                best_match_depth = match_depth
 
         if repo_alias is None:
             # Path is inside workspace but doesn't match a repo — use default
--- a/tests/unit/server/test_mcp_workspace_detection.py
+++ b/tests/unit/server/test_mcp_workspace_detection.py
@@ -1,0 +1,61 @@
+"""Tests for MCP workspace member detection."""
+
+from pathlib import Path
+
+import pytest
+
+from repowise.server.mcp_server._server import _detect_workspace
+
+
+@pytest.fixture
+def workspace(tmp_path: Path) -> Path:
+    """Create a workspace with a root repo and two declared members."""
+    (tmp_path / "backend").mkdir()
+    (tmp_path / "frontend").mkdir()
+    (tmp_path / "test-repos" / "microdot").mkdir(parents=True)
+
+    (tmp_path / ".repowise-workspace.yaml").write_text(
+        "version: 1\n"
+        "default_repo: repowise\n"
+        "repos:\n"
+        "- path: .\n"
+        "  alias: repowise\n"
+        "  is_primary: true\n"
+        "- path: backend\n"
+        "  alias: backend\n"
+        "- path: frontend\n"
+        "  alias: frontend\n",
+        encoding="utf-8",
+    )
+    return tmp_path
+
+
+@pytest.mark.parametrize(
+    ("member_path", "expected_alias"),
+    [
+        ("backend", "backend"),
+        ("frontend", "frontend"),
+    ],
+)
+def test_detect_workspace_prefers_most_specific_member(
+    workspace: Path,
+    member_path: str,
+    expected_alias: str,
+) -> None:
+    ws_root, ws_config, repo_alias = _detect_workspace(str(workspace / member_path))
+
+    assert ws_root == workspace.resolve()
+    assert ws_config is not None
+    assert repo_alias == expected_alias
+
+
+def test_detect_workspace_root_uses_primary(workspace: Path) -> None:
+    _, _, repo_alias = _detect_workspace(str(workspace))
+
+    assert repo_alias == "repowise"
+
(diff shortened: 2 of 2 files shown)


Ayush7614 (approved):
Verified locally: fix correctly prefers deepest (most-specific) workspace repo match so `backend`/ `frontend` are not shadowed by `.` — `uv run pytest tests/unit/server/test_mcp_workspace_detection.py tests/unit/server/test_mcp_workspace.py -q` 33 passed, `ruff check` clean. New regression tests cover declared members, workspace root, and non-member fallback. Health gate passed (1.0→1.8). LGTM — no issues found.

RaghavChamadiya (approved):
Thanks @cipher-v. Ranking the containing entries by `len(entry_abs.parts)` instead of taking the first match is the right shape: every entry that matches is a genuine ancestor of the path, so the deepest one is the most specific by construction, and it stops the result depending on the order repos happen to be listed in the workspace file. The three tests pin the parts that matter, a declared member, the workspace root falling to primary, and a directory inside the workspace that is not a member.

PR #1877, opened 2026-08-23
fix(generation): stamp editor files with the indexed commit, not live HEAD

### LINKED ISSUE #1874: [Bug] Generated CLAUDE.md stamps live HEAD instead of the commit the index was built against
## Summary

The `Last indexed` stamp in a generated `CLAUDE.md` reports whatever `git HEAD`
was when the file was *written*, not the commit the index it describes was built
from. The two routinely disagree, nothing self-heals the drift, and every fact
under the stamp inherits it.

## Mechanism

`EditorFileDataFetcher.fetch` shells out at
`generation/editor_files/fetcher.py:73`:

```python
            indexed_commit=_get_head_short_sha(self._repo_path),
```

```python
def _get_head_short_sha(repo_path: Path) -> str:
    """Return the short SHA of HEAD, or empty string if not a git repo."""
    try:
        result = subprocess.run(
            ["git", "rev-parse", "--short", "HEAD"],
```

The value it wants already exists on the row it just loaded. `repo` is in scope
at line 65, and `Repository.head_commit` is documented for exactly this
(`persistence/crud/repository.py:52-58`):

> `head_commit` records the git commit the index was built against - the value
> the MCP `_meta` freshness check compares to the live HEAD.

## Why the drift is routine, not a corner case

A rebase fires the post-commit hook once per replayed commit. Each firing
rewrites `CLAUDE.md` with the then-current HEAD while the index underneath is
unchanged, so the stamp walks forward on its own. There is no path that pulls it
back.

This is also why the mismatch is easy to misread: the same commit before and
after a rebase has two SHAs, so the stamp and the index can look divergent when
only the stamp moved.

## Blast radius

Everything below the stamp is a static snapshot taken at the same moment and
carries the same drift, including the code-health "worst performer" line. The
stamp is the only thing claiming to date them, so if it is wrong they are all
silently misdated.

## Fix

Read the stored commit, falling back to the shell call only when the column is
empty (pre-backfill indexes):

```python
            indexed_commit=(repo.head_commit or "")[:7] or _get_head_short_sha(self._repo_path),
```

Worth confirming the short-SHA length matches what `git rev-parse --short`
produces on this repo, so the rendered stamp does not change width.

## Tests

The existing assertion only checks the `YYYY-MM-DD` shape of `indexed_at`, so it
survives untouched. A new test should set `head_commit` to a known value, leave
the working tree on a different commit, and assert the stamp reports the stored
one.

Exberg:
## Summary

- The "Last indexed" stamp in generated `CLAUDE.md` / `AGENTS.md` shelled out for the live `git HEAD` at write time instead of using `Repository.head_commit` — the commit the index was actually built against. A rebase fires the regeneration hook once per replayed commit, so the stamp walked forward on its own while the index underneath stayed unchanged, silently misdating every fact rendered below it.
- The stamp now reads the stored `head_commit` (shortened to 7 chars to match the width of `git rev-parse --short`) and falls back to the shell-out only when the column is empty (pre-backfill indexes).

## Related Issues

Fixes #1874

## Test Plan

- New regression test sets `head_commit` to a known full SHA, points the fetcher at a non-git directory (so any live-HEAD shell-out would resolve to `""`), and asserts the stamp reports the stored commit's short form.
- Companion test covers the fallback: with `head_commit` NULL the legacy shell-out path is preserved unchanged.
- Confirmed RED before the fix (`assert '' == 'a1b2c3d'`), GREEN after.

- [x] Tests pass (`pytest`) — full unit suite: 14,339 passed. One pre-existing failure in `tests/unit/cli/test_agent_target_baseline.py::test_integration_writes_match_baseline` reproduces identically on clean `main` (platform-dependent baseline: expects the Windows `AppData` path, gets macOS `Library/Application Support`); unrelated to this change.
- [ ] Web build passes (`npm run build`) *(if frontend changes)* — N/A, no frontend changes

## Checklist

### DIFF
--- a/packages/core/src/repowise/core/generation/editor_files/fetcher.py
+++ b/packages/core/src/repowise/core/generation/editor_files/fetcher.py
@@ -67,10 +67,18 @@ class EditorFileDataFetcher:
 
         kg_layers, kg_tour = await self._get_kg_data()
 
+        # The stamp must report the commit the index was built against, not
+        # the live HEAD: a rebase fires the regeneration hook per replayed
+        # commit while the index underneath is unchanged, so a live-HEAD
+        # stamp walks forward on its own and silently misdates everything
+        # below it. Fall back to the shell-out only when the column is empty
+        # (pre-backfill indexes).
+        stored_commit = ((repo.head_commit or "")[:7]) if repo else ""
+
         return EditorFileData(
             repo_name=repo_name,
             indexed_at=datetime.now(UTC).strftime("%Y-%m-%d"),
-            indexed_commit=_get_head_short_sha(self._repo_path),
+            indexed_commit=stored_commit or _get_head_short_sha(self._repo_path),
             architecture_summary=await self._get_architecture_summary(),
             key_modules=await self._get_key_modules(),
             entry_points=await self._get_entry_points(),
--- a/tests/unit/generation/test_editor_file_fetcher.py
+++ b/tests/unit/generation/test_editor_file_fetcher.py
@@ -372,3 +372,45 @@ async def test_fetch_indexed_at_is_date_string(session, repo, tmp_path):
     import re
 
     assert re.match(r"^\d{4}-\d{2}-\d{2}$", data.indexed_at)
+
+
+async def test_fetch_indexed_commit_reports_stored_head_commit(session, tmp_path):
+    """The stamp must report the commit the index was built against.
+
+    Regression test for #1874: the stamp used to shell out for the live HEAD,
+    which drifts from the indexed commit (e.g. a rebase rewrites CLAUDE.md
+    with the then-current HEAD while the index underneath is unchanged).
+    """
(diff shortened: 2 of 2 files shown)


RaghavChamadiya (approved):
Thanks @Exberg for writing down the mechanism in the code. The comment is the part that stops this regressing: the reason a live-HEAD stamp is wrong is not obvious from the line itself, and "a rebase fires the regeneration hook per replayed commit while the index underneath is unchanged" is exactly the sentence the next person needs.

Two other PRs arrived for #1874 with the same one-line change, so to say why this is the one I took: it is the only one that covers the pre-backfill case. `head_commit` can be NULL on an index built before that column was populated, and the `or _get_head_short_sha(...)` fallback exists precisely for that, so a test that pins it is what keeps the fallback from being deleted as dead later. Using a real 40-character sha and asserting on `[:7]` also tests the truncation rather than assuming it.

Full matrix green on the head commit. Merging.

PR #1881, opened 2026-08-23
fix: support trailing globs in path search

### LINKED ISSUE #1871: [Bug] search_codebase path mode cannot resolve a trailing glob, because escape_like does not strip it
## Summary

`search_codebase(query="tool_overview*", mode="path")` returns nothing. Every
other path query shape works: bare stem, partial path, extension, full path. The
asterisk is the only one that fails, and it fails silently as "no such file"
rather than as a rejected query.

## Mechanism

`search_paths_single` builds the LIKE pattern from the raw query at
`tool_search_symbols.py:274-284`:

```python
    qnorm = query.strip().lower().replace("\\", "/")
    ...
                Page.target_path.ilike(f"%{escape_like(qnorm)}%", escape=LIKE_ESCAPE),
```

`escape_like` (`persistence/sql.py:17`) handles the three characters SQL LIKE
cares about and nothing else:

```python
    return value.replace("\\", "\\\\").replace("%", "\%").replace("_", "\_")
```

`*` is not a LIKE metacharacter, so it is neither escaped nor translated. It
survives into the pattern as a literal asterisk, `%tool_overview*%` matches only
paths that literally contain one, and no path does.

## Why stripping rather than translating

The match is already `%...%` on both sides, so it is a substring search. `foo*`
and `foo` therefore have identical intended semantics here, and the same holds
for `?`. Translating `*` to `%` would also work but adds a second escaping
concern to a function whose whole job is to remove one.

```python
    qnorm = query.strip().lower().replace("\\", "/").strip("*?")
```

Worth deciding whether to strip only at the ends (above) or everywhere. A
mid-string `*` (`src/*/main.py`) is a real glob whose semantics a substring match
cannot reproduce, so stripping it silently would answer a different question than
the one asked. Leading and trailing only is the conservative read.

## Not in scope

Real glob support in path mode. That is a feature; this is a query shape that
looks supported, is not, and reports its failure as an empty result.

mikemikimike:
## Summary

- Treat trailing `*` and `?` in path-mode searches as boundary glob markers.
- Add regression coverage for both trailing glob forms against a real SQLite-backed search.

## Related Issues

Fixes #1871

## Implementation

Normalize the path query by stripping only leading/trailing `*` and `?` after path normalization. This preserves the existing substring matching behavior and deliberately avoids interpreting mid-path glob syntax as a broader glob language.

## Compatibility

Queries without trailing glob markers are unchanged. A query containing only glob markers continues to return no results.

## Test Plan

- `PYTHONPATH=packages/server/src;packages/core/src python -m pytest tests/unit/server/mcp/test_search.py -k trailing_glob -q` — 2 passed
- `python -m ruff check packages/server/src/repowise/server/mcp_server/tool_search_symbols.py tests/unit/server/mcp/test_search.py` — passed
- `git diff --check` — passed
- The full `test_search.py` module was attempted but exceeded the local wait window and was stopped; the focused regression tests completed successfully.
- Frontend build not run; this change is server-side Python only.


RaghavChamadiya (approved):
Green on the rebase, and thanks for the patience @mikemikimike, this went round more times than it should have for a one-line change.

Verified against current main: `search_paths_single` at `tool_search_symbols.py:274` still normalises with `query.strip().lower().replace("\\", "/")` and nothing downstream strips glob markers, so the fix applies cleanly and the `qnorm` assignment is intact on `025ad6b7`. Stripping only at the boundaries is the right scope. Path mode is a substring match, so a leading or trailing `*` carries no information and a mid-string one cannot be honoured, and the comment says exactly that rather than leaving the next reader to work out why `src/*/main.py` is left alone.

The parametrised table is what made this easy to take: `tool_*_overview` asserting an empty result pins the deliberate non-support, and `"  *tool_overview?  "` pins that whitespace is stripped before the markers are. `escape_like` still runs on the result, so a stripped marker cannot turn into a SQL wildcard.

Thanks also @Georgefifth for catching the folded-line breakage on the earlier head and verifying the fix on `c2d503b4`. Two people checking the same blob independently is why that never reached main.

Merging.

PR #1889, opened 2026-08-23
fix(mcp): renumber deduped guided tour steps

### LINKED ISSUE #1872: [Bug] guided_tour skips step numbers after deduplication, so the walk its own hint promises is not walkable
## Summary

`get_overview(include=["tour"])`'s `guided_tour` runs 1, 2, 4, 5 with 3 missing.
Its own hint says "read these page_ids in order... Each step builds on the
previous", so a gap reads as a step the response failed to serve rather than one
that was deliberately folded away.

## Mechanism

`_dedupe_tour_steps` (`tool_overview.py:820-836`) collapses a run of consecutive
steps sharing `(kind, reason)`. That is correct and well motivated: a stretch of
identical "The X layer's anchor" steps teaches a fresh agent nothing after the
first.

But the numbers are carried through untouched by `_build_guided_tour`
(`tool_overview.py:859`):

```python
                    "order": s.get("order"),
```

`order` is the position the step held in the persisted tour, before dedupe. Fold
out step 3 and the emitted sequence keeps 1, 2, 4, 5.

## Fix

Renumber in the emit loop in `_build_guided_tour`, which is where the deduped
list becomes the response:

```python
        for n, s in enumerate(tour, start=1):
            ...
                    "order": n,
```

## The trap

**Do not renumber in the persisted `overview_page.metadata_json`.**
`packages/ui/src/present/build-present-model.ts:34` and
`editor_files/fetcher.py:476` read `guided_tour` out of that metadata
independently and never call `_dedupe_tour_steps`. Renumbering at the source
would give them a sequence that no longer matches their own step count. The fix
belongs in the MCP projection only.

## Not in scope

Whether the dedupe should happen at all. It should; the docstring's reasoning is
sound.

mikemikimike:
## Summary

- Fixes #1872.
- Renumber `guided_tour` steps after consecutive duplicate steps are removed.
- Adds a regression test covering persisted orders `1, 2, 4` becoming `1, 2, 3` in the MCP response.

## Implementation

The persisted overview metadata is left unchanged. Renumbering happens only while building the MCP projection, so other consumers of the stored tour retain their original ordering.

## Compatibility

No API fields were added or removed. The `order` values are now contiguous for deduplicated MCP tour responses, matching the response hint's sequential-walk semantics.

## Validation

- `python -m py_compile packages/server/src/repowise/server/mcp_server/tool_overview.py tests/unit/server/mcp/test_overview_outline.py` — passed locally.
- Full pytest/CI validation was not run locally because the repository clone was interrupted by the environment's network timeout; GitHub Actions should run the repository test matrix.


RaghavChamadiya (approved):
Green, and this is now the test-only PR I asked for. Thanks for turning it around @mikemikimike, and for taking "your fix lost, your test won" in the spirit it was meant.

Checked what it is actually asserting against main. `_build_guided_tour` at `tool_overview.py:919` renumbers with `enumerate(tour, start=1)` after `_dedupe_tour_steps` has collapsed the run, and `_dedupe_tour_steps` keys on `(kind, reason)` for *consecutive* steps only. Both of your renumbering tests feed a tour that genuinely collapses and then pin the surviving `target_path` list, not just the numbers, which is the part that proves the renumbering follows the steps that survived rather than merely producing a tidy `1, 2, 3`. `test_guided_tour_renumbers_steps_after_deduplication` is the sharper of the two, because its four steps have four distinct paths and it is the shared `(layer, same)` pair that collapses, so it pins the dedupe key as well as the ordering. Asserting `json.loads(page.metadata_json) == metadata` in both covers a mutation nobody would have gone looking for until it happened.

`test_guided_tour_is_omitted_when_not_requested` is new coverage rather than a variation, and it pins both `guided_tour` and `guided_tour_hint` being absent, which is the pair that has to move together.

Non-blocking, and only worth doing if you are back in this file: the two renumbering tests assert the same output and sit at opposite ends of the module, one above `_row` and one after `test_section_sort_key_is_numeric_and_puts_unplaced_last`. They earn their keep for different reasons, so I am not asking you to drop either, but they read better adjacent.

Merging.

PR #1896, opened 2026-08-23
fix(ingestion): parse C++ types with export macros

### LINKED ISSUE #1866: [Bug] A C++ type declared with an export macro produces no symbol, so it is invisible to the graph and reads as dead code
**Summary**

`struct LEVELDB_EXPORT WriteOptions { ... }` yields no `WriteOptions` symbol at
all. The type cannot be a call target, a heritage parent or a `type_use` target,
calls to its constructor bind to whatever forward declaration is in scope, and
the definition has no inbound edge so it reports as dead code.

Seen on leveldb: `db/db_test.cc:2047` calls `WriteOptions()` and the edge lands
on `include/leveldb/db.h:23`, which is only `struct WriteOptions;`. The real
definition at `include/leveldb/options.h:168` is never indexed.

**Mechanism**

tree-sitter cannot tell `struct A B {...}` (a macro'd type definition) from a
function returning `struct A` named `B`. It picks the second reading:

```
function_definition       'struct MYLIB_EXPORT WriteOptions { ... }'
  struct_specifier        'struct MYLIB_EXPORT'
    struct
    type_identifier       'MYLIB_EXPORT'        <-- read as the type name
  identifier              'WriteOptions'        <-- the real name, on a node no query reads
  compound_statement      '{ ... }'
```

The `struct_specifier` names the macro, the type name sits on a bare
`identifier`, and the whole node is a `function_definition` rather than a struct.
Neither the struct pattern nor the function pattern in `queries/cpp.scm`
captures it.

Confirmed against leveldb's index. Only one file has a `WriteOptions` symbol and
it is the forward declaration:

```
files whose symbol table has WriteOptions:
    include/leveldb/db.h -> include/leveldb/db.h::WriteOptions
decl_to_def entries mentioning WriteOptions:
    (none)
```

With no definition indexed, `_link_declarations` has nothing to pair, so
`_redirect_to_definition` cannot move the edge off the declaration.

**Repro**

One file, two structs, one of them macro'd:

```cpp
#define MYLIB_EXPORT
struct MYLIB_EXPORT WriteOptions {
  WriteOptions() = default;
  bool sync = false;
};
struct PlainOptions {
  PlainOptions() = default;
};
```

Symbols extracted:

```
SYMBOL options.h | MYLIB_EXPORT  | variable | is_declaration=False
SYMBOL options.h | PlainOptions  | struct   | is_declaration=False
SYMBOL options.h | PlainOptions  | method   | is_declaration=False
```

`PlainOptions` is captured as both struct and constructor. `WriteOptions` does
not appear at all, and `MYLIB_EXPORT` is captured as a variable in its place.

**Impact**

Export macros are near-universal in shipped C/C++ libraries. An indicative scan
for `(struct|class) <ALLCAPS> <Name>` across some widely-used C++ repositories:

| repo | occurrences | files |
|---|---|---|
| fmt | 57 | 5 |
| abseil-cpp | 41 | 29 |
| leveldb | 27 | 15 |
| nlohmann-json | 26 | 1 |
| pybind11 | 6 | 4 |

That is a regex scan rather than an exact defect count, so treat it as a size
signal. Not every C++ project uses the pattern: Crow, seastar and aria2 read
zero, checked rather than assumed.

**Done looks like**

`queries/cpp.scm` recognises a `function_definition` whose declarator is a bare
`identifier` and whose type is a field-less `struct_specifier`, treating the
identifier as the type name rather than the macro.

A parser test pinning the repro above: `WriteOptions` extracted as a struct with
its constructor, `MYLIB_EXPORT` not extracted as a variable, and `PlainOptions`
unchanged so the ordinary path is not regressed.

Excelius-Wang:
## Summary

- Recover C++ class and struct definitions that tree-sitter parses as function definitions when an export macro appears before the real type name.
- Preserve constructor parentage and suppress the participating empty export macro as a false variable symbol.
- Add regression coverage for macro-decorated struct/class definitions and the unchanged ordinary struct path.

## Related Issues

Fixes #1866

## Test Plan

- [x] `uv run pytest -q tests/unit/ingestion/parser/test_cpp.py` — 4 passed
- [x] `uv run pytest -q tests/unit/ingestion/parser tests/unit/ingestion/test_cpp_*.py` — 347 passed
- [x] `uv run pytest -q tests/unit tests/integration/test_cpp_dead_code.py` — 14,444 passed, 12 skipped, 2 xfailed; one unrelated existing mascot tagline assertion failed
- [x] `uv run ruff check .`
- [x] `npm run lint`
- [x] `npm run type-check`
- [ ] Web build passes (`npm run build`) — not applicable; no frontend files changed

## Checklist

- [x] All tests relevant to this change pass

### DIFF
--- a/packages/core/src/repowise/core/ingestion/parser.py
+++ b/packages/core/src/repowise/core/ingestion/parser.py
@@ -210,6 +210,17 @@ def _is_bodiless_cpp_type(language: str, node_type: str, def_node: Node) -> bool
     return parent is None or parent.type != "type_definition"
 
 
+def _cpp_export_macro_parent(node: Node, parent_names: dict[int, str]) -> str | None:
+    """Return the real type name for a member inside a macro-decorated C++ type."""
+    ancestor = node.parent
+    while ancestor is not None:
+        parent_name = parent_names.get(ancestor.id)
+        if parent_name is not None:
+            return parent_name
+        ancestor = ancestor.parent
+    return None
+
+
 @cache
 def _load_compiled_query(lang: str, grammar_tag: str | None = None) -> object | None:
     """Process-wide cache of compiled tree-sitter Query objects.
@@ -617,6 +628,32 @@ class ASTParser:
         if file_info.language in _TS_JS_LANGUAGES:
             ts_deferred_exports = ts_deferred_export_names(src)
 
+        # tree-sitter-cpp parses ``struct EXPORT Name { ... }`` as a
+        # function_definition whose return type is ``struct EXPORT`` and whose
+        # bare declarator is the real type name. cpp.scm marks those matches so
+        # the specifier can keep its class/struct kind while the outer recovery
+        # node acts as the type body for nested members.
+        cpp_export_type_defs: dict[int, tuple[Node, str]] = {}
+        cpp_export_type_parents: dict[int, str] = {}
+        cpp_export_macro_names: set[str] = set()
+        if file_info.language == "cpp":
+            for capture_dict in matches:
+                type_nodes = capture_dict.get("symbol.cpp_export_type", [])
+                def_nodes = capture_dict.get("symbol.def", [])
+                name_nodes = capture_dict.get("symbol.name", [])
+                macro_nodes = capture_dict.get("symbol.cpp_export_macro", [])
+                if not type_nodes or not def_nodes or not name_nodes:
+                    continue
+                type_name = _node_text(name_nodes[0], src)
+                if not type_name:
+                    continue
+                outer_node = type_nodes[0]
+                cpp_export_type_defs[def_nodes[0].id] = (outer_node, type_name)
+                cpp_export_type_parents[outer_node.id] = type_name
+                cpp_export_macro_names.update(_node_text(node, src) for node in macro_nodes)
+
+        cpp_export_type_parent_ids = frozenset(cpp_export_type_parents)
+
         for capture_dict in matches:
             def_nodes = capture_dict.get("symbol.def", [])
             name_nodes = capture_dict.get("symbol.name", [])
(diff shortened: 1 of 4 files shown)


RaghavChamadiya (approved):
Thanks @Excelius-Wang. Naming the grammar behaviour precisely is what makes this reviewable: tree-sitter-cpp parses `struct MYLIB_EXPORT WriteOptions { ... }` as a `function_definition` whose specifier is named for the macro and whose real type name lands as a bare declarator. Once that is stated, every part of the fix follows from it.

I measured this rather than reading it, since dead types are the kind of thing that quietly distorts the whole graph. Same seven C++ sources through the parser on current main and on your branch:

| source | main | this PR |
|---|---|---|
| `struct MYLIB_EXPORT` + `class MYLIB_EXPORT` with bodies | 1 symbol, a `variable` named `MYLIB_EXPORT` | `WriteOptions` struct, `ClientOptions` class, `go` method parented correctly |
| `#define MYLIB_EXPORT __declspec(dllexport)` + macro'd struct | 1 `variable` | `Thing` struct |
| macro'd class containing a nested class and a method | 1 `variable` | `Outer`, `Inner` parented to `Outer`, `m` parented to `Outer` |
| plain `struct` / `class`, no macro | 2 symbols | identical |
| `struct Foo makeFoo() { ... }` and `Foo makeFoo2() { ... }` | 2 `function` symbols | identical |

The last row is the one I most wanted to see. `(function_definition type: (class_specifier ... !body) declarator: (identifier) body: (compound_statement))` looks like it could swallow an ordinary function returning a struct by value, and it does not: a real function's declarator is a `function_declarator`, not a bare `identifier`, so the pattern cannot reach it. The `!body` on the specifier does the other half by excluding `struct Foo { } x;`. The narrowing commit is doing real work.

The nested case is the strongest argument for merging. Recovering `Inner` with `parent_name = "Outer"` means `_cpp_export_macro_parent` is walking to the right container rather than to the nearest name, and that parentage is what the graph actually consumes. Threading `ignored_node_ids` through `_has_callable_ancestor` with a `frozenset()` default is the right way to get there without touching any other language's behaviour.

Suppressing the participating `#define` is correct too. An empty export macro is declaration scaffolding, and leaving it as a `variable` symbol gives the graph a node that has no runtime existence.

Two things unchanged from main, so not yours and not blocking, but worth writing down because they sit on the same surface. A macro'd forward declaration with no body (`struct MYLIB_EXPORT Fwd;`) still yields no type symbol and still leaks `MYLIB_EXPORT` as a `variable`, since the suppression set is only populated when the body-form pattern matches. And `struct Inline { int a; } gInline;` yields nothing at all on both. Both are the same class of bug as #1866 rather than regressions from this; I will file the forward-declaration one separately.

Full matrix green on the head commit. Merging. This closes #1866.

PR #1906, opened 2026-08-24
fix(cli): prevent stale file_page rows from surviving repowise update (#1744)

### LINKED ISSUE #1744: [Bug] stale structural `file_page` rows survive `repowise update` when repo is already up to date*
## Describe the Bug

`repowise doctor` reports stale structural `file_page` rows, but `repowise update` does not refresh them when the repository is already at the current sync commit.

In this state:

- `repowise update` reports `Already up to date`
- `repowise doctor` continues to report stale pages
- `repowise generate --stale` cannot repair them because `file_page` is structural and `generate` only regenerates model-written pages

This appears to leave no normal command that clears the stale structural pages.

---

## Steps to Reproduce

1. Run:

```bash
repowise doctor
```

and observe:

```text
Stale pages    FAIL    94 stale
```

2. Confirm the stale page types:

```bash
sqlite3 .repowise/wiki.db \
"SELECT page_type, freshness_status, COUNT(*)
 FROM wiki_pages
 WHERE freshness_status IN ('stale','expired')
 GROUP BY page_type, freshness_status
 ORDER BY COUNT(*) DESC;"
```

Result:

```text
file_page|stale|94
```

3. Run:

```bash
repowise update
```

Result:

```text
Already up to date.
```

4. Run:

```bash
repowise doctor
```

The same 94 stale pages remain.

5. I also tried:

```bash
repowise generate --stale --cascade full
```

It successfully regenerated the model-written pages:

```text
Plan: 13 pages (12 module_page, 1 repo_overview) · cascade: full
Generated 13 pages
```

but `repowise doctor` still reports:

```text
Stale pages    FAIL    94 stale
```

and the SQLite query still returns:

```text
file_page|stale|94
```

---

## Expected Behavior

Because `file_page` is a structural page type that is refreshed by `repowise update`, running `repowise update` should reconcile already-stale structural pages even when there are no new Git commits.

Alternatively, if a different command is required to repair stale structural pages, `repowise doctor` should provide an actionable command for doing so.

---

## Actual Behavior

`repowise update` exits early with:

```text
Already up to date.
```

while 94 structural `file_page` rows remain marked stale.

`repowise generate --stale` does not clear them because structural pages are outside the generation scope.

The rest of the index appears healthy:

```text
Database              OK
SQL ↔ Vector Store    OK
SQL ↔ FTS Index       OK
Coordinator drift     OK   SQL=398, Vector=398, Drift=0.0%
```

---

## Environment

- OS: macOS
- Python version: Python 3.14.7
- Repowise version: repowise, version 0.44.0
- Installation method: uv tool

---

## Additional Context

This looks similar in shape to previous cases where the `Already up to date` early-return happens before a structural staleness/re-render sweep.

Related issues that appear adjacent but not identical:

- #850 — cascade budget can leave dependent pages stale
- #851 — stale-page prioritisation
- #847 — stale prose becoming unreachable after sync-pointer advancement

In this case, all stale rows are specifically:

```text
file_page|stale|94
```

and the repository itself is otherwise clean and synchronized.

kunnalsinngh14:
## Issue
Fixes #1744 - Stale structural `file_page` rows survive `repowise update`

## Root Cause
The `repowise update` command exits early with "Already up to date" when Git HEAD 
is unchanged, even if the database contains stale structural pages. This skips the 
reconciliation pass that should clear the stale status.

## Solution
Before the early exit, check if the database contains stale structural pages. If so, 
proceed to reconciliation even though Git HEAD hasn't changed. Stale pages are added 
to the regeneration list and marked fresh after reconciliation.

## Changes
- Added `load_stale_structural_file_paths()` helper in `deterministic.py`
- Updated early-return gate in `command.py` to check for stale pages
- Updated `check_repo_staleness()` in `workspace/update.py` to recognize database staleness

## Testing
- Added 6 comprehensive unit tests
- All 248 tests pass (6 new + 242 regression)
- Verified E2E: doctor detects stale pages, update reconciles them

## Verification
Reproduces the reported case: 94 stale `file_page` rows are now cleared by 
`repowise update` even when Git HEAD is unchanged.


shmindmaster (commented):
Blocking findings against head d47fe4fc2e83b041d6a13a3257d86a9465a748c6:

1. P1 — packages/cli/src/repowise/cli/commands/update_cmd/command.py:917 still returns UpdateOutcome.NOOP unconditionally after stale_db_paths is loaded. A nonempty stale_db_paths therefore never reaches stale_extra around line 1208, so this PR does not reconcile the stale file_page rows from #1744.

2. The new tests exercise stale-path loading and workspace scoping, but none invokes run_update at unchanged HEAD and proves a stale file_page row transitions to fresh. Please add that end-to-end regression test.

3. Once the P1 return is bypassed, dry_run reaches try_acquire_update_lock around line 930 and the pending-marker cleanup around line 981 before the later dry-run return. Add a stale-row dry-run case asserting DB/state bytes, page freshness, lock files, and pending markers remain unchanged.

The repository-scoping and package-layering changes are directionally correct, but the main behavior remains unreachable until the no-op control flow is fixed.

RaghavChamadiya (approved):
Verified end to end; merging.

PR #1913, opened 2026-08-24
fix(ingestion): parse C++ export-macro forward declarations

### LINKED ISSUE #1901: [Bug] A C++ type forward-declared through an export macro yields no symbol, and leaks the macro as a variable
#1896 recovered C++ types declared through an export macro **with a body**. The bodiless form is still lost, and it takes a phantom symbol with it.

## What happens

```cpp
#define MYLIB_EXPORT
struct MYLIB_EXPORT Fwd;
class MYLIB_EXPORT Fwd2;
```

Parsed against current main, the whole file yields exactly one symbol:

```
L1   kind=variable   parent=None   decl=False   MYLIB_EXPORT
```

`Fwd` and `Fwd2` produce nothing, and `MYLIB_EXPORT` is emitted as a `variable`.

## Why

Two mechanisms, both inherited from the shape of #1896's fix rather than broken by it.

The recovery patterns added to `cpp.scm` require `body: (compound_statement)`:

```scheme
(function_definition
  type: (struct_specifier
    name: (type_identifier) @symbol.cpp_export_macro
    !body
  ) @symbol.def
  declarator: (identifier) @symbol.name
  body: (compound_statement)
) @symbol.cpp_export_type
```

A forward declaration has no compound statement, so it never matches and the type is not recovered.

The macro leak follows from the same miss. `cpp_export_macro_names` is only populated from matches of those patterns, and the `preproc_def` suppression in `_extract_symbols` checks membership in that set. With no match there is no name in the set, so the `#define` survives as a symbol.

## Why it matters

Two costs, and the second is the worse one. Headers that forward-declare through an export macro contribute no type nodes, so anything referring to those types has nothing to resolve against. And every such header contributes a `variable` node for the macro that has no runtime existence, which is a wrong node rather than a missing one.

## Notes toward a fix

The bodiless form is a different parse from the body form, so this probably wants its own pattern rather than relaxing `body:` on the existing ones (relaxing it risks reaching the ordinary-function shape that the current pattern is careful to exclude). Whatever the shape, the recovered type should come out with `is_declaration = True`, matching how `_is_bodiless_cpp_type` treats an ordinary forward declaration.

The macro suppression should probably not depend on a successful type recovery at all, since the `#define` is scaffolding either way.

Related: #1866, #1896.

Excelius-Wang:
## Summary

- Recover C++ `class` and `struct` forward declarations decorated by a locally active empty object
  macro.
- Mark recovered types as declarations and suppress only the macro definition that proves the
  recovery is safe.
- Preserve the existing export-macro type-body recovery introduced by #1896.
- Reject ambiguous declarations when the macro state cannot be proven.

## Why

Tree-sitter parses the following two forms with the same basic shape:

```cpp
#define API
struct API Fwd;
```

```cpp
struct API Fwd;
```

Both contain a bodiless `struct_specifier` named `API` followed by a declarator named `Fwd`.

In the first example, preprocessing produces `struct Fwd;`, so `Fwd` should be emitted as a
`struct` declaration. In the second example, there is no evidence that `API` is an export macro.
Treating every matching declarator as a type would incorrectly reinterpret ordinary
tag-and-declarator syntax.

The parser now recovers `Fwd` only when the local preprocessing history proves that `API` is an
active empty object macro at that exact source position.

## Implementation

- Add Tree-sitter query captures for top-level and class-scoped export-macro forward declarations.
- Track empty, non-empty, undefined, and unknown macro states in source order.
- Handle redefinitions, `#undef`, compatible conditional branches, and exact active-definition
  suppression.
- Treat `#include`, `#include_next`, and `#import` as conservative state barriers.
- Model direct `push_macro` and `pop_macro` stacks.
- Detect `_Pragma` and MSVC `__pragma` wrappers, including aliases, parameter relays, and
  object-like wrappers used inside expressions.
- Invalidate stale macro-stack snapshots after includes or opaque wrapper invocations.
- Include the complete template declaration in the recovered symbol range.

## Scope and Limitations

- A declaration such as `struct API Fwd;` is recovered as `struct Fwd;` only when `API` can be
  proven to be an active empty object macro. Otherwise, `Fwd` is not guessed to be a type.
- For the ordinary tag-and-declarator form, the existing parser also does not emit a symbol for the
  tag `API`. This behavior is unchanged from `origin/main` and is outside the scope of #1901.
- Macros supplied only by included files or compiler command-line options remain conservative false
  negatives because `ASTParser` operates on one source file without a complete preprocessing
  environment.
- Locally defined non-empty export or attribute macros are not recovered by this forward-declaration
  path. Expanding those macros safely would require preprocessing context beyond the syntax tree, so
  they remain conservative false negatives.
- Macro states that cannot be safely merged across conditional branches are also rejected. A
  missing recovered symbol is preferred over an incorrect graph node.

## Test Plan

- [x] Focused C++ parser regression tests — 77 passed
- [x] Parser, C++/Rust ingestion, dead-code, reachability, C++ integration, and generation tests —
  2,053 passed
- [x] GCC macro-state matrix — 780 generated define/undefine/push/pop sequences, zero mismatches
- [x] GCC/Clang adversarial cases — 11 passed
- [x] `origin/main` differential — 24 unchanged cases identical and six intended recoveries passed
- [x] Ruff 0.15.6 and Ruff 0.6.9
- [x] Changed test-file formatting
- [x] `git diff --check`

Targeted Mypy reports only two unchanged baseline errors in `parser_helpers.py` and the existing
`TypeReference` construction path in `parser.py`.

## Related Issues

Fixes #1901

## Checklist

- [x] The implementation follows the existing parser and Tree-sitter query pipeline.
- [x] Existing #1896 behavior is preserved.
- [x] High-risk boundary and compiler-backed regression tests are included.
- [x] No user-facing documentation changes are required.

### DIFF
--- a/packages/core/src/repowise/core/ingestion/parser.py
+++ b/packages/core/src/repowise/core/ingestion/parser.py
@@ -399,10 +399,6 @@ def _build_cpp_macro_facts(matches: list[dict], src: str) -> _CppMacroFacts:
 
     known_names = {name for _, name in macro_definitions.values()}
     known_names_by_length = tuple(sorted(known_names, key=len, reverse=True))
-    has_include_like = bool(include_nodes) or any(
-        _cpp_preproc_call_parts(_node_text(node, src))[0] in _CPP_INCLUDE_LIKE_DIRECTIVES
-        for node in preproc_calls.values()
-    )
 
     # A macro whose replacement contains a pragma stack operation is itself a
     # state hazard when invoked. Resolve simple wrapper aliases transitively;
@@ -455,11 +451,12 @@ def _build_cpp_macro_facts(matches: list[dict], src: str) -> _CppMacroFacts:
     }
 
     # Object-like macros expand wherever their identifier token appears, not
-    # only as a standalone expression. Walk the existing tree only when such
-    # a local wrapper exists; a global identifier query would add matches and
-    # extraction work to every C/C++ file, including files with no candidate.
+    # only as a standalone expression. Walk the existing tree only when a
+    # local definition proves that the name wraps a stack operation. An include
+    # is a barrier at its own position, but without preprocessing context it is
+    # not evidence that every later identifier is an imported wrapper.
     possible_invocations: dict[int, Node] = {}
-    if (object_hazard_names or has_include_like) and macro_definitions:
+    if object_hazard_names:
         root = next(iter(macro_definitions.values()))[0]
         while root.parent is not None:
             root = root.parent
@@ -470,7 +467,7 @@ def _build_cpp_macro_facts(matches: list[dict], src: str) -> _CppMacroFacts:
                 continue
             if node.type == "identifier":
                 name = _cpp_normalize_identifier(_node_text(node, src))
-                if name in object_hazard_names or has_include_like:
+                if name in object_hazard_names:
                     possible_invocations[node.id] = node
             pending_nodes.extend(node.children)
 
@@ -545,11 +542,6 @@ def _build_cpp_macro_facts(matches: list[dict], src: str) -> _CppMacroFacts:
             direct_pragma_call_ids.add(node.id)
             add_stack_operation(node, _node_text(node, src), direct=True)
 
-    include_positions = sorted(barriers)
-
-    def has_prior_include(position: int) -> bool:
-        return bisect_left(include_positions, position) > 0
-
     def add_indirect_hazards(node: Node, text: str, target_text: str = "") -> None:
         if add_stack_operation(node, text, direct=False):
             return
@@ -568,11 +560,6 @@ def _build_cpp_macro_facts(matches: list[dict], src: str) -> _CppMacroFacts:
             # The wrapper may push or pop even when its final macro value is
             # conservatively represented as UNKNOWN. Taint the stack too, so
             # a later direct pop cannot restore a stale local snapshot.
-            barriers.add(node.start_byte)
-        # Included headers may define pragma wrappers that are invisible to
-        # this one-file parser. Once one is invoked, a later local definition
-        # alone no longer proves the macro state.
-        if has_prior_include(node.start_byte):
             barriers.add(node.start_byte)
 
     for node, target_text in call_sites.values():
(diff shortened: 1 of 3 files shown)


RaghavChamadiya (approved):
Thanks @Excelius-Wang. `fix(ingestion): narrow C++ macro stack hazards` is exactly the fix, and I checked it by running the four cases rather than reading the diff.

Against your head, with the real parser:

```
1  #define API / struct API Fwd;                          -> [('Fwd', 'struct')]
2  #include <vector> / #define API / struct API Fwd;      -> [('Fwd', 'struct')]
3  #include <vector> / #define API / int helper(...)      -> [('helper', 'function'), ('Fwd', 'struct')]
   / struct API Fwd;
4  #include <vector> / struct Tag Fwd;                    -> []
```

Case 3 is the one that mattered. It was emitting nothing before, and it is the shape of nearly every real header, so this is the difference between the new path firing on our C++ corpus and never firing at all. Case 4 still declines, so narrowing the hazard set did not cost the negative case, which was the thing worth being careful about.

I also poked at the decline paths to make sure the narrowing did not turn "cannot prove it" into "assume it":

```
#undef API before the declaration     -> [('API', 'variable')]
#define API(x) x  (function-like)     -> [('API', 'function')]
#define API __declspec(dllexport)     -> [('API', 'variable')]
```

All three decline the recovery and emit no `Fwd`, and the `API` symbol in each is the macro definition itself coming through the ordinary path, not a misparsed declarator. That is the right shape: the recovery suppresses the macro definition only in the cases where it has proven the macro is an active empty object-like macro, and leaves it alone everywhere else. `tests/unit/ingestion/parser/test_cpp.py` is 77 passed locally, and the full matrix is green.

The safety argument was already the hard half and it held up on the first read. Good work getting the reach to match it without loosening it.

The two non-blocking notes from last time stand and I am not holding on either. `_build_cpp_macro_facts` is still a large function with the replay state machine inline, and splitting that out would make it testable on its own; worth doing when something next needs changing in there rather than as its own churn now. And the description still says the ordinary tag-and-declarator form emits no symbol for the tag, which is not quite what the parser does, though it is pre-existing behaviour and only the wording.

Merging.

PR #1920, opened 2026-08-25
Add tree-sitter pattern for three-part qualified calls (#1918)

### LINKED ISSUE #1918: [Bug] A three-part C++ qualified call like ns::util::fn() produces no call site at all
**Summary**

`cpp.scm` has two patterns for a scoped call and both require the qualified
identifier's `name` field to be an `identifier`. tree-sitter nests a three-part
qualifier left-recursively, so in `ns::util::toHex(x)` the outer node's `name`
field is itself a `qualified_identifier`. Neither pattern matches, and the
parser mints no call site for that line at all.

This is not a call site with a missing target. There is no call site, so the
call never reaches the resolver and cannot be resolved, refused, or counted.

**Where**

`packages/core/src/repowise/core/ingestion/queries/cpp.scm`, the two patterns
under the `Calls` heading commented as the scoped call. Both spell
`name: (identifier) @call.target`.

The grammar shape, for `ns::util::toHex`:

```
qualified_identifier
  scope: (namespace_identifier)   ns
  name:  qualified_identifier          <-- an identifier is expected here
           scope: (namespace_identifier)  util
           name:  (identifier)           toHex
```

**Repro**

```cpp
namespace ns { namespace util { int toHex(int x); } }

int main() {
  util::toHex(1);      // captured: target toHex, scope util
  ns::util::toHex(2);  // no call site produced at all
  return 0;
}
```

Parsing that through `ASTParser` yields exactly one call site, for line 4.

**Expected**

A third pattern for the nested shape, capturing the innermost `identifier` as
the target and its adjacent scope as the qualifier, so a three-part call is
captured the same way a two-part one already is.

**Impact**

Counted from `call_expression` nodes only, so an out-of-line definition cannot
inflate the number:

| repo | two-part, captured | three-part, invisible |
|---|---|---|
| seastar | 10,064 | **1,132** |
| aria2 | 8,331 | 168 |
| leveldb | 839 | 28 |

It concentrates in code that nests namespaces, which is ordinary modern C++.
An invisible call site is worse than an unresolved one: an unresolved call is
still visible to every downstream consumer as a call we declined to bind, while
this one is not there at all.

**Done looks like**

A third pattern in `cpp.scm`, a test with a three-part call asserting the target
and the qualifier, and two controls that must keep passing:

- the existing two-part `util::toHex(x)` still matches with its scope
- an unqualified `::free()` still matches

The second control matters because a tree-sitter field is mandatory once named,
which is why there are two patterns rather than one today. Adding the capture to
an existing pattern would stop the no-scope shape matching. The parser dedupes
call records and keeps the one carrying a scope, so an additional overlapping
pattern is safe.

**Note for whoever picks this up**

This adds call sites, so it can add call edges, and the edges it adds should be
spot checked against the source before merge rather than assumed correct. A
three-part qualifier names a real, written scope, so the target should be
checkable by reading the call site.

Aman-goel-04:
## Summary

- `cpp.scm`'s two scoped-call patterns require `name: (identifier)`, but tree-sitter nests a three-part qualifier left-recursively (`ns::util::fn()`), so the outer node's `name` field is itself a `qualified_identifier`. Neither pattern matched, so these calls produced no call site at all — not an unresolved one, just absent, so they never reached the resolver.
- Adds a third pattern that captures the innermost identifier as `@call.target` and its adjacent scope as `@call.scope`, matching how the existing two-part pattern already behaves.
- Verified against real leveldb source: 19 genuine three-part call sites now captured (including leading-`::` global-qualified variants like `::crc32c::Extend(...)`, which hit the same left-recursion shape), with zero unexpected overlap against the existing two-part pattern.

## Related Issues

Fixes #1918

## Test Plan

- Added `TestThreePartQualifiedCallProducesACallSite` to `tests/unit/ingestion/test_cpp_scoped_calls.py`, covering:
  - a three-part call (`ns::util::toHex()`) now resolves to the function
  - the existing two-part control (`util::toHex()`) is unaffected
  - the unqualified control (`free_thing()`) is unaffected
- Verified the new tree-sitter pattern directly (outside pytest) against the exact grammar shape, confirming no collision/dedup dependency with the existing two-part pattern.
- Ran the new pattern against a real `google/leveldb` checkout: 19 genuine three-part call sites found and manually spot-checked against source, all correct.
- Full suite: `pytest tests/` — 2654 passed, 5 skipped, 1 pre-existing unrelated failure (`test_repo_totals_churn.py::test_a_rebased_branch_in_the_history_falls_back`, a local git-rebase fixture issue, not touched by this change).

- [x] Tests pass (`pytest`)
- [x] Web build passes (`npm run build`) *(if frontend changes)* — N/A, no frontend changes

## Checklist

- [x] I have updated documentation if needed - Not needed

### DIFF
--- a/packages/core/src/repowise/core/ingestion/parser.py
+++ b/packages/core/src/repowise/core/ingestion/parser.py
@@ -1317,9 +1317,12 @@ class ASTParser:
         for call in calls:
             key = (call.line, call.target_name, call.receiver_name)
             existing = deduplicated.get(key)
-            # Two patterns match a scoped call (one keeps the qualifier, one does
-            # not) and both dedup to this key, so the richer record has to win
-            # whichever order they arrive in.
+            # Two of the three scoped-call patterns can match the same two-part call
+            # (one keeps the qualifier, one does not) and both dedup to this key, so
+            # the richer record has to win whichever order they arrive in. The
+            # three-part pattern (ns::util::fn()) never collides here, it's the
+            # only pattern that can match a nested qualified_identifier, so it
+            # always lands as a fresh key.
             if (
                 existing is None
                 or (existing.receiver_call is None and call.receiver_call is not None)
--- a/packages/core/src/repowise/core/ingestion/queries/cpp.scm
+++ b/packages/core/src/repowise/core/ingestion/queries/cpp.scm
@@ -237,6 +237,25 @@
   arguments: (argument_list) @call.arguments
 ) @call.site
 
+; The same call once more, for a three-part qualifier: ns::util::fn(args).
+; The grammar nests qualified_identifier left-recursively (see the two-level
+; qualified function definition above), so the outer node's name field is
+; itself a qualified_identifier rather than an identifier, neither pattern
+; above can match it, and no call site was produced at all for this shape
+; (#1918). Capturing the innermost identifier as the target and its adjacent
+; scope as the qualifier mirrors how a two-part call is already captured;
+; the deeper namespace prefix (ns) is dropped the same way the class-name
+; extractor already drops it for definitions.
+(call_expression
+  function: (qualified_identifier
+    name: (qualified_identifier
+      scope: (namespace_identifier) @call.scope
+      name: (identifier) @call.target
+    )
+  )
+  arguments: (argument_list) @call.arguments
+) @call.site
+
 ; Chained call: obj.method1().method2(args)
 (call_expression
   function: (field_expression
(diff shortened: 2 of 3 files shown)


RaghavChamadiya (approved):
Green across the matrix, and the verification is in my comment above: main misses both `ns::util::toHex()` and `::crc32c::Extend()`, this branch captures both, and the existing two-part shape still produces exactly one call site rather than two. Merging.

Issue #327, opened 2026-05-31
post-commit hook extremely slow


yairEO:
Hi, I love Repowise but using it on a real company's project, without many thousands of files, is extremely slow.
every time I commit the smallest change I must wait at least a whole minute for the post-commit hook to finish indexing. this is unbearable when I need to commit many files and slows me down to a point I'm considering removing the hook completely...

yairEO:
Update: it was really slow, but now it's less slow so it's bearable. I don't know what changed

RaghavChamadiya:
Hey @yairEO , thanks for flagging this. I'd love to actually pin down what was eating that minute
A few quick things would help a lot:
1.  repowise --version
2. Roughly how big is the repo? (git ls-files | wc -l and the main languages)
3. When you commit, does the hook print a line like …update complete in Xs? If you can paste one of those, the number and the wording tell me whether it's doing the light index refresh or the heavier doc regeneration.
4. OS + whether it's a laptop on battery / spinning disk vs SSD, just to rule out the obvious.

If it's easy, running repowise update once manually right after a commit and pasting the full output would be very helpful too

Ideally post-commit indexing should be at most 30s for incremental changes (most often <15s)

joptimus:
I kind of experienced something similar previously so I took a quick look at the current hook/update code and I wonder if there are two things happening here.

  The current post-commit hook is supposed to run `repowise update` in the background, so the actual `git commit` should not have to wait for indexing to finish. If the commit is blocking for around a minute, it might be worth checking whether the repo still has an older hook installed.

  A quick thing to try would be:

  ```bash
  repowise hook install
```

  That should refresh the installed hook. After that, the Repowise block in .git/hooks/post-commit should say it runs in the background/non blocking.

  There may still be a separate performance issue with repowise update itself. From what I saw, the incremental update path still re-traverses and reparses the repo to rebuild the graph so even a small commit can make the background update take a while on a real project.

  So maybe the issue splits into two follow ups

  1. Make repowise hook status detect an outdated hook, not just whether a Repowise hook exists.

  2. Look into making incremental updates lighter so small commits do less whole-repo work.

henrikekblad:
Hi! I faced the exact same issue in VS Code where the UI would freeze for ~20-30 seconds on every commit because the editor was waiting on the background subshell process group to finish.

repowise, version 0.41.0

[...]

PR #1926, opened 2026-08-25
perf(ingestion): one repo walk per update, not one per worker

azhard:
## Summary

- **`_build_file_info` ran one full repo walk per worker.** `build_repo_graph` maps it over every path with ~2x cpu_count workers, and it reaches `FileTraverser._console_script_tables()`, whose lazy init is an unsynchronised `if self._console_scripts is None`. Every worker arriving before the first one assigns starts its own `_collect_console_scripts` — each a full `iter_glob` walk. Measured on a 17k-file repo with 28 workers: **263.67s vs 8.06s** when the value is computed once, for identical `FileInfo` output. Fixed with double-checked locking, plus the same treatment for the `_get_dir_ignore` cache on the same hot path (`functools.cached_property` is not a substitute — 3.12 dropped its internal lock).
- **The package scan answered from files that are not in the index.** `_primary_language_in` and `_find_entry_points_in` walked each package separately, and neither applied the traverser's own ignore semantics. `_BLOCKED_DIRS` already contains `dist`/`build`/`target`/`vendor`, so those directories are never indexed — committed or not — yet they could decide a package's language. A committed `dist/bundle.js` that is *not* gitignored and *not* indexed made a TypeScript package report `javascript`. Merged into one `_scan_package_dir` taking the `is_pruned` predicate `scan_package_roots` already uses: **30.89s → 1.12s**, and the answers get corrected (on the repo I measured, a bench crate stopped reporting `dlang` from its `target/criterion` HTML, and package entry points went 1,452 → 11, all 1,441 dropped paths untracked).
- **Eight resolver scans each walked the whole repo during `GraphBuilder.build()`**, costing 339,025 `realpath` calls and 4.1M `lstat`s. Four now route through the existing `glob_via` seam and share `ResolverContext.walk_snapshot`. Net: ingestion pipeline **322.73s → 26.36s**, `repowise update --since HEAD~1` **8m41s → 3m27s**, with the graph unchanged (77,493 nodes / 215,904 edges, edge sets equal element by element).

**One deliberate behaviour change:** `discover_cmake_reactor`'s orphan sweep never took a `prune_nested_git` argument, so it hardcoded `True` and ignored the repo's setting. Served from the snapshot it now inherits `ctx.prune_nested_git`, so a repo indexed with `--include-nested-repos` surfaces `CMakeLists.txt` files under nested checkouts that were previously skipped. This brings cmake in line with what `ruby.py` and `swift_spm.py` already did, and it has a test either way. Flagging it because it is the one thing here that changes output rather than only cost.

**One case gets less informative, deliberately:** a package whose sources are *all* ignored now reports `unknown` rather than a language read off files nothing indexes.

## Related Issues

Refs #327 — the "post-commit hook extremely slow" thread, where the diagnosis was *"the incremental update path still re-traverses and reparses the repo so even a small commit can make the background update take a while."* This is that, measured and fixed. No open issue tracks it.

## Test Plan

- [x] `uv run pytest tests/unit/ingestion tests/unit/pipeline` — 2711 passed, 5 skipped. One failure, `test_repo_totals_churn.py::test_a_rebased_branch_in_the_history_falls_back`, reproduces on a pristine checkout of the same commit (its `git rebase -q master` fails where `init.defaultBranch` is `main`) and is unrelated to these files.
- [x] `uv run pytest tests/unit/ingestion/test_traverser.py tests/unit/ingestion/external_systems/test_cmake.py` — 114 passed
- [x] `uv run ruff check` — clean on all changed files
- [x] `uv run repowise risk main..HEAD` — *"about as risky as a typical commit in this repo"*, 47th percentile

Nine tests added, and I checked which of them actually fail without the fix rather than assuming:

| test | without the fix |
|---|---|
| `test_console_script_tables_is_collected_once` | **fails** — `_collect_console_scripts` runs once per worker |
| `test_dir_ignore_cache_publishes_one_spec_per_directory` | **fails** — two threads get different spec objects for one directory |
| `test_pre_seeded_root_entry_survives_concurrent_readers` | passes either way — pins the root entry `setdefault` keeps |
| `test_gitignored_output_no_longer_decides_the_language` | **fails** |
| `test_entry_points_come_only_from_indexed_directories` | **fails** |
| `test_a_committed_build_dir_is_excluded_too` | **fails** |
| `test_a_package_with_nothing_indexed_reports_unknown` | pins the shape that gets less informative |
| `test_orphan_glob_reads_the_snapshot_when_given_one` | covers snapshot/live equivalence |
| `test_orphan_glob_honours_the_snapshot_nested_repo_setting` | covers the behaviour change above |

## Checklist

- [x] I have updated documentation if needed *(no user-facing docs affected; rationale is in the docstrings)*

## Rejected alternatives

- **Pre-computing the console-script tables before the pool.** Fixes the measured case only; the next caller off the hot path pays it again, and the invariant that makes it safe is invisible at the call site.
- **Pruning `build`/`dist`/`out` in `PRUNED_DIRS`.** `fs_walk.py` says explicitly not to use the derived list for manifest discovery, "where a module legitimately rooted at `build/` or `coverage/` must still be found". The traverser's own ignore test is the right boundary and already exists.
- **A `WalkSnapshot.walk()` method** so `ts_workspace._get_repo_scan` could share too — that needs an API addition rather than an existing seam, so it is left out.

## Not fixed here

Four repo walks remain per build. `read_go_modules` runs before the `ResolverContext` exists, so sharing the snapshot means threading it down from `GraphBuilder`; `ts_workspace._get_repo_scan` uses `walk_repo` directly and needs the method above. Happy to follow up on either if you want them in scope.

### DIFF
--- a/packages/core/src/repowise/core/ingestion/external_systems/cmake.py
+++ b/packages/core/src/repowise/core/ingestion/external_systems/cmake.py
@@ -31,6 +31,8 @@ from pathlib import Path, PurePosixPath
 
 import structlog
 
+from repowise.core.fs_walk import WalkSnapshot
+
 from ..languages.specs.cpp import INCLUDE_FRAGMENT_EXTENSIONS
 from .base import ExternalSystemRecord
 
@@ -595,6 +597,7 @@ def discover_cmake_reactor(
     repo_root: Path,
     *,
     max_files: int = 2000,
+    snapshot: WalkSnapshot | None = None,
 ) -> list[CMakeFile]:
     """Walk the repo from ``repo_root`` following ``add_subdirectory`` links.
 
@@ -637,9 +640,9 @@ def discover_cmake_reactor(
     skip_dirs = {".git", "build", "_build", "out", "_deps", "cmake-build-debug",
                  "cmake-build-release", "node_modules", ".venv", "venv"}
     if len(out) < max_files:
-        from repowise.core.fs_walk import iter_glob
+        from repowise.core.fs_walk import glob_via
 
-        for cml in iter_glob(repo_root, "CMakeLists.txt"):
+        for cml in glob_via(snapshot, repo_root, "CMakeLists.txt"):
             try:
                 rel = cml.resolve().relative_to(repo_root).as_posix()
             except ValueError:
--- a/packages/core/src/repowise/core/ingestion/resolvers/cpp_workspace.py
+++ b/packages/core/src/repowise/core/ingestion/resolvers/cpp_workspace.py
@@ -373,7 +373,11 @@ def build_cpp_workspace_index(ctx: ResolverContext) -> CppWorkspaceIndex:
     # is what that path did before.
     source_map = getattr(ctx, "source_map", None)
 
-    cmake_files = discover_cmake_reactor(repo_path)
+    # Same reason as ``source_map`` above: the stand-in context has no
+    # snapshot, and None means the live walk this always did.
+    cmake_files = discover_cmake_reactor(
+        repo_path, snapshot=getattr(ctx, "walk_snapshot", None)
+    )
     file_api_targets = parse_cmake_file_api_reply(repo_path)
     bazel_files = discover_bazel_packages(repo_path) if is_bazel_repo(repo_path) else []
 
--- a/packages/core/src/repowise/core/ingestion/resolvers/ruby.py
+++ b/packages/core/src/repowise/core/ingestion/resolvers/ruby.py
@@ -67,16 +67,26 @@ def _scan_gem_metadata(ctx: ResolverContext) -> tuple[frozenset[str], tuple[str,
     names: set[str] = set()
     lib_roots: set[str] = set()
     if ctx.repo_path is not None:
-        from repowise.core.fs_walk import iter_glob
+        from repowise.core.fs_walk import glob_via
 
         repo = ctx.repo_path.resolve()
         candidates: list = []
         try:
             candidates.extend(
-                iter_glob(ctx.repo_path, "Gemfile", prune_nested_git=ctx.prune_nested_git)
+                glob_via(
+                    ctx.walk_snapshot,
+                    ctx.repo_path,
+                    "Gemfile",
+                    prune_nested_git=ctx.prune_nested_git,
+                )
             )
             candidates.extend(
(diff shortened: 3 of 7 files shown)


RaghavChamadiya (approved):
Thanks @azhard. 263.67s to 8.06s on a 17k-file repo with 28 workers, for identical `FileInfo` output, is the kind of number that only shows up when someone actually profiles the phase instead of assuming it is I/O bound.

The diagnosis is the good part. An unsynchronised `if self._console_scripts is None` looks harmless right up until the caller is `_build_file_info` under a pool that `build_repo_graph` maps over every path with ~2x cpu_count workers. Then every worker that reaches the check before the first one assigns starts its own `_collect_console_scripts`, and each of those is a full `iter_glob` walk of the repo. The cost is not the lazy init, it is that the lazy init was written for a single-threaded caller and the caller changed underneath it.

Both fixes read correctly to me:

The double-checked lock on `_console_script_tables` keeps the fast path lock-free once assigned, and the assignment publishes a finished tuple so a racing reader sees either `None` or the whole object. The note that `functools.cached_property` is not a substitute is right and worth having written down, since it is the obvious reach: 3.12 removed its internal lock precisely because that lock serialised across instances, so it now permits the duplicate computation this exists to prevent.

`_get_dir_ignore` using `setdefault` under the lock rather than assignment is the subtler of the two, and your comment gives the reason: a racer's spec is equivalent, but keeping the first published one means every caller holding a reference sees the same object. Reading outside the lock and writing under it is the right trade when a miss costs two `exists()` and a compile rather than a walk.

The `_scan_package_dir` merge is more than a de-duplication and I want to name the behaviour change so it is on the record rather than buried in a perf PR. Folding `_primary_language_in` and `_find_entry_points_in` into one walk halves the passes, but adding `is_pruned` also means the scan stops descending into gitignored trees. That is a change in what the answer is, and it is the more correct answer: `_BLOCKED_DIRS` already keeps `dist`/`build`/`target`/`vendor` out of the index, so a package's primary language was being decided partly by artifacts nothing indexes. Your point that `_detect_language` *opens* every file whose extension it does not recognise, for the shebang check, makes it a correctness and a cost argument at once, and a gitignored build tree is exactly where those files are.

The fixture-tie fix in the last commit is a good catch on your own test: a package with one `.ts` and one `.js` makes `max(counts, key=...)` return whichever the dict happened to order first, so the test would have passed or failed on iteration order rather than on the behaviour.

PR #1945, opened 2026-08-27
fix(core): sweep page snapshots when deleting a repository

### LINKED ISSUE #1944: [Bug] DELETE /api/repos/{id} fails with FOREIGN KEY constraint failed once a page has been regenerated
### What happens

`DELETE /api/repos/{id}` returns **HTTP 500** for any repository whose pages have been
regenerated at least once. The repo cannot be removed from the dashboard or the API.

```
sqlalchemy.exc.IntegrityError: (sqlite3.IntegrityError) FOREIGN KEY constraint failed
[SQL: DELETE FROM repositories WHERE repositories.id = ?]
```

Hit on 0.46.0 removing two duplicate registrations pointing at the same folder. Both
attempts 500'd; the traceback is only visible if the server's stderr is redirected,
since `serve` writes no log file.

### Root cause

`delete_repository` relies entirely on the FK cascade:

```python
# packages/core/src/repowise/core/persistence/crud/repository.py
await session.delete(repo)   # "CASCADE handles all child ORM tables"
```

Every direct child of `repositories` does declare `ondelete="CASCADE"` — but the
*grandchild* does not:

```python
# packages/core/src/repowise/core/persistence/models.py:195
page_id: Mapped[str] = mapped_column(Text, ForeignKey("wiki_pages.id"), nullable=False)
```

So cascading the delete into `wiki_pages` strands every `wiki_page_versions` row that
references those pages, and SQLite aborts the whole statement. A `PageVersion` is
created on the *second* upsert of a page, which is why the failure only shows up once
a repo has been regenerated — a freshly indexed repo deletes fine.

The pipeline already knows about this. All four page-pruning paths in
`pipeline/persist.py` (lines 1066, 1151, 1217, 1312) pre-delete snapshots before
deleting pages:

```python
await session.execute(delete(PageVersion).where(PageVersion.page_id.in_(batch)))
await session.execute(delete(Page).where(...))
```

Repository deletion is the one path that does not.

### Why no test caught it

`tests/unit/persistence/conftest.py` builds its engine with `create_async_engine`
directly rather than the project's `create_engine`, so `PRAGMA foreign_keys=ON` is
never issued. SQLite's default is *off*, which means **no foreign key is enforced
anywhere in the persistence suite** and referential bugs are structurally invisible
to it. `test_delete_repository_success` passes precisely because it deletes a repo
with no pages, under a connection that would not have complained either way.

### Reproduce

Index a repo, regenerate its pages once (so a snapshot exists), then `DELETE
/api/repos/{id}` → 500. Or against any existing store:

```sql
-- non-zero means this repo cannot currently be deleted
SELECT count(*) FROM wiki_page_versions
 WHERE page_id IN (SELECT id FROM wiki_pages WHERE repository_id = '<repo_id>');
```

### Workaround

Delete the snapshots first, then call the API:

```sql
DELETE FROM wiki_page_versions
 WHERE page_id IN (SELECT id FROM wiki_pages WHERE repository_id = '<repo_id>');
```

### Fix

PR attached: `delete_repository` sweeps the repository's `PageVersion` rows before
deleting, matching the pipeline's existing convention, plus regression tests that run
with production pragmas so this class of bug is catchable in future.

Note that a schema-level fix alone would not help anyone's existing index —
`_reconcile_schema` is additive-only and explicitly does not reconcile FK constraint
changes on existing columns, and local SQLite stores never run Alembic.

### Environment

repowise 0.46.0 (pipx), Windows 11, Python 3.12, SQLite backend.

jgt87:
Fixes #1944.

`DELETE /api/repos/{id}` returns HTTP 500 for any repository whose pages have been
regenerated at least once:

```
sqlalchemy.exc.IntegrityError: (sqlite3.IntegrityError) FOREIGN KEY constraint failed
[SQL: DELETE FROM repositories WHERE repositories.id = ?]
```

### Why

`delete_repository` leans entirely on the cascade. Every direct child of
`repositories` declares `ondelete="CASCADE"`, but the grandchild does not —
`PageVersion.page_id` (`models.py:195`) is a plain `ForeignKey("wiki_pages.id")`.
Cascading into `wiki_pages` therefore strands the snapshot rows and SQLite aborts the
statement. Snapshots are written on the second upsert of a page, so a freshly indexed
repo deletes fine and a regenerated one can never be deleted.

### The change

`delete_repository` now sweeps the repository's `PageVersion` rows first. This is the
convention the codebase already follows: all four page-pruning paths in
`pipeline/persist.py` (1066, 1151, 1217, 1312) pre-delete snapshots before deleting
pages for exactly this reason. Deletion was the one path that missed it.

I deliberately fixed this in the application layer rather than the schema. Adding
`ondelete="CASCADE"` to the model plus a migration would leave every existing user
broken: `_reconcile_schema` is additive-only and documents that FK constraint changes
on existing columns are not reconciled, and local SQLite stores never run Alembic. So
the declared FK stays as-is and the sweep is explicit — happy to add the schema change
on top as a follow-up if you'd prefer both.

Scoped by `repository_id` (a plain column on `PageVersion`, always populated from
`existing.repository_id`), so it is a single statement with no page-id fan-out.

### Tests

`tests/unit/persistence/test_delete_repository_fk.py` — three tests: the FK pragma is
actually on, a repo with archived snapshots deletes cleanly, and the sweep does not
touch another repo's snapshots.

They need their own fixture. `tests/unit/persistence/conftest.py` builds its engine
with `create_async_engine` directly, so `PRAGMA foreign_keys=ON` is never issued and
SQLite defaults to off — **no foreign key is enforced anywhere in the persistence
suite**, which is why this shipped. The new fixture uses the project's own
`create_engine` so the pragmas match a real `.repowise/wiki.db`. That is also the
reason `test_delete_repository_success` passes today: it deletes a repo with no pages,
on a connection that would not have complained either way.

Worth considering separately whether the shared fixture should adopt `create_engine`
too — it would make the whole suite FK-aware, and may well surface other latent
referential bugs.

### Verification

- New tests fail with `IntegrityError: FOREIGN KEY constraint failed` on `main`, pass with the fix.
- `uv run pytest tests/unit/persistence/ tests/unit/server/test_repos.py` → 665 passed.
- `ruff format --check` and `ruff check` clean.
- Confirmed against a real 0.46.0 store: the two repos that 500'd delete cleanly.

### DIFF
--- a/packages/core/src/repowise/core/persistence/crud/repository.py
+++ b/packages/core/src/repowise/core/persistence/crud/repository.py
@@ -10,12 +10,13 @@ import json
 from datetime import datetime
 from pathlib import Path
 
-from sqlalchemy import select
+from sqlalchemy import delete, select
 from sqlalchemy.ext.asyncio import AsyncSession
 
 from ..models import (
     GenerationJob,
     Page,
+    PageVersion,
     Repository,
     WebhookEvent,
     _new_uuid,
@@ -193,10 +194,18 @@ async def delete_repository(session: AsyncSession, repo_id: str) -> bool:
 
     NOTE: The caller should clean up the FTS index *before* calling this,
     since the CASCADE will delete Page rows and we lose the page IDs.
+
+    ``PageVersion`` is swept explicitly first. Its ``page_id`` foreign key is
+    declared without ``ondelete="CASCADE"``, so the cascade that removes the
+    repository's ``Page`` rows leaves the snapshots pointing at rows that no
+    longer exist and SQLite aborts the whole statement with ``FOREIGN KEY
+    constraint failed``. The pipeline's page-pruning paths already pre-delete
+    snapshots for exactly this reason; deletion is the one path that did not.
     """
     repo = await session.get(Repository, repo_id)
     if repo is None:
         return False
+    await session.execute(delete(PageVersion).where(PageVersion.repository_id == repo_id))
     await session.delete(repo)
     await session.flush()
     return True
--- a/tests/unit/persistence/test_delete_repository_fk.py
+++ b/tests/unit/persistence/test_delete_repository_fk.py
@@ -1,0 +1,104 @@
+"""Repository deletion under real foreign-key enforcement.
+
+The shared ``async_engine`` fixture builds its engine with
+``create_async_engine`` directly, which never issues ``PRAGMA foreign_keys=ON``
+— SQLite's default is *off*, so foreign keys are not enforced anywhere in the
+persistence suite. Production engines come from
+:func:`repowise.core.persistence.database.create_engine`, which turns the
+pragma on for every connection. A referential bug is therefore invisible to
+the existing tests and fails only against a real store, which is how
+``delete_repository`` shipped unable to delete any repository whose pages had
+ever been regenerated.
+
+These tests use the project's own ``create_engine`` so the pragmas match what
+a user's ``.repowise/wiki.db`` actually runs with.
+"""
+
+from __future__ import annotations
+
+import pytest
+from sqlalchemy import func, select, text
+from sqlalchemy.ext.asyncio import AsyncSession, async_sessionmaker
+
+from repowise.core.persistence.crud import (
+    delete_repository,
+    get_page_versions,
+    get_repository,
+    upsert_page,
+)
+from repowise.core.persistence.database import create_engine, init_db
+from repowise.core.persistence.models import Page, PageVersion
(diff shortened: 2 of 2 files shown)


RaghavChamadiya (approved):
Thanks @jgt87, and thanks for amending the trailer without argument. The blocker was authorship, never the work.

The finding underneath this is worth more than the fix: the persistence suite never enables `PRAGMA foreign_keys`, so a whole class of cascade bug can pass its tests. That is the kind of thing that stays broken for a year because every test agrees with it.

Green across 3.11, 3.12 and 3.13. Merging.

Issue #1935, opened 2026-08-27
Security as an intelligence layer, not a byproduct — scoping check before we build


tacopiek:
## Where this comes from

We run repowise across a workspace of TypeScript/Next.js repos. Digging into the
security surface, we found the plumbing is genuinely good and the analysis
behind it is thin. Before we spend weeks on it, we would like to know which
parts of the following you would take, and which you would rather we keep in a
fork.

## What already exists (and works well)

- `SecurityScanner` + the `security_findings` table, idempotent across re-runs
  via the `(repository_id, file_path, kind, line_number, commit_sha)` constraint
- `repowise security scan --history` with commit provenance for secrets that
  were later removed
- `GET /api/repos/{repo_id}/security` with severity/path/history filters
- Per-request line verification through `check_finding_line`, so a finding does
  not silently drift when the file changes
- The findings table on the code-health page

None of that needs replacing. Everything below feeds those same surfaces.

## The gap

**1. The pattern registry is Python-shaped.** Eleven regexes: `eval`, `exec`,
`pickle.loads`, `subprocess(shell=True)`, `os.system`, hardcoded password/secret,
f-string SQL, concatenated SQL, `verify=False`, weak hashes. On a Python
codebase that is a reasonable starting set. On a Next.js codebase almost none of
it can fire. There is no pattern for `dangerouslySetInnerHTML`,
`child_process.exec`, template-literal SQL, `NEXT_PUBLIC_`-prefixed secrets,
permissive CORS on mutating handlers, or route handlers with no authorization
check.

Measured on a 55-file Next.js/TypeScript app with 14 API route handlers talking
to SQL Server, the working-tree scan produces **one** finding — a `low` on a
test file. Manual review of the same routes finds considerably more, including a
route that writes the client-supplied actor name and timestamp straight into an
audit log.

**2. There is no aggregation.** `security` is the only surface that returns a
flat list. `health` has a 1-10 score with defect-backed bands; `risk` has an
attributable 0-10 breakdown. Security has row count. There is no way to ask
"is this repo better or worse than last month", and no gate a CI job can fail on.

**3. Agents cannot reach it.** There is a tool for health, risk, change risk,
blast radius, dead code, architecture and rationale. There is none for security
findings. The UI can see them; the thing the product exists to inform cannot.

**4. It is not a layer.** `docs/layers/` documents every intelligence layer —
health, change risk, dead code, decisions, graph, refactoring, test
[...]

PR #1947, opened 2026-08-28
fix(security): stop reporting RegExp.exec, and catch the constant secret form

tacopiek:
Refs #1935. Two bugs in the existing pattern registry, found by running the scan over a 17-repository corpus (1109 JS/TS/Python files) and adjudicating every hit by hand.

## `exec_call` fires on every regex match in JavaScript

`exec` is not a global in JavaScript — the name belongs to `RegExp.prototype.exec` — so the receiver-chain prefix in the pattern is exactly what makes `re.exec(expr)`, `/x/.exec(s)` and `cellPattern.exec(xml)` match. On the corpus all 9 hits were XLSX and format-parsing code, none was a process spawn, and each was reported at `high`.

Outside Python the kind now requires the file to name `child_process`, which is where the dangerous call actually comes from. On the same corpus the finding set swaps completely:

| | before | after |
|---|---|---|
| `exec_call` hits | 9, all `RegExp.exec` in parsing code | 10, all real `execSync` in build scripts and `next.config.ts` |

The gate is per file, so a file that both spawns and parses still reports every `exec(` in it. That residual is deliberate and covered by a test rather than left implicit.

## The secret patterns were case-sensitive

`hardcoded_password` and `hardcoded_secret` matched only the lowercase spelling, so the constant form a pinned credential is usually written in — `API_KEY = "..."`, `SECRET = "..."` — never fired. A live `N8N_API_KEY = '...'` sat unreported in two repositories of the corpus.

One wrinkle worth knowing for anyone touching the registry next: the fix is a scoped `(?i:...)` group rather than an `re.IGNORECASE` flag. `_ANY_PATTERN` is built by concatenating the patterns' source text, which drops per-pattern flags — with a flag the prefilter stays case-sensitive and rejects the line before the pattern loop ever runs. The first version of this fix looked correct and changed nothing. There is now a test asserting the prefilter accepts the constant form.

## Effect

Corpus totals, before → after: **16 → 19 findings**, with the 9 false-positive `exec_call` hits replaced by 10 true ones and two committed credentials surfaced that the registry previously walked past.

## Tests

`tests/unit/analysis/` passes (693). The `exec\n  (payload);` case in `test_fallback_eval_exec_call_corpus` changed from `{"exec_call"}` to `set()` — that is the behaviour change, not an oversight — and two regex-parsing cases were added alongside it. A new corpus test covers the `child_process` gate in both directions.

`ruff check` and `ruff format --check` are clean on both files. `mypy` reports one pre-existing `Result[Any].rowcount` error in this file that is present on `main` and untouched here.

### DIFF
--- a/packages/core/src/repowise/core/analysis/security_scan.py
+++ b/packages/core/src/repowise/core/analysis/security_scan.py
@@ -46,13 +46,36 @@ _CALL_PATTERNS: list[tuple[re.Pattern, str, str]] = [
 ]
 _CALL_KINDS = frozenset(kind for _, kind, _ in _CALL_PATTERNS)
 
+# ``exec`` is not a global in JavaScript; the name belongs to
+# ``RegExp.prototype.exec``. The receiver-chain prefix above therefore matches
+# ``re.exec(str)``, ``/x/.exec(s)`` and ``pattern.exec(xml)`` — idiomatic parsing
+# code, reported at ``high``. Measured on a 17-repo TypeScript corpus, every
+# ``exec_call`` hit outside Python was a regex match and none was a process
+# spawn, so the kind was pure noise on those languages.
+#
+# The dangerous call in JavaScript comes from ``child_process``, so that is what
+# gates it: a file that never names the module cannot spawn one. ``eval`` needs
+# no such gate — it is a genuine global there.
+_CHILD_PROCESS_IMPORT = re.compile(r"child_process|node:child_process")
+_JS_EXEC_CALL = re.compile(r"(?<![\w$])(?:[A-Za-z_$][\w$]*\s*\.\s*)*exec(?:File)?(?:Sync)?\s*\(")
+
 _PATTERNS: list[tuple[re.Pattern, str, str]] = [
     *_CALL_PATTERNS,
     (re.compile(r"pickle\.loads"), "pickle_loads", "high"),
     (re.compile(r"subprocess\..*shell\s*=\s*True"), "subprocess_shell_true", "high"),
     (re.compile(r"os\.system"), "os_system", "high"),
-    (re.compile(r"password\s*=\s*['\"]"), "hardcoded_password", "high"),
-    (re.compile(r"(?:api_?key|secret)\s*=\s*['\"]"), "hardcoded_secret", "high"),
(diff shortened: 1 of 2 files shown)


RaghavChamadiya (approved):
Thanks @tacopiek. Running the scan over a 17-repo corpus and adjudicating every hit by hand is what makes this a fix rather than a guess, and the before/after table showing the `exec_call` finding set swapping completely, 9 regex hits out for 10 real `execSync` hits in, is the clearest possible statement of the problem.

Verified the mechanism against the tree before merging, because the gate only works if nothing else emits the kind. It holds: `scan_file`'s per-line loop does `if kind in _CALL_KINDS: continue`, so `_call_findings` is the only producer of `eval_call` and `exec_call` and your `is_python` branch there is the whole story. Gating on the file naming `child_process` also picks up `execSync`, `execFile` and `execFileSync`, which the old `exec\s*\(` pattern could not match at all, so this gains real detections while removing the noise rather than only trading one for the other.

The `(?i:...)` scoped group is the detail I would have got wrong. `_ANY_PATTERN` is built as `"|".join(f"(?:{p.pattern})" for p, _, _ in _PATTERNS)`, so it concatenates pattern *source text* and drops per-pattern flags. An `re.IGNORECASE` flag on the compile would have left the prefilter case-sensitive, the line would have been rejected at `if not _ANY_PATTERN.search(line)`, and the pattern would never have run. Your test asserts `_ANY_PATTERN.search(source)` as well as the finding, so the trap is pinned and not just avoided. Writing the reason into the comment matters here too, because the next person's instinct will be to "tidy" that into a flag.

Keeping the residual false positive, a file that both spawns and parses regexes, explicit and covered by its own test case is the right call. The gate is per file and cannot be otherwise without dataflow, so saying so in a test beats leaving it implicit and having someone rediscover it as a bug.

Reworking your own comment after the bot flagged that it matched the pattern it was explaining is a good catch on yourself, and it is a genuinely funny failure mode for a scanner that reads raw source.

I have left a note on #1948: its "expect `exec_call` to be entirely noise" section documents the behaviour this PR just removed, so that page wants a rebase and a rewrite of that section before it lands.

PR #1948, opened 2026-08-28
docs(layers): document the security scan, floor and all

tacopiek:
Refs #1935. `docs/layers/` has a page for every intelligence layer — health, change risk, dead code, decisions, graph, refactoring, test intelligence, bug history — and none for the security surface, which matches how it is built. This adds one, covering the local scan only.

Written to say where the floor is rather than to make the layer sound larger than it is, so roughly half the page is about what it cannot do.

Rebased onto main, so the page now describes the registry as #1947 left it.

## What it covers

- **The eleven patterns and three severities**, plus the details that are easy to miss: `eval`/`exec` are resolved through the AST in Python and lexically elsewhere, `exec_call` outside Python additionally requires the file to name `child_process`, and `subprocess_shell_true` gets a second whole-source pass so a call opening on one line with `shell=True` on another is still seen.
- **What the registry does not catch** — most languages, since the patterns are Python-shaped; anything needing framework semantics or dataflow; real secret detection, which is a named assignment in any case and nothing more, no entropy, key formats or provider prefixes; dependencies.
- **The false positives to expect**, including the two structural ones: the `child_process` gate is per file rather than per call, so a file that both spawns and parses text with regexes still reports every `exec(`; and the per-line patterns run on raw source, so a comment that spells out a credential assignment reports itself.
- **Working tree versus `--history`**, and why history mode defaults to the secret kinds — a commit that once called `eval()` is history doing what history does, whereas a committed secret stays valid until someone rotates it.
- **The three outcomes of `check_finding_line`**, and why `security_sensitive_symbol` is checked in place rather than relocated on its snippet.
- **How to read the surface**, ending on the point the page exists to make: a low finding count means the registry did not match, and on most repositories in most languages that is all it means.

It also links the page from the layer table in `docs/README.md`.

Docs only, no code changes.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -58,6 +58,7 @@ your agent in under five minutes, with no API key.
 | [layers/GRAPH.md](layers/GRAPH.md) | The dependency graph: what is in it, how every edge is resolved, and how much to trust each one |
 | [layers/DECISIONS.md](layers/DECISIONS.md) | Architectural decisions mined from your repo and from your own agent sessions |
 | [layers/DEAD_CODE.md](layers/DEAD_CODE.md) | Unreachable files, unused exports, and zombie packages by confidence tier |
+| [layers/SECURITY.md](layers/SECURITY.md) | The local pattern scan: what the eleven patterns catch, what they do not, and how far to trust the result |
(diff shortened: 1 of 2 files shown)


RaghavChamadiya (approved):
Green after the rebase. The page says where the floor is instead of making the layer sound larger than it is, which is what this one needed.

PR #1953, opened 2026-08-28
fix(providers): add codex_cli to the server provider catalog (#1530)

### LINKED ISSUE #1530: [Bug] web dashboard's provider list disagrees with the CLI (codex_cli missing)

sloemo01:
## What

Fixes #1530. The web dashboard's provider list disagreed with the CLI: `codex_cli` is a built-in provider in the registry (`packages/core/.../llm/registry.py`) but was missing from the server's `PROVIDER_CATALOG`, which is what the dashboard renders. The catalog had `claude_cli` and `opencode` but not `codex_cli`.

## Root cause

The server catalog in `packages/server/src/repowise/server/provider_config.py` was maintained by hand, and the drift test guarding it only compared `env_keys` for providers that already existed in both places — it could not catch a *missing* entry.

## Changes

- Add a `codex_cli` entry to `PROVIDER_CATALOG` (keyless, `requires_key: false`, sentinel model `codex_cli/default`; the runtime model list comes from the authenticated codex catalog, and `list_provider_status` appends the resolved active model when it isn't cataloged).
- Harden `test_server_provider_catalog_agrees_with_the_registry` to assert every built-in provider (except `mock`) is present in the server catalog, so a future provider addition can't silently skip the dashboard again.

## Tests

`uv run pytest tests/unit/test_providers/test_provider_env_registry.py -q` → 26 passed. The hardened test fails on unpatched main.

### DIFF
--- a/packages/server/src/repowise/server/provider_config.py
+++ b/packages/server/src/repowise/server/provider_config.py
@@ -142,6 +142,20 @@ PROVIDER_CATALOG: list[dict[str, Any]] = [
         "requires_key": False,
     },
     {
+        "id": "codex_cli",
+        "name": "Codex (Local CLI)",
+        "default_model": "codex_cli/default",
+        "models": [
+            # The CLI's model list comes from the authenticated codex catalog
+            # at runtime (see core/providers/llm/codex_cli.py), so the static
+            # catalog only names the sentinel default; list_provider_status
+            # appends the resolved active model when it isn't cataloged.
+            "codex_cli/default",
+        ],
+        "env_keys": [],
+        "requires_key": False,
+    },
+    {
         "id": "opencode",
         "name": "OpenCode (Local CLI)",
         "default_model": "opencode/default",
--- a/tests/unit/test_providers/test_provider_env_registry.py
+++ b/tests/unit/test_providers/test_provider_env_registry.py
@@ -105,7 +105,18 @@ def test_every_provider_that_can_be_auto_detected_is_in_the_order():
 
 
 def test_server_provider_catalog_agrees_with_the_registry():
+    from repowise.core.providers.llm.registry import _BUILTIN_PROVIDERS
     from repowise.server.provider_config import PROVIDER_CATALOG
+
+    catalog_ids = {entry["id"] for entry in PROVIDER_CATALOG}
+    # Every provider the CLI can resolve must appear in the server catalog so
+    # the web dashboard's provider picker never silently disagrees with what
+    # the CLI can resolve. `mock` is the one deliberate exception — it is a
+    # test/fallback provider, not something a user should pick in the UI.
+    missing = set(_BUILTIN_PROVIDERS) - catalog_ids - {"mock"}
+    assert not missing, (
+        f"server/provider_config.py is missing catalog entries for: {sorted(missing)}"
+    )
 
     mismatched = {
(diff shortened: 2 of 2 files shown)


RaghavChamadiya (approved):
Thanks @sloemo01. The catalog entry is the fix for #1530, but the drift test is worth more than the fix and is the reason this goes in as-is. `set(_BUILTIN_PROVIDERS) - catalog_ids - {"mock"}` means the next provider added to the registry cannot silently miss the dashboard picker, which is how this one got here. Naming `mock` as the single deliberate exception in a comment, rather than filtering it silently, is what keeps that assertion readable in two years.

I checked the guard against current main rather than trusting the run from the 28th, since the registry has taken churn since. On `origin/main` the difference is exactly `{codex_cli}` and nothing else, so this PR closes the set and the assertion passes with no follow-up needed:

```
builtin:  anthropic claude_cli codex_cli deepseek edenai gemini kimi litellm mock ollama openai opencode openrouter
catalog:  anthropic claude_cli           deepseek edenai gemini kimi litellm      ollama openai opencode openrouter
```

Pinning the static entry to the `codex_cli/default` sentinel and letting `list_provider_status` append the resolved model is the right call too. Freezing a model list that the authenticated catalog decides at runtime would have been stale on the first release.

Green on a7529ccb.

PR #1955, opened 2026-08-28
fix(coverage): flag severe path-mapping loss instead of treating it as success (#1746)

### LINKED ISSUE #1746: [Bug] coverage add treats severe path-mapping loss as a successful note

sloemo01:
## What

Fixes #1746. `repowise coverage add` treated severe path-mapping loss as a successful note: a report whose files mostly failed to map to the repo tree still ingested, reported "Ingested coverage for N file(s)", and exited 0 — while `get_coverage_summary` computed `covered/total` over only the mapped subset, so one mapped file at 90% reported the repository at 90%. That number feeds the health scores and the Tests tab, making the fragment look like repo-wide coverage.

## Root cause

`ResolvedCoverage` tracked `matched_exact`/`matched_suffix`/`unmatched`/`ambiguous` but nothing flagged the aggregate as a fragment. `get_coverage_summary`'s denominator was the mapped subset by construction, and no writer or reader carried a "this is partial" signal.

## Changes

- **`mapping_partial` on the resolver** (`discovery.py`): true when fewer than half the report's files mapped to the tree (`matched * 2 < report_file_count`). Computed on report-file count, not matched keys, so merged multi-reports can't shrink the denominator. Deliberately not a hard ratio on the matched side: monorepos legitimately ingesting one package's report stay unflagged.
- **Persisted**: new `coverage_files.mapping_partial` column (model + Alembic `0056`; SQLite stores pick it up via the additive schema reconciler). Threaded through `save_coverage_files`, `get_coverage_summary`, the `IndexStore` interface + SQL adapter, the pipeline phase (now returns the flag), and the health report.
- **CLI**: `coverage add` prints a red fragment warning and **exits non-zero** on a partial ingest, so scripts can't mistake a fragment for complete coverage. `coverage status` shows an inline partial note. Pipeline progress surfaces the same warning.
- **Server/MCP**: `mapping_partial` flows through the summary dict; the empty-summary shape includes it.

## Tests

- `tests/unit/health/test_coverage_discovery.py`: majority-maps → False, majority-unmapped → True, merged-report denominator, end-to-end `build_coverage_map` flag.
- `tests/unit/persistence/test_coverage_narrow_read.py`: legacy ingests read `False`, partial ingests read `True` through the summary.
- Full coverage suite: 167 passed; entire unit suite: 14,896 passed.

### DIFF
--- a/packages/cli/src/repowise/cli/commands/coverage_cmd.py
+++ b/packages/cli/src/repowise/cli/commands/coverage_cmd.py
@@ -181,6 +181,7 @@ def coverage_add(
             # --- Per-file aggregate coverage (lcov / cobertura / clover / json).
             agg_matched = 0
             unmapped = 0
+            mapping_partial = False
             if agg_paths:
                 resolved, errors = build_coverage_map(
                     repo_path,
@@ -198,6 +199,7 @@ def coverage_add(
                 # for the one run where every single one of them was.
                 skipped = resolved.unmatched + resolved.ambiguous
                 unmapped = len(skipped)
+                mapping_partial = resolved.mapping_partial
                 if resolved.files:
                     await save_coverage_files(
                         session,
@@ -205,6 +207,7 @@ def coverage_add(
                         resolved.files,
                         source_format=resolved.source_format or "lcov",
                         ingested_commit_sha=head_sha,
+                        mapping_partial=mapping_partial,
                     )
                     agg_matched = resolved.matched
                     console.print(
@@ -216,6 +219,14 @@ def coverage_add(
                         console.print(
                             f"[yellow]{len(skipped)} report file(s) did not map to the "
                             f"repo tree[/yellow] (e.g. {sample})."
+                        )
+                    if mapping_partial:
+                        console.print(
+                            "[red]More than half the report did not map to the repo "
+                            "tree.[/red] The stored coverage describes only a fragment "
+                            "of the repository — fix [cyan]coverage.strip_prefix[/cyan] "
+                            "or [cyan]coverage.path_prefix[/cyan] in "
+                            ".repowise/config.yaml and re-run."
                         )
 
             # --- Per-test map, from any report that carries contexts.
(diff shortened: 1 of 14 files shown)


RaghavChamadiya (approved):
Thanks @sloemo01. Measuring the loss against the report rather than against the matched set is the thing that makes this correct, and `test_mapping_partial_measured_on_report_files_not_matched_keys` is the test I would have asked for if it were not already here. Merging the same key across two reports shrinks `by_key` but not the denominator, and a ratio taken off `coverage_map` would have quietly called that case partial.

I checked the arithmetic against `resolve_reports` on main. Every `fc` lands in exactly one of `matched_exact`, `matched_suffix`, `ambiguous` or `unmatched`, with a `continue` on the miss branch, so the majority test is well defined and no report file escapes the count.

The split between the two call sites is the right shape and worth stating so it does not get flattened later. `_build_pipeline_coverage` warns and carries on, because an indexing run should not die over coverage. `repowise coverage add` returns False and exits non-zero, because that command's whole contract is that its exit status tells a complete ingest from an incomplete one. That is the behaviour #1746 is actually about, so I want it in the release notes as a change rather than as a fix: a majority-unmapped report that used to exit 0 will now exit 1 without `--strict`.

One non-blocking thing, for whenever you are next in this file. `report_file_count` is exactly the existing `ResolvedCoverage.total` property, since `total` is `matched + len(unmatched) + len(ambiguous)` and those are the only three destinations. `result.mapping_partial = result.matched * 2 < result.total` after the loop drops the counter and the two lines that maintain it. Not worth a push on its own.

`0056` is additive with a `server_default` and the `getattr(latest, "mapping_partial", False)` read means legacy rows report complete rather than partial, which is the safe direction. Full matrix green on f71d1314.

PR #1956, opened 2026-08-28
fix(doctor): explain stale pages and why --repair cannot clear them (#1710)

### LINKED ISSUE #1710: [Bug] `repowise doctor --repair` fails with strange `404` error
## Describe the Bug

`repowise doctor --repair` fails:

Image

## Steps to Reproduce

1. Run `repowise doctor --repair`
2. See error

## Expected Behavior

To work well or at least better explain the error because I have no idea how to fix it

## Actual Behavior

Strange 404 error for an unkown IP (Repowise serve is running in background at `http://127.0.0.1:3000`)

## Environment

- OS: `Windows 11`
- Python version: `3.12.1`
- Repowise version: `0.44.0`

## `config.yaml`

```yml
editor_files:
  claude_md: false
provider: ollama
model: maternion/lfm2.5:latest
embedder: ollama
reasoning: auto
distill:
  commands:
    enabled: false
wiki_style: caveman
commit_limit: 200
exclude_patterns:
- analysis_temp
- db-data
```

sloemo01:
## What

Fixes #1710. `repowise doctor --repair` reported `N stale` with no explanation of what stale means or what clears it — the 3rd run in the report ends "Nothing to repair" while 191 pages stay stale, which reads as the repair failing to do its job.

## Root cause

Stale pages are content lag, not store drift: `mark_stale_pages` decays pages hit by a change cascade but beyond the regeneration budget. `--repair` only reconciles SQL↔vector↔FTS mismatches, so it correctly leaves them alone — but neither the check detail nor the repair summary said so.

## Changes

- The "Stale pages" check detail now explains what stale means (content lags the code — the change cascade exceeded the regeneration budget), that `repowise update --full` regenerates them, and that `--repair` only fixes store drift.
- `--repair` now says the same explicitly: after repairing drift it notes stale pages are untouched by design; when there's no drift but stale pages remain, it says so instead of "Nothing to repair".

## Tests

`uv run pytest tests/unit/cli/test_doctor_*.py` → 37 passed. Full unit suite: 14,896 passed.

### DIFF
--- a/packages/cli/src/repowise/cli/commands/doctor_cmd/repo_checks.py
+++ b/packages/cli/src/repowise/cli/commands/doctor_cmd/repo_checks.py
@@ -231,7 +231,19 @@ def _run_repo_checks(
                 return 0
 
             stale_count = run_async(_check_stale())
-            checks.append(_check("Stale pages", stale_count == 0, f"{stale_count} stale"))
+            if stale_count:
+                checks.append(
+                    _check(
+                        "Stale pages",
+                        False,
+                        f"{stale_count} stale — pages whose content lags the code "
+                        "(change cascade exceeded the regeneration budget). "
+                        "`repowise update --full` regenerates them; "
+                        "`--repair` cannot, it only fixes store drift.",
+                    )
+                )
+            else:
+                checks.append(_check("Stale pages", True, "0 stale"))
         except Exception:
             checks.append(_check("Stale pages", True, "Could not check"))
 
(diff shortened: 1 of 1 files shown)


RaghavChamadiya (approved):
Thanks @sloemo01. The framing is what makes this land: stale is content lag and `--repair` reconciles store drift, so the old report was two unrelated numbers printed next to each other and read as a failure. Saying that in the check detail and again in the repair summary fixes the actual confusion in #1710.

I checked the `stale_count` reference in the repair branches, since it is assigned inside a `try` at the check site. It is initialized to 0 at `repo_checks.py:207` ahead of that block, so the "Could not check" path cannot reach the new lines with the name unbound.

PR #1957, opened 2026-08-28
fix(claude-md): rank workspace contract links by confidence before the cap (#1588)

### LINKED ISSUE #1588: [Bug] Workspace CLAUDE.md contract links are cut in insertion order, not by confidence
## Describe the Bug

The workspace `CLAUDE.md` caps its cross-repo contract links but doesn't rank them first, so which ones survive the cut is whatever order they happened to be deduped into.

`claude_md_cmd.py:198`:

```python
return list(best.values())[:_MAX_CONTRACT_LINKS]
```

`best` is a dict built while scanning the raw contract links, so `list(best.values())` is insertion order. Ten lines up in the same file, the co-change list does it correctly:

```python
co_changes = sorted(co_changes, key=lambda c: c.get("frequency", 0), reverse=True)
```

`ContractLink` carries `confidence`, and nothing reads it here.

## Steps to Reproduce

1. Set up a workspace with more provider/consumer contract links than `_MAX_CONTRACT_LINKS`.
2. Run `repowise workspace claude-md` (or whatever regenerates the workspace file).
3. The contract links section holds the first N found, not the N with the strongest evidence.

## Expected Behavior

The links that make the file are the highest-confidence ones, the same way co-changes are the highest-frequency ones.

## Actual Behavior

Insertion order decides. A high-confidence link discovered late in the scan gets dropped in favour of a weak one found early.

## Additional Context

This file is loaded as context by every agent session in the workspace, which makes it about the highest-visibility list we cut anywhere. It's also a small fix: one `sorted(...)` with a tiebreak on a stable field so the output doesn't shuffle between runs.

Heads up for whoever picks this up: there's active work on the workspace contract path right now (`feat/workspace-consumer-recall`), so it's worth checking that branch has landed before starting, or you'll be rebasing on top of changes to the same file.

sloemo01:
## What

Fixes #1588. The workspace `CLAUDE.md` caps its cross-repo contract links but didn't rank them first — `_curate_contract_links` returned `list(best.values())[:_MAX_CONTRACT_LINKS]` in insertion order, so which links survived the cut was whatever order the scan deduped them into. A high-confidence link discovered late got dropped for a weak one found early. This file is loaded as context by every agent session in the workspace, making it one of the highest-visibility lists we cut anywhere.

## Root cause

`claude_md_cmd.py:198` sliced a dict's values (insertion order) instead of sorting by evidence. `ContractLink` carries `confidence`, and nothing read it. The co-change list ten lines up already sorted by frequency — the contract list just never got the same treatment.

## Changes

- Sort the curated links by `confidence` descending before applying `_MAX_CONTRACT_LINKS`.
- Tiebreak on stable fields (`contract_id`, `consumer_repo`, `consumer_file`) so identical-confidence rows don't shuffle between runs.
- Links without a `confidence` field (pre-existing data) sort to the end but stay stable.

## Tests

Two new tests in `tests/unit/cli/test_claude_md_contract_curation.py`: weak-first-but-strong-wins across the cap, and no-confidence links rank last but stay run-stable. 7 passed; full unit suite: 14,896 passed.

### DIFF
--- a/packages/cli/src/repowise/cli/commands/claude_md_cmd.py
+++ b/packages/cli/src/repowise/cli/commands/claude_md_cmd.py
@@ -195,7 +195,22 @@ def _curate_contract_links(links: list[dict]) -> list[dict]:
             "alembic/versions/" in (current.get("provider_file") or "") and not is_migration
         ):
             best[key] = link
-    return list(best.values())[:_MAX_CONTRACT_LINKS]
+    curated = list(best.values())
+    # The cap cuts on evidence, not insertion order: rank by confidence so a
+    # high-confidence link discovered late in the scan is never dropped for a
+    # weak one found early (the co-change list already does the same by
+    # frequency). Tiebreak on stable fields so the output doesn't shuffle
+    # between runs.
+    curated.sort(
+        key=lambda link: (
+            float(link.get("confidence", 0.0) or 0.0),
+            link.get("contract_id") or "",
+            link.get("consumer_repo") or "",
+            link.get("consumer_file") or "",
+        ),
+        reverse=True,
+    )
+    return curated[:_MAX_CONTRACT_LINKS]
 
 
 def _generate_workspace(
--- a/tests/unit/cli/test_claude_md_contract_curation.py
+++ b/tests/unit/cli/test_claude_md_contract_curation.py
@@ -16,6 +16,7 @@ def _link(
     contract_id,
     provider_file="src/models.py",
     consumer_file="src/client.ts",
+    confidence=None,
 ):
     return {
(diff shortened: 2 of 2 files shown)


RaghavChamadiya (approved):
Thanks @sloemo01. The point that this file is loaded as context by every agent session in the workspace is the right argument for caring about which links survive the cut, and the co-change list ten lines up already sorting by frequency makes the inconsistency hard to defend.

Tiebreaking on `contract_id` / `consumer_repo` / `consumer_file` rather than leaving equal-confidence rows to dict order is the part that keeps this from trading one nondeterminism for another, and the no-confidence test pins that.

PR #1958, opened 2026-08-28
fix(config): surface broken repo config instead of silently using defaults (#852)

### LINKED ISSUE #852: [Bug] Missing warnings for swallowed configuration errors

sloemo01:
## What

Fixes #852. Configuration errors in the update path were silently swallowed: a malformed `.repowise/config.yaml` (or unreadable `.env`) resolved as "no config", so the user's provider/model/coverage settings silently vanished and every run used defaults — with no warning anywhere.

## Root cause

`load_repo_config` in `packages/core/src/repowise/core/repo_config.py` only caught `ImportError` (a YAML parse error propagated raw), and the server's `_load_repo_context` wrapped every failure in `except Exception: cfg = {}` — indistinguishable from "no config file". The CLI callers that don't wrap it (e.g. `update --full`) crash raw on a malformed file; the ones that do (server, workspace) go silent.

## Changes

- New typed `RepoConfigError` raised by `load_repo_config` (malformed YAML, non-mapping top level) and `load_repo_env` (existing-but-unreadable `.env`), with messages naming the file and the underlying parse error.
- The server's `_load_repo_context` now catches `RepoConfigError` explicitly and logs a warning naming the repo, instead of silently returning defaults. Absent config files still return `{}` — only *broken* files warn.

## Tests

`tests/unit/test_repo_config_errors.py` (4 tests): malformed YAML raises, non-mapping raises, absent config stays empty, unreadable `.env` raises. Config + server + provider suites: 86 passed; full unit suite: 14,896 passed.

### DIFF
--- a/packages/cli/src/repowise/cli/helpers.py
+++ b/packages/cli/src/repowise/cli/helpers.py
@@ -21,7 +21,7 @@ from repowise.core.reasoning import (
 from repowise.core.reasoning import (
     resolve_reasoning as resolve_core_reasoning,
 )
-from repowise.core.repo_config import CONFIG_FILENAME, load_repo_config
+from repowise.core.repo_config import CONFIG_FILENAME, RepoConfigError, load_repo_config
 
 # Update lock — coordinates concurrent `repowise update` invocations and lets
 # the augment hook suppress stale-wiki warnings while a refresh is in flight.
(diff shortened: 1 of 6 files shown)


RaghavChamadiya (approved):
Thanks @sloemo01. Introducing `RepoConfigError` rather than widening what the callers catch is the right shape: "no config file" and "the config file is broken" were the same empty dict, and a distinct type is what lets each call site pick its own volume. Keeping the warning on `err_console` so `--format json` stays parseable is the detail I would have expected to have to ask for.

I checked the `except ImportError` fallback survives. It sits ahead of the new `except Exception`, so a tree without PyYAML still gets the flat line parser instead of a hard error. That was the thing I most expected to be broken and it is not.

Two things, both non-blocking, and I am merging over them rather than sending you round again.

The `RepoConfigError` raised for a non-mapping YAML is caught by the function's own `except Exception` and re-wrapped, so the path appears twice. I ran it:

```
LIST   -> Could not parse ...\.repowise\config.yaml: Could not parse ...\.repowise\config.yaml: expected a YAML mapping, got list
BROKEN -> Could not parse ...\.repowise\config.yaml: while parsing a flow sequence
```

The malformed case reads fine, only the non-mapping one doubles. An `except RepoConfigError: raise` ahead of the broad clause fixes it.

Second, `config_path.read_text()` is above the `try`, so an unreadable `config.yaml` still leaves as a bare `OSError` and `load_config` in `helpers.py` only catches `RepoConfigError`. That is the exact case you just fixed for `.env`, in the opposite direction: unreadable `.env` now warns, unreadable `config.yaml` still traces back. It is pre-existing rather than something this PR broke, which is why it is not holding the merge, but the asymmetry is worth closing while the file is fresh.

Both fit in one small follow-up. Full matrix green on cbea7b85.

PR #1959, opened 2026-08-28
fix(dead-code): clamp unused-export confidence on unresolved dynamic use (#1910)

### LINKED ISSUE #1910: [Bug] Dynamic-use uncertainty can still produce high-confidence dead-code findings
## Summary

Dead-code analysis can represent "the static graph found no reference", but it cannot consistently represent "a dynamic-use mechanism may be the missing reference".

That distinction matters at the top confidence tier. A callback, registry value, decorated entry point, plugin, reflective lookup or string-dispatched symbol can have no ordinary import or call edge and still be live. When the defining path does not happen to match an existing runtime-risk filename heuristic, the finding can retain high confidence and appear deletion-ready.

The desired behavior is not to declare every dynamically-shaped symbol live. It is to distinguish:

- a proven static reference;
- a positive dynamic-use refutation;
- unresolved dynamic-use risk;
- a genuinely unsupported candidate.

## Contributor difficulty

**This is not a beginner-friendly issue.** It crosses graph evidence, confidence scoring, persistence, MCP serialization and several forms of dynamic dispatch. A contributor should be comfortable designing conservative static-analysis rules, preserving compatibility, and proving both false-positive removals and true-positive controls. It is deliberately labelled `help wanted`, not `good first issue`.

## Current mechanism

`DeadCodeFindingData` currently carries:

```python
confidence
reason
evidence
safe_to_delete
risk_factors
```

The existing risk-factor layer is deliberately conservative and useful, but it is primarily path-based. It recognizes config, bootstrap, database, environment, script and runtime-asset naming patterns and caps findings that look likely to be loaded outside the static graph.

That leaves dynamic use whose defining file has an ordinary path:

```python
registry["serializer"] = JsonSerializer
register_handler(on_payment)
@framework.command("sync")
plugin_class = getattr(module, configured_name)
callbacks.append(after_commit)
```

If the graph does not model the use, none of these forms necessarily contributes a risk factor. "No static reference found" can therefore become stronger than the evidence supports.

The response contract has the same compression problem. Summary counts identify findings, but do not distinguish:

- candidates examined;
- findings supported by positive unused evidence;
- findings downgraded because use is unresolved;
- candidates refuted by a dynamic reference;
- candidates suppressed entirely.

## Existing related work

This should not absorb the individual resolver bugs already tracked elsewhere.

- #640 covers Lambda string handlers and a Rust field-type case.
- #1554 covers guarded JSX false negatives.
- Previously fixed cases such as dynamic imports, FastAPI routes and Go function arguments remain regression controls, not work to reimplement.

Those issues establish concrete missing edges. This issue is about the confidence contract applied after graph evidence and refutation checks are assembled.

## Proposed shape

Introduce an explicit dead-code evaluation result for each candidate, with enough information to distinguish:

```text
candidate
positive unused evidence
refutations
unresolved dynamic-use risks
final confidence
suppressed or emitted
safe-to-delete decision
```

The exact internal representation is open, but three properties are required.

### 1. Refutations are first-class

A registry, callback, decorator, plugin declaration, reflection target or string-dispatch reference that can be resolved to the candidate must prevent a high-confidence unused verdict.

A refutation is stronger than a filename risk factor. It says a use was found, not merely that one may exist.

### 2. Unresolved dynamic use is monotonic

When a dynamic mechanism is present but cannot be resolved precisely, it may lower confidence and must clear `safe_to_delete`. It must never increase confidence or turn an existing review-only finding into a deletion-ready one.

### 3. Counts describe different populations

Expose distinct totals for candidates, emitted findings, supported findings, refuted candidates and suppressed/downgraded candidates. A refutation count must not be presented as an evidence count.

## Sealed fixture

The fixture should contain at least:

- a function passed as a callback;
- a class stored as a registry value;
- a decorator-registered entry point;
- a plugin or framework entry point;
- a reflective or string-dispatched reference;
- an exported symbol with no reference of any kind;
- an ordinary statically referenced control;
- two same-named symbols where only one is dynamically referenced.

For each dynamic form, include both:

- a resolvable case that refutes the finding;
- an unresolved case that caps confidence but does not invent a reference.

## Done looks like

- No known callback, registry, decorator, plugin or string-dispatch fixture appears as a high-confidence or deletion-ready false positive.
- A genuinely unused control still appears at the expected confidence.
- Refutation can only remove or downgrade a finding; it cannot create a new high-confidence finding.
- Candidate, evidence, refutation, downgrade, suppression and emitted totals are internally consistent.
- Single-repository and workspace responses preserve repository identity.
- Existing risk-factor behavior remains intact.
- The exact MCP payload is tested, not only the internal analyzer report.
- Live dogfood samples every removed or downgraded population and hand-reads controls that did not move.

## The trap

Do not implement this as a broad allowlist such as "decorated functions are live" or "anything stored in a dictionary is live."

Those rules would trade visible false positives for invisible false negatives. Admission must be tied to a defined dynamic-use mechanism, and unresolved cases should be downgraded rather than silently suppressed.

## Not in scope

- Implementing the Lambda configuration resolver from #640
- Implementing Rust field-type edges
- Guarded JSX reachability from #1554
- Changing MCP response budgets
- Deleting code reported by the analyzer
- Recalibrating unrelated risk scores

sloemo01:
## What

Fixes #1910. Dynamic-use uncertainty could still produce high-confidence dead-code findings: the unused-exports pass had no unresolved-dynamic-use clamp.

## Root cause

The unreachable-files pass caps confidence at `RISK_CAP_CONFIDENCE` (0.4) when a finding's *package* contains dynamic-dispatch machinery (`analyzer.py:986-988`), but `_detect_unused_exports` had no equivalent. A public symbol in a package whose source uses runtime dispatch (`importlib`, entry points, reflection, `getattr` keyed by a configured string) can be wired up without any static edge — a registry value, a callback, a plugin looked up by a name the graph never sees. The existing `file_dynamically_loaded` rescue only covered the *resolved* case (a `dynamic_uses`/`framework` edge the extractors minted); the unresolved case shipped at high confidence with `safe_to_delete=True`.

## Changes

- In `_detect_unused_exports`, clamp confidence to `RISK_CAP_CONFIDENCE` when the symbol's parent package is in `_dynamic_import_dirs` — the same package-level question the unreachable-files pass asks.
- Add matching evidence text naming the dynamic-import mechanism, so the downgrade is explainable in the finding.
- The finding stays in the report as a review candidate; `safe_to_delete` is cleared. Genuinely unused symbols in packages with no dynamic machinery keep their original confidence (verified by control cases).

## Tests

`tests/unit/dead_code/test_dynamic_use_uncertainty.py` — 6 cases: registry-value target downgraded, clean-package control keeps 0.7/`safe_to_delete`, callback registration downgraded, resolved `dynamic_uses` edge still refutes, same-named pair stays identity-aware, marker-vs-edge monotonicity. 6 passed; full unit suite 14,890 passed on the branch.

### DIFF
--- a/packages/core/src/repowise/core/analysis/dead_code/analyzer.py
+++ b/packages/core/src/repowise/core/analysis/dead_code/analyzer.py
@@ -1420,6 +1420,24 @@ class DeadCodeAnalyzer:
                 if risk_factors:
                     confidence = min(confidence, RISK_CAP_CONFIDENCE)
 
+                # Unresolved dynamic-use clamp — the same package-level question
+                # the unreachable-files pass asks (see ``_make_unreachable_finding``).
+                # A public symbol in a package whose source uses runtime-dispatch
+                # machinery (``importlib``, entry points, reflection, ``getattr``
+                # keyed by a configured string) can be wired up without any static
+                # edge: a registry value, a callback, a plugin looked up by a name
+                # the graph never sees. ``file_dynamically_loaded`` above only
+                # rescues the *resolved* case — a ``dynamic_uses``/``framework``
+                # edge the extractors actually minted. This clamp covers the
+                # unresolved case, where the mechanism is present but the target
+                # is not pinned: it may lower confidence and must clear
+                # ``safe_to_delete``, and it must never raise either. The finding
+                # stays in the report as a review candidate (the cap lands exactly
+                # on the default ``min_confidence``), exactly like the sibling
+                # clamp on unreachable files.
+                if self._dynamic_import_dirs and str(Path(node).parent) in self._dynamic_import_dirs:
+                    confidence = min(confidence, RISK_CAP_CONFIDENCE)
+
                 safe = confidence >= SAFE_CONFIDENCE_THRESHOLD
 
                 git_meta = self.git_meta_map.get(str(node), {})
@@ -1428,6 +1446,11 @@ class DeadCodeAnalyzer:
                 risk_line = risk_evidence(risk_factors)
                 if risk_line:
                     evidence.append(risk_line)
+                if self._dynamic_import_files and confidence <= RISK_CAP_CONFIDENCE:
+                    evidence.append(
+                        "Package uses dynamic imports or runtime-resolved edges, "
+                        "so the absence of a static reference does not establish disuse"
+                    )
 
                 findings.append(
                     DeadCodeFindingData(
--- a/tests/unit/dead_code/test_dynamic_use_uncertainty.py
+++ b/tests/unit/dead_code/test_dynamic_use_uncertainty.py
@@ -1,0 +1,386 @@
+"""Issue #1910 — dynamic-use uncertainty must cap unused-export confidence.
+
+The unreachable-files pass already clamps a finding whose *package* uses
+runtime-dispatch machinery (``importlib``, reflection, ``getattr`` keyed by a
+configured string — see ``test_dynamic_import_dirs.py``). The unused-exports
+pass had no equivalent: it only rescued a file a ``dynamic_uses``/``framework``
+graph *edge* pinned to it (the resolved case). When the mechanism was present
+but the target was not pinned — a registry value, a callback, a plugin looked
+up by a name the graph never sees — the symbol shipped at high confidence with
+``safe_to_delete=True``.
+
+Every unresolved case carries a control: a genuinely unused symbol in a package
+with no dynamic machinery must keep its original confidence, so a fix that
+clamps everything fails the suite.
+
+The scenario shapes come from the issue's sealed fixture:
+
+- a class stored as a registry value (``registry['serializer'] = cls``);
+- a function passed as a callback (``callbacks.append(after_commit)``);
+- a reflective / string-dispatched reference
+  (``getattr(module, configured_name)``) — the name never appears literally,
+  so the name-occurrence clamp (``name_occurrences``) cannot see it;
+- an exported symbol with no reference of any kind (the control);
+- two same-named symbols where only one sits in a dynamically-loaded package.
+"""
+
+from __future__ import annotations
+
+from pathlib import Path
+from types import SimpleNamespace
+
+import pytest
+
+from repowise.core.analysis.dead_code import (
+    DeadCodeAnalyzer,
+    DeadCodeKind,
+)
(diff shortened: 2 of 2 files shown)


RaghavChamadiya (approved):
Thanks @sloemo01. This is the right fix and the reason is that it is not a new rule, it is an existing rule applied to the pass that was missing it.

Checked both against the tree. The unreachable-files pass has carried this since it was written:

```python
# analyzer.py:987
if self._dynamic_import_dirs and str(Path(node).parent) in self._dynamic_import_dirs:
    confidence = min(confidence, RISK_CAP_CONFIDENCE)
# analyzer.py:1007
if self._dynamic_import_files and confidence <= RISK_CAP_CONFIDENCE:
    evidence.append(...)
```

and your addition to `_detect_unused_exports` is the same two conditions on the same two fields, in the same order. `_dynamic_import_dirs` is derived from `_dynamic_import_files` in `__init__`, so both passes are asking one question of one piece of state. That symmetry is worth more than the fix itself: the next person changing how dynamic use is detected changes it in one place and both passes follow.

The distinction the comment draws is the one that matters and I want it on the record, because it is the thing a future reader will otherwise re-litigate. `file_dynamically_loaded` above already rescues the **resolved** case, where an extractor actually minted a `dynamic_uses` or `framework` edge. What had no handling was the **unresolved** case: the machinery is present, the target is not pinned, and no static edge exists to find. A registry value, a callback appended to a list, a `getattr(module, configured_name)` where the name never appears literally in source. That last one also slips the `name_occurrences` clamp, so nothing else in the pipeline was catching it.

Clamping rather than suppressing is the correct severity. The finding stays in the report as a review candidate and only loses `safe_to_delete`, and capping at exactly the default `min_confidence` means it sits at the boundary rather than vanishing. For a tool whose worst failure is telling someone to delete live code, "I am not sure" is the honest answer here and it is the one this now gives.

The test file earns its 386 lines. Pairing every unresolved case with a control, a genuinely unused symbol in a package with no dynamic machinery that must keep its original confidence, is what stops the obvious wrong fix, clamping everything, from passing. The two same-named symbols where only one sits in a dynamically-loaded package is the case I would have missed writing this myself.

Green across the matrix, and the production change is 24 lines against 386 lines of test. Merging.

Issue #1709, opened 2026-08-18
[Bug] Running `repowise update --full` not rendering progress in CLI


yairEO:
## Describe the Bug

`repowise update --full` not showing any progress 

## Steps to Reproduce

1. Run `repowise update --full`
2. wait forever
3. See no progress and be confused what's going on while CPU is high and obviously the local LLM is working hard.

## Expected Behavior

Understand what's happening and see progress 

## Actual Behavior

Nothing interesting is shown. I have no idea how long will I be waiting. The high CPU usage greatly affects my work and I want to plan my day and not run the update only in the night, so I want to know if this will take an hour or 12 hours.. and how much progress was made already. 

It would already be reassuring to add a message that it's safe to abort at any time and the work will safely resume itself whenever `repowise update --full` will run again.

<img src="https://github.com/user-attachments/assets/7bcadb6b-8e3f-4554-835e-f3f72075d218" />

<img src="https://github.com/user-attachments/assets/12c40a09-e016-4c92-b8bc-3ce24e22abd0" />

## Environment

- OS: `Windows 11`
- Python version: `3.12.1`
- Repowise version: `0.44.0`
- Installation method: `uv tool upgrade repowise`

RaghavChamadiya:
Thanks @yairEO, the CPU screenshot is what made this quick to pin down, and the question about whether it's safe to abort turned out to be the important one.

First, a correction to the framing, because the number the CLI showed you is misleading. That "3661 pages" is every page in the plan, and the large majority of them never reach a model. `_render_file_page` in `packages/core/src/repowise/core/generation/page_generator/core.py:311` says it plainly: "Render a file page from structure. The only renderer for this type. Built straight from the assembled context via a Jinja template: no provider call, no tokens." Symbol spotlights go the same way. On this repo's own index, 4,184 of 4,282 pages are rendered from structure and only 98 are model written, about 2.3%.

So your run is roughly 90 to 100 model calls, not 3,661. They are the expensive ones though. A module page budgets 45k input and 6k output tokens, so ~92 of them is several million tokens through a local model, which is why llama-server sits pinned for hours. It is working, not wedged.

[...]

PR #1961, opened 2026-08-28
fix(generation): persist a subject reuse key so interrupted runs can be topped up (#1089)

### LINKED ISSUE #1089: A run interrupted by a provider outage can never be topped up: pages are always regenerated, and the missing ones are unreachable
# A run interrupted by a provider outage can never be topped up: pages are always regenerated, and the missing ones are unreachable

**Version:** 0.34.0 (behaviour re-checked against `main` — see below, both causes are still there)
**Provider:** `codex_cli`, later `opencode`
**Repo:** ~1,180 files, TypeScript/SQL, 768 pages generated

## What happened

A full `repowise init` ran for 106 minutes. After ~95 minutes the Codex subscription session died (`codex exec exited with 1`, then `403 Forbidden` on `wss://chatgpt.com/backend-api/codex/responses`). The remaining **105 pages failed** — and because the failures came at the end, they were the whole upper half of the wiki:

| page type | failed |
|---|---|
| file_page | 58 |
| module_page | 22 |
| layer_page | 8 |
| infra_page | 6 |
| onboarding | 5 |
| scc_page | 4 |
| repo_overview | 1 |
| architecture_diagram | 1 |

The wiki now contains only `file_page` and `symbol_spotlight`. No overview, no onboarding, no module or layer pages.

Trying to fill that gap surfaced two separate problems.

## Problem 1 — `repowise update` cannot fill the gap

```
$ repowise update --dry-run
Already up to date.
```

`update` diffs *files changed since the last sync*. No file changed, so it does nothing — even though 105 planned pages were never persisted. There is no path that says "generate what is missing". The only way to obtain those pages is `init`, which brings us to:

## Problem 2 — `init` regenerates every existing page

`--force` is documented as *"Regenerate all pages, ignoring existing"*, which implies that without `--force` existing pages are kept. In practice they are regenerated every time.

Measured: an `init` on an unchanged repo regenerated pages at ~40 s each, re-spending on all 768 pages to reach the 105 that were missing.

The cross-run reuse gate is in `core/generation/page_generator/core.py`:

```python
if prior is not None and prior.model_name == self._provider.model_name:
    reuse = bool(content_hash) and prior.content_hash == content_hash
    if not reuse:
        reuse = prior.source_hash == compute_source_hash(user_prompt + source_salt)
```

Two things keep it from ever firing across processes:

1. **`prior.content_hash` is always empty.** `load_prior_pages()` (`core/persistence/crud/pages.py`) constructs `PriorPage` without `content_hash`, and the `wiki_pages` table has no such column, so the preferred reuse key can never be populated from the database. Only the fallback remains.
2. **The fallback compares the rendered prompt hash**, and the prompt embeds RAG context that is rebuilt each run. The code's own comment says this "almost never matches on a reindex" — which matches what we observed.

Additionally, `prior.model_name == self._provider.model_name` means **any provider change invalidates the whole wiki**. This is exactly the situation an outage creates: the original provider is gone, so the user must switch — and then 0 of 768 pages are reusable, including the ones the dead provider successfully wrote.

`main` still has both causes: the reuse block now checks only `model_name` + prompt `source_hash` (the `content_hash` branch is gone), and `load_prior_pages` still does not carry a content hash.

### Suggested fix

- Persist `content_hash` on `wiki_pages` and populate it in `load_prior_pages`, so the stable key actually works across processes.
- Treat a model mismatch as a reason to *re-verify*, not to discard — or at least let a run top up only the pages that are absent, independent of which model wrote the others.
- Make the missing-page set reachable: have `init`/`update` diff the planned page set against the persisted one and generate only the difference.

## Problem 3 — `--language` does not reach deterministic tier-2 pages

On a repo configured with `language: de`, 623 of 768 pages are the deterministic tier-2 pages rendered from `templates/file_page_tier2.j2`. That template hardcodes English (`## Overview`, `## Public API`, `## Dependencies`, `## Usage Notes`, `It exposes N public symbols`, `Imported by N files…`). No model is involved, so `--language` never touches them — about 80 % of the wiki stays English while the LLM pages are translated.

### Suggested fix

Pass the resolved language into the template render context and select labels from a per-language map, keeping English as the fallback. In our copy, one line in `_render` plus a label map in the template was enough for the whole set of tier-2 pages to follow `language`.

## Problem 4 (minor) — tier-2 "Dependencies" lists the file's own symbols

`context/assembler.py:140` fills `dependencies` from `graph.successors(path)`. A file node also points at the symbols it defines, so every tier-2 page lists its own symbols as dependencies:

```
## Dependencies
- `lib/api/country-rules.ts::CACHE_TTL_MS`
- `lib/api/country-rules.ts::normalizeCC`
- `lib/api/country-rules.ts::__module__`
```

Affects all 623 tier-2 pages here. Filtering successors whose node id is prefixed with the file's own path would fix it.

## Problem 5 (minor) — agent-CLI providers absorb the repo's agent instructions

`codex_cli` and `opencode` run the agent inside the repository (`codex exec --cd <repo>`, `opencode run --dir <repo>`), so the agent loads `AGENTS.md` / `CLAUDE.md` and treats them as task instructions. Two effects show up in the stored pages:

- narration of its own tool use: `"I'll examine the actual source file to write an accurate spotlight page. Let me check…"` before the first heading — 25 of 145 LLM pages here.
- adoption of unrelated project workflow: `"I'm using the repository's Beads workflow guidance to ensure this one-shot documentation task follows the project's tracking conventions."`

A documentation render is not a coding task in that repo. Instructing the provider to ignore repository agent-instruction files for this call — and/or trimming everything before the first heading — would keep the pages clean.

## Why this combination hurts

Each problem alone is survivable. Together they mean: **a long run that dies partway through cannot be repaired.** The pages that were paid for cannot be kept, the pages that are missing cannot be requested, and switching away from the provider that died guarantees a full re-spend.

sloemo01:
## What

Fixes #1089. A run interrupted by a provider outage can never be topped up: pages are always regenerated, and the missing ones are unreachable.

## Root cause

The cross-run reuse gate in `PageGenerator._call_provider` keyed exclusively on the freshly-rendered prompt's `source_hash`. The prompt drifts run to run, because RAG context is rebuilt and populated concurrently each run, so on a full re-run the gate almost never matched. Every unchanged page was re-billed, and the pages a failed run left behind were never reachable as a cheap top-up: a stub row was either regenerated from scratch or hidden entirely. The maintainer's guidance in the issue thread named the shape of the fix, a stable subject key plus routing `repowise generate` through the same reuse path.

## Changes

- **Persisted subject reuse key**: new `wiki_pages.content_hash` column (Alembic `0067`, chained after `0065`). Computed as `sha256(subject_material : renderer_fingerprint)`: the file's byte hash, a group's member list, or the repo name, folded with the prompt-template source, system prompt, language, style, and a hand-bumped generation version. Unlike `source_hash` it survives RAG-context drift, so an unchanged page is reused instead of re-billed.
- **Reuse gate prefers the subject key**: `_call_provider` matches on `content_hash` when both sides have one (falling back to the prompt hash for pre-key pages), and never reuses a row a model did not write. Rows with `provider_name == 'template'` (keyless stubs, or stubs a failed provider call left) always re-call the model, so a top-up actually completes the missing pages.
- **`repowise generate` feeds prior pages into the gate**: `scoped_generation` loads prior pages and passes them through, so scoped refills reuse unchanged siblings instead of re-billing. A load failure degrades to no reuse, never a failed run.
- Structural pages get the same stable-key treatment via `_structural_content_hash`, and `--force` / workspace / upgrade call sites pass the key through.

## Rebased onto main

Main moved a long way under this branch, and two conflicts needed real decisions, both in the CLI.

- `init_cmd/command.py`, `init_cmd/generation.py`, `init_cmd/workspace.py`: main has since added `timings` and `warnings` parameters to the same signatures this branch touched. Both sides are kept.
- `upgrade_flow.py`: main replaced the buffered `run_generation` call with `run_generation_with_persistence` (issue #1709), and that wrapper already owns `prior_pages` and exposes `reuse_prior_pages`. Main's wrapper is kept, and this branch's manual prior-pages load is dropped as redundant.

The migration is renumbered to `0067`, chained after main's head `0065`. An earlier commit on this branch also added `0056_coverage_mapping_partial.py` purely to give `0057` a valid parent. Main now ships that same migration as `0060_coverage_mapping_partial.py`, so the branch's copy is deleted rather than renumbered.

On numbering: main's head is `0065` and PR #2340 adds `0066`, so this branch takes `0067` rather than colliding on a revision id. Two migrations based on the same parent do leave two Alembic heads if both merge, which is worth chaining at merge time.

## Tests

- `test_reuse_gate.py` (6): subject-key hit, template-stub never reused, fallback to prompt hash, fingerprint change forces regen, empty material falls back.
- `test_content_hash_roundtrip.py` (3): column persists, round-trips through `load_prior_pages`, legacy rows default empty.
- `test_scoped_generation_prior_pages.py` (2): `generate` passes prior pages in, load failure degrades.
- `test_generation_persist.py` (5): key survives persist.

After the rebase, those files plus `test_structural_salt.py` give 38 passed. Reverting `crud/pages.py` to main fails 3 of the roundtrip and pipeline tests, and restoring it leaves the tree clean. The broader CLI suite is 2568 passed with 1 failed, that failure being the pre-existing `test_agent_target_baseline.py::test_integration_writes_match_baseline`, whose baseline was recorded on Windows.

### DIFF
--- a/packages/cli/src/repowise/cli/commands/init_cmd/command.py
+++ b/packages/cli/src/repowise/cli/commands/init_cmd/command.py
@@ -266,6 +266,7 @@ def _run_generation_phase(
     embedder_name_resolved: str,
     resume: bool,
     test_run: bool,
+    reuse_prior_pages: bool = True,
 ) -> tuple[bool, bool]:
     """Run the LLM generation phase for a single-repo init.
 
@@ -273,6 +274,9 @@ def _run_generation_phase(
     and the caller should return immediately; ``cost_declined`` is True when the
     user declined the cost gate (generation skipped, index still saved). Mutates
     ``result`` in place with the generated pages, vector store, and enriched KG.
+
+    ``reuse_prior_pages=False`` is ``--force``: every page is regenerated
+    instead of reused from the prior run (issue #1089).
     """
     from repowise.core.cost_estimator import STRUCTURAL_PAGE_TYPES
     from repowise.core.generation import GenerationConfig
@@ -395,6 +399,7 @@ def _run_generation_phase(
         resume=resume,
         verbose=True,
         test_run=test_run,
+        reuse_prior_pages=reuse_prior_pages,
     )
     return False, False
 
@@ -1411,6 +1416,9 @@ def init_command(
             # say nothing, so a template wiki is never a run to continue.
             resume=resume and _prior_docs_mode != "deterministic",
             test_run=test_run,
+            # ``--force`` regenerates every page: the reuse gate is off, so
+            # nothing is carried over from the prior run (issue #1089).
+            reuse_prior_pages=not force,
         )
         if gen_stop:
             return
--- a/packages/cli/src/repowise/cli/commands/init_cmd/generation.py
+++ b/packages/cli/src/repowise/cli/commands/init_cmd/generation.py
@@ -259,12 +259,17 @@ def run_repo_generation(
     resume: bool,
     verbose: bool,
     test_run: bool = False,
+    reuse_prior_pages: bool = True,
 ) -> list[Any]:
     """Generate wiki pages for one repo and enrich its knowledge graph.
 
     Builds the embedder + vector store + cost tracker, runs the resume-friendly
     generation wrapper, enriches the KG, and flushes buffered cost rows in one
     transaction (kept out of the contended generation window, issue #326).
+
+    ``reuse_prior_pages=False`` (``repowise init --force``) disables the
+    cross-run reuse gate so every page is regenerated, as the flag's help
+    text has always promised (issue #1089).
 
     Mutates ``result`` in place with ``generated_pages``, ``preserved_page_ids``
     (the pages a resumed run skipped, which persistence must not sweep) and
@@ -342,6 +347,7 @@ def run_repo_generation(
                 progress=gen_callback,
                 resume=resume,
                 preserved_page_ids=preserved_page_ids,
+                reuse_prior_pages=reuse_prior_pages,
                 cost_tracker=cost_tracker,
                 generation_config=gen_config,
                 # In-memory curated modules: on a fresh init the
--- a/packages/cli/src/repowise/cli/commands/init_cmd/workspace.py
+++ b/packages/cli/src/repowise/cli/commands/init_cmd/workspace.py
@@ -97,6 +97,7 @@ def _run_workspace_generation(
     onboarding: bool = True,
     wiki_style: str = DEFAULT_STYLE,
     language: str = "en",
+    reuse_prior_pages: bool = True,
 ) -> list[Any]:
     """Run LLM generation for a single repo in the workspace init flow.
 
@@ -104,6 +105,9 @@ def _run_workspace_generation(
     the user declines the cost gate (caller persists the index without docs);
     other errors propagate so the caller can log per-repo failures without
     aborting the whole workspace run.
+
+    ``reuse_prior_pages=False`` is ``--force``: every page is regenerated
+    instead of reused from the prior run (issue #1089).
     """
     from repowise.core.generation import GenerationConfig
 
@@ -165,6 +169,7 @@ def _run_workspace_generation(
         resume=resume,
         verbose=False,
         test_run=test_run,
+        reuse_prior_pages=reuse_prior_pages,
     )
 
 
@@ -415,6 +420,7 @@ def _ingest_and_generate_repo(repo: Any, idx: int, total: int, ctx: _WorkspaceCt
                 onboarding=ctx.onboarding,
                 wiki_style=ctx.wiki_style,
                 language=ctx.language,
+                reuse_prior_pages=not ctx.force,
             )
             result.generated_pages = generated_pages
             # (result.vector_store is set inside _run_workspace_generation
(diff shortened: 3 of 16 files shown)


sloemo01:
Rebased onto main (`b78b88052`). Two things changed in the process: the migration is now 0066 (it was 0057, which collided with main's 0057) and chains after 0065, and the branch no longer carries its own copy of the coverage-mapping migration since main has that as 0060. The upgrade path also stopped threading prior pages itself, because `run_generation_with_persistence` on main already loads and passes them, so only the subject-key changes remain there.
Reuse-gate, content-hash roundtrip, scoped-generation, generation-persist and structural-salt tests: 38 passed. Migration chain resolves to a single head off 0065.

sloemo01:
The red leg is a timing flake rather than this change.

`tests/unit/server/mcp/test_number_precision.py::test_real_tool_payload_rounds_a_seeded_raw_double` asserts the `get_risk` payload carries no raw doubles, and the value it caught is `_meta.timing_ms`. That is a wall-clock measurement, so it only fails when a run happens to land on a non-integral millisecond; here it was 1077.85.

Three things point the same way:

- The file passes 33/33 locally on this branch, three runs in a row.
- This PR touches 14 files and none of them are on the MCP payload path.
- The 3.12 and 3.13 legs are green on this run.

I cannot retrigger the job without admin rights, so a re-run should clear it.

Worth noting separately: the assertion is timing-dependent by construction, and `timing_ms` is the one field in that payload that cannot be made stable. If you would rather it stopped flaking on others, rounding `timing_ms` at the source would fix it for everyone.

PR #1962, opened 2026-08-28
feat(ingestion): index Razor/Blazor markup (.razor, .cshtml) into the graph (#1404)

### LINKED ISSUE #1404: [Bug] Razor / Blazor markup is not indexed, causing missing graph edges and false-positive dead code
## Summary

Razor and Blazor markup files (`.razor` and `.cshtml`) do not appear to be indexed as code files.

For example:

```text
get_context("Components/Orders.razor")
→ Target not found
```

while the corresponding `.razor.cs` file is indexed.

This is more than a documentation issue: dependencies and usages that exist only in Razor markup are absent from the graph, which can cause false-positive dead-code findings and incomplete context/impact analysis.

## Why this matters

A typical Blazor page using Radzen may contain:

```razor
@inject OrderService OrderService

<RadzenAlert AlertStyle="AlertStyle.Warning">
    Some orders require attention.
</RadzenAlert>

<RadzenDataGrid Data="@orders" TItem="Order">
    <Columns>
        <RadzenDataGridColumn TItem="Order" Property="Name" Title="Name" />
    </Columns>
</RadzenDataGrid>

<RadzenButton Text="Save" Click="@SaveOrders" />
```

This markup contains real relationships:

```text
Orders.razor → RadzenAlert
Orders.razor → RadzenDataGrid
Orders.razor → RadzenDataGridColumn
Orders.razor → OrderService
Orders.razor → SaveOrders
Orders.razor → Order.Name
```

If `.razor` files are not parsed, these relationships are absent.

As a result, C# methods, properties, services, or components used only from Razor markup may incorrectly appear unused.

The existing `.razor`, `.razor.cs`, `.cshtml`, and `.cshtml.cs` never-flag patterns protect the markup files themselves, but they do not restore outgoing dependency edges from markup.

## Expected behavior

Ideally `.razor` and `.cshtml` files should participate in the index and dependency graph, including relationships such as:

* `<RadzenDataGrid />` → component type
* `@inject IService` → service/interface
* `Click="@Handler"` → handler
* `@bind-Value="Property"` → property
* `Service.Method()` → C# method/type
* `.razor` ↔ `.razor.cs`

This would improve `get_context`, dead-code detection, dependency graphs, and impact analysis for ASP.NET / Blazor applications.

sloemo01:
## What

Fixes #1404. Razor/Blazor markup (`.razor`, `.cshtml`) is now indexed: the files enter the graph with symbols and call edges, and the dead-code false positives they produced are gone.

## Root cause

`.razor`/`.cshtml` had no language spec — the extension→language map is built from `languages/specs/__init__.py` + `models.py`'s `LanguageTag`, and neither contained razor. The traverser classified these files `unknown`, the parser returned empty `ParsedFile`s, and the `*.razor`/`*.cshtml` never-flag globs in `dead_code/constants.py` suppressed the resulting false positives while restoring **zero edges**. Verified on main: `EXTENSION_TO_LANGUAGE[".razor"]` missing; the issue's `Orders.razor` parsed to 0 symbols / 0 calls.

## Changes

Follows the maintainer's map from the issue thread (first-PR scope):

- **`ingestion/sfc_source.py`** — a byte-scanning `Locator` (there is no tree-sitter-razor; an HTML grammar mis-parses `List<Order>` as a tag). Brace-matches `@code`/`@functions`/`@{ }` interiors (skipping `@@` escapes), records PascalCase tags as component calls, skips `<` inside C# regions so generics never mint bogus edges. `Locator` gained a `byte_scan` shape; `_cached_scan` dispatches on it, so the parser, complexity walker, dataflow parsing and duplication tokenizer all pick it up via `prepare_source`.
- **`languages/specs/razor.py`** (new) — `LanguageSpec`: `.razor`+`.cshtml`, `shares_grammar_with="csharp"`, `scm_file="csharp.scm"`, C#'s builtins/heritage/manifest/lock data, `import_support="none"` (honest: `@using` isn't projected yet). Registered after `_CSHARP` (load order is load-bearing for `shares_grammar_with`).
- **`models.py`** — `"razor"` added to `LanguageTag` (drives traverser + `EXTENSION_TO_LANGUAGE`).
- **`language_configs.py`** — razor aliased to the C# config.
- **`synthetic_symbols`** — razor joins the SFC component-symbol pass (file = component, stem as name).
- **`complexity/languages.py` + `perf/dialects/__init__.py`** — razor aliased to the C# node map / perf dialect (dataflow skipped — C# has no dialect).
- **`traverser.py`** — removed `.razor`/`.cshtml` from `_REFERENCE_BEARING_EXTENSIONS` (they now have a spec).
- **`docs/layers/LANGUAGE_SUPPORT.md`** — new "Razor / Blazor markup" section.

Result on the issue's exact example: `symbols: [Orders/class]`, `calls: [SaveOrdersAsync@21, RadzenAlert@4, RadzenDataGrid@8, RadzenDataGridColumn@10, RadzenButton@14]`, `parse_errors: []` — and no false `Order` edge. `get_context("Components/Orders.razor")` now resolves.

## Known scope (deferred per maintainer)

`@code` members are call edges but not yet symbols (the class-wrapper is the maintainer's explicitly-deferred second PR); `@using`/`@inject`/`@bind`/attribute expressions aren't projected; dead-code never-flags untouched per the maintainer's explicit instruction.

## Tests

- `test_sfc_source.py` — new `TestRazor*` blocks (offset invariants, blanking, fencing, component tags, degradation): 90 passed.
- `test_razor_extractors.py` (new, mirrors `test_svelte_extractors.py`): end-to-end symbols/calls, `.cshtml`, health walker: 9 passed.
- Affected suites (ingestion + dead-code + health): 3,945 passed, 1 pre-existing environmental failure (`test_repo_totals_churn` — fails on clean main too). Ruff clean on all changed files.

### DIFF
--- a/README.md
+++ b/README.md
@@ -473,7 +473,7 @@ the orchestrators. Full matrix and the contributor recipe:
 
 ## Supported languages
 
-**19 languages parsed to AST · 35 on a five-rung ladder · framework-aware across
+**20 languages parsed to AST · 36 on a five-rung ladder · framework-aware across
 all of them.**
 
 "Do you support X" has five useful answers, not two, so languages land on a
@@ -507,6 +507,7 @@ ladder and every rung says what it buys you.
   <img src="https://img.shields.io/badge/Delphi-EE1F35?style=flat-square&logo=delphi&logoColor=white" alt="Object Pascal / Delphi" />
   &nbsp;<strong>· Partial &nbsp;</strong>
   <img src="https://img.shields.io/badge/Luau-00A2FF?style=flat-square&logo=lua&logoColor=white" alt="Luau" />
+  <img src="https://img.shields.io/badge/Razor-512BD4?style=flat-square&logo=blazor&logoColor=white" alt="Razor / Blazor" />
 </p>
 
 Below those two rungs the ladder keeps going, and a language on a lower rung is
@@ -516,7 +517,7 @@ still doing real work rather than being ignored:
 |---|---|---|
 | **Full** (13) | Python · TypeScript · JavaScript · Svelte · Vue · Java · Kotlin · Go · Rust · C++ · C# · Scala · Ruby | The whole pipeline: AST symbols, import resolution, a resolved call graph, heritage, docstrings, framework edges, **and code-health markers** |
 | **Good** (5) | C · Swift · PHP · Dart · Object Pascal | All of the above except the full health suite |
-| **Partial** (1) | Luau / Roblox | AST symbols and `require()` resolution, Rojo and `.luaurc` aware |
+| **Partial** (2) | Luau / Roblox · Razor / Blazor | Luau: AST symbols and `require()` resolution, Rojo and `.luaurc` aware. Razor: component symbols, `@code` and component-tag call edges, C# health markers; no import resolution yet |
 | | | ⎯⎯ *tree-sitter parsing stops here; the rungs below come from git and imports* ⎯⎯ |
 | **Lightweight** (7) | Elixir · Clojure · Haskell · Lean 4 · Erlang · F# · HTML | A real file-to-file import graph, and no symbol-level claims |
 | **Structural** (9) | Objective-C · R · Zig · Julia · Elm · OCaml · Crystal · Nim · D | Git history: blame, hotspots, co-change, ownership, bug history |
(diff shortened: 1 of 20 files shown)


RaghavChamadiya (approved):
Hardened on the branch and validated on eShopOnWeb; see the comment above.

PR #1971, opened 2026-08-29
fix(risk-semantics): correct temporal_hotspot contract to unbounded (…

### LINKED ISSUE #1970: [Bug] direct_risks.temporal_hotspot is an unbounded sum, but its own scale contract declares it a 0-1 ratio
## Summary

`BlastRadiusResponse.structural_impact_scale` declares
`direct_risks.temporal_hotspot` to be a `normalized_component_score`, unit
`ratio`, range 0.0–1.0. The value emitted in that field is an unbounded sum of
per-commit churn. On this repository 40.2% of tracked files exceed 1.0 and the
maximum is 23.2, so the published contract is wrong for two files in five.

## Mechanism

The field is computed as a decayed sum in
`ingestion/git_indexer/file_history.py:359-367`:

```python
        _ln2 = math.log(2)
        temporal_score = 0.0
        for c in commits:
            age_days = max((now.timestamp() - c.ts) / 86400.0, 0.0)
            weight = math.exp(-_ln2 * age_days / HOTSPOT_HALFLIFE_DAYS)
            lines = min((c.added + c.deleted) / 100.0, 3.0)
            temporal_score += weight * lines
        meta["temporal_hotspot_score"] = temporal_score
```

Each commit contributes up to `3.0`, and nothing divides by a count or a
maximum. With `HOTSPOT_HALFLIFE_DAYS = 180.0` a file touched repeatedly with
large diffs accumulates without bound.

`pr_blast.py:203-212` passes it through with only rounding:

```python
            temporal = float(getattr(meta, "temporal_hotspot_score", 0.0) or 0.0)
            ...
                    "temporal_hotspot": round(temporal, 4),
```

while `analysis/risk_semantics.py:279-283` publishes the contradicting claim:

```python
                "direct_risks.temporal_hotspot": {
                    "kind": "normalized_component_score",
                    "unit": "ratio",
                    "range": {"minimum": 0.0, "maximum": 1.0},
                },
```

## Scale of the mismatch

Measured over this repository's own index, restricted to files that exist in the
working tree (3,638 files with a `git_metadata` row):

| | |
|---|---|
| `temporal_hotspot_score` > 1.0 | 1,464 files (40.2%) |
| > 5.0 | 83 files |
| > 10.0 | 11 files |
| maximum | 23.214 (`mcp_server/tool_health.py`) |

This is not a tail effect. The declared range is exceeded by a factor of 23 at
the top and by any amount at all for two files in five.

## Why the contract is the thing that broke

`temporal_hotspot_score` has always been an unbounded decayed sum, and its two
existing consumers read it correctly as one: `enrich.meets_hotspot_floors`
compares it against `HOTSPOT_MIN_TEMPORAL_SCORE`, and
`enrich.compute_percentiles` uses it only as a sort key before deriving the
genuinely normalised `churn_percentile`. Neither treats it as a ratio.

The defect is that the scale metadata added alongside `structural_impact_scale`
described it as one. Publishing the contract is what turned a private
intermediate into a public claim, and the claim does not match the value.

## Blast radius

`structural_impact_score` multiplies by this field
(`structural_score = pagerank * (1 + temporal_hotspot)`, `pr_blast.py:221-223`),
so a client that trusts the declared range will under-estimate the file term by
whatever factor the sum has run past 1.0 — up to 24x on the measured maximum.
Any consumer rendering the field as a percentage is wrong by the same factor.

## Fix

Two options, and they are not equivalent:

1. **Correct the contract** to describe what is emitted: `kind:
   "raw_decayed_churn"`, `unit: "decayed_line_hundreds"`, `range: {minimum: 0.0,
   maximum: None}` — matching how `direct_risks.structural_score` already
   declares itself unbounded.
2. **Normalise the value**, e.g. emit `churn_percentile` (already computed, and
   genuinely 0–1) in this field instead.

Option 2 changes `structural_impact_score` for every repository and would need
the band thresholds re-checked against the new range, since the file term
currently rides on this multiplier. Option 1 is the honest, non-breaking change
and is what the sibling `structural_score` field already does.

akshatmalik-bruh:
## Summary

- `direct_risks.temporal_hotspot` was declared `kind: normalized_component_score / unit: ratio / range: 0–1` but the emitted value is an unbounded exponentially-decayed churn sum (40.2% of files exceed 1.0; max observed: 23.2)
- Corrects the `structural_impact_scale` contract in `risk_semantics.py` to `kind: raw_decayed_churn / unit: decayed_line_hundreds / range: max: None`, mirroring how the sibling `direct_risks.structural_score` field already declares itself unbounded
- No computation paths changed (`file_history.py`, `pr_blast.py`, `enrich.py` untouched) — the contract was the only thing wrong

## Related Issues

Fixes #1970

## Test Plan

- [x] Tests pass (`pytest`) — 12 new pure-unit contract tests in `tests/unit/analysis/test_temporal_hotspot_contract.py`, all green; pre-existing `test_dynamic_use_confidence` failure is unrelated and present on `main`

## Checklist

- [x] I have added tests for new functionality — `test_temporal_hotspot_contract.py` holds the corrected contract and explicitly asserts the old `ratio`/`1.0` values are gone, so CI catches any future regression
- [x] I have updated documentation if needed — the `note` field in the contract points consumers to `churn_percentile` for a normalised 0–1 alternative


RaghavChamadiya (approved):
Fixed the contract rather than the value, which was the call on the issue, and all three surfaces move together: the runtime dict, the `DirectRiskEntry` JSDoc, and the CHANGE_RISK.md table row. A consumer reading any one of them now gets the same answer.

The regression class asserting the old values are gone is worth keeping. `normalized_component_score` / `ratio` / `maximum: 1.0` is exactly the shape someone re-adds by pattern-matching the neighbouring fields, and pointing the note at `churn_percentile` gives them the normalised alternative in the same breath.

PR #1979, opened 2026-08-29
fix(core): validate REPOWISE_EMBEDDING_DIMS at parse points (#826)

jyotirmya17:
Hi @RaghavChamadiya, thanks for the guidance! As requested, I closed #842 and opened this fresh branch off `main` to address the remaining hole.

This focuses entirely on validating the numeric env vars at their parse points:
1. `_server.py` and `app.py`: `REPOWISE_EMBEDDING_DIMS` is now properly validated. If it's negative or unparseable, it falls back to 768 and prints a warning to `stderr` (matching the `resolve_embedding_timeout` pattern).
2. Direct-constructor guards added for `OllamaEmbedder(dimensions=-5)` to raise `ValueError`.
3. Added `test_embedding_dims_validation.py` to cover both the parse fallback behaviour and the constructor rejection.

Let me know if this hits the mark!

### DIFF
--- a/packages/core/src/repowise/core/providers/embedding/ollama.py
+++ b/packages/core/src/repowise/core/providers/embedding/ollama.py
@@ -74,7 +74,20 @@ class OllamaEmbedder:
         env_dimensions = os.environ.get("OLLAMA_EMBEDDING_DIMS") or os.environ.get(
             "REPOWISE_EMBEDDING_DIMS"
         )
-        self._requested_dimensions = dimensions or (int(env_dimensions) if env_dimensions else None)
+        resolved_dims: int | None = None
+        if dimensions is not None:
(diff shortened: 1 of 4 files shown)


RaghavChamadiya (approved):
Scoped as asked. It guards `_server.py`'s `_embedder_kwargs`, the same hole in `app.py:_build_embedder`, and the Ollama constructor, and it does not reinstate the `base.py` raise that #826 ruled out. Falling back to 768 with a warning rather than raising is the right shape here: a malformed env var should not take down a server that has a perfectly good default.

Sorry this sat unreviewed since 29 August. That is on me, not on the PR.

Non-blocking: the `import sys` inside the else branch in `app.py` could go to the top of the module with the others.

PR #1990, opened 2026-08-29
fix: map boolean reasoning config values to valid modes

### LINKED ISSUE #1989: [Bug]: reasoning: false in config.yaml raises ValueError instead of disabling reasoning
## Describe the Bug

Writing `reasoning: false` in `.repowise/config.yaml` crashes with a `ValueError` instead of disabling reasoning. Same thing happens with `reasoning: true`.

## Steps to Reproduce

1. Create `.repowise/config.yaml`:

   ```yaml
   reasoning: false
   ```

2. Run any command that reads the config:

   ```bash
   repowise init
   ```

3. You get:

   ```
   ValueError: reasoning must be one of: auto, off, none, minimal, low, medium, high, xhigh, max
   ```

## Expected Behavior

`reasoning: false` should just work and map to `"off"`. Same for `reasoning: true` mapping to `"auto"`. It's the most obvious way a user would try to disable reasoning.

## Actual Behavior

Crashes with `ValueError`. YAML parses `reasoning: false` as a Python bool `False`, the config loader reads the raw string `"false"` but doesn't do anything with it, and then `normalize_reasoning()` rejects it because `"false"` isn't a valid mode.

## Environment

- OS: Windows 11
- Python version: 3.13.1
- Repowise version: 0.46.0
- Installation method: pip

## Additional Context

The relevant code is in `repo_config.py:27-30` and `reasoning.py:46-48`. The bool detection path reads the raw string but never maps it to valid modes.

soumojit-D48:
## Summary

- Fix `reasoning: false` and `reasoning: true` in `.repowise/config.yaml` raising `ValueError` instead of mapping to valid reasoning modes
- Map `"false"` → `"off"` and `"true"` → `"auto"` in the config loader's bool-detection path
- Add defense-in-depth boolean aliases in `normalize_reasoning()` so they work everywhere

## Related Issues

Fixes #1989

## Test Plan

- [x] Tests pass (`pytest`)

## Checklist

- [x] I have added tests for new functionality (14 tests in `test_repo_config_reasoning.py`)

### DIFF
--- a/packages/core/src/repowise/core/reasoning.py
+++ b/packages/core/src/repowise/core/reasoning.py
@@ -43,6 +43,10 @@ def normalize_reasoning(
         return default
 
     normalized = str(value).strip().lower()
+    # Accept common boolean strings so ``reasoning: true/false`` works
+    # everywhere, not just when round-tripped through repo_config.
+    boolean_aliases: dict[str, ReasoningMode] = {"true": "auto", "false": "off"}
+    normalized = boolean_aliases.get(normalized, normalized)
     if normalized not in REASONING_MODES:
         choices = ", ".join(REASONING_MODES)
         raise ValueError(f"reasoning must be one of: {choices}")
--- a/packages/core/src/repowise/core/repo_config.py
+++ b/packages/core/src/repowise/core/repo_config.py
@@ -27,7 +27,11 @@ def load_repo_config(repo_path: Path | str) -> dict[str, Any]:
         if isinstance(result, dict) and isinstance(result.get("reasoning"), bool):
             raw_reasoning = _read_flat_scalar(text, "reasoning")
             if raw_reasoning:
-                result["reasoning"] = raw_reasoning
+                # Map common boolean strings to valid reasoning modes so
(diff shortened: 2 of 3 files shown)


RaghavChamadiya (approved):
Thanks @soumojit-D48, and sorry this waited nearly two weeks. The fix is in the right place: normalize_reasoning is where str(value) turns YAML's False into the string 'false' and then rejects it, so mapping the boolean spellings there is what makes reasoning: false work on every path rather than only the one that round-trips through repo_config. Correcting my own earlier comment on #1989 while I am here: I said repo_config.py never touches the value, and that was wrong — load_repo_config does carry an isinstance(..., bool) branch that recovers the raw text (repo_config.py:46-49). Your change to it is therefore real rather than redundant, and mapping there too leaves the returned config holding a valid mode instead of a string that only happens to normalize later. One non-blocking follow-up for whoever is next in that file: resolve_reasoning's raw = reasoning or os.environ.get(...) at :57 still discards an explicit falsy first argument, which is a second wrong answer for the same input and is untouched by this.

PR #2002, opened 2026-08-30
refactor(ingestion): collapse _derive_exports dead branch and clarify LanguageConfig metadata

### LINKED ISSUE #1999: [Refactor]  _derive_exports: dead if-branch silently renders export_node_types config field inert
## Describe the Bug / Design Gap

While reading the ingestion layer I found a dead code branch in
`ASTParser._derive_exports` ([`parser.py:2044–2046`][derive_exports]) that
makes the `export_node_types` field on `LanguageConfig` a no-op at runtime,
even though the field carries meaningful values for TypeScript and JavaScript.

[derive_exports]: https://github.com/repowise-dev/repowise/blob/main/packages/core/src/repowise/core/ingestion/parser.py#L2044-L2046

## What the code currently does

```python
def _derive_exports(self, symbols, config, src) -> list[str]:
    """Derive the list of exported names from parsed symbols."""
    if config.export_node_types:
        return [s.name for s in symbols if s.visibility == "public" and s.parent_name is None]
    return [s.name for s in symbols if s.visibility == "public" and s.parent_name is None]
```

Both branches return the **identical expression**. The `if` branch is dead —
it can never produce a result different from the `return` below it.

## What was clearly intended

`export_node_types` is populated for exactly two languages in
[`language_configs.py`][langconfigs]:

| Language | `export_node_types` |
|---|---|
| `typescript` | `["export_statement"]` |
| `javascript` | `["export_statement"]` |
| all others | `[]` |

[langconfigs]: https://github.com/repowise-dev/repowise/blob/main/packages/core/src/repowise/core/ingestion/language_configs.py

The `if` branch was clearly meant to implement a **TS/JS-specific derivation
path** — deriving the export list from AST-captured `export_statement` nodes
rather than from the visibility heuristic. Two further signals support this:

1. The `src: str` parameter is passed to `_derive_exports` but **never used**
   inside the method — a strong indication that a text-scan or AST-walk
   approach was planned but never written.

2. The field name `export_node_types` implies node-type-based filtering, not
   visibility-based filtering.

## Current blast radius

**Zero user-visible impact today.** `ts_visibility()` already sets
`visibility="public"` from the presence of the `export` keyword, so the
fallback expression coincidentally produces the correct export list for
TypeScript and JavaScript. No test failure, no wrong output.

## The forward risk (why it still matters)

- `export_node_types` is a config field that **appears meaningful** but is
  silently ignored at runtime. A developer adding a new language and setting
  `export_node_types` will expect narrower export derivation and get none.
- The unused `src` parameter adds confusion: it looks like it drives some
  logic but drives nothing.
- The gap means there is **no future path** to implementing proper
  AST-node-based export filtering for TS/JS without first untangling this.

## Questions for the maintainer before I claim this

I want to understand the intended direction before starting work, since there
are at least three valid fixes of different scope:

**Option A — Simplest: collapse to one branch, document the reasoning**

Remove the `if` branch entirely, collapse to one `return`, and add a comment
explaining that TS/JS export correctness is guaranteed by `ts_visibility()`
upstream rather than by a separate derivation path here. Also remove the
unused `src` parameter. Low risk, no behaviour change.

**Option B — Correct: implement the intended TS/JS path**

Make the `if config.export_node_types:` branch actually use the AST captures
to filter — e.g. only include symbols whose `def_node` is a direct child of
an `export_statement`. This is the architecturally correct answer and would
make `export_node_types` mean something, but it requires more care to avoid
false-negatives for `export default` and re-export shapes.

**Option C — Remove `export_node_types` from `LanguageConfig` entirely**

If the field's only downstream consumer is this method and it is currently
inert, it may be cleaner to remove the field and document that export
derivation is responsibility of the visibility extractor. This would be a
breaking change for any code that reads `config.export_node_types` externally.

---

Which of these aligns with the design intent? Happy to take whichever direction
makes sense and write the PR once I have your input.

## Environment

- Repowise version: current `main`
- File: `packages/core/src/repowise/core/ingestion/parser.py`
- Related: `packages/core/src/repowise/core/ingestion/language_configs.py`

akshatmalik-bruh:
## Summary

- Collapsed redundant `if config.export_node_types:` branch in `ASTParser._derive_exports()` to a single return statement.
- Dropped unused `src` parameter from `_derive_exports()` and updated its call site in `parser.py`.
- Added docstring explaining that TS/JS export visibility is resolved upstream by `refine_ts_visibility()`, and updated `language_configs.py` comments to clarify `import_node_types` and `export_node_types` are descriptive metadata not read at runtime.

## Related Issues

Closes #1999

## Test Plan

- [x] Tests pass (`uv run pytest tests/unit/ingestion` — 2,803 passed)
- [x] Lint passes (`uv run ruff check .` and `uv run ruff format --check .`)

## Checklist

### DIFF
--- a/packages/core/src/repowise/core/ingestion/language_configs.py
+++ b/packages/core/src/repowise/core/ingestion/language_configs.py
@@ -39,10 +39,10 @@ class LanguageConfig:
     # Maps tree-sitter node type → our canonical SymbolKind string
     symbol_node_types: dict[str, str]
 
-    # tree-sitter node types that carry import information (doc purposes)
+    # tree-sitter node types that carry import information (descriptive metadata; not read at runtime)
     import_node_types: list[str]
 
-    # tree-sitter node types that export symbols (doc purposes)
+    # tree-sitter node types that export symbols (descriptive metadata; not read at runtime)
     export_node_types: list[str]
 
     # (name: str, modifier_texts: list[str]) → "public" | "private" | ...
@@ -372,7 +372,7 @@ LANGUAGE_CONFIGS: dict[str, LanguageConfig] = {
             # flagged as a follow-up, not attempted here.
             "declType": "class",
             "declProc": "function",  # signature only (interface decl / forward decl)
-            "defProc": "function",   # full definition with body
+            "defProc": "function",  # full definition with body
             # Matches C#'s choice for the same concept (property_declaration ->
             # "variable"). Rust is the one language that keeps fields under a
             # distinct "property" kind; everywhere else a field and a callable
@@ -381,7 +381,7 @@ LANGUAGE_CONFIGS: dict[str, LanguageConfig] = {
             "declProp": "variable",
         },
         import_node_types=["declUses"],
-        export_node_types=[],       # Pascal has no explicit re-export syntax
+        export_node_types=[],  # Pascal has no explicit re-export syntax
         # No pascal_visibility exists yet. Pascal visibility is per-*section*
         # (`strict private`/`protected`/`public`/`published` governs every
         # declaration until the next section keyword, i.e. extractors/
--- a/packages/core/src/repowise/core/ingestion/parser.py
+++ b/packages/core/src/repowise/core/ingestion/parser.py
@@ -1056,7 +1056,7 @@ class ASTParser:
             else []
         )
         heritage = extract_heritage(matches, config, file_info, src)
-        exports = self._derive_exports(symbols, config, src)
+        exports = self._derive_exports(symbols, config)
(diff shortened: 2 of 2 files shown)


RaghavChamadiya (approved):
Thanks @akshatmalik-bruh. Both arms of that `if` returned the identical expression, so the branch was reading as a meaningful distinction that did not exist. Deleting it and saying in the docstring where TS/JS export visibility is actually decided, `refine_ts_visibility()` upstream, puts the explanation where the next reader will look for it.

Retitling the `import_node_types` / `export_node_types` comments from "doc purposes" to "not read at runtime" is a small change that carries most of the value here.

PR #2018, opened 2026-08-30
fix(generation): emit file pages for test infrastructure, fixtures, and e2e helpers (#1497)

### LINKED ISSUE #1497: e2e specs are ingested but get no wiki page, so get_answer cannot answer test-infrastructure questions

Nithin0620:
## Summary

* Closes #1497
* Fixes wiki `file_page` generation so test infrastructure and fixture files are included, allowing `get_answer` to retrieve test conventions and infrastructure documentation.
* Updates test-file selection, scoring, and TypeScript/JavaScript test fixture detection.

## Related Issues

Closes #1497

## Test Plan

* [x] Tests pass (`pytest`)
* [x] Lint passes (`ruff check .`)
* [ ] Web build passes (`npm run build`) *(not applicable — no frontend changes)*

### Verification

* [x] Ran `pytest tests/unit/generation/test_selection_contract.py tests/unit/test_test_paths.py` — **187 tests passed**
* [x] Ran the full generation test suite — **1,456 tests passed in 38.29s**

## Checklist

* [x] My code follows the project's code style
* [x] I have added tests for new functionality
* [x] All existing tests still pass
* [x] I have updated documentation if needed


sloemo01:
Both red tests are the same cause, and it needs a decision rather than a fix.

```
FAILED TestParityGoldens::test_test_fixture_stems_match_historical_set
  assert {'conftest','spec_helper','test_helper'} == {'fixture','fixtures','setup-tests','conftest','spec_helper','test_helper'}

FAILED TestParityGoldens::test_test_stem_suffixes_match_historical_set
  assert {'.test','.spec','_test','_spec','_unittest'} == {'_test','_spec','_unittest'}
```

The specs now carry `test_fixture_stems=("fixtures","fixture","setup-tests")` and the `(".test",".spec")` suffixes, and `TestParityGoldens` exists to notice exactly that. Those names are the tell. They are deliberate anchors, not oversights, and the surrounding comments say so: the suffix golden carries a note that `_unittest` "was a conscious addition to the historical set". So a vocabulary change is expected to update the anchor in the same PR and explain why the set grew.

That gives two honest paths:

- **if widening it is the point**, and it reads like it is given the PR is about giving fixture and test-support files pages, update both goldens here and say in the body why the set grew
- **if it was a side effect** of the page-selection work, narrow the spec change and leave the goldens alone

Either way I would not edit the goldens quietly, since the next person to touch test detection reads them as the contract.

For a fast local loop, the two tests are pinned to `tests/unit/ingestion/test_language_capabilities.py`, so `uv run pytest tests/unit/ingestion/test_language_capabilities.py -q` is enough to see both.

Nithin0620:
Thanks @RaghavChamadiya and @sloemo01 for Suggesting These changes !

 I've updated the branch with the following changes:

1. **Reverted global spec additions**: Removed `test_fixture_stems` and `test_stem_suffixes` from the JavaScript and TypeScript specs,
  restoring the historical golden sets and fixing `TestParityGoldens` (`test_test_fixture_stems_match_historical_set` and
  `test_test_stem_suffixes_match_historical_set`).
2. **Contextual test fixture recognition**: Added `_SUPPORT_STEMS_IN_TEST_TREE` in `test_paths.py` so stems like `fixtures`, `fixture`,
  and `setup-tests` are only recognized as test support when located inside a test directory tree (such as `apps/console/e2e/fixtures.ts`),
  avoiding global classification leaks.
3. **Added negative test cases**: Added test cases in `tests/unit/test_test_paths.py` verifying that `src/fixtures.ts`, `src/fixtures.py`, and `lib/fixtures.go` remain classified as production code.

    All unit tests across ingestion, generation, and test paths are now green locally.

PR #2030, opened 2026-08-31
docs(init): document --resume and expand its help text (#113)

### LINKED ISSUE #113: UPDATE Documentation by --resume case for long-running `init` operations
# Checkpoint/Resume for long-running `init` operations

## Problem

When running `repowise init` on large repositories, the process can take many hours. If the process is interrupted (timeout, crash, user termination, network issues), **all progress is lost** and the init must restart from scratch.

**My experience:**
- Repository: ~553 pages to generate
- Process ran for 8+ hours (416/553 pages completed)
- Process timed out after ~11,218 seconds
- **Result:** All 416 pages generated were lost - SQL database showed 0 pages

The pages existed in LanceDB (`_transactions/`, `_versions/`, `data/` folders with 416 fragments) but were not committed to SQL and were not recoverable via standard commands. I had to manually write a Python script to extract data from LanceDB and import it to SQL.

I'm frustrated when I spend 8 hours waiting for indexing only to lose everything because of a timeout or interruption.

## Proposed Solution

**1. Periodic checkpoints** - Commit to SQL database every N pages (configurable)
```bash
repowise init . --checkpoint-interval 50  # Save every 50 pages
```

**2. Auto-resume on re-run** - `repowise init` should detect partial state and continue:
```bash
$ repowise init .
Resuming from checkpoint: 416/553 pages already generated
Generating pages... ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━╺━━━━━━━━━ 137/553
```

**3. Graceful shutdown** - SIGTERM/SIGINT should trigger final checkpoint before exit

**4. Recovery command** (optional, for cases where checkpoint failed):
```bash
repowise recover .  # Scan LanceDB and import uncommitted pages to SQL
```

## Alternatives Considered

**Workaround I used:**
1. Manually extracted data from `.repowise/lancedb/wiki_pages.lance/` using Python + lancedb
2. Created `state.json` manually with page count
3. Inserted pages into SQL via sqlite3
4. Rebuilt FTS index manually

This is too complex for average users.

**Other approaches:**
- **External process manager** (systemd, supervisord) - doesn't help, data loss occurs within repowise
- **Smaller batch sizes** - still loses all progress if interrupted mid-batch
- **`repowise update` instead of `init`** - only works for git changes, not for resume after crash

## Additional Context

**Environment:**
- repowise: latest
- Repository size: 553 pages
- Runtime before interruption: ~8 hours
- Provider: litellm with zai/glm-5

**Why this matters:**
Large codebases (1000+ files, enterprise repos) are common. For these users, repowise is unusable in production environments where long-running processes may be interrupted. A single timeout wastes hours of compute time and API costs.

**Relevant code locations:**
- `generation_jobs` table tracks progress but doesn't seem to implement checkpoints
- LanceDB has uncommitted transactions in `_transactions/` folder
- `repowise doctor` detects "Coordinator drift" but can't fix it automatically

sloemo01:
Closes #113.

**Problem:** `--resume` exists (checkpoint controller) but is undocumented — an interrupted 8-hour `init` looks like a total loss.

**Changes:**
1. README: "Resuming an interrupted index" subsection under the init usage section — one paragraph + the snippet, per the greenlight.
2. `init --help`: `--resume` text expanded from "Resume from last checkpoint." to explain the vector-store skip semantics and the safe no-op on a fully indexed repo.
3. Failure-mode note: pages are written to LanceDB incrementally while the SQL `generation_jobs` row finalizes only at the end, so a hard interrupt can leave LanceDB ahead of SQL — `--resume` is the supported recovery path and `repowise doctor` flags the drift.

**Verified:** `repowise init --help` renders the new text; CLI command tests pass (31). CLI_REFERENCE.md already documents `--resume` fully — no change needed there.

### DIFF
--- a/README.md
+++ b/README.md
@@ -618,6 +618,21 @@ already rendered from structure at index time.
 Or pick the provider for the first index directly with `repowise init --provider
 gemini|anthropic|openai`.
 
+**Resuming an interrupted index.** If `init` is interrupted (timeout, crash,
+Ctrl+C), re-run it with `--resume` and it continues from where it stopped —
+pages already written to the vector store are skipped, and only the missing
+ones are generated:
+
+```bash
+repowise init . --resume
+```
+
+`--resume` is a safe no-op on a fully indexed repo, so it is the right thing to
+reach for whenever a long run is cut short. It works because pages are written
+to LanceDB incrementally, while the SQL `generation_jobs` row only finalizes at
+the end — a hard interrupt can leave LanceDB ahead of SQL, and `--resume` is
+the supported recovery path (`repowise doctor` flags the drift).
+
 **3. Connect your agent.** Step 2 already did this for Claude Code: `init`
 writes a repo-root `.mcp.json` unconditionally and, unless you passed
 `--no-editor-setup`, also registers repowise with `~/.claude/settings.json`.
--- a/packages/cli/src/repowise/cli/commands/init_cmd/command.py
+++ b/packages/cli/src/repowise/cli/commands/init_cmd/command.py
@@ -436,7 +436,15 @@ def _run_generation_phase(
     "--dry-run", is_flag=True, default=False, help="Show generation plan without running."
 )
 @click.option("--yes", "-y", is_flag=True, default=False, help="Skip cost confirmation prompt.")
-@click.option("--resume", is_flag=True, default=False, help="Resume from last checkpoint.")
+@click.option(
+    "--resume",
+    is_flag=True,
+    default=False,
+    help=(
+        "Skip pages already generated in the vector store and continue from "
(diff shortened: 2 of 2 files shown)


RaghavChamadiya (approved):
Thanks @sloemo01. An interrupted long index reading as a total loss is a real cost, and the flag already existed, so this is the cheapest possible fix for it.

I checked the failure-mode paragraph against the tree before taking it, since it is the kind of claim that ages badly in a README. `doctor` does compute the SQL-versus-vector drift and surface it (`repo_checks.py`, the `drift_detail` line), so pointing people there is accurate.

PR #2032, opened 2026-08-31
fix(cli): resolve --embedder auto through the repo pin

azhard:
## What

`repowise reindex` ignores the repo's pinned embedder. `--embedder auto` — the default — resolves from the environment only, so a repo with `embedder: ollama` in `config.yaml` aborts with:

```
No real embedder available. Set a real embedder key, configure Ollama, or pass --embedder mock for test vectors.
```

`--embedder ollama` then works, which is the tell that the pin was fine and only the resolution was wrong.

## Root cause

`reindex_cmd._reindex` resolves `auto` via `_resolve_embedder(None)`, which reads `REPOWISE_EMBEDDER` and provider API keys. It never consults `config.yaml`, despite the `--embedder` help text saying *"'auto' detects from env vars / config."*

`resolve_embedder_for_repo` already exists for this, and its docstring states the invariant plainly:

> The pinned `embedder` in `config.yaml` is what wrote the table, so it is what can query it. Readers used to resolve from the environment while writers read the pin, which is how a repo indexed keyless ends up queried with whatever API key happens to be exported: the 1536-wide query vector finds nothing in an 8-wide table, and search just returns nothing.

`reindex` is on the *writing* side of exactly that split, and was resolving from the environment.

## Consequences

- A repo pinned to a keyless embedder cannot reindex without restating the pin on the command line.
- Worse, and silent: a repo pinned to one embedder, reindexed in a shell exporting a different provider's key, has its table **rewritten** by that other provider. Every reader honours the pin, so they then query vectors of the wrong width and search returns nothing — with no error at write time.

## Related Issues

No open issue tracks this. Hit it directly: a repo pinned to `embedder: ollama` refused to reindex, and `--embedder ollama` on the same repo worked, which is what pointed at resolution rather than configuration.

## Changes

`auto` now resolves through `resolve_embedder_for_repo(repo_path)`. `REPOWISE_EMBEDDER` still wins, since that is the documented escape hatch after a manual re-embed; the pin now takes precedence over ambient API keys instead of losing to them.

## Tests

`test_reindex_auto_honours_the_repo_pinned_embedder` writes `embedder: ollama` to a repo, exports an unrelated `OPENAI_API_KEY`, and asserts the embedder built is the pinned one.

On unpatched main:

```
E   AssertionError: assert 'openai' == 'ollama'
```

CLI suite: 2253 passed, 1 skipped, 2 xfailed. `test_integration_writes_match_baseline` fails both with and without this branch — it compares against a Windows baseline (`AppData/Roaming/...` vs `Library/Application Support/...`) and is unrelated.

## One thing worth flagging separately

The new test clears `REPOWISE_EMBEDDER` explicitly. That is correct on its own terms, since the variable outranks the pin and a test about pin precedence must control it — but it is also load-bearing, because the variable leaks between tests today.

`test_embedder_resolution.py::test_the_reader_prefers_the_repo_pin` calls `monkeypatch.delenv("REPOWISE_EMBEDDER", raising=False)` and then invokes `serve_cmd._load_local_provider_config()`, which sets that variable in `os.environ`. When the variable is already absent, `delenv(raising=False)` records nothing, so monkeypatch has no saved state to restore and the value the production code set survives into every later test in the process. It is observable directly:

```
pytest tests/unit/cli/test_embedder_resolution.py tests/unit/cli/test_reindex_cmd.py
```

I have not fixed that here — it is a separate concern from this command and I did not want to widen the diff. Happy to send a follow-up if you would like one.

## Follow-up commit

Addresses @sloemo01's review note about the mock pin. The abort guard compared the resolved embedder against the flag the user typed, so once `auto` read the pin, a repo with `embedder: mock` resolved to mock and was then rejected as unconfigured. `mock` reaches that guard two ways — asked for, or fallen back to when nothing is configured — and only the second is an error. The guard now tests the resolved name, and `requested_embedder` is gone with it.

`test_reindex_does_not_abort_on_a_deliberately_pinned_mock_embedder` covers it, and fails against the previous guard.

Also ran the three suites the health bot flagged as importing the changed file — `test_dead_code_integration.py`, `test_persistence.py`, `test_mock_provider.py`: 51 passed.

### DIFF
--- a/packages/cli/src/repowise/cli/commands/reindex_cmd.py
+++ b/packages/cli/src/repowise/cli/commands/reindex_cmd.py
@@ -48,7 +48,7 @@ async def _reindex(repo_path, embedder_name: str, batch_size: int) -> None:
     from sqlalchemy import select
     from sqlalchemy.ext.asyncio import AsyncSession, async_sessionmaker
 
-    from repowise.cli.providers.embedders import build_embedder
+    from repowise.cli.providers.embedders import build_embedder, resolve_embedder_for_repo
     from repowise.core.persistence.database import create_engine, init_db
     from repowise.core.persistence.models import Page
     from repowise.core.providers.embedding.base import MockEmbedder
@@ -56,9 +56,16 @@ async def _reindex(repo_path, embedder_name: str, batch_size: int) -> None:
     # --- Resolve embedder ---
     requested_embedder = embedder_name
     if embedder_name == "auto":
-        from repowise.cli.commands.init_cmd import _resolve_embedder
-
-        embedder_name = _resolve_embedder(None)
+        # Resolve through the repo, not the environment. `_resolve_embedder(None)`
+        # reads REPOWISE_EMBEDDER and API keys only, so `auto` never saw the
+        # `embedder` pin in config.yaml despite this flag's help text promising
+        # "env vars / config". A repo pinned to a keyless embedder therefore
+        # aborted with "No real embedder available", and a repo pinned to one
+        # embedder in a shell exporting a different provider's key had its table
+        # REWRITTEN by that other provider — the exact writer/reader split
+        # `resolve_embedder_for_repo` was added to prevent, and one this command
+        # is on the writing side of.
+        embedder_name = resolve_embedder_for_repo(repo_path)
 
     embedder_impl = build_embedder(embedder_name, repo_path)
     if isinstance(embedder_impl, MockEmbedder) and requested_embedder != "mock":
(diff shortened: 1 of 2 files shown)


sloemo01 (approved):
Fix is correct and minimal —  now resolves through the repo pin like every reader already does, closing the silent table-rewrite. The mock-pin edge case I noted is a suggestion, not a blocker; happy to follow up separately.

sloemo01 (approved):
Fix is correct and minimal — auto now resolves through the repo pin like every reader already does, closing the silent table-rewrite. The mock-pin edge case I noted is a suggestion, not a blocker; happy to follow up separately.

RaghavChamadiya (approved):
Thanks @azhard, the writeup is the useful part here. Naming the silent case, a repo pinned to one embedder having its table rewritten by whatever key happens to be exported, is what makes this worth fixing rather than a help-text correction.

I checked `resolve_embedder_for_repo` and its docstring states the exact invariant you quote, so `reindex` was on the writing side of the split it was added to close. Moving the mock guard onto the resolved name is the right call too: with the pin now in play, `embedder: mock` in `config.yaml` is as deliberate as passing the flag, and guarding on the flag would have aborted those repos.

PR #2033, opened 2026-08-31
feat(cli): add --verbose to augment, wired to configure_cli_logging (#930)

sloemo01:
The `augment` slice of #930 — the last long-running command without a `--verbose` flag.

**Changes:**
- `--verbose`/`-v` on `repowise augment`, calling `configure_cli_logging(verbose=verbose)` before any hook work — mirroring the reindex/export/restyle/claude-md/watch/coverage/health/workspace slices.
- Tests: quiet/verbose/`-v` forwarding through `CliRunner` (asserting logging is configured before `_run_augment`), plus help-text coverage.
- CLI reference: flag table for the augment section.

**Verified:** 48 augment-related tests pass. Note: the `test_plugin_content` CI failure is pre-existing on main (v0.47.0 release hand-edited a generated plugin file without regenerating from `plugins/shared/`) — unrelated to this change.


RaghavChamadiya (approved):
Thanks @sloemo01. Matching what the reindex/export/restyle slices already do keeps `#930` converging rather than accumulating one-off shapes.

I did stop on the new module-level import, since this file's operational invariants call out a cold-start budget and `augment` runs on every hook fire. It is fine: `repowise/cli/_setup.py` has no module-level imports of its own, deferring `logging` and `structlog` inside the function body, so it is not in the class of heavy imports that note is about. Asserting that logging is configured *before* `_run_augment` rather than just that the flag parses is the right thing for the test to pin.

PR #2038, opened 2026-08-31
fix(update): show live progress and persist incrementally on --full (#1709)

### LINKED ISSUE #1709: [Bug] Running `repowise update --full` not rendering progress in CLI

sloemo01:
## What

Fixes #1709 — the reporting/safety halves of Raghav's diagnosis.

## Root cause

- `update --full` passed `progress=None` into generation — every progress callback no-oped, so a long run showed a dead screen while the local model sat pinned for hours.
- Every page was buffered in memory and written once at the end — a Ctrl-C at hour eight cost the entire run (the issue's real question: is aborting safe?).
- The estimate line never called `structural_page_summary()`, so "3661 pages" read as 3661 model calls when ~98 of them are.

## Changes

1. Generation now routes through `run_generation_with_persistence` (init's wrapper): pages are written to the DB the instant each completes, prior pages are reused from the store, and a real Rich progress bar advances. Interrupting now loses at most the in-flight page.
2. The estimate line calls `structural_page_summary()`, printing the "N more pages rendered from the code itself, no model, no cost" breakdown.

## Tests

Regression guards (structural, mirroring the strict-branch pins in test_coverage_cmd.py): the call site must use the persistence wrapper, a live `RichProgressCallback`, and the structural summary — all three fail on the pre-fix code (verified via sabotage run).

### DIFF
--- a/packages/cli/src/repowise/cli/commands/upgrade_flow.py
+++ b/packages/cli/src/repowise/cli/commands/upgrade_flow.py
@@ -69,6 +69,7 @@ def _gate_cost(
             COST_GATE_USD,
             cost_gate_declined,
             format_cost,
+            structural_page_summary,
         )
         from repowise.core.cost_estimator import build_generation_plan, estimate_cost
 
@@ -80,6 +81,9 @@ def _gate_cost(
 
     pages = sum(p.count for p in plans)
     console.print(f"Estimated: [bold]{pages}[/bold] pages, [bold]{format_cost(est)}[/bold].")
+    structural = structural_page_summary(plans)
+    if structural:
+        console.print(f"  [dim]{structural}[/dim]")
 
     # Nothing to ask on a run that cannot be asked. Without this the confirm
     # reads EOF, raises, and the caller reports a successful run that generated
@@ -197,7 +201,7 @@ async def _run_upgrade(
         upsert_pages_from_generated,
         upsert_repository,
     )
-    from repowise.core.pipeline import rehydrate_graph_builder, run_generation
+    from repowise.core.pipeline import rehydrate_graph_builder
 
     url = get_db_url_for_repo(repo_path)
     engine = create_engine(url)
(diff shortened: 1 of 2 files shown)


RaghavChamadiya (approved):
Thanks @sloemo01. Routing `--full` through init's wrapper rather than growing a second persistence path is the right move, and the two properties I would have worried about both hold when you read the wrapper:

Reuse cannot swallow the upgrade. `--full` exists to turn a template wiki into a written one, so "reuse every prior page" sounds alarming, but the prior-page hit requires `prior.model_name == self._provider.model_name` (`page_generator/core.py:365-368`), and a template page never carries a real model name. Template pages get regenerated; only genuinely model-written pages with a matching prompt hash are reused, which is the behaviour you want on a resumed upgrade.

The double write is harmless and deliberate. The wrapper persists each page as it lands and `_run_upgrade` still calls `upsert_pages_from_generated` at the end; the wrapper's own docstring covers exactly this, since `upsert_page` no-ops when content, prompt hash and model are unchanged, so no version bump and no duplicate history. The sessions in `_run_upgrade` are all short `async with` blocks, so the wrapper's separate engine is not writing against an open transaction.

The structural-pages line is a good addition on this path specifically. "3661 pages" on an upgrade reads as 3661 model calls to anyone who has not seen the estimator internals.

Merging. One note, not a blocker but I want it said plainly since it will decide whether this stays fixed: all three new tests are `inspect.getsource` string checks. They pin the call site, and they will pass a code path that raises at runtime, which matters more here than usual because `_run_upgrade` is not exercised end to end anywhere. A `--full` run that dies inside the new `with Progress(...)` block would be green on CI. The precedent you cite is real, so I am not asking you to build a harness for it now; the useful version later is a test that calls `_run_upgrade` with a stub provider and asserts pages exist in the DB before the function returns.

PR #2041, opened 2026-08-31
feat(workspace): surface supporting evidence for cross-repo co-change pairs (#483)

### LINKED ISSUE #483: Surface supporting evidence (authors, matched commits, time gaps) for cross-repo co-change rows
## Background

Follow-up to #475. The cross-repo co-change page now frames itself honestly as a temporal, same-author hint (#481) and shows `frequency` and `last_date` per pair. A remaining ask from #475 is to show *why* each row exists: the actual commits, authors, and time gaps behind a co-change pair, so a user can judge whether a relationship is meaningful instead of trusting a bare score.

## Proposed change

For each co-change pair, surface supporting evidence on demand (for example an expandable row or an inspector panel):

- The authors involved.
- A few example matched commit pairs (the cross-repo commit pair that created the relationship), with short SHAs and dates.
- The time gap between the paired commits.

## Why this is not a "good first issue"

The current pipeline does not persist this. `detect_cross_repo_co_changes` in `packages/core/src/repowise/core/workspace/cross_repo.py` accumulates `strength`, `frequency`, and `last_date` per file pair, but discards the underlying commit identities and authors. Delivering this needs:

1. Core: capture and persist a bounded sample of evidence per pair (author email, commit SHAs, timestamps) into `CrossRepoCoChange` / the overlay JSON, with a cap so the artifact does not blow up on large workspaces.
2. Server: expose the evidence through the workspace co-changes endpoint.
3. Types + UI: extend `WorkspaceCoChangeEntry` and render the evidence.

## Pointers

- Core: `packages/core/src/repowise/core/workspace/cross_repo.py` (`detect_cross_repo_co_changes`, `CrossRepoCoChange`, `CrossRepoOverlay`).
- Type: `WorkspaceCoChangeEntry` in `packages/types/src/workspace.ts`.
- UI: `packages/ui/src/workspace/co-change-table.tsx`.

## Acceptance

- A co-change row can reveal its supporting evidence (authors, example matched commits, time gaps).
- The persisted overlay stays bounded in size (evidence is sampled/capped).
- Tests cover the new evidence capture in core.

sloemo01:
## What

Fixes #483 — the evidence layer Raghav asked for in the thread ("the sample bound decides the schema").

## Root cause

Cross-repo co-change pairs carried only a strength/frequency score — no authors, no example commits, no time-gap signal — so a reviewer couldn't tell a real co-evolution from a coincidental co-edit.

## Changes

**Core** (`detect_cross_repo_co_changes`):
- `git log` now captures `%H`; `_GitCommit.sha` added
- `CoChangeEvidence`: distinct authors + example matched commit pairs (`source_sha`, `target_sha`, `date`, `gap_hours`) + max time gap. **Bounded: 3 commit pairs and 5 authors per pair** (the cap Raghav's scoping asked to settle first) — a pair backed by hundreds of commits cannot bloat the overlay
- One example pair per co-session, deduped by SHA pair
- Overlay version 2 → 3; round-trip tested, pre-v3 overlays load with `evidence=None`

**Server:** `WorkspaceCoChangeEvidence` schema, `WorkspaceCoChangeEntry.evidence`, router exposes it.

**Types + UI:** mirrors in `packages/types` and `packages/api-client`; the co-change table gains an evidence disclosure per row (authors, commit pairs with gap, max-gap summary), dash when absent.

## Tests

Core evidence capture (pair shape, bounded sample, overlay round-trip) — 66 workspace tests pass; 145 server workspace tests pass. The UI vitest suite could not run locally (no `node_modules` in this checkout); the type change is consistent with the updated test fixture.


sloemo01:
Rebased onto `main` and pushed. The branch was `dirty` because `main` moved underneath the whole co-change area, so most of this is a port rather than a conflict resolution.

**What main changed, and what moved.** #2530 split the workspace code into modules and rewrote the read path. `cross_repo.py` no longer scans package manifests (`manifests.py` now owns that, `_MAX_PACKAGE_DIAGNOSTICS` with it), the router no longer builds co-change rows itself, and the UI table was rewritten around memoised rows. Five files conflicted.

**Resolutions, file by file:**

- `packages/core/.../cross_repo.py`: one hunk, in the constants block. Main moved the package-diagnostics constant out, so that deletion is taken and only the two evidence caps stay. The evidence capture itself (`CoChangeEvidence`, the `%H` sha in the git-log format, the per-pair bounded sample) auto-merged and is unchanged.
- `packages/core/.../reads.py`: not a conflict, but this is where the port lands. Rows are built here now, so `list_co_changes` passes `evidence` through with the rest of the row. Nothing else moved.
- `packages/server/.../routers/workspace.py`: both hunks go to main. The router delegates to `reads.list_co_changes` and no longer constructs `WorkspaceCoChangeEntry` rows, so the inline builder this branch added there is gone; evidence arrives through `reads.py` instead.
- `packages/server/.../schemas/__init__.py`: main rewrote the module to PEP 562 lazy loading, so that body is taken as-is and `WorkspaceCoChangeEvidence` is registered in all three places the new structure has (`TYPE_CHECKING`, `_EXPORTS`, `__all__`). Checked the lazy attribute resolves, not just the import.
- `packages/ui/.../co-change-table.tsx`: main replaced the virtualized table with a memoised-row table. `EvidenceCell` and its column are ported onto that shape, so the disclosure now sits in a `CoChangeRow` and uses the shared `InfoTip` main introduced rather than a bare `title`.

**Two type errors the rebase surfaced, both fixed.** `evidence` is required on `WorkspaceCoChangeEntry`, so the pair factories in `co-change-ai-prompt.test.ts` and `co-change-pair-drawer.test.tsx` no longer satisfied the type. Both now set `evidence: null`, which is what a pre-evidence overlay produces.

**Verified on `2fdd756`:**

```
tests/unit/workspace/test_cross_repo.py      66 passed
tests/unit/server/test_workspace_router.py   73 passed
vitest packages/ui/__tests__/workspace/     178 passed (19 files)
type-check: ui / types / api-client          clean
```

End-to-end check that the wiring is real rather than merely type-clean: `reads.list_co_changes` returns the `evidence` dict on a row that carries it, returns `None` for one that does not, and `WorkspaceCoChangeEvidence` validates it through the lazy import.

One caveat on the UI suite: `npm run test --workspace @repowise-dev/ui` reports 66 failures both on this head and on a clean `main` at `3ffce9993` (8 files, mostly `localStorage is not available` in this environment). Those are pre-existing and unrelated; the workspace directory is otherwise green.

PR #2042, opened 2026-08-31
fix(ui): scope performance context all badge

### LINKED ISSUE #2036: [Bug] The All context tab badge is not cross-filtered, so it disagrees with the tabs beside it and with the page it opens
**Summary**

`ContextTabs` builds five badges from two different scopes.

`packages/ui/src/health/performance/filters.tsx:58-69`

```tsx
const tabs = [
  { id: "all", label: "All", badge: total },
  ...contexts.map((context) => ({
    id: context,
    label:
      collapsed && context === "production" ? "Production & tooling" : contextLabel(context),
    badge:
      collapsed && context === "production"
        ? countOf(counts, "production") + countOf(counts, "tooling")
        : countOf(counts, context),
  })),
];
```

`total` is `repositoryTotal` (`packages/ui/src/health/performance-view.tsx:157, 232`), which is
`summary.repository_total` and never narrows. The four `countOf` badges read `facets.context`,
which the server cross-filters by every *other* active filter.

With `actionability=plan_ready` set, the row reads `All 743 | Production 22 | Tooling 5 |
Test 12 | Unknown 1`. The four sum to 40. Clicking "All" returns a page whose `total` is 40, and
`ScopeLine` then prints "40 opportunities under 1 filter, from 743 in the repository". The badge
that said 743 is the control the user pressed to get 40.

**Mechanism**

`_facets` excludes a dimension from its own filter but applies all the others
(`packages/server/src/repowise/server/services/performance_health.py:335-345`):

```python
if any(
    values is not None
    and other != name
    and (row[dimensions[other]] or "none") not in values
    for other, values in selected.items()
):
    continue
```

So `facets.context` is already scoped to boundary, confidence and actionability.
`repository_total` is scoped to nothing. The two are mixed into one tab row.

The per-context badges are the correct ones. The "All" badge is the outlier, and it is the one
carrying the largest number, so the row reads as though the contexts fail to account for most of
the repository.

**Repro**

1. Open the performance tab on any repo with more than one actionability state.
2. Note the "All" badge.
3. Set Actionability to any single value.
4. The four context badges drop. "All" does not, and no longer equals their sum.
5. Click "All". The resulting page total matches the sum of the four, not the badge.

**Fix**

Badge "All" with the sum of the context facet, falling back to `total` when the server sends no
`facets.context`. That fallback matters: a server predating context scoping sends no
`repository_total` either, and `repositoryTotal` is then the only number available.

```tsx
const scopedTotal = counts?.reduce((sum, entry) => sum + entry.total, 0) ?? total;
```

Worth a test in `packages/ui/__tests__/health/` asserting the All badge equals the sum of the
context badges under an active narrowing filter, and that it falls back to `total` when
`facets.context` is absent.

Aman-goel-04:
## Summary

- Scope the Performance health “All” badge to the sum of the current context facets.
- Preserve the repository total as a fallback when facets.context is unavailable.
- Add regression tests covering narrowed, collapsed, fallback, and empty-facet states.

## Related Issues

Fixes #2036 

## Test Plan

- [x] Tests pass (npm --workspace @repowise-dev/ui test -- __tests__/health/context-tabs.test.tsx)

## Checklist

### DIFF
--- a/packages/ui/__tests__/health/context-tabs.test.tsx
+++ b/packages/ui/__tests__/health/context-tabs.test.tsx
@@ -1,0 +1,85 @@
+/**
+ * The "All" tab in ContextTabs.
+ *
+ * total (repository_total) and facets.context are counted over two
+ * different scopes: the former never narrows, the latter is cross-filtered by
+ * every other active selection. Badging "All" with the sum of the context
+ * facet keeps the row internally consistent -- the number on the button a
+ * user presses has to match what pressing it returns. The fallback to
+ * total exists only for a server that predates context scoping and sends
+ * no facets.context at all.
+ */
+import { beforeAll, describe, expect, it, vi } from "vitest";
+import { render, screen } from "@testing-library/react";
+import { ContextTabs } from "../../src/health/performance/filters.js";
+import type { PerformanceFacets } from "@repowise-dev/types/health";
+
+// jsdom has no layout engine; ViewTabs' active-tab scroll is a no-op here.
+beforeAll(() => {
+  Element.prototype.scrollIntoView = vi.fn();
+});
+
+const scopedCounts = [
+  { value: "production", total: 22 },
+  { value: "tooling", total: 5 },
+  { value: "test", total: 12 },
+  { value: "unknown", total: 1 },
+];
+
+describe("ContextTabs -- the All badge", () => {
+  it("equals the sum of the context facet when another filter has narrowed it", () => {
+    render(
+      <ContextTabs
+        value="all"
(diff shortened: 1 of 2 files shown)


RaghavChamadiya (approved):
Thanks @Aman-goel-04. The framing in the test file header is exactly right: the number on a button has to match what pressing it returns, and `repository_total` and `facets.context` were being counted over two different scopes.

`counts?.reduce(...) ?? total` gets the empty case right too, which is easy to get wrong here. An empty `facets.context` array reduces to 0 and badges 0, while a missing one falls through to the repository total, and those need to stay distinct. Good that the fourth test pins it.

PR #2043, opened 2026-08-31
feat(mcp): add set_finding_status — the durable write half of findings triage (#1535)

### LINKED ISSUE #1535: Findings need a disposition lifecycle: declined suggestions are re-minted as open on every run

sloemo01:
## What

Fixes #1535 (track 2 — the disposition lifecycle). A new opt-in specialist MCP tool, `set_finding_status` (safety: `mutating`).

## Root cause

#1986 (merged Aug 29) already gave refactoring plans a versioned identity and real lifecycle: stable content-derived `public_id`, the `open|acknowledged|resolved|false_positive` vocabulary, a finalizer that never re-emits `false_positive` plans, and `update_refactoring_suggestion_status` as the single writer. The web UI can PATCH a status. What #1535 still names as missing is the **agent-facing write**: `get_health` and the generated task prompts tell agents to flag false positives, but there was nowhere to record that verdict from inside the agent loop.

## Changes

- Resolves a plan by storage id, content `public_id`, or the display id a deep link carries (same fallback `generate_refactoring_code` uses).
- Writes **only** through `update_refactoring_suggestion_status` — the single lifecycle owner — so a `false_positive` recorded here is suppressed by the finalizer on later runs, exactly like the REST PATCH path. No second writer.
- `reason` is a free-text note stored on the row for audit.
- Statuses: `open`, `acknowledged`, `resolved`, `false_positive`.

## Tests

4 tool-layer tests (vocabulary validation, missing-plan error, shared-writer routing with reason, display-id fallback) — 20 pass with the lifecycle suite. All 27 MCP surface/drift guards pass.

### DIFF
--- a/docs/agent/INTEGRATIONS.md
+++ b/docs/agent/INTEGRATIONS.md
@@ -73,10 +73,10 @@ name for a host at this tier, which is the point of the tier.
 
 ## The MCP surface
 
-repowise registers **seventeen MCP tools**. A single-repo server advertises **ten**
+repowise registers **eighteen MCP tools**. A single-repo server advertises **ten**
 of them by default, and workspace mode adds one more automatically for **eleven**. A
-further **six** are off by default, enabled through the `mcp.tools` config block or
-`--tools +name`. The `lean` profile trims the default surface to **six** tools
+further **seven** are off by default, enabled through the `mcp.tools` config block
+or `--tools +name`. The `lean` profile trims the default surface to **six** tools
 (seven in workspace mode) for agents on a tight context budget.
 
 Per-tool detail: [MCP_TOOLS.md](MCP_TOOLS.md).
--- a/docs/agent/MCP_TOOLS.md
+++ b/docs/agent/MCP_TOOLS.md
@@ -2,7 +2,7 @@
 
 repowise exposes a curated set of tools via the [Model Context Protocol](https://modelcontextprotocol.io) (MCP). These tools give AI coding assistants (Claude Code, Codex, Cursor, Cline, Windsurf) structured access to your codebase intelligence: dependency graph, git history, documentation, and architectural decisions.
 
-17 tools are registered in total. A single-repo server advertises 10 by default: exactly the canonical tools. Workspace mode adds the `list_repos` discovery utility, for 11. 6 specialist tools are opt-in where eligible. The surface is configurable; see [Configuring the tool surface](#configuring-the-tool-surface).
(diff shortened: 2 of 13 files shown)


RaghavChamadiya (approved):
Two things make this safe to take as a new MCP surface. default=False with tier="specialist" keeps the advertised set at ten, so nobody pays context for a tool they have not asked for. And it writes only through update_refactoring_suggestion_status, so a false_positive recorded by an agent is suppressed by the same finalizer that suppresses one recorded from the web UI, rather than being a second lifecycle that drifts from the first. I checked the repo scoping too, since this is the mutating one: get_refactoring_suggestion, the display-id fallback and the write are all keyed on repository.id off the resolved context, so a workspace agent cannot reach another member's rows. The display-id fallback matching generate_refactoring_code is the right detail — a deep link carries that id and an agent following one should not have to know which id shape it holds.

PR #2044, opened 2026-08-31
fix(ui): clarify performance filter counts

### LINKED ISSUE #2037: [Bug] Two filter labels claim the counts are repository-wide when they are cross-filtered
**Summary**

Two strings in the performance filter surface tell the reader the counts beside each control are
repository-wide. They are scoped to every other active filter.

`packages/ui/src/health/performance/states.tsx:78-80`

```tsx
The counts beside each filter come from the whole repository, so a combination can still
be empty.
```

`packages/ui/src/health/performance/filters.tsx:102`

```tsx
<span className="tabular-nums">on all {only.total.toLocaleString()}</span>
```

**Mechanism**

`_facets` (`packages/server/src/repowise/server/services/performance_health.py:335-345`) skips a
row only when it fails a filter *other than* the dimension being counted:

```python
if any(
    values is not None
    and other != name
    and (row[dimensions[other]] or "none") not in values
    for other, values in selected.items()
):
    continue
```

Every count therefore already accounts for the rest of the selection.

`filters.tsx` states this correctly in its own module docstring, 50 lines above the second
string (`filters.tsx:26-30`):

```
Every option and every count comes from the server's facets, which are counted over the
base result, so choosing a value never removes its own alternatives from the list.
```

That is the accurate claim, and it is narrower than the two strings above. "Comes from the whole
repository" does not hold.

The first string is also self-defeating. Because the counts are cross-filtered, selecting any
value a control offers returns exactly that many rows, so `FilteredEmpty` is effectively
unreachable from the controls themselves. It is reached from a restored view state whose
combination no longer matches the index. The sentence offered at that moment explains the empty
result with a mechanism that does not exist, at the one point the user is trying to work out
what happened.

`SingleValueFact` reads "Confidence high on all 12" in a repository holding 743, because
`only.total` is the narrowed count and "all" is not.

**Repro**

Set two filters that cannot co-occur (a boundary and a confidence with no shared rows) and read
the empty-state copy. Then set one filter that collapses another dimension to a single value and
read the "on all N" line against the repository total in the lede.

**Fix**

Copy only, no logic change.

The empty state should say the counts already account for the other filters, and that this
combination came from a restored selection rather than from a choice made here.

`SingleValueFact` should drop "all", or qualify it, when any other filter is active.

Aman-goel-04:
## Summary

- Clarify that performance filter counts account for other active filters.
- Update single-value filter text to avoid implying repository-wide counts.
- Explain that empty combinations may come from restored view state.

## Related Issues

Fixes #2037 

## Test Plan

- [x] Tests pass (npm --workspace @repowise-dev/ui test)

## Frontend screenshots before and after the fix:
- Before: 
Screenshot 2026-08-31 at 10 22 09 AM

- After: 
Screenshot 2026-08-31 at 10 29 38 AM

The before/after screenshots were captured from different index snapshots (9c162335 and 361a7647), so the opportunity count changed from 413 to 411 after re-indexing. The PR only changes the displayed filter copy; it does not change analysis or counting logic.

## Checklist

### DIFF
--- a/packages/ui/__tests__/health/performance-view.test.tsx
+++ b/packages/ui/__tests__/health/performance-view.test.tsx
@@ -68,7 +68,7 @@ describe("PerformanceView queue", () => {
     render(<PerformanceView adapter={adapter()} />);
     await rows();
     expect(screen.queryByLabelText("Evidence confidence")).toBeNull();
-    expect(screen.getByText("on all 7")).toBeTruthy();
+    expect(screen.getByText("on 7 matching")).toBeTruthy();
   });
 
   it("states the rendered scope, the filtered total, and the repository total", async () => {
--- a/packages/ui/src/health/performance/filters.tsx
+++ b/packages/ui/src/health/performance/filters.tsx
@@ -99,7 +99,7 @@ function SingleValueFact({ facet, only }: { facet: PerformanceFacetKey; only: Pe
       <span className="text-[var(--color-text-secondary)]">
         {facetValueLabel(facet, only.value)}
       </span>{" "}
-      <span className="tabular-nums">on all {only.total.toLocaleString()}</span>
+      <span className="tabular-nums">on {only.total.toLocaleString()} matching</span>
     </p>
   );
 }
@@ -205,4 +205,3 @@ export function ContextHint({ context }: { context: PerformanceContextFilter }) 
     </p>
   );
 }
-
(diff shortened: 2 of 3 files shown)


RaghavChamadiya (approved):
Thanks @Aman-goel-04. Pairing this with #2036 is the right instinct: once the counts are cross-filtered, the old empty-state text ("the counts beside each filter come from the whole repository") is not just unhelpful, it is false, and "on all 7" claims a scope the number no longer has.

I merged #2042 a few minutes ago, so the scoping half is on main now. This one is green apart from a `test_rewrite_perf.py` wall-clock budget that failed at a 208ms median against a 200ms limit, in a file your diff does not touch. That is a CI timing flake and not yours, so I have re-run the job rather than holding you on it.

PR #2045, opened 2026-08-31
fix(server): support git logs in worktrees

### LINKED ISSUE #2024: [Bug] get_why drops its live git-log layer in a worktree, because .git is a file there
**Summary**

`_git_archaeology_fallback` mines three layers when no decision governs a file.
Layer 3 shells out to `git log`, and it is gated on the repository's `.git`
being a directory.

`packages/server/src/repowise/server/mcp_server/tool_why.py:1871-1873`

```python
local_path = getattr(repository, "local_path", None)
if local_path and (Path(local_path) / ".git").is_dir():
    git_log_results = await _run_git_log(local_path, file_path, stem)
```

In a linked worktree `.git` is a gitdir pointer file, not a directory, so the
gate is false and the layer never runs. The other two layers read the database
and still answer, so `git_archaeology` comes back populated with one layer
quietly missing and the wire layer omits `git_log_total` entirely.

**Mechanism**

The guard's premise does not hold at this call site.

Two other places use the same `is_dir()` check and are right to.
`mcp_server/_meta.py:51` and `core/persistence/crud/repository.py:107` both go
on to read `.git/HEAD` as a path, which cannot work through a pointer file, and
`repository.py:105-108` documents the choice as deliberate so the written
commit and the read-back comparison stay symmetric.

`_run_git_log` does none of that. It shells out, with the repo as cwd
(`tool_why.py:1911-1915`):

```python
subprocess.run(
    ["git", "log", "--follow", "--format=%H\t%an\t%ai\t%s", "-20", "--", file_path],
    cwd=repo_path, ...
)
```

`git` resolves a worktree perfectly well on its own. So the guard is refusing a
call that would have succeeded. It reads as a check copied from a reader that
needs a real `.git` directory into one that only needs a working tree.

**Repro**

```
git worktree add ../rw-probe main
cd ../rw-probe

ls -la .git
# -rw-r--r-- 1 ... 65 ... .git        <- a file, not a directory

git log --follow --format=%H -3 -- packages/server/src/repowise/server/mcp_server/tool_why.py
# prints three commits, exit 0

pytest tests/unit/server/mcp/test_recoverable_caps_wire_contracts.py::test_why_fallback_archaeology_and_rationale_wire_recover_annotated_tails
```

```
tests/unit/server/mcp/test_recoverable_caps_wire_contracts.py:870: in test_why_fallback_archaeology_and_rationale_wire_recover_annotated_tails
    assert result["git_archaeology"]["git_log_total"] == 23
E   KeyError: 'git_log_total'
```

That test sets `repository.local_path = str(Path.cwd())`, so it reproduces the
production shape rather than mocking around it. It passes in an ordinary
checkout. The two asserts immediately above it, `file_commits_total` and
`cross_references_total`, pass in both, which is the tell: the dict is present,
one layer is gone.

**Impact**

Anyone whose indexed `local_path` is a worktree. `get_why` still answers, and
still answers from decisions where a decision governs the file. What is lost is
the live-git layer of the no-decision fallback, which is the path most files
take. The response does not report the degradation. Nothing beyond that is
claimed.

Worth knowing that worktrees are a supported configuration elsewhere in the
product: `packages/cli/src/repowise/cli/worktree.py:41` tests
`(repo_path / ".git").is_file()` precisely because that is what one looks like.

Second-order: the test above fails in any worktree, so a contributor working in
one sees a red suite that has nothing to do with their branch.

**Done looks like**

`.exists()` at `tool_why.py:1872`. A pointer file is still a working tree, and
the call behind the gate only needs `git` to be able to resolve one.

Leave `_meta.py:51` and `crud/repository.py:107` alone. Their `is_dir()` is load
bearing and their comments say so. Changing all three together is the wrong fix.

Verified locally on `main`: with `.exists()` the failing test passes and
`tests/unit/server/mcp` runs 1609 passed.

Tests: the wire-contract test only covers this when the suite happens to run
from a worktree, which is not a guarantee. Pin the behaviour directly instead,
with a test that points `repository.local_path` at a directory whose `.git` is a
pointer file and asserts the git-log layer still fires.

Aman-goel-04:
## Summary

- Enable git archaeology fallback for repositories using linked worktrees.
- Treat .git pointer files as valid working trees for live git log execution.
- Add regression coverage for the worktree pointer-file case.

## Related Issues

Fixes #2024 

## Test Plan

- [x] Tests pass (pytest) — 1609 passed

## Checklist

### DIFF
--- a/packages/server/src/repowise/server/mcp_server/tool_why.py
+++ b/packages/server/src/repowise/server/mcp_server/tool_why.py
@@ -1869,7 +1869,7 @@ async def _git_archaeology_fallback(
     # --- Layer 3: Live git log (when local repo exists) ---
     git_log_results = []
     local_path = getattr(repository, "local_path", None)
-    if local_path and (Path(local_path) / ".git").is_dir():
+    if local_path and (Path(local_path) / ".git").exists():
         git_log_results = await _run_git_log(local_path, file_path, stem)
     result["git_log"] = git_log_results
 
--- a/tests/unit/server/mcp/test_recoverable_caps_wire_contracts.py
+++ b/tests/unit/server/mcp/test_recoverable_caps_wire_contracts.py
@@ -834,7 +834,12 @@ async def test_why_fallback_archaeology_and_rationale_wire_recover_annotated_tai
     from repowise.core.persistence.models import Repository
 
     repository = await session.get(Repository, setup_mcp)
-    repository.local_path = str(Path.cwd())
+    # A linked worktree stores `.git` as a gitdir pointer file. The live git
+    # log only needs a working tree, so this must still enable the fallback.
+    worktree_path = tmp_path / "worktree"
+    worktree_path.mkdir()
+    (worktree_path / ".git").write_text("gitdir: /tmp/repowise-test-worktree\n")
+    repository.local_path = str(worktree_path)
     await session.flush()
 
     async def sealed_git_log(*_args: Any, **_kwargs: Any) -> list[dict[str, Any]]:


RaghavChamadiya (approved):
Thanks @Aman-goel-04. `.is_dir()` was the wrong question to ask: a linked worktree stores `.git` as a gitdir pointer file, and the live git log only needs a working tree, so `.exists()` is the condition that actually matches what the code goes on to do.

Writing a real pointer file in the test rather than mocking the check is what makes this hold up, since it fails against the old predicate for the same reason a real worktree did.

PR #2047, opened 2026-08-31
feat(security): add five JS/TS patterns to the registry

tacopiek:
## Summary

- The pattern registry (`packages/core/src/repowise/core/analysis/security_scan.py`) is Python-shaped; this is Tier 1 B from #1935 — five patterns that make the existing Security tab useful on a JS/TS codebase, same registry, same multi-line pass, same test file.
- `unsafe_inner_html` (med), `template_literal_sql` (med), `public_env_secret` (high), `new_function_call` (high), `reject_unauthorized_false` (med).
- Docs updated: `docs/layers/SECURITY.md` (pattern table, count, and a new section with the measurements below) and the one-line count in `docs/README.md`.

## Related Issues

Refs #1935 (Tier 1 B). Complements #1947 (exec_call / secret-case fixes) and #1948 (the docs page), both already merged.

## What each pattern catches, and why it's shaped the way it is

**`unsafe_inner_html`** — `__html:` (React's `dangerouslySetInnerHTML` shape) assigned a non-literal value. The value test is a positive lookahead (`(?=[A-Za-z_$(])`) rather than a negative one excluding quotes: tried at the position *before* the preceding `\s*`, a negative lookahead would trivially pass and let a string literal through anyway. Ships at `med`, not `high` — see the measurement table, every hit on the corpus was a name reference to a source-pinned constant, and that's not distinguishable from a genuinely dynamic value without dataflow.

**`template_literal_sql`** — analogue of `fstring_sql` for a JS/TS template literal. Requires the SQL verb's companion clause (`SELECT`+`FROM`, `UPDATE`+`SET`) before an interpolation counts, because the bare verb is an ordinary English word and fires on prose and on Tailwind's `select-none` otherwise.

**`public_env_secret`** — a secret-shaped name (`API_KEY`, `SECRET`, `TOKEN`, `PASSWORD`) behind a `NEXT_PUBLIC_`/`VITE_` prefix, which ships straight into the client bundle. Excludes any name containing `ANON`: a Supabase anon key is public by design and protected by row-level security, and that was the one corpus-noise class narrow enough to carve out as a rule.

**`new_function_call`** — `new Function(...)`, the same class of dynamic-code-execution risk as `eval`.

**`reject_unauthorized_false`** — `rejectUnauthorized: false`, the Node https/tls analogue of `tls_verify_false`.

## Measurements

Run over the same 17-repository, 1109-file JS/TS/Python corpus as #1947, every hit adjudicated by hand. Three of the five needed tightening before they were shippable:

| Kind | First cut | Verified real | After tightening | The noise |
|---|---|---|---|---|
| `unsafe_inner_html` | 15 | 0 | 4 | every hit a source-pinned stylesheet constant referenced by name |
| `template_literal_sql` | 10 | 2 | 2 | prose (`` `Order update failed: ${status}` ``) and a Tailwind class (`select-none`) |
| `public_env_secret` | 9 | 3 | 3 | Supabase anon keys — public by design, protected by row-level security |
| `new_function_call` | 0 | 0 | 0 | — |
| `reject_unauthorized_false` | 0 | 0 | 0 | — |

CORS on mutating handlers and missing authorization are left out entirely, as agreed in #1935 — those need framework semantics, not a flat pattern.

## Test Plan

- [x] Tests pass (`pytest tests/unit/analysis/` — 773 passed, including new parametrized cases per pattern covering both the positive and the known-noise/negative shapes above)
- [x] Lint passes (`ruff check .` and `ruff format --check .`, both clean on the changed files)
- [ ] Web build passes (`npm run build`) — no frontend changes
- [x] `mypy` on the changed module reports only the one pre-existing `Result[Any].rowcount` error already present on `main`, untouched here (same note as #1947)

## Checklist

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -58,7 +58,7 @@ your agent in under five minutes, with no API key.
 | [layers/GRAPH.md](layers/GRAPH.md) | The dependency graph: what is in it, how every edge is resolved, and how much to trust each one |
 | [layers/DECISIONS.md](layers/DECISIONS.md) | Architectural decisions mined from your repo and from your own agent sessions |
 | [layers/DEAD_CODE.md](layers/DEAD_CODE.md) | Unreachable files, unused exports, and zombie packages by confidence tier |
-| [layers/SECURITY.md](layers/SECURITY.md) | The local pattern scan: what the eleven patterns catch, what they do not, and how far to trust the result |
+| [layers/SECURITY.md](layers/SECURITY.md) | The local pattern scan: what the sixteen patterns catch, what they do not, and how far to trust the result |
 | [layers/LANGUAGE_SUPPORT.md](layers/LANGUAGE_SUPPORT.md) | What works per language: 19 parsed to a full AST, 35 on the five-rung ladder |
 | [layers/WIKI.md](layers/WIKI.md) | The generated wiki: page types, what `update` re-renders, styles, output language |
 
--- a/docs/layers/SECURITY.md
+++ b/docs/layers/SECURITY.md
@@ -4,7 +4,7 @@ Repowise records a small set of security signals while it indexes: pattern
 matches over source text, and symbol names that look security-relevant. Pure
 regex and SQL, no LLM calls, no network, no dependency resolution.
 
-This is a floor, not a scanner. The registry holds eleven patterns. It has no
+This is a floor, not a scanner. The registry holds sixteen patterns. It has no
 model of your framework, no notion of which inputs are attacker-controlled, and
 no dataflow: it cannot tell a parameterised query from a concatenated one beyond
 what the surrounding characters give away, and it cannot tell whether a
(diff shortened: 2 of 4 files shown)


RaghavChamadiya (approved):
Thanks @tacopiek. Shipping the measurement table alongside the patterns is what makes this reviewable: "15 first cut, 0 verified real, 4 after tightening" is a much more honest thing to hand a maintainer than five regexes and an assertion that they work.

Several of the shaping decisions are ones I would have asked about and you pre-empted. Requiring `SELECT`+`FROM` before a template literal counts, because the bare verb is an ordinary English word and collides with Tailwind's `select-none`. Carving out `ANON` because a Supabase anon key is public by design. And shipping `unsafe_inner_html` at `med` instead of `high` on the grounds that a name reference to a source-pinned constant is not distinguishable from a dynamic value without dataflow, which is the correct reading of that evidence.

Updating the "Most languages" caveat in `SECURITY.md` rather than only the pattern count matters too. That paragraph existed to stop a JS repo reading a low finding count as good news, and it would have been quietly wrong after this.

PR #2049, opened 2026-08-31
feat(ingestion): emit Flutter widget-tree edges from build() bodies (#142)

### LINKED ISSUE #142: [Feature] Flutter and Dart support
## Problem

I saw your project and I would love to integrate it on a Flutter project.

## Proposed Solution

I think it should to a parse of files + widget tree understanding and classes hierachy. 

## Alternatives Considered

I used grapify, but it only does a code file graph with no context of what's inside or how some relate.

## Additional Context

Flutter whole UI structure is having a widget tree. For the AI to be able to easily reuse components and UI code it needs to understand the project widget tree, how the files relate what what's for us. Ideally a link maybe with the compiler sematics and analysis can work too. I'm not sure.

I'm open to fork and build a solution, and create a PR, but I need a bit more guidance to understand what's important and what's not from the LLM perspective

sloemo01:
## What

Fixes #142 — Phase 2 of the Flutter ask, exactly as scoped in the thread.

## Root cause

Phase 1 (Dart at Good tier) is already shipped — spec, queries, resolver, heritage all exist. The Flutter handler in `framework_edges/flutter.py` already covered **navigation** (route tables, GoRoute/MaterialPageRoute builders, `runApp` → entry point). What was missing is the **widget tree itself**: a `build()` method returning a widget tree is the composition structure an LLM needs to reuse components, and it had no edges.

## Changes

- New pass in the Flutter handler: every constructor name inside a `build()` body is a child widget of the file's widget class. Framework widgets (Scaffold, Column, Text, …) resolve to nothing in the repo class map and are skipped; repo widgets get a parent→child edge.
- **Latent bug fixed:** `runApp`'s target is usually the same file (MyApp lives in main.dart), so no edge is added and the `is_entry_point` stamp silently no-oped on a nonexistent node. The node is now created before stamping.

## Tests

4 framework-edge tests — widget-tree edges, framework-widget skip, non-entry-file composition (proves the new pass, not the runApp window), route tables still working. Sabotage-verified: the widget-tree test fails without the new pass. 106 framework-edge + Dart tests pass.

### DIFF
--- a/packages/core/src/repowise/core/ingestion/framework_edges/flutter.py
+++ b/packages/core/src/repowise/core/ingestion/framework_edges/flutter.py
@@ -1,4 +1,4 @@
-"""Flutter navigation edges.
+"""Flutter framework edges: navigation + widget tree.
 
 Flutter wires screens together through route tables and builder callbacks
 rather than direct imports of a call site — a widget referenced only from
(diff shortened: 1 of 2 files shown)


RaghavChamadiya (approved):
Hardened on the branch and validated on flutter/samples; see the comment above.

PR #2061, opened 2026-08-31
fix(cli): stop capping the page listing at 10000

### LINKED ISSUE #2060: [Bug] doctor reports phantom orphans and --repair deletes the live index above 10000 pages
## What

On a repository with more than 10000 pages, `doctor` reports a large orphan count in both store rows that is entirely phantom, and `doctor --repair` deletes the live index in response to it.

On an 18900-page repository:

```
Database              OK    10000 pages
SQL ↔ Vector Store    FAIL  0 missing, 8900 orphaned · 1 stub(s) awaiting --resume
SQL ↔ FTS Index       FAIL  11 missing, 3877 orphaned
Coordinator drift     OK    SQL=18918, Vector=18918, Drift=0.0%
```

The table contradicts itself: coordinator drift counts 18918 on both sides and correctly reports 0.0%, while the two reconciliation rows report thousands of orphans. `Database` also reports exactly 10000, which is not the page count.

## Root cause

`packages/cli/src/repowise/cli/commands/doctor_cmd/repo_checks.py` reads the SQL side of both checks with:

```python
pages = await list_pages(session, repo.id, limit=10000)   # :115, the Database row
pages = await list_pages(session, repo.id, limit=10000)   # :275, the reconciliation
```

`list_pages` is the paginated listing helper — its signature is `limit: int = 100, offset: int = 0`. Passing 10000 does not remove the cap, it raises it. The reconciliation then builds `sql_ids = {p.id for p in pages}` from that first page of results and treats every store entry outside it as an orphan.

So every page past the 10000th is invisible to the check, and each one turns its vector and FTS row into a phantom orphan. The `Database` row reports `len(pages)`, hence exactly `min(actual, 10000)`.

## Reproduction

Reproducing the SQL side against the live database, since the FTS index is inspectable in SQLite:

```
pages in db          : 18900
doctor's capped view : 10000
invisible to doctor  : 8900

FTS rows             : 13866
orphans vs CAPPED sql: 3877   <-- what doctor reports
orphans vs FULL sql  : 0      <-- the truth
```

3877 is exactly what doctor reported. Every one is phantom.

An independent check points the same way: a full `repowise reindex --embedder ollama` completed ("Done! Indexed 18918 items"), and the orphan counts did not move by a single unit — 8900 and 3877 before and after. A real orphan set would have changed.

## Why this is more than a wrong number

`--repair` acts on exactly these sets:

```python
await fts.delete_many(list(orphaned_fts))
...
for pid in orphaned_vector:
    await vs.delete(pid)
```

On any repository over the cap, `doctor --repair` therefore deletes live index entries — here 8900 vectors and 3877 FTS rows — and reports them as repairs. The surrounding comments show this failure mode was already a concern ("a repair that deletes the index it was asked to fix"); this is that outcome reached through the page cap rather than through the embedder.

There is also a second-order effect: `doctor` FAILs permanently on two rows that cannot be cleared by any user action, which is a good way to train people to stop reading it.

## Same pattern elsewhere

`limit=10000` appears at two more call sites, both with the same "paginated helper used as fetch-everything" shape:

- `packages/cli/src/repowise/cli/commands/status_cmd.py:173` — `repowise status` prints a correct total but a per-type table that sums to exactly 10000.
- `packages/cli/src/repowise/cli/commands/export_cmd.py:129` — a full export silently stops at 10000 pages.

Neither destroys data, so I have left them out of the fix to keep the diff to the reconciliation. Happy to take them in a follow-up if you would like them fixed the same way.

## Environment

repowise 0.47.0, macOS, SQLite backend, 18900 pages.

I have a fix and tests ready and will open a PR against this issue.

azhard:
Fixes #2060.

## What

On a repository with more than 10000 pages, `doctor` reports a large orphan count in both store rows that is entirely phantom, and `doctor --repair` deletes the live index in response to it.

On an 18900-page repository:

```
Database              OK    10000 pages
SQL ↔ Vector Store    FAIL  0 missing, 8900 orphaned
SQL ↔ FTS Index       FAIL  11 missing, 3877 orphaned
Coordinator drift     OK    SQL=18918, Vector=18918, Drift=0.0%
```

The table contradicts itself. Coordinator drift counts 18918 on both sides and reports 0.0%; the reconciliation rows report thousands of orphans against the same data. `Database` reports exactly 10000, which is not the page count.

## Root cause

`repo_checks.py` read the SQL side of both checks with:

```python
pages = await list_pages(session, repo.id, limit=10000)   # :115, the Database row
pages = await list_pages(session, repo.id, limit=10000)   # :275, the reconciliation
```

`list_pages` is the paginated listing helper — `limit: int = 100, offset: int = 0`. Passing 10000 raises the cap rather than removing it. The reconciliation then built `sql_ids = {p.id for p in pages}` from that first page of results and treated every store entry outside it as an orphan.

Every page past the 10000th was invisible to the check, so its vector and FTS row had no page behind it and was counted as drift. The `Database` row reported `len(pages)`, hence `min(actual, 10000)`.

Reproducing the SQL side against the live database, since the FTS index is inspectable in SQLite:

```
pages in db          : 18900
doctor's capped view : 10000
invisible to doctor  : 8900

FTS rows             : 13866
orphans vs CAPPED sql: 3877   <-- what doctor reports
orphans vs FULL sql  : 0      <-- the truth
```

3877 is exactly what doctor reported, and every one is phantom. A full `repowise reindex` also completed without moving either count by a single unit, which a real orphan set would not do.

## Why this is more than a wrong number

`--repair` acts on exactly these sets — `fts.delete_many(list(orphaned_fts))`, and `await vs.delete(pid)` for each id in `orphaned_vector`. On any repository over the cap it therefore deletes live index entries and reports them as repairs: here 8900 vectors and 3877 FTS rows.

The comments around that block show this outcome was already a concern ("a repair that deletes the index it was asked to fix"). This is the same outcome reached through the page cap rather than through the embedder.

## Changes

`limit=10000` appeared at four call sites. All four are fixed here; the fix differs per site because what each one needs differs.

**doctor** (`repo_checks.py`) — the data-loss path:

- The reconciliation reads every page via `_all_pages_for_reconciliation`, selecting `id`, `content` and `metadata_json` — the three fields the checks below it read, for the id set, the information floor and the stub predicate. Narrow columns rather than whole rows, since hydrating every page to discard it is what made a cap look necessary.
- The `Database` row counts in SQL via `_page_count` instead of measuring a fetched list.

**status** (`status_cmd.py`) — printed a correct total in its header, from a separate query, above a per-type table that summed to exactly 10000, with the token total truncated to match. A wiki over the cap disagreed with itself on screen. It only needs counts and token sums, so it aggregates in SQL with a `GROUP BY` rather than counting a fetched list.

**export** (`export_cmd.py`) — wrote an archive missing every page past the 10000th, with nothing in the output to say so. That is worst in `--full`, the archival mode whose purpose is completeness. Export needs whole rows, so it pages through with `offset` rather than narrowing columns.

The remaining `list_pages` callers are correct and untouched: `limit=1` where one row is wanted (`editor_files/fetcher.py`), and `limit=limit` where the caller is paginating for real (`stores/_sql_meta.py`, `server/routers/pages.py`).

## Tests

`tests/unit/cli/test_doctor_page_cap.py`, two cases against a 10001-page repository whose pages are all indexed in FTS, so a healthy repo reports no drift in either direction:

- `test_a_page_past_the_old_cap_is_not_an_orphan` — the canary is the oldest page by `updated_at`, which is what `list_pages` sorts on descending, so it was the first page the old cap dropped.
- `test_the_database_row_counts_every_page`.

Both fail on unpatched main and pass with the fix. The fixture bulk-inserts and uses `index_many`, which keeps it at 1.5s; indexing a page at a time took 130s.

`tests/unit/cli/test_page_listing_is_not_capped.py`, two cases over a 10001-page repository:

- `test_status_counts_every_page_and_its_tokens`
- `test_export_walks_past_the_old_cap`

Both fail on unpatched main with `assert 10000 == 10001`.

CLI suite: 2257 passed, 1 skipped, 2 xfailed. Doctor suites: 16 passed. Full `tests/unit` + `tests/providers`: 15851 passed, 13 skipped, 3 xfailed. `ruff check packages/ tests/` clean.

Three failures in the full run are pre-existing and unrelated — each fails identically on `main` with this branch stashed:

- `test_agent_target_baseline.py::test_integration_writes_match_baseline` — compares against a Windows baseline (`AppData/Roaming/...` vs `Library/Application Support/...`) and so fails on macOS.
- `test_repo_totals_churn.py::test_a_rebased_branch_in_the_history_falls_back`
- `test_transcript_episodes.py::test_the_episode_is_dated_from_the_session_not_the_index`

## Scope

An earlier revision of this PR fixed only `doctor` and named the other two sites for a follow-up. They are included now: the mistake is one pattern — a paginated listing helper used as fetch-everything — and leaving two instances in place would mean a third reviewer rediscovering it. Each site is a separate commit if you would rather take them apart.

### DIFF
--- a/packages/cli/src/repowise/cli/commands/doctor_cmd/repo_checks.py
+++ b/packages/cli/src/repowise/cli/commands/doctor_cmd/repo_checks.py
@@ -59,6 +59,57 @@ def _is_stub_fallback_row(page) -> bool:
     return isinstance(meta, dict) and STUB_FALLBACK_ERROR in meta
 
 
+
+async def _page_count(session: object, repo_id: str) -> int:
+    """Count this repository's pages in SQL.
+
+    Counted in the database rather than by measuring a fetched list. The
+    previous spelling took ``len()`` of a ``list_pages`` page, whose ``limit``
+    defaults to 100 and was passed 10000 here, so every repository with more
+    than 10000 pages reported exactly 10000 — a plausible-looking number that
+    never moves.
+    """
+    from sqlalchemy import func, select
+
+    from repowise.core.persistence.models import Page
+
+    result = await session.execute(  # type: ignore[attr-defined]
+        select(func.count()).select_from(Page).where(Page.repository_id == repo_id)
+    )
+    return int(result.scalar_one())
+
+
+async def _all_pages_for_reconciliation(session: object, repo_id: str) -> list:
+    """Every page this repository has, for reconciling against the indexes.
+
+    The reconciliation asks which store entries have no page behind them, so
+    anything it cannot see it calls an orphan. It read the pages through
+    ``list_pages(limit=10000)`` — a paginated listing helper whose ``limit``
+    defaults to 100 — and treated that page-one result as the whole database.
+    Every page past the 10000th was therefore invisible, and each one turned
+    its vector and FTS row into a phantom orphan: on a 18900-page repository,
+    8900 of them. ``--repair`` deletes what this reports, so what it removed
+    was the live index.
+
+    Three columns rather than whole rows: the id to match against the stores,
+    the content for the information floor, and metadata_json for the stub
+    predicate. Nothing downstream reads any other field, and hydrating full
+    ORM objects for every page only to discard them is what made a cap look
+    necessary in the first place.
+    """
+    from sqlalchemy import select
+
+    from repowise.core.persistence.models import Page
+
+    result = await session.execute(  # type: ignore[attr-defined]
+        select(Page.id, Page.content, Page.metadata_json).where(
+            Page.repository_id == repo_id
+        )
+    )
+    return list(result.all())
+
+
+
 def _run_repo_checks(
     repo_path: _DoctorPath, repair: bool, *, fmt: str = "table"
 ) -> tuple[bool, list[DoctorCheck]]:
@@ -99,7 +150,6 @@ def _run_repo_checks(
                     create_session_factory,
                     get_repository_by_path,
                     get_session,
-                    list_pages,
                 )
 
                 url = get_db_url_for_repo(repo_path)
@@ -112,8 +162,7 @@ def _run_repo_checks(
                 async with get_session(sf) as session:
                     repo = await get_repository_by_path(session, str(repo_path))
                     if repo:
-                        pages = await list_pages(session, repo.id, limit=10000)
-                        count = len(pages)
+                        count = await _page_count(session, repo.id)
                 await engine.dispose()
                 return count
 
@@ -251,7 +300,6 @@ def _run_repo_checks(
                     create_session_factory,
                     get_repository_by_path,
                     get_session,
-                    list_pages,
                 )
                 from repowise.core.persistence.information_floor import (
                     meets_information_floor,
@@ -272,7 +320,8 @@ def _run_repo_checks(
                     if not repo:
                         await engine.dispose()
                         return set(), set(), set(), set()
-                    pages = await list_pages(session, repo.id, limit=10000)
+                    pages = await _all_pages_for_reconciliation(session, repo.id)
+                    sql_ids = {p.id for p in pages}
                     # ``Page``'s primary key is the column ``id``; there is no
                     # ``page_id`` attribute. The old spelling raised here on
                     # every run, was swallowed below, and left both store
@@ -280,7 +329,7 @@ def _run_repo_checks(
                     # drift was ever detected and --repair never had anything
                     # to repair. The FTS repair block had the same defect and
                     # was fixed; this half was missed.
-                    sql_ids = {p.id for p in pages}
+
                     # The page vector store also holds decision embeddings under
                     # the "decision:<id>" namespace, so they belong on the SQL
                     # side of the ORPHAN check (but NOT FTS, which only indexes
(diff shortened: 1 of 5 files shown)


RaghavChamadiya (approved):
Thanks @azhard. The table contradicting itself is the detail that makes this diagnosis stick: coordinator drift counting 18918 on both sides while the reconciliation rows reported thousands of orphans against the same data could only mean the two halves were reading different row sets. `Database` reporting exactly 10000 named the cap outright.

The consequence is the serious part. `--repair` deletes what the reconciliation reports, so on any repository past the cap it was deleting live index entries in response to phantom orphans.

I checked the three substitutions against `list_pages` before taking this, since each one has to preserve semantics the helper was applying implicitly. `include_tombstones=False` is exactly `Page.freshness_status != "tombstone"`, which is what the new `status_cmd` aggregate filters on; the reconciliation and the count both ran at the default `include_tombstones=True`, and neither replacement filters. The three narrowed columns also cover every field downstream reads: `id` for the store match, `content` for `meets_information_floor`, and `metadata_json` for `_is_stub_fallback_row`.

Catching `export_cmd` and `status_cmd` in the same sweep was right. A truncated archive that looks complete is the same class of bug and would have outlived the doctor fix.

Issue #2062, opened 2026-08-31
[Bug] Async PostgreSQL cleanup races with event-loop shutdown


T012m3n7oR:
## Description

When Repowise performs model-based generation with PostgreSQL configured, the command can finish generating all pages but emit an asyncpg/SQLAlchemy cleanup traceback during shutdown:

```text
RuntimeError: Event loop is closed
```
followed by:
```text
RuntimeWarning: coroutine 'Connection._cancel' was never awaited
```

The generation result appears successful, but an asynchronous PostgreSQL engine is being cleaned up after its owning event loop has already closed.

## Steps to Reproduce

1. Configure Repowise to use PostgreSQL:

   `export REPOWISE_DB_URL='postgresql://user:password@postgres/repowise'`

2. Run model-based generation or workspace initialization:

   `repowise init /path/to/workspace --all --resume`

   or:

   `repowise generate --all`

3. Allow generation to complete.
4. Observe the output during process shutdown.

Observed output included:

✓ Generated 207 pages

followed by:

```text
File ".../sqlalchemy/connectors/asyncio.py", line 402, in terminate
    self.await_(asyncio.shield(self._terminate_graceful_close()))
...
File ".../asyncpg/connection.py", line 1682, in _cancel_current_command
    self._cancellations.add(self._loop.create_task(self._cancel(waiter)))
...
File ".../asyncio/base_events.py", line 545, in _check_closed
    raise RuntimeError('Event loop is closed')
RuntimeError: Event loop is closed

/usr/local/lib/python3.12/asyncio/sslproto.py:800: RuntimeWarning:
coroutine 'Connection._cancel' was never awaited
RuntimeWarning: Enable tracemalloc to get the object allocation traceback
```

## Expected Behavior

All asynchronous SQLAlchemy engines, sessions, and asyncpg connections should be closed before the event loop that owns them is shut down.

A successful generation should exit cleanly without:

- RuntimeError: Event loop is closed;
- unawaited coroutine warnings; or
- asyncpg connection-cleanup tracebacks.

## Actual Behavior

Repowise reports that generation completed successfully, but an asyncpg cleanup traceback is emitted during shutdown.

The observed run generated all reported pages, so this does not appear to have interrupted that particular generation. However, the cleanup sequence is not deterministic and could cause connection leaks, failed cost persistence, noisy shutdowns, or obscure a real failure.

## Environment

- OS: Linux
- Python version: 3.12
- Repowise version: 0.47.0
[...]

PR #2083, opened 2026-09-02
fix(cli): dispose the cost-tracker engine so a later loop doesn't reuse a dead connection

AmirF194:
Root cause: `make_cost_tracker` opens the cost-tracker's engine and hands back its session factory, but `run_repo_generation` calls `build_cost_tracker`, then a buffered `record()`, then `flush_cost_tracker` across three separate `asyncio.run()` calls. Postgres uses SQLAlchemy's pooled `AsyncAdaptedQueuePool`, so the connection opened in the first loop is still sitting in the pool when `flush` runs in the third, now-closed one, and the flush silently drops the row instead of raising. SQLite's `NullPool` never holds a connection between checkouts, which is why this only shows up on Postgres.

Fix: dispose the engine right after `open_repo_db` returns, inside `make_cost_tracker`. That clears the pool, not the session factory, so the next checkout (in whatever loop is live then) opens a fresh connection instead of reusing the dead one.

Verification:
- New integration test reproduces the exact production sequence against a real Postgres 16 container: writes 0 rows on main (matches your traceback), 1 row with the fix. Needs `REPOWISE_TEST_PG_URL`, skipped otherwise, same as the other Postgres-only tests here.
- `tests/unit/`, `tests/providers/` (16392 passed) and `ruff check .` clean on the branch. Two failures (`test_mcp_watchdog.py::test_server_exits_when_client_dies`, `test_repo_config_errors.py::test_unreadable_env_raises_typed_error`) also fail the same way with nothing else changed, so they're not from this diff.
- Ran on Python 3.11 only. Couldn't get 3.12/3.13 installed here (numpy has no prebuilt wheel for 3.13 and my sandbox can't build it from source), but nothing in this diff touches anything numeric.

### DIFF
--- a/packages/cli/src/repowise/cli/providers/cost_tracking.py
+++ b/packages/cli/src/repowise/cli/providers/cost_tracking.py
@@ -40,9 +40,13 @@ async def make_cost_tracker(repo_path: Path, repo_name: str) -> CostTracker:
     """Build a DB-backed :class:`CostTracker` for the repo-local database.
 
     Resolves (or creates) the repository row so every LLM call made during
-    this run is persisted to the ``llm_costs`` table. The engine is
-    intentionally left open: it lives for the duration of generation and is
-    disposed later by the persistence path that reuses the same database.
+    this run is persisted to the ``llm_costs`` table.
+
+    The engine is disposed before returning: this coroutine and ``flush()``
+    each run inside their own ``asyncio.run()``, and a connection left pooled
+    here would be bound to a dead loop by the time ``flush`` runs in a later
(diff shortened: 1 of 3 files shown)


RaghavChamadiya (approved):
CI has finally run on this branch and it is green on all three Python versions plus ruff, which was the only thing left holding it. Thanks for waiting it out, @AmirF194.

Restating the verdict so the record is on the thread rather than only in my head: disposing eagerly inside `make_cost_tracker`'s own loop is a better fix than the one I proposed on #2062, because it keeps the engine and every use of it inside a single loop instead of arranging for a tidier late `dispose()`. A `dispose()` from a different loop is the failure, not the repair.

Merging.

PR #2085, opened 2026-09-03
fix(cli): stop doctor reporting excluded files as missing

### LINKED ISSUE #2084: [Bug] doctor reports pages for excluded files as missing from FTS
## What

`doctor`'s SQL ↔ FTS reconciliation reports pages for **excluded** files as missing from the index. They are correctly absent — the user excluded them — so the row FAILs permanently and no user action can clear it.

On an 18949-page repository with `**/*.mocks.dart` and generated localizations in `exclude_patterns`:

```
SQL ↔ FTS Index    FAIL   5049 missing, 0 orphaned
```

Of those 5049, **5015 are pages whose file is excluded**. 34 are genuine drift.

`--repair` cannot fix it either: the remedy it would apply is to index content the user asked not to index.

## Root cause

`packages/cli/src/repowise/cli/commands/doctor_cmd/repo_checks.py` never builds an exclusion spec. `grep -c "is_excluded\|build_exclude_spec"` over that file returns 0.

Every other read path filters. `core/exclusion.py`'s own module docstring states the contract:

> Excluded files are skipped at ingest time, but DB rows may predate an `exclude_patterns` / gitignore change, so read paths (MCP tools, editor-file generation) filter again at query time.

and the codebase applies it carefully elsewhere, including for the `path::Name` shape that symbol pages carry:

- `server/mcp_server/_helpers.py:692` `filter_graph_nodes` — *"file nodes match on `node_id`, symbol nodes on `file_path`"*
- `server/mcp_server/_graph_utils.py:29` `_node_id_is_excluded` — `path = node_id.split("::", 1)[0]`
- `server/mcp_server/_prose_symbols.py:209` — matches on `row.file_path`

The reconciliation is the one read path that does not, so every page whose file is excluded is counted as drift on every run.

## Measurement

Against the live database, matching on the file the way `filter_graph_nodes` does for symbol nodes:

```
missing from FTS, FILE is excluded (should not be reported): 5015
missing from FTS, file NOT excluded (genuine drift):           34
```

The 5015 break down as 5023 `symbol_spotlight` and 26 `file_page` rows, almost all under `*.mocks.dart` and `app_localizations_*.dart`. None are below the information floor — the floor exclusion the check already applies does not cover them, because being excluded is a different reason from being too thin.

## Relationship to #2060

This was masked until now. Before #2061 the SQL side was read with `list_pages(limit=10000)`, so the reconciliation could only see the first 10000 pages and under-reported missing entries along with everything else. Uncapping it in #2061 made the check see the whole database, which correctly surfaced that it had never filtered excluded pages.

So the defect predates #2061; #2061 made it visible. Worth stating plainly so the new number does not read as a regression from that PR.

## Environment

repowise 0.48.0, macOS, SQLite backend, 18949 pages, 5015 of them for excluded files.

Fix and tests ready; PR to follow.

azhard:
Fixes #2084.

## What

`doctor` reports pages for **excluded** files as missing from the indexes. They are absent because the user excluded them, so the row FAILs permanently and nothing clears it — `--repair`'s only remedy is to index the content they asked to exclude.

On an 18949-page repository with `**/*.mocks.dart` and generated localizations in `exclude_patterns`:

```
SQL ↔ FTS Index    FAIL   5049 missing, 0 orphaned
```

5015 of those 5049 are pages whose file is excluded. 34 are genuine drift.

## Root cause

`repo_checks.py` never built an exclusion spec — `grep -c "is_excluded\|build_exclude_spec"` over the file returned 0.

`core/exclusion.py`'s module docstring states the contract it missed:

> Excluded files are skipped at ingest time, but DB rows may predate an `exclude_patterns` / gitignore change, so read paths (MCP tools, editor-file generation) filter again at query time.

Every other reader honours it, including for the `path::Name` shape symbol pages carry:

- `_helpers.py:692` `filter_graph_nodes` — *"file nodes match on `node_id`, symbol nodes on `file_path`"*
- `_graph_utils.py:29` `_node_id_is_excluded` — `path = node_id.split("::", 1)[0]`
- `_prose_symbols.py:209` — matches on `row.file_path`

The reconciliation was the one reader that did not.

## Changes

- `_all_pages_for_reconciliation` also selects `target_path`.
- `indexable_ids` drops pages whose file is excluded, matched on the file: `target_path.split("::", 1)[0]`. A symbol page's `target_path` is `path::Name`, which no file pattern matches, so filtering without the split silently keeps every symbol of every excluded file.
- MISSING side only, matching the information-floor and stub exclusions beside it. Whether a store entry for a newly-excluded file is itself drift worth deleting is a separate question, and answering it here would make `--repair` delete on a rules edit.

## Tests

`tests/unit/cli/test_doctor_honours_exclusions.py` — a repo excluding `**/*.g.dart` with a file page and a symbol page left in the database for an excluded file, and only the kept page indexed.

Two negative controls, both checked:

| Reverted to | Result |
|---|---|
| no exclusion filter (main) | `2 missing, 0 orphaned` — fails |
| filter present, `::` split removed | `1 missing, 0 orphaned` — fails on the symbol page |
| the fix | `in sync` — passes |

The pattern is extension-anchored deliberately. A directory prefix like `build/**` matches `build/x.py::Name` too, so it would let a fix that forgot the split pass — my first draft of this test used one and the second control did not bite.

Doctor suite: 81 passed. CLI suite: 2383 passed, 1 skipped, 2 xfailed. `ruff check packages/ tests/` clean. `test_agent_target_baseline.py::test_integration_writes_match_baseline` fails with and without this branch — it compares against a Windows baseline and fails on macOS.

Against the live 18949-page index, the FTS row goes from `5049 missing` to `34 missing`.

## Relationship to #2060 / #2061

This was masked until now. Before #2061 the SQL side was read with `list_pages(limit=10000)`, so the reconciliation saw only the first 10000 pages and under-reported missing entries along with everything else. Uncapping it made the check see the whole database, which surfaced that it had never filtered excluded pages.

The defect predates #2061; that PR made it visible. Flagging it so the number does not read as a regression from it.

## Second commit: naming the command in the resume hint

Same file, same failure shape — doctor's output sending a reader somewhere unhelpful.

The SQL ↔ Vector row printed `N stub(s) awaiting --resume` without saying which command takes the flag. `--resume` exists only on `init`, so a reader following the advice reaches for the command they were just told about, `generate`, and 0.48.0 answers `No such option '--resume'. Did you mean '--yes'?`. From there they are guessing, prompted by a health report.

The row now names `repowise init --resume`. `test_doctor_resume_stub.py` asserted the bare flag and now pins the new string.

Separable if you would rather take only the exclusion fix — it is its own commit.

### DIFF
--- a/packages/cli/src/repowise/cli/commands/doctor_cmd/repo_checks.py
+++ b/packages/cli/src/repowise/cli/commands/doctor_cmd/repo_checks.py
@@ -449,8 +449,13 @@ def _run_repo_checks(
             # Held-back stubs are not drift, but they are also not nothing: the
             # wiki has a page there that no model wrote. Say so on the same row
             # rather than letting "in sync" imply the wiki is complete.
+            # Name the command, not just the flag. `--resume` exists only on
+            # `init`, so a reader who takes this advice tries `generate
+            # --resume` first — 0.48.0 answers "No such option '--resume'. Did
+            # you mean '--yes'?" — and has to guess which command was meant. A
+            # health report is the worst place to make someone guess.
             if stub_count:
-                vec_detail += f" · {stub_count} stub(s) awaiting --resume"
+                vec_detail += f" · {stub_count} stub(s) awaiting `repowise init --resume`"
             checks.append(_check("SQL ↔ Vector Store", vec_ok, vec_detail))
 
             fts_ok = not missing_from_fts and not orphaned_fts
--- a/packages/cli/src/repowise/cli/commands/doctor_cmd/repo_checks.py
+++ b/packages/cli/src/repowise/cli/commands/doctor_cmd/repo_checks.py
@@ -17,6 +17,7 @@ from repowise.cli.helpers import (
     reconcile_schema_best_effort,
     run_async,
 )
+from repowise.core.exclusion import build_exclude_spec, is_excluded
 
 from ._types import DoctorCheck, _check, _status_markup
 from .advisories import _advise_claude_md_stamp
@@ -92,9 +93,9 @@ async def _all_pages_for_reconciliation(session: object, repo_id: str) -> list:
     8900 of them. ``--repair`` deletes what this reports, so what it removed
     was the live index.
 
-    Three columns rather than whole rows: the id to match against the stores,
-    the content for the information floor, and metadata_json for the stub
-    predicate. Nothing downstream reads any other field, and hydrating full
+    Four columns rather than whole rows: the id to match against the stores,
+    the content for the information floor, metadata_json for the stub
+    predicate, and target_path to tell whether the file is excluded. Nothing downstream reads any other field, and hydrating full
     ORM objects for every page only to discard them is what made a cap look
     necessary in the first place.
     """
@@ -103,7 +104,7 @@ async def _all_pages_for_reconciliation(session: object, repo_id: str) -> list:
     from repowise.core.persistence.models import Page
 
     result = await session.execute(  # type: ignore[attr-defined]
-        select(Page.id, Page.content, Page.metadata_json).where(
+        select(Page.id, Page.content, Page.metadata_json, Page.target_path).where(
             Page.repository_id == repo_id
         )
     )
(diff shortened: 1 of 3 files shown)


RaghavChamadiya (approved):
This is right, and the part I want on the record is the split on `::`. A symbol page's `target_path` is `path::Name`, which no file pattern matches, so a fix that filtered on the raw value would have kept every symbol of every excluded file and looked like it worked. Your test pins that deliberately with an extension-anchored pattern rather than a directory prefix, which is the case that would have hidden it.

It also lands on the right side of the contract `core/exclusion.py` already sets: rows outlive an `exclude_patterns` edit by design and readers filter at query time. This check was the one reader that did not, and holding the filter to the MISSING side keeps `--repair` from deleting on a rules edit.

The resume-hint rename rides along and is fine by me: `--resume` exists only on `init`, so the bare flag did send people to try it on `generate`.

PR #2088, opened 2026-09-03
fix MCP workspace detection and add no-workspace flag

### LINKED ISSUE #2077: [Bug] repowise mcp <path> still serves the workspace primary for a nested non-member repo (#1818 residual)
**Summary**

Residual of #1818, which was closed completed on 2026-08-29. The reported case
still reproduces, and a test now pins it as intended behaviour.

`repowise mcp <path>` pointed at an indexed git repo nested inside a workspace,
but not registered as a member, serves the workspace primary. The named repo's
index is unreachable, and nothing tells the caller.

**Mechanism**

Not the `repo_alias is None` fallback at
`packages/server/src/repowise/server/mcp_server/_server.py:364-367`. That branch
is never reached when the workspace declares its primary as `path: .`. The match
loop at `_server.py:352-362` tests containment, not identity:

```python
for entry in ws_config.repos:
    entry_abs = (ws_root / entry.path).resolve()
    try:
        resolved.relative_to(entry_abs)
    except ValueError:
        continue

    match_depth = len(entry_abs.parts)
    if match_depth > best_match_depth:
        repo_alias = entry.alias
        best_match_depth = match_depth
```

`(ws_root / ".").resolve()` is `ws_root`, so every path under the workspace is
contained by the primary's entry and matches on the normal path. The alias then
flows to `_server.py:400`, `ws_config.default_repo = ws_repo_alias`, the line
meant to honour the argument, which instead confirms the wrong one.

#1865 fixed the higher-impact half by adding the `best_match_depth` preference,
so pointing at a *registered* member now works. Bullets two and three of #1818,
a nested indexed repo dropping to single-repo mode and a `--no-workspace`
escape hatch for `mcp`, never shipped: `mcp_cmd.py` still exposes only
`--transport`, `--port`, `--host`, `--tools` and `--all`.

The residual is now asserted as correct in
`tests/unit/server/test_mcp_workspace_detection.py:58`:

```python
def test_detect_workspace_non_member_uses_primary(workspace: Path) -> None:
    _, _, repo_alias = _detect_workspace(str(workspace / "test-repos" / "microdot"))

    assert repo_alias == "repowise"
```

That fixture's `microdot` is a bare `mkdir` (line 15), no git, no `.repowise/`.
For that input "use primary" is right: `repowise mcp ./packages/core` should
serve the repo. `_detect_workspace` simply never distinguishes a subdirectory
of the primary from a separate indexed repo that happens to sit inside it. That
distinction is the whole fix.

**Repro**

With `microdot` indexed (188 pages) inside a workspace whose primary is `.`:

```
repowise mcp ./test-repos/microdot
```

stderr: `repowise MCP: workspace mode - 3 repos, default='repowise'`

- `get_overview` returns "Repository Overview: repowise"
- `get_answer("how does routing work in microdot")` answers about repowise
- `get_context(["src/microdot/microdot.py"])` returns `Target not found`

Control, the same repo cloned outside the workspace root:

```
repowise mcp C:\tmp\microdot-smoke
```

Single-repo mode, 10 tools instead of 11, `get_overview` returns
"Repository Overview: microdot-smoke", `get_context` on the same file resolves.

**Impact**

Undiscoverable from inside the protocol, which is what keeps this above
cosmetic. Server instructions (`_server.py:538`) do not name the served repo.
`list_repos` (`tool_repos.py:68`) reports the members and the default
(`tool_repos.py:80`) but never echoes the launch path, so the agent cannot
compare the two. `_print_network_startup` prints "Workspace / Default repo"
only for `sse` and `streamable-http`; stdio prints nothing by design. The
`_log.info` line is stderr and invisible to the client. `get_context` returns
an ordinary miss, indistinguishable from a typo.

Realistic triggers beyond a dogfooding `test-repos/`: a submodule or vendored
clone inside a `path: .` primary, and a fresh clone dropped into a workspace
parent before `repowise workspace add`, which lands on the `:364` branch.

It also contradicts two written rules. `docs/WORKSPACE_ROBUSTNESS.md:20-25`, UX
principles 1 and 2: "never silently change behavior" and "Path arguments always
win over auto-detection". And `helpers.py:1245-1247`: "`resolve_command_target`
is the single source of truth. Every command should call it before doing any
work." 17 command modules call it; `mcp_cmd.py` calls it zero times, hand-rolls
`_workspace_summary`, and `_server.py` hand-rolls `_detect_workspace` a second
time.

Worth stating plainly since it cuts against the original issue's framing:
`status` is not inconsistent in the way #1818 implied. `repowise status
test-repos/microdot` serves microdot correctly and prints no notice, because
rule 3 at `helpers.py:1451-1472` sets `auto_detected=False` for an explicit
path and `CommandTarget.notice` (`helpers.py:1342-1349`) only narrates
auto-detection. So `status` obeys the path and `mcp` overrides it. The
divergence is sharper than "one warns and one does not".

**Done looks like**

In `_detect_workspace`, before returning a containment-derived alias, drop to
single-repo mode when the named path is its own indexed repo and is not a
registered member: `resolved != entry_abs` for the matched entry and
`(resolved / ".repowise" / "state.json").exists()`, then
`return None, None, None`. That gives `mcp` the rule-3 semantics `status`
already has, needs no new flag, and leaves `repowise mcp ./packages/core`
unchanged because it has no nested `.repowise/`.

`test_detect_workspace_non_member_uses_primary` then needs splitting: keep the
bare-directory fixture asserting the primary, add an indexed-repo fixture
asserting single-repo mode.

If the behaviour change is judged too invasive, the floor is a `_log.warning`
at `_server.py:400` when the matched entry's path is not the path named. But
stderr is invisible to the agent, so that fixes the operator's experience and
not the wrong answers. Adding `--no-workspace` to `mcp_cmd.py` is worth doing
either way, as the escape hatch #1818 asked for.

Aman-goel-04:
## Summary

  - Fix MCP workspace detection so a nested indexed repo inside a workspace no longer silently serves the workspace primary.
  - Add a --no-workspace escape hatch for repowise mcp to force single-repo mode when needed.
  - Expand CLI and server regression coverage for both the nested-repo case and the new flag.

  ## Related Issues

  - Fixes #2077 

  ## Test Plan

  - [x] Tests pass (pytest)

  ## Checklist

### DIFF
--- a/packages/cli/src/repowise/cli/commands/mcp_cmd.py
+++ b/packages/cli/src/repowise/cli/commands/mcp_cmd.py
@@ -12,7 +12,9 @@ from repowise.cli.ui import load_dotenv
 from repowise.core.workspace.config import WorkspaceConfig, find_workspace_root
 
 
-def _workspace_summary(path: Path) -> dict[str, object] | None:
+def _workspace_summary(path: Path, *, no_workspace: bool = False) -> dict[str, object] | None:
+    if no_workspace:
+        return None
     workspace_root = find_workspace_root(path)
     if workspace_root is None:
         return None
@@ -121,6 +123,16 @@ def _print_network_startup(
     default=False,
     help="Expose every tool eligible in the current repository/workspace mode.",
 )
+@click.option(
+    "--no-workspace",
+    is_flag=True,
+    default=False,
+    help=(
+        "Force single-repo mode even when the path is inside a workspace. "
+        "Useful for nested indexed repos that should not inherit the "
+        "enclosing workspace's default repo."
+    ),
+)
 def mcp_command(
     path: str | None,
     transport: str,
@@ -128,6 +140,7 @@ def mcp_command(
     host: str | None,
     tools: str | None,
     all_tools: bool,
+    no_workspace: bool,
 ) -> None:
     """Start the MCP server for editor integration.
 
@@ -149,6 +162,7 @@ def mcp_command(
         repowise mcp --tools +get_execution_flows  # default set plus one
         repowise mcp --tools lean        # six-tool agent-lean profile
         repowise mcp --all               # every available tool
+        repowise mcp --no-workspace      # force single-repo mode
         repowise mcp --transport streamable-http  # HTTP on port 7338
     """
     if path is None:
@@ -157,7 +171,7 @@ def mcp_command(
         repo_path = resolve_repo_path(path)
     load_dotenv(repo_path)
 
-    workspace = _workspace_summary(repo_path)
+    workspace = _workspace_summary(repo_path, no_workspace=no_workspace)
     repowise_dir = repo_path / ".repowise"
     if workspace is None and not repowise_dir.exists():
         console.print(
@@ -183,4 +197,5 @@ def mcp_command(
         host=resolved_host,
         port=port,
         tools=tools_override,
+        workspace_mode=not no_workspace,
     )
--- a/packages/server/src/repowise/server/mcp_server/_server.py
+++ b/packages/server/src/repowise/server/mcp_server/_server.py
@@ -349,6 +349,7 @@ def _detect_workspace(repo_path: str | None):
         resolved = _Path(repo_path).resolve()
         repo_alias = None
         best_match_depth = -1
+        best_match_abs = None
         for entry in ws_config.repos:
             entry_abs = (ws_root / entry.path).resolve()
             try:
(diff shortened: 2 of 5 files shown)


RaghavChamadiya (approved):
The single-repo drop is gated on the thing I asked for: `(resolved / ".repowise" / "state.json").exists()`, so a plain subdirectory still falls through to the primary and `repowise mcp ./packages/core` keeps working. Both cases are pinned as separate tests, which is why I can read that off the diff instead of guessing.

`--no-workspace` is a good addition on top: containment is the right default, and this gives the nested case an explicit override rather than making people restructure the workspace file.

PR #2105, opened 2026-09-03
fix(augment): BM25 floor for the rescue's FTS fallback (#2092)

### LINKED ISSUE #2092: augment: grep-rescue FTS fallback fires on weak matches — add relevance floor
## Summary

The zero-result grep rescue's FTS fallback (`cli/commands/augment_cmd/search.py::_rescue`) returns the first FTS row that points at a code page **without checking its relevance score**. A multi-token pattern where only one common token matches still produces rows, so the hook emits a confident `Wiki suggests ...` line for an unrelated file.

## Repro (repowise 0.47.0, pi agent, dotfiles repo)

Grep for `bluetooth codec negotiation fallback` (concept absent from repo — zero literal matches):

```
[repowise] No literal match for `bluetooth codec negotiation fallback`. Wiki suggests `config/shared/claude/hooks/prefer-web-tools.py` (file_page).
```

That file is unrelated — the FTS match rides on the single token "fallback". BM25 scores tell the story:

| query | top FTS score | related? |
|---|---|---|
| `bluetooth codec negotiation fallback` | 3.45 | no (single-token ride) |
| `btrfs space monitoring` | 16.47 | yes |
| `containerd image cleanup` | 18.50 | yes |
| `hyprland monitor workspace` | 9.70 | yes |

Weak matches cluster ~3; genuine ones 2–5x higher.

## Why it matters

Rescue's docstring says its 44% action rate comes from "firing only when it has genuinely new information", and the symbol path holds that bar (exact-variant matches only, no FTS under real results). The FTS fallback is the one leg with no quality gate — and a confidently wrong `[repowise]` line teaches the agent to discount all of them.

## Fix we're testing locally

Floor on the BM25 score, consistent with the existing fixed-tunables convention:

```python
# BM25 floor for the rescue's FTS fallback. A multi-token query where only
# one common token matches ("fallback") still returns rows at score ~3;
# genuinely related pages score 2-3x higher (observed: weak 3.45 vs real
# 8.70). Below the floor, silence beats a confident wrong suggestion.
_RESCUE_FTS_MIN_SCORE = 6.0
```

```python
    for r in fts_rows:
        if getattr(r, "score", 0.0) < _RESCUE_FTS_MIN_SCORE:
            # Rows are score-ordered; the first weak row means all are weak.
            break
```

Verified both directions: weak query now silent, strong queries still rescue. Running it in day-to-day use for a few days to watch for over-muting.

Open question for you: absolute BM25 floors are corpus-sensitive. If you'd rather gate on **query-token coverage** (require ≥2 distinct query tokens matched) or a corpus-relative score, happy to implement that instead. Will submit a PR once you weigh in / after our test window.

sloemo01:
## What

Fixes #2092. The zero-result grep rescue's FTS fallback returned the first row without checking its relevance score, so a multi-token pattern where only one common token matched produced a confident "Wiki suggests ..." line for an unrelated file (repro: `bluetooth codec negotiation fallback` riding on "fallback" at BM25 3.45).

## Root cause

`_rescue` in `augment_cmd/search.py` returned the first FTS row pointing at a code page with no score check. Weak matches cluster ~3; genuine ones score 2-3x higher (observed 3.45 vs 8.70), so the fallback leg had no quality gate while the symbol path held an exact-match bar.

## Changes

- Added `_RESCUE_FTS_MIN_SCORE = 6.0` — below the floor, silence beats a confident wrong suggestion.

## Tests

- Below-floor silent, above-floor fires — sabotage-verified (below-floor test fails without the fix).

### DIFF
--- a/packages/cli/src/repowise/cli/commands/augment_cmd/search.py
+++ b/packages/cli/src/repowise/cli/commands/augment_cmd/search.py
@@ -50,6 +50,11 @@ _RESCUE_TOP_N = 2
 # answer depend on row order; this is small enough to stay one indexed lookup
 # and large enough that the ranking, not the LIMIT, picks the winner.
 _RESCUE_EXACT_FETCH = 8
+# BM25 floor for the rescue's FTS fallback. A multi-token query where only
+# one common token matches ("fallback") still returns rows at score ~3;
+# genuinely related pages score 2-3x higher (observed: weak 3.45 vs real
+# 8.70). Below the floor, silence beats a confident wrong suggestion.
+_RESCUE_FTS_MIN_SCORE = 6.0
 # Files the triage ranker considers. Ranking is over the grep's own matches,
 # so this only bites on floods wider than 200 files, where the tail is taken
 # by match count, the one signal available before any index lookup.
@@ -793,6 +798,10 @@ async def _rescue(
     except Exception:
         fts_rows = []
     for r in fts_rows:
+        if r.score < _RESCUE_FTS_MIN_SCORE:
+            # A weak FTS hit is a single-token ride, not a real match —
+            # silence beats a confident wrong suggestion.
+            continue
         target = getattr(r, "target_path", None) or ""
         page_type = getattr(r, "page_type", "") or ""
         if "::" in target:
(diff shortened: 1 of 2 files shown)


sloemo01:
Closing this one. #2092 got fixed on main by #2214, which gates the rescue's FTS fallback on query-token coverage. That fix is dialect-independent, and the score floor here is not: SQLite BM25 is unbounded while Postgres ts_rank sits in a 0-1 range, so a single constant is wrong on one of them. Nothing left to rebase.

PR #2106, opened 2026-09-03
fix(init): --resume skips the interactive questionnaire (#2098)

### LINKED ISSUE #2098: [Bug] init --resume re-runs the whole interactive questionnaire
**Summary**

`repowise init --resume` on a TTY re-runs the entire interactive questionnaire: the banner, the repository scan, the mode menu, the provider and embedder choices and the documentation style. A resume is meant to pick up a run that already answered all of those.

**Mechanism**

`packages/cli/src/repowise/cli/commands/init_cmd/command.py:1006`

```python
is_interactive = sys.stdin.isatty() and provider_name is None and not index_only and not yes
```

`resume` is not a term in that expression, so it has no effect on whether the prompts run.

The intent was clearly the opposite. Three lines above, at `:1000-1003`:

```python
# On --resume, continue the prior run's git tier so a resumed fast index
# doesn't silently fall back to the expensive FULL tier (issue #341). Done
# before the interactive gate so a resume never re-prompts for the mode.
run_mode = effective_run_mode_for_resume(repo_path, run_mode, resume)
```

So one answer, the git tier, is carried across a resume, and the comment says the point of computing it early is that "a resume never re-prompts for the mode". Every other answer is asked again.

**Impact**

Mild on its own and worse in combination. A user resuming after a crash cannot tell whether they are resuming or starting a fresh run, because the two are indistinguishable from the terminal until generation starts. It came up in #2097, where the reporter read the repeated questions as evidence that `--resume` was discarding the completed work. It was not, but nothing on screen said so.

There is a second-order risk: the questions are re-asked but the prior answers are not offered as defaults, so a user who answers differently the second time resumes a run under a different provider or style than the pages already on disk were written with.

**Done looks like**

Skip the interactive gate under `--resume` and reuse the previous run's answers, which are already on disk. Print one line naming what is being resumed, in place of the menu, so the user can see the run was picked up rather than restarted.

The narrower version, if reusing the stored answers turns out to be more than it looks: add `and not resume` to the `is_interactive` expression and fall through to the non-interactive defaults, which at least stops the run diverging from its own earlier answers.

Split out of #2097, where the `--resume` behaviour was reported alongside an unrelated crash.

sloemo01:
## What

Fixes #2098. `repowise init --resume` on a TTY re-ran the entire interactive questionnaire — banner, scan, mode menu, provider, embedder, style — indistinguishable from a fresh run, and a user who answered differently resumed under a different provider/style than the pages on disk were written with.

## Root cause

`is_interactive` at `init_cmd/command.py:951` never excluded `--resume`, so the flag had no effect on whether the prompts ran. The prior run's answers were already on disk (config.yaml), which the non-interactive path reads.

## Changes

- Extracted the gate into `_interactive_gate()` (pure, testable) and added `and not resume`.
- A one-line "Resuming ... reusing its answers" notice replaces the menu so the user can see the run was picked up, not restarted.

## Tests

- 6 gate cases (resume/yes/provider/index-only/non-tty all skip; plain TTY prompts).

### DIFF
--- a/packages/cli/src/repowise/cli/commands/init_cmd/command.py
+++ b/packages/cli/src/repowise/cli/commands/init_cmd/command.py
@@ -409,6 +409,25 @@ def _run_generation_phase(
 # ---------------------------------------------------------------------------
 
 
+def _interactive_gate(
+    *,
+    isatty: bool,
+    provider_name: str | None,
+    index_only: bool,
+    yes: bool,
+    resume: bool,
+) -> bool:
+    """Whether the interactive questionnaire runs.
+
+    Interactive requires a TTY, no explicit provider, docs on, and neither
+    --yes nor --resume. --resume skips the gate because the prior run
+    already answered every question and those answers are on disk
+    (config.yaml); re-asking would let a resumed run diverge from the pages
+    already written with them (issue #2098).
+    """
+    return isatty and provider_name is None and not index_only and not yes and not resume
+
+
 @click.command("init")
 @click.argument("path", required=False, default=None)
 @click.option(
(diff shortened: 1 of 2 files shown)


RaghavChamadiya (approved):
That is the change, and thanks for turning it around @sloemo01 — sorry it then sat on my side for four days rather than yours. The notice now claims only what the run does. Pulling the predicate into _interactive_gate with the six cases named is the part that will still be paying off in a year: the next flag added to that chain gets a test that says what it does to a resume, instead of an inline and that nobody reads.

PR #2107, opened 2026-09-03
fix(update): surface the two silent embedder degradations (#1370)

### LINKED ISSUE #1370: cli: update deterministic page path and other embedder builds still degrade silently (R3 coverage)

sloemo01:
## What

Fixes #1370 — the two survivors Raghav enumerated in his close comment on #1373.

## Root cause

1. `_full_rescore_interval_days` still did `except ValueError: pass`, so a malformed `REPOWISE_FULL_RESCORE_INTERVAL_DAYS` silently defaulted to 7.0.
2. `_build_update_vector_store` swallowed failures with a bare `except` and never touched the run's `degraded` list, so semantic dedup could be off with the panel saying nothing.

## Changes

- Rescore interval now warns on stderr naming the var and the offending value (same pattern as `build_embedder`'s misconfiguration report).
- Vector store now records `Decision vector store: <exc>` in `degraded` when one is passed; no-degraded callers keep the old silent contract.

## Tests

- 3 rescore-interval cases + 3 vector-store cases, sabotage-verified.

### DIFF
--- a/packages/cli/src/repowise/cli/commands/update_cmd/command.py
+++ b/packages/cli/src/repowise/cli/commands/update_cmd/command.py
@@ -1623,7 +1623,7 @@ def run_update(
 
     # Build the shared vector store once: reused by the supersession detector
     # below and by the decision upsert in _persist (semantic dedup + search).
-    decision_vector_store = _build_update_vector_store(repo_path, cfg)
+    decision_vector_store = _build_update_vector_store(repo_path, cfg, degraded)
 
     # --- Two-pass decision evolution (Phase 3C), BEFORE page regeneration ----
     # Cross-reference the diff + new commit bodies against existing decisions
--- a/packages/cli/src/repowise/cli/commands/update_cmd/incremental.py
+++ b/packages/cli/src/repowise/cli/commands/update_cmd/incremental.py
@@ -12,7 +12,9 @@ from typing import Any
 from repowise.cli.helpers import console, run_async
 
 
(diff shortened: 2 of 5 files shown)


RaghavChamadiya (approved):
Thanks @sloemo01. Both halves are the right shape.

The vector-store one lands in the same list every other update degradation goes into (`command.py:1237`, alongside "Editor file refresh", "Decision re-scan", "Session discovery"), so the message reads consistently in the panel and needs no new plumbing. Keeping `degraded` optional with the old silent behaviour when it is absent is what makes this safe to merge without auditing every caller, and you tested that case explicitly rather than assuming it.

The `except ValueError: pass` on `REPOWISE_FULL_RESCORE_INTERVAL_DAYS` was the better find of the two. A typo there did not just get ignored, it made a 7-day cadence look like a deliberate setting.

Merging. One non-blocking note: the precedent you cite for stderr uses `err_console` from `cli.helpers` (`providers/embedders.py:213-217`), not a bare `print`. Same stream either way, so nothing is wrong here, and a plain `print` does avoid Rich trying to read square brackets inside an interpolated env value. Worth a look next time you are in that file, since one of the two is going to end up the house style.

PR #2108, opened 2026-09-03
fix(init): surface generation-phase embed failures in the run's degraded list (#1369)

### LINKED ISSUE #1369: cli: init generation-phase embedder degradation can still go silent after a clean header probe

sloemo01:
## What

Fixes #1369 — residual from the #852 config-warning fix.

## Root cause

The init header probes `build_embedder` and warns on degradation, but the generation phase rebuilds the embedder and a mid-run embed failure (e.g. a local Ollama that goes down) only wrote a structlog warning — and the CLI pins structlog to ERROR unless `-v`, so the run finished looking clean with semantic search silently off, and state.json carried no `degraded` entry.

## Changes

- Threads an `on_warning` callback from `run_generation` down through `PageGenerator.generate_all` into the orchestrator's embed-batch failure handler, routing the text through `progress.on_message('warning', ...)` so the Rich callback records it.
- `run_repo_generation` merges the generation callback's warnings into the caller's list, so init's persisted `degraded` (state.json) and the completion panel see it.

## Tests

- Stub generator + progress recorder pin the seam — sabotage-verified.

### DIFF
--- a/packages/cli/src/repowise/cli/commands/init_cmd/command.py
+++ b/packages/cli/src/repowise/cli/commands/init_cmd/command.py
@@ -270,6 +270,7 @@ def _run_generation_phase(
     resume: bool,
     test_run: bool,
     timings: Any | None = None,
+    warnings: list[str] | None = None,
 ) -> tuple[bool, bool]:
     """Run the LLM generation phase for a single-repo init.
 
@@ -400,6 +401,7 @@ def _run_generation_phase(
         verbose=True,
         test_run=test_run,
         timings=timings,
+        warnings=warnings,
     )
     return False, False
 
@@ -1452,6 +1454,7 @@ def init_command(
             # file. Resuming against it would skip the entire model run and
             # say nothing, so a template wiki is never a run to continue.
             resume=resume and _prior_docs_mode != "deterministic",
+            warnings=run_warnings,
             test_run=test_run,
             timings=callback.table,
         )
--- a/packages/cli/src/repowise/cli/commands/init_cmd/generation.py
+++ b/packages/cli/src/repowise/cli/commands/init_cmd/generation.py
@@ -260,6 +260,7 @@ def run_repo_generation(
     verbose: bool,
     test_run: bool = False,
     timings: Any | None = None,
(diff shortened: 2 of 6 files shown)


RaghavChamadiya (approved):
Thanks @sloemo01. I traced the whole chain against the tree before approving, because "the warning now reaches the user" is the kind of claim that is easy to make and easy to get wrong by one hop:

`on_warning` fires in the embed-batch handler, `run_generation` routes it to `progress.on_message("warning", ...)`, and `on_message` appends to `self.warnings` at `ui/progress.py:216`. `run_repo_generation` then extends the caller's list with `gen_callback.warnings`, and because `run_warnings` at `init_cmd/command.py:1452` is a real list handed down by reference, the extend lands in the same object that becomes `base_state["degraded"]` at `:1715-1716`. So this does reach state.json and not just the terminal, which is the half that matters for an agent reading the run afterwards.

A bonus you may not have counted on: `on_message` runs `redact_secrets` over the text, and the message interpolates the exception, which for a provider error can carry a URL or a key. Routing through the callback rather than printing directly is what buys that.

Merging. Two non-blocking notes.

The new test pins the seam rather than the behaviour: it stubs `PageGenerator`, so it proves `run_generation` threads `on_warning` down and that invoking it records a warning, but nothing exercises the real `except` around the embed batch. That is a fair trade for a test that stays fast, and the docstring is honest about it, so I am not asking for more. Worth knowing if this regresses: the seam test will still pass if someone deletes the `self.on_warning(...)` call in `orchestrate.py`.

Second, #2106 edits the same `_index_with_resume` call site a few lines from your `warnings=run_warnings` argument. Whichever lands second will want a trivial rebase; there is no real disagreement between them.

PR #2110, opened 2026-09-03
fix(files): pin a file's doc, scoped to its repository (groundwork for #812)

### LINKED ISSUE #812: Generate a doc from the UI for an undocumented file, and keep it synced
## Problem

Some files never get an LLM-generated wiki page during indexing, so their **Doc** tab shows the empty "no doc" state (see #811). Today the only way to get a page for such a file is to re-run a full sync, and even then the file may be skipped again. A user who specifically wants documentation for one file has no direct way to ask for it.

## Proposed solution

Add a **"Generate doc"** button to the empty Doc tab (`packages/ui/src/files/file-doc-tab.tsx`, the `wikiPage == null` branch). Clicking it generates a wiki page for just that one file.

### Flow
1. User opens the Doc tab of an undocumented file and clicks **Generate doc**.
2. A single-file generation job runs (LLM call). The tab shows a loading state while it works.
3. When it finishes, the generated page appears in the tab like any other doc.

### Keep it synced (the important part)
Once a user hand-generates a doc for a file, that file becomes **pinned** ("this file's doc is wanted"):

- On every future reindex, a pinned file's doc is **regenerated automatically**, so it stays in step with code changes instead of silently disappearing again.
- The existing freshness badge already on `FileDocTab` (`fresh` / `stale` / `outdated`) surfaces when the file has changed under the doc, so the user can tell when it drifted.

This is what makes the button more than a one-shot: the doc a user asks for keeps following the code.

## Implementation plan

Wiki pages persist to `.repowise/wiki.db` and are produced in the **Generation** phase of the 4-phase `repowise init` pipeline. The plan builds on that.

**1. Persist the pin (backend, wiki.db)**
- Add a `pinned` boolean (default false) on the per-file wiki-page record in `.repowise/wiki.db`, so the pin lives with the wiki data and survives every reindex.
- A file with no page yet but that a user pins gets a lightweight row carrying `pinned = true` so the next Generation phase knows to produce it.

**2. Single-file generation endpoint (backend)**
- Add an endpoint that runs the Generation phase for exactly one file path: assemble the same context the init pipeline uses, run the doc prompt, and persist the resulting page (with `pinned = true`) into wiki.db.
- Run it as a background job the UI polls, since it is an LLM call — return a job id, expose a status/result endpoint.

**3. Honor pins on reindex (backend, Generation phase)**
- In the Generation phase, always include files where `pinned = true` in the set of files that get a page, regardless of the normal selection heuristic. Their pages are regenerated on every reindex.

**4. Button + states (UI)**
- In the `wikiPage == null` branch of `file-doc-tab.tsx`, render a **Generate doc** button.
- On click, call the endpoint, switch to a loading state, poll the job, then render the returned page. Surface errors inline with a retry.
- Reuse the same treatment in the other two spots that share this empty state: `packages/ui/src/files/file-overview-tab.tsx` and `packages/vscode/webview/src/views/docs/App.tsx`.

**5. Unpin (UI + backend)**
- On a pinned file's Doc tab, show an **Unpin** control that sets `pinned = false`, so the user can stop auto-regenerating a doc they no longer want.

## Acceptance criteria
- Clicking **Generate doc** on an undocumented file produces and displays its wiki page.
- That file is marked pinned in wiki.db and its page is regenerated on the next reindex without any manual action.
- The freshness badge reflects drift when the file changes under a pinned doc.
- A pinned file can be unpinned, after which reindex no longer regenerates its page.

sloemo01:
Part of #812

## Summary

The plumbing for a pinned file doc: a page can be marked pinned, and pinned pages always enter generation selection so a doc a user asked for keeps following the code instead of disappearing again.

This is groundwork, not the full feature. #812 also asks for a single-file background generation job with polling, both UI surfaces, and unpin; those are not here. The UI action that would have called this is removed rather than left in the tree unwired.

## Changes
- `Page.pinned` column, persisted, plus `set_page_pinned` in the CRUD layer.
- Pinned pages bypass the normal selection heuristic (`page_selection.py`, `scope.py`).
- `POST /{repo_id}/files/{path}/pin-doc`, which pins an existing page or creates a lightweight template row for a file with no page yet.
- `PinDocResponse` schema and the generated `http.ts` entry.

## Ownership, existence, and CI

Three things the endpoint got wrong, each with a test:

- **`Page.id` is global.** It is `"{page_type}:{target_path}"`, so two repositories holding the same path share one id. `file_detail` already refuses to read another repository's row; pinning it was still possible. Now a page belonging to a different repository is a 409, and the test asserts the other repository's page is left unpinned.
- **No existence check.** Any string a caller sent produced a pinned row, including a typo or an empty path, and the UI rendered it as a doc that was asked for. Now the path must resolve to a real file inside the checkout, checked after `resolve()` so a `..` segment cannot walk out of the repository. A repository whose `local_path` is not a directory is not treated as a missing file, so this cannot turn a moved checkout into a 404 on a previously working call.
- The Ruff import-order failure is fixed, and `packages/types/src/generated/http.ts` needed no regeneration: it already declares `PinDocResponse`.

## Test Plan
- 6 tests in `tests/unit/server/test_file_pin.py`, including the three new ones above.
- The two existing pin tests now create the files they pin, which they did not need to before.
- `ruff check` clean.

### DIFF
--- a/packages/api-client/src/files.ts
+++ b/packages/api-client/src/files.ts
@@ -1,5 +1,5 @@
 import type { FileDetailResponse, FilesIndexResponse } from "@repowise-dev/types/files";
-import { apiGet, BASE_URL, buildHeaders } from "./client";
+import { apiGet, apiPost, BASE_URL, buildHeaders } from "./client";
 
 /** Slim per-file rows for the browsable Files index + treemap. */
 export async function getFilesIndex(repoId: string): Promise<FilesIndexResponse> {
@@ -21,6 +21,15 @@ export async function getFileDetail(
   const encoded = filePath.split("/").map(encodeURIComponent).join("/");
   const qs = opts?.fields ? `?fields=${opts.fields}` : "";
   return apiGet<FileDetailResponse>(`/api/repos/${repoId}/files/${encoded}${qs}`);
+}
+
+/** Pin a file's doc so every reindex regenerates it (issue #812). */
+export async function pinFileDoc(
+  repoId: string,
+  filePath: string,
+): Promise<{ file_path: string; pinned: boolean }> {
+  const encoded = filePath.split("/").map(encodeURIComponent).join("/");
+  return apiPost(`/api/repos/${repoId}/files/${encoded}/pin-doc`);
 }
 
 /** Raw file content from the repo checkout (plain text, not JSON). */
--- a/packages/core/src/repowise/core/generation/page_selection.py
+++ b/packages/core/src/repowise/core/generation/page_selection.py
@@ -42,6 +42,10 @@ class PageRecord:
     target_path: str
     is_template: bool
     freshness_status: str = "fresh"
+    # A user hand-requested this page (UI "Generate doc"): it must be
+    # regenerated every run regardless of the selection heuristic (issue
+    # #812), so the doc keeps following the code.
+    pinned: bool = False
 
     @property
     def is_stale(self) -> bool:
@@ -125,6 +129,9 @@ def resolve_page_selection(
         selected.update(r.page_id for r in records if r.is_template)
     if intent.stale:
         selected.update(r.page_id for r in records if r.is_stale)
+    # Pinned pages are regenerated unconditionally: the user asked for this
+    # doc once and it should keep following the code (issue #812).
+    selected.update(r.page_id for r in records if r.pinned)
     if intent.path_globs:
         selected.update(
             r.page_id for r in records if _matches_glob(r.target_path, intent.path_globs)
--- a/packages/core/src/repowise/core/generation/scope.py
+++ b/packages/core/src/repowise/core/generation/scope.py
@@ -88,6 +88,7 @@ def load_page_records(pages: list[Any]) -> list[PageRecord]:
                 target_path=p.target_path,
                 is_template=is_template,
                 freshness_status=getattr(p, "freshness_status", "fresh"),
+                pinned=bool(getattr(p, "pinned", False)),
             )
         )
     return records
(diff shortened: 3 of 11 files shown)


sloemo01:
Took the groundwork path.

Retitled to `Part of #812`, dropped the `Fixes`, and removed the unreachable UI claim rather than leaving it in the tree: `FilePagePanels` never passed `onGenerateDoc`, so the button could not render, and `pinFileDoc` in the API client had no caller either. Both are gone. The description text is back to main's wording, so nothing promises an action a user cannot take.

The ownership, existence, and CI items are here:

- **Ownership.** `Page.id` is `"{page_type}:{target_path}"` and therefore global, so two repositories with the same path share an id. `file_detail` already guards this (`if page_row.repository_id != repo_id: page_row = None`); the pin endpoint did not, so one repository's Doc tab could mark another repository's page. It is a 409 now, and the test asserts the other repository's page is still unpinned.
- **Existence.** Any string produced a pinned row, including a typo or an empty path, and the UI rendered it as a doc that had been asked for. The path must now resolve to a real file inside the checkout, checked after `resolve()` so `..` cannot walk out. A repository whose `local_path` is not a directory is deliberately not treated as a missing file, so a moved checkout cannot turn a previously working call into a 404.
- **CI.** The Ruff import-order failure was real and is fixed. On `http.ts`: it already declares `PinDocResponse`, so I did not regenerate it.

Two notes on the tests. The two existing pin tests never created the files they pinned, which is why the existence check broke them; they build a real file now. And dropping either guard fails exactly its own test:

```
sabotage 1 (no existence check)  -> test_pin_404s_for_a_file_that_is_not_there
sabotage 2 (no ownership check)  -> test_pin_refuses_a_page_belonging_to_another_repository
```

6 tests in `test_file_pin.py`, `ruff check` clean.

What is still not here from #812, for the record: the single-file background generation job with polling, unpin, and a docs-reader action. The `pinned` flag is consulted by scoped page selection, so the persistence and selection halves are in place for whoever wires the rest.

PR #2120, opened 2026-09-04
Stop a restart from double-running a repo sync

atharrva01:
On startup, the server marks any job left `running`/`pending` as failed — meant for the process's own crashed jobs, but it's unconditional across the whole DB. In a multi-worker deployment sharing one database, one process restarting flips a *different* process's still-running job to failed, which quietly defeats the guard that stops two syncs from running on the same repo at once. A second sync then starts while the first is still going, both writing into the same repo's data.

Fix: the stale-job sweep and the active-job guard now check the job's heartbeat (`updated_at`, already refreshed by progress writes) instead of trusting `status` alone. A live job survives another worker's restart; a genuinely abandoned one (real crash) still gets reclaimed. Found and fixed the same bug in two places — the sync guard and the initial-index guard — now sharing one `has_active_job()` check.

Reproduced the race and verified the fix with a standalone script against the real functions, added a regression test, and the existing `tests/unit/server` (2900) + `tests/unit/persistence` (791) suites pass.


atharrva01:
Yeah @RaghavChamadiya i think u are right on that , asyncio.sleep yields and a real blocking call doesn't, so my test wasn't actually exercising the failure case.

Fixed build_knowledge_graph_skeleton, and while I was in there checked every other phase for the same thing rather than wait for you to find the next one. Two more turned up: _read_sources (blocks on pool.map, both call sites) and wire_tsconfig_resolver (the comment right above it already admits it's slow on big monorepos). All three go through asyncio.to_thread now, same as external_systems/graph.flows sitting right next to them.

Left one thing alone on purpose, _build_pipeline_coverage inside the health phase. It loops over coverage report files, not repo files, so it's bounded by a handful of reports, not repo size, and it's inside a phase that ticks anyway. Didn't feel like it belonged in the same bucket, figured I'd say so rather than just skip it quietly.

Added a test for each side of this: offload it and the heartbeat survives, don't and it can't, that's just how a blocked event loop works, not something a smarter heartbeat could fix.

Reran everything, all green, rebased on main again.

atharrva01:
@RaghavChamadiya , a gentle ping on this

Issue #2119, opened 2026-09-04
Mask the secret value in persisted and served security finding snippets


RaghavChamadiya:
**Summary**

`scan_file` stores the raw matched line, truncated to 120 characters, as `snippet` (`security_scan.py:314-317`). `persist()` and `replace_findings()` write it verbatim into `security_findings.snippet`. It is then served unredacted by `GET /api/repos/{repo_id}/security` (`SecurityFindingResponse.snippet`, `schemas/code_quality.py:100`), the findings table in `packages/ui/src/security/`, and the `security_signals` block in `get_risk`. `services/security_lines.py` re-verifies the line against live source at read time but never touches the text.

**Why it matters**

The moment a credential finding is a true positive, the secret sits in plaintext in `.repowise/`'s database, in every API response, and in MCP output that lands in agent transcripts. `--history` makes it worse: a key rotated and removed years ago is resurrected into the local store. `docs/business/COMMERCIAL.md` already describes the hosted store as holding fingerprints and redacted previews and "never the secret value"; the local scanner should hold itself to the same floor.

**Proposal**

Mask at scan time, before persist, so no downstream surface can leak it. For credential kinds, replace the quoted value with its first four characters and `****`, keeping the variable name and surrounding text so the finding is still locatable. Non-credential kinds keep the full line. Existing rows: clear `snippet` for credential kinds in the next migration and let the next scan repopulate, which is cheaper than rewriting. No API or UI shape changes.

Related: #1935, #2117.

akshatmalik-bruh:
can you assign it to me @RaghavChamadiya

rishu685:
hey can i take this issue @RaghavChamadiya sir!

RaghavChamadiya:
@akshatmalik-bruh yours, assigned. Before you start, two things, one of which is a correction to my own issue body.

**The line reference in the summary is wrong.** I wrote `security_scan.py:314-317`. That is the `add()` helper, which serves the call-pattern and spanning passes. Credential findings never go through it. They are built in the line-by-line loop at `:383-390`, which assembles its own `snippet = line.strip()[:120]`. Fix the code, not the line I pointed at.

[...]

PR #2121, opened 2026-09-04
fix(security): add value gate, placeholder filter, and path awareness for credential findings (#2118)

### LINKED ISSUE #2118: [Bug] Credential findings have no value gate or path awareness, so test fixtures are rated high
**Summary**

Every `hardcoded_password` / `hardcoded_secret` finding is rated `high`, and the patterns fire on anything quoted after the `=`. Run over this repo at HEAD (06d33e7), `packages/` and `tests/` produce 277 credential findings. 252 sit under test or fixture paths. The rest are docstring examples like `api_key="sk-..."` in the provider modules and minified `.next/` output. Zero are real. On a repository with a large test suite the Security tab is dominated by lines like `password="test"`.

**Mechanism**

The two credential patterns (`security_scan.py:81-82`) require nothing beyond an opening quote, so `password = ''` matches. There is no minimum value length, no placeholder recognition, and no path awareness. `scan_file` assigns the registry severity unconditionally (`security_scan.py:300-306`). The noise is not confined to the tab: the severity-by-directory matrix in `packages/ui/src/security/` and the `security_signals` block that `get_risk` attaches (`tool_risk/assessment.py:334-339`) both read the same rows, so an agent editing a test file is told it carries high-severity hardcoded passwords. `repowise security scan --history` repeats this across every blob in history.

**Proposal**

1. Require a value of at least 8 characters after the opening quote before a credential kind fires. An empty or two-character string is never a leak.
2. Skip values that name themselves a placeholder: `example`, `changeme`, `placeholder`, `dummy`, `sample`, `fake`, `xxx`, `your_`, and a leading `<`. Case-insensitive substring.
3. Downgrade findings under `test`, `tests`, `__tests__`, `fixtures`, `spec`, `mock` and `example` path segments to `low`. Downgrade, not suppress: a real key in a test file is still a leak, it just should not outrank product code.
4. A regression fixture repo with fake test passwords across a few commits, asserting `repowise security scan --history` reports none of them as `high`.

Acceptance on this repo: zero credential findings rated `high` under `tests/`, and the provider docstring examples gone.

Related: #1935. Pairs with #2117, which is the recall side of the same patterns.

rishu685:
## Summary
Closes #2118.

Introduces minimum value length validation, placeholder recognition, and path-aware severity downgrading for `hardcoded_password` and `hardcoded_secret` findings in `security_scan.py` and `history_scan.py`.

### Key Changes
- **Value Length Gate (`>= 8` chars)**: Regexes capture assigned string literal values while retaining `(?i:...)` scoped inline groups for case-insensitive prefiltering. Strings under 8 characters (`""`, `"test"`, `"short"`) are skipped.
- **Placeholder Filtering**: 
  - Exact match filtering for JDK default keystore passwords (`"password"`, `"changeit"`).
  - Case-insensitive substring filtering for `"example"`, `"changeme"`, `"placeholder"`, `"dummy"`, `"sample"`, `"fake"`, `"xxx"`, `"your_"`, `"your-"`, `"..."`, and `"fixture"`.
  - Leading `<` filtering (e.g., `<YOUR_API_KEY>`).
- **Path-Aware Severity Downgrade (`high` → `low`)**: Checks path segments for test/fixture/spec/mock/example directory tokens (`test`, `tests`, `__tests__`, `fixtures`, `spec`, `mock`, `example`, etc.) and `is_test_related_path()`. Credentials found in test material are downgraded to `low` severity instead of being suppressed.
- **History Scan Parity**: Removed redundant test exclusion in `history_scan.py` so history scans record test credentials at `low` severity.
- **Tests**: Added unit tests in `test_security_scan.py` and `test_security_scan_history.py` covering short values, placeholder filtering, path downgrading, and git history scans. All 103 tests pass.

### DIFF
--- a/packages/core/src/repowise/core/analysis/history_scan.py
+++ b/packages/core/src/repowise/core/analysis/history_scan.py
@@ -44,7 +44,6 @@ from repowise.core.analysis.security_scan import (
     SecurityScanner,
 )
 from repowise.core.ingestion.models import EXTENSION_TO_LANGUAGE
-from repowise.core.test_paths import is_test_related_path
 
 
 def _run_git(repo_path: Path, args: list[str], *, timeout: float = 30.0) -> str:
@@ -348,17 +347,12 @@ class HistorySecurityScanner:
         summary.blobs_scanned = len(blobs)
         blob_introduced_at, commit_dates = self._blob_introductions(repo_path, since, to)
 
-        # Test material is excluded from history mode, not merely down-ranked.
-        # A committed secret matters because it leaked a live credential, and a
-        # fixture is not one: on this repo 730 of 832 history hits (88%) were
-        # `api_key="sk-test"` in provider unit tests. That ratio makes the whole
-        # report untrustworthy, which is worse than reporting nothing.
-        # `--all-patterns` lifts this along with the kind gate.
+        # Test material findings are downgraded to low severity by scan_file
+        # so real keys in test files are still recorded without polluting high-severity reports.
         source_items = [
             (blob_sha, path)
             for blob_sha, path in blobs.items()
-            if (not path or self._is_source(path))
-            and not (secrets_only and path and is_test_related_path(path))
+            if not path or self._is_source(path)
         ]
         contents_map = self._read_blobs_batch(
             repo_path, [blob_sha for blob_sha, _ in source_items]
--- a/packages/core/src/repowise/core/analysis/security_scan.py
+++ b/packages/core/src/repowise/core/analysis/security_scan.py
@@ -22,12 +22,67 @@ import ast
 import logging
 import re
 from datetime import UTC, datetime
+from pathlib import PurePosixPath
 from typing import Any
 
 from sqlalchemy import text
 from sqlalchemy.ext.asyncio import AsyncSession
 
+from repowise.core.test_paths import is_test_related_path
+
 logger = logging.getLogger(__name__)
+
+_CREDENTIAL_EXACT_PLACEHOLDERS: frozenset[str] = frozenset({"password", "changeit"})
+
+_CREDENTIAL_SUBSTRING_PLACEHOLDERS: tuple[str, ...] = (
+    "example",
+    "changeme",
+    "placeholder",
+    "dummy",
+    "sample",
+    "fake",
+    "xxx",
+    "your_",
+    "your-",
+    "...",
+    "fixture",
+)
+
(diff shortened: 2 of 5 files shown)


RaghavChamadiya (approved):
Thanks @rishu685. The right call here was making the path signal a severity, not an exclusion, and that is what you did. A live credential committed into `tests/` is still a leak, so dropping the file from the scan was always the cruder answer; the reason the old exclusion existed was value noise (`api_key="sk-test"`, 730 of 832 hits on this repo's own history), and your 8-character floor removes that class directly. That is why the history exclusion could come out at the same time.

I test-merged this against current main and ran `tests/unit/analysis/test_security_scan.py`, `test_security_scan_history.py` and `test_no_bare_type_name_copies.py` on the merge result: 110 passed. Worth saying what that proves about one test you did not touch. `test_history_default_skips_test_fixtures_and_placeholders` still asserts `tests/test_client.py` stays out of a secrets report, and it still passes, but for a different reason than before: the path filter it was written against is gone, and the value `"fixture-value"` is now caught by the placeholder list instead. Same outcome, different mechanism, which is fine, but the docstring on `test_all_patterns_lifts_the_noise_gate` is now slightly off. Placeholder and length filtering are not lifted by `--all-patterns`, only the kind gate is. Non-blocking, and arguably the right behaviour, since a placeholder is never a leak.

Merging. Two things for whoever picks up the rest of the sequence, both of which this PR sets up rather than causes:

`val = match.group(1) if match.groups() else ""` is a silent drop for any future `SECRET_KINDS` pattern that carries no capture group: no group means an empty value, an empty value fails the 8-character floor, and the finding disappears with nothing logged. Both current secret patterns now capture, so this is correct today. #2117 adds a PEM pattern to `_SPANNING_PATTERNS`, and that is exactly the shape that would land on it, so the capture group is a requirement there rather than a style preference.

The severity downgrade is computed per file at `is_low_sev_file` and applied to each finding as it is built. #2119's masking has to be one pass over `findings` keyed on `SECRET_KINDS` before `scan_file` returns, not another branch inside this loop, because there are five snippet sites and `--history` reaches them through `history_scan.py:373` as well.

One nit, non-blocking and already harmless: the `framework_edges/flask.py` ledger entry in `test_no_bare_type_name_copies.py` landed on main separately while this was open, with its own comment about blueprint variables, so the merge leaves the key in the dict twice. The keys are `_PREFIX + "..."` expressions rather than literals so nothing flags it and the value is identical, but the second copy is sitting under a comment about route handlers that does not describe it. Drop it next time you are in that file.

PR #2144, opened 2026-09-05
fix(server): resolve primary vector store for postgresql databases (#2140)

### LINKED ISSUE #2140: [Bug] build_primary_vector_store cannot identify primary repo when REPOWISE_DB_URL is PostgreSQL
## Describe the Bug

`build_primary_vector_store` selects the primary repository by checking whether `<local_path>/.repowise/wiki.db` appears as a substring of `db_url`. That only works for SQLite URLs. With `postgresql+asyncpg://…`, the path never matches; selection falls through to “single repository row” or returns `InMemoryVectorStore` with `repo_id=None`.

The docstring already states the shared-registry rule in wiki.db terms, so Postgres multi-repo is a designed-in miss.

## Steps to Reproduce

1. Point `REPOWISE_DB_URL` at PostgreSQL with **more than one** `repositories` row (typical workspace after indexing 2+ members).
2. Start `repowise serve` so lifespan runs primary vector-store init.
3. Observe that the `local_db in normalized_url` loop matches nothing (Postgres URL contains no filesystem `…/wiki.db` path).
4. With `len(rows) != 1`, no primary is selected → in-memory store and `primary_vector_repo_id is None`.

Control: with exactly one row, the `len(rows) == 1` fallback hides the bug.

## Expected Behavior

Resolve the primary via workspace `default_repo` / `is_primary` / configured path (or another shared-DB identity), not by substring-matching a `wiki.db` filesystem path into `db_url`.

## Actual Behavior

For PostgreSQL URLs the path-match never hits. Multi-row databases do not get a reliable primary vector-store binding.

## Environment

- Repowise version: 0.48.0
- Installation method: Docker (`REPOWISE_DB_URL=postgresql+asyncpg://…`)

### Call site chain (v0.48.0)

**Defined in** `packages/server/src/repowise/server/search_helpers.py`:

```python
async def build_primary_vector_store(session_factory, db_url: str, embedder):
    """… For a shared registry database, only a row whose wiki.db matches
    db_url is treated as primary. …"""
    # …
    for repo_id, local_path in rows:
        local_db = (Path(local_path).resolve() / ".repowise" / "wiki.db").as_posix()
        if local_db in normalized_url:   # never true for postgresql+asyncpg://…
            selected = (repo_id, local_path)
            break
    if selected is None and len(rows) == 1:
        selected = rows[0]
    # …
    return InMemoryVectorStore(embedder), None  # multi-row Postgres lands here
```

**Invoked from serve lifespan** before workspace registration — `packages/server/src/repowise/server/app.py` ~215–233:

```python
vector_store, primary_vector_repo_id = await build_primary_vector_store(
    session_factory,
    db_url,
    embedder,
)
app.state.vector_store = vector_store
app.state.primary_vector_repo_id = primary_vector_repo_id
```

`db_url` here is already `resolve_db_url()` output (Postgres when env is set), so the wiki.db substring check cannot succeed.

### Same substring trick elsewhere

In the same lifespan, when a `wiki.db` *does* exist, primary FTS reuse does:

```python
# app.py ~309–312
if db_url and db_url_posix in db_url.replace("\\", "/"):
    app.state.workspace_fts[repo_id] = fts
```

That branch is also dead on Postgres (and members are skipped earlier for missing `wiki.db` anyway). Worth fixing with the same identity helper if touching this area.

### Suggested fix shape

Prefer, in order:

1. Workspace config `default_repo` / `is_primary` → resolve that entry’s absolute path → `crud.get_repository_by_path`.
2. Else `get_configured_db_url()` + explicit primary path from serve/MCP cwd.
3. Else keep `len(rows) == 1` fallback.

Then open LanceDB under that repo’s `local_path` via existing `_build_repo_vector_store` (disk layout is still per-repo even when SQL is shared).

### Related

- Same Postgres / `wiki.db` assumption family as #1034; different module than #1853
- #2138 — workspace serve registration / search fan-out (empty `workspace_fts`); fixing primary id helps seed caches that `resolve_repo_vector_store` expects
- #2139 — `/api/repos` status semantics under Postgres

rishu685:
Closes #2140

### Problem
In `packages/server/src/repowise/server/search_helpers.py`, `build_primary_vector_store` previously identified the primary repository by checking if each repository's SQLite `.repowise/wiki.db` path was a substring of `db_url`.

When `REPOWISE_DB_URL` points to a shared PostgreSQL database (e.g., `postgresql+asyncpg://...`), this substring check always fails. When multiple repositories exist in the database, `build_primary_vector_store` fails to identify any primary repository and falls back to an empty `InMemoryVectorStore`. Consequently, semantic search queries (`/api/search`) search an empty in-memory store and return `[]` instead of using the persisted LanceDB index or falling back to lexical search.

### Solution
1. **Implemented 3-Tier Resolution Ordering**:
   - **Workspace Config**: Looks up the workspace root via `find_workspace_root()` and resolves `ws_config.get_primary()` (`default_repo` / `is_primary`) against registered repository rows.
   - **Process CWD**: Matches the current working directory (`Path.cwd().resolve()`) against registered repository paths.
   - **Single Repository Fallback**: Falls back to `rows[0]` if and only if exactly one repository exists (`len(rows) == 1`).
2. **Removed SQLite-Specific Matching**:
   - Deleted the `local_db in normalized_url` substring test.
   - Updated docstrings in `search_helpers.py` to eliminate SQLite `wiki.db` assumptions.
3. **Prevent Empty Directory Creation (`create=False`)**:
   - Added `create: bool = False` parameter to `build_primary_vector_store` so server startup does not manufacture empty LanceDB directories for unindexed repositories.

### Changes
- `packages/server/src/repowise/server/search_helpers.py`: Updated `build_primary_vector_store` resolution chain, docstring, and `create=False` behavior.
- `tests/unit/server/test_search_persistence.py`: Added regression tests for PostgreSQL URLs with multiple repositories (workspace config & CWD matching), single-repo fallback, and `create=False` behavior.

### Verification
- Ran regression unit tests:
  ```bash
  pytest tests/unit/server/test_search_persistence.py
  # 6 passed in 3.07s

### DIFF
--- a/packages/server/src/repowise/server/search_helpers.py
+++ b/packages/server/src/repowise/server/search_helpers.py
@@ -61,9 +61,10 @@ async def build_primary_vector_store(
 ) -> tuple[Any, str | None]:
     """Build the primary repo's durable vector store when it can be identified.
 
-    Identifies the primary repo by checking workspace configuration, matching
-    against the current working directory, or falling back to the single
-    configured repository. The returned repo id lets startup seed the per-repo
+    A repo-local SQLite database normally contains one repository row. For a
+    shared registry database, only a row whose ``wiki.db`` matches ``db_url`` is
+    treated as primary. Identifies the primary repo by checking ``db_url``,
+    workspace configuration, current working directory, or single-row fallback.
     cache so jobs and searches reuse the same connection.
     """
     from sqlalchemy import select
@@ -81,22 +82,29 @@ async def build_primary_vector_store(
         logger.debug("primary_vector_repo_lookup_failed", exc_info=True)
 
     selected: tuple[str, str] | None = None
+    normalized_url = db_url.replace("\\", "/")
+    for repo_id, local_path in rows:
+        local_db = (Path(local_path).resolve() / ".repowise" / "wiki.db").as_posix()
+        if local_db in normalized_url:
+            selected = (repo_id, local_path)
+            break
 
-    try:
-        from repowise.core.workspace.config import WorkspaceConfig, find_workspace_root
+    if selected is None:
+        try:
+            from repowise.core.workspace.config import WorkspaceConfig, find_workspace_root
 
-        ws_root = find_workspace_root()
-        if ws_root is not None:
-            ws_config = WorkspaceConfig.load(ws_root)
-            primary_entry = ws_config.get_primary()
-            if primary_entry is not None:
-                primary_path = (ws_root / primary_entry.path).resolve()
-                for repo_id, local_path in rows:
-                    if Path(local_path).resolve() == primary_path:
-                        selected = (repo_id, local_path)
-                        break
-    except Exception:
-        logger.debug("workspace_config_primary_lookup_failed", exc_info=True)
+            ws_root = find_workspace_root()
+            if ws_root is not None:
+                ws_config = WorkspaceConfig.load(ws_root)
+                primary_entry = ws_config.get_primary()
+                if primary_entry is not None:
+                    primary_path = (ws_root / primary_entry.path).resolve()
+                    for repo_id, local_path in rows:
+                        if Path(local_path).resolve() == primary_path:
+                            selected = (repo_id, local_path)
+                            break
+        except Exception:
+            logger.debug("workspace_config_primary_lookup_failed", exc_info=True)
 
     if selected is None:
         try:
--- a/packages/server/src/repowise/server/search_helpers.py
+++ b/packages/server/src/repowise/server/search_helpers.py
@@ -53,13 +53,17 @@ def _build_repo_vector_store(repo_path: Path, embedder: Any, *, create: bool) ->
 
 
 async def build_primary_vector_store(
-    session_factory, db_url: str, embedder: Any
+    session_factory,
+    db_url: str,
+    embedder: Any,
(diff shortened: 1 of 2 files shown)


RaghavChamadiya (approved):
Thanks @rishu685, this is exactly what I asked for and the ordering is right. The `db_url` match is back at the top, which matters because it is the only primary signal that does not depend on where the process happens to be running, then workspace config, then cwd, then the single-row fallback. Both new branches are wrapped so a malformed workspace file cannot take startup down with it.

You also answered the `create` question the right way round. `_build_repo_vector_store`'s own docstring says search-only callers should not create a store, and startup is a search-only caller, so `create=False` there stops an unindexed repo getting an empty `.repowise/lancedb` written under it. I checked the write path is untouched: `job_executor.py:378` resolves its own store through `resolve_repo_vector_store(..., create=True)`, so jobs still land in the durable location.

Green on all eight jobs. Merging.

One non-blocking thing to pick up next time you are in this file: the docstring edit left a dangling clause. The line "The returned repo id lets startup seed the per-repo" was replaced, so it now reads "...or single-row fallback." followed by "cache so jobs and searches reuse the same connection." with nothing in front of "cache". Worth one line to repair.

PR #2145, opened 2026-09-05
Benchmark scope selection cost

### LINKED ISSUE #2099: [Bug] update never marks concept pages stale, so model-written prose drifts with no signal
**Summary**

`repowise update` marks file pages stale and never marks a module, layer, cycle or repo-overview page stale, so the model-written pages drift from HEAD with nothing recording that they have. `doctor` reports a clean stale count throughout and `generate --stale` cannot find them, because nothing has set the bit it reads.

**Mechanism**

Two markers exist in `packages/core/src/repowise/core/pipeline/persist.py`, and they differ in exactly one respect: which page ids they can address.

`mark_stale_pages` (`:192`) maps a set of changed paths onto `file_page:<path>` ids:

```python
return await mark_page_ids_stale(session, repo_id, (f"file_page:{path}" for path in paths))
```

`mark_page_ids_stale` (`:221`) takes arbitrary ids, and its docstring states the case for its own existence:

> Scoped generation's cascade needs to mark dependents of every type - module, SCC, layer and the repo-wide overview / architecture / onboarding pages - stale when the run regenerated a file but not its summaries.

Callers:

- `mark_stale_pages` is called from `update_cmd/persistence.py:743` and `update_cmd/deterministic.py:280`, both on the `update` path, both passing `decay_paths`, which are file paths.
- `mark_page_ids_stale` is called from `scoped_generation.py:266` and nowhere else.

So the marker that can express "this module page is now behind its files" is reachable only from `generate`'s own cascade. An `update`-driven workflow, which is what the post-commit hook, the file watcher and the server's polling fallback all drive, never reaches it.

**Why it is invisible**

Every freshness surface reads `freshness_status`, so they all agree with each other and all of them are wrong in the same direction. `doctor`'s stale-page check, `get_stale_pages`, the coverage view and `generate --stale` report zero drift on a repository whose module and overview pages were written once and have not been touched since. There is no signal anywhere until a reader opens a page and finds it describing code that moved. Reported in #2029 against a repository of ~6,500 files whose module pages diverged over weeks of commits.

**Constraint on the fix**

`AUTO_SYNC.md:225` promises that server sync does not regenerate wiki pages and costs no LLM spend. Marking is free and does not touch that promise; enqueueing regeneration would. So the fix is to make freshness truthful and leave the spend decision with the operator, who can then act on it with `generate --stale`.

**The open question, which is why this is an issue and not a PR**

`update` has `decay_paths`, a set of file paths. It does not have the cascade's non-file dependents, which is what `scoped_generation` builds as `plan.stale_ids`. So the work is deciding where an incremental update gets the module, layer and overview pages that depend on the files it touched, and whether that derivation is cheap enough to run on every hook-driven update. The page tree and the related-pages backfill both already know something about page-to-page structure at that point in `update_cmd/persistence.py`, and that is the first place I would look.

Split out of #2029, which asked whether the server's `check_staleness` job should enqueue regeneration. It should not, and correcting that job's docstring is tracked there. This is the gap that remains once it is corrected.

Related: #1744 (stale structural `file_page` rows surviving `update`) and #1845 (prioritising oldest stale pages under budget, which assumes marking already happens).

Aman-goel-04:
## Summary
- `repowise update` marked file pages stale on every commit but never marked module, SCC, or repo-overview pages, so model-written summaries silently drifted from HEAD with no signal anywhere (`doctor`, `get_stale_pages`, and `generate --stale` all reported clean).
- Reuses the existing scoped-generation cascade (`build_dependencies` + `expand_cascade`) to derive those dependents from `decay_paths` during `update`, seeding with `mode="none"` so dependents are marked stale without being regenerated, keeping `AUTO_SYNC.md`'s no-LLM-spend-on-sync promise intact.
- Adds a benchmark script (`scripts/benchmark_scope_selection.py`) used to validate this is cheap enough to run on every hook-driven update (~15ms/1.2k files, ~138ms/11k files for the full selection); the numbers are also recorded as a comment at the call site so a future reader doesn't mistake it for an accidental full re-selection.

## Related Issues
Fixes #2099

## Test Plan
- [x] Tests pass (`pytest`)

Added `test_update_decay_paths_use_cascade_dependents`, which pins the cascade contract directly: seeds are the `file_page:` ids from `decay_paths`, `mode` is `"none"` (not `"dependents"`, so containers are never regenerated), and the resulting stale set includes the module/SCC/repo-overview pages. 

## Checklist

### DIFF
--- a/packages/cli/src/repowise/cli/commands/update_cmd/persistence.py
+++ b/packages/cli/src/repowise/cli/commands/update_cmd/persistence.py
@@ -156,6 +156,33 @@ async def _persist_partial_health(session: Any, repo_id: str, report: Any) -> No
     await persist_partial_health(session, repo_id, report)
 
 
+async def _load_page_rows(session: Any, repo_id: str) -> list[Any]:
+    """Lightweight page rows used to derive stale structural dependents."""
+    from sqlalchemy import select
+    from sqlalchemy.orm import load_only
+
+    from repowise.core.persistence.models import Page
+
+    result = await session.execute(
+        select(Page)
+        .options(
+            load_only(
+                Page.id,
+                Page.page_type,
+                Page.target_path,
+                Page.provider_name,
+                Page.metadata_json,
+                Page.freshness_status,
+            )
+        )
+        .where(
+            Page.repository_id == repo_id,
+            Page.freshness_status != "tombstone",
+        )
+    )
+    return list(result.scalars())
+
+
 async def _persist_incremental_commits(session: Any, repo_id: str, repo_path: Any) -> None:
     """Capture + upsert ``git_commits`` rows for commits new since the last index.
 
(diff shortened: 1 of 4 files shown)


RaghavChamadiya (commented):
LGTM

swati510 (approved):

PR #2150, opened 2026-09-06
fix(vocabulary): skip fenced code blocks when harvesting headings (#2142)

### LINKED ISSUE #2142: [Bug] The heading harvest never skips fenced code blocks, so shell comments are mined as house vocabulary
## Summary

The house-vocabulary harvest finds markdown headings with `_HEADING.finditer(text)` over the raw document (`vocabulary.py:731`) and never tracks fenced code blocks. Every `#` comment inside every fenced shell, Python, Ruby, YAML or Dockerfile block in a repository is therefore a heading candidate, indistinguishable from a real one.

`_NOT_PROSE` (`vocabulary.py:467`) already knows about ` ``` `, but it is consulted only by the *definition* scan, at `:510` and `:536`, which decides what prose follows a heading. Nothing consults it when deciding whether the heading itself is real.

Split out of #2097, where the crash was a code comment made only of hyphens inside a fenced block in @nail's docs. The crash is fixed separately in #2101. This is the gap underneath it, and it does not crash, which is why nobody has noticed it.

## Demonstration

Against current main (6d74db2):

```python
text = '''# Real Heading

Some prose about the thing.

```bash
# Install dependencies
npm install
# Build Pipeline
npm run build
```
'''
for m in _HEADING.finditer(text):
    t = _normalise(m.group(1))
    print(repr(m.group(1)), '->', repr(t), 'useful=', _is_useful(t))
```

```
'Real Heading'         -> 'Real Heading'         useful= True
'Install dependencies' -> 'Install dependencies' useful= True
'Build Pipeline'       -> 'Build Pipeline'       useful= True
```

`_HEADING` is `^\s{0,3}#{1,4}\s+(.+?)\s*#*\s*$` with `re.MULTILINE`, so a shell comment needs only a space after the `#` to qualify, which is the convention everybody writes.

## Why it matters beyond noise

Two effects, and the second is the one I care about.

A phantom term can survive corroboration. Nothing about "Build Pipeline" is malformed; if those words appear in code, it becomes a house term with the same standing as one mined from a real heading, and it then shapes concept-tree naming.

More reliably, phantoms displace real terms. `_harvest` caps the candidate pool and **returns early** when it is full (`vocabulary.py:750`):

```python
if limit is not None and len(order) >= limit:
    return [candidates[k] for k in order]
```

`max_terms` defaults to 200 (`:936`, `:1000`). A README with a long install section can spend that budget on shell comments before the walk ever reaches the documents carrying the repository's actual vocabulary. The loss is silent and there is no signal anywhere that it happened.

## Fix shape

Strip fenced regions before the heading scan, or track fence state while walking lines, so `_HEADING` only ever sees prose. Worth doing once and sharing: `_release_note_ratio` at `:569` runs `_HEADING.findall(text)` over raw text too, so a document whose fenced blocks are full of `# v1.2.0` style comments can be misjudged as release notes and dropped wholesale.

Note that indented code blocks and `~~~` fences exist as well. `_NOT_PROSE` only carries the backtick form today, so whatever this uses should not simply reuse that tuple.

## Tests

`tests/unit/generation/test_vocabulary_house_terms.py`. The cases worth pinning: a `#` comment inside a ` ``` ` block yields no term, a real heading immediately after a fenced block still does, and a heading whose text contains a backtick span is unaffected.

akshatmalik-bruh:
## Summary

- Strips fenced code blocks (both ``` and ~~~) before harvesting headings in `vocabulary.py` so that shell/script comments (`# Install dependencies`) inside code examples are not mined as project terminology.
- Replaces code block content lines with empty strings rather than deleting them, preserving character count and byte offsets so `_definition_after_heading` remains accurate.
- Applies the same stripping in `_is_release_notes()` to prevent docs containing code examples with version comments (e.g. `# v1.0.0`) from being misclassified as release notes.

## Related Issues

Fixes #2142

## Test Plan

- [x] Tests pass (`uv run pytest tests/unit/generation/test_vocabulary_house_terms.py -v`)
- [x] Output fixture test `test_extract_terms_output_is_byte_identical` passes
- [x] Added 5 regression unit tests for #2142 covering backtick fences, tilde fences, real headings after code blocks, inline code spans, and version comments in code blocks.

## Checklist

### DIFF
--- a/packages/core/src/repowise/core/generation/concept_tree/vocabulary.py
+++ b/packages/core/src/repowise/core/generation/concept_tree/vocabulary.py
@@ -68,13 +68,10 @@ _FENCE_OPEN = re.compile(r"^[ ]{0,3}(`{3,}|~{3,})", re.MULTILINE)
 
 
 def _strip_fenced_blocks(text: str) -> str:
-    """Return *text* with every fenced code block's content replaced by blank
-    lines of the same length.
+    """Return *text* with lines inside fenced code blocks replaced with empty strings.
 
-    Lines inside a fenced block are replaced with empty strings so that the
-    total character count and every character offset are preserved.  This keeps
-    ``_definition_after_heading``'s ``m.end()`` byte offsets valid even after
-    stripping.
+    Lines inside a fenced block are replaced with empty strings so that code
+    and shell comments are stripped while keeping line breaks intact.
 
     Both backtick fences (``` … ```) and tilde fences (~~~ … ~~~) are
     supported.  The fence marker on the *opening* line determines which marker
@@ -106,8 +103,7 @@ def _strip_fenced_blocks(text: str) -> str:
                 inside = False
                 result.append(line)  # keep the closing fence line itself
             else:
-                # Replace content with an empty string of the same byte length
-                # so every subsequent offset stays correct.
+                # Replace content line with an empty string to strip code/comments.
                 result.append("")
 
     return "\n".join(result)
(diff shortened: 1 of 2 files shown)


RaghavChamadiya (approved):
That is it, @akshatmalik-bruh. Both the finditer and the slice now run on `prose`, so the offsets agree with each other and the class of bug I flagged is gone rather than moved. Checked it directly rather than taking the diff's word for it:

```
headings on raw text: ['Ledger', 'Blast Radius', 'Install dependencies', 'Cache Layer']
headings on prose:    Ledger, Blast Radius, Cache Layer
'Cache Layer' -> 'A memoized store in front of the provider.'
```

`Cache Layer` is the one that matters. It is the heading *below* the fence, and it was the case whose definition previously landed mid-code-block; it now resolves correctly. Tilde fences and the `_is_release_notes` path are both covered too.

One non-blocking leftover for whenever you are next in this file: `entries += [(m.group(1), None) for m in _BOLD_TERM.finditer(text)]` still reads the raw text, so a `**Bolded Term**` inside a fenced block is mined even though a `#` comment beside it is not. Same bug, smaller surface, and it does not hold this up.

PR #2159, opened 2026-09-08
fix(core): consolidate DOC_EXTENSIONS copies and add drift guard (#1489)

### LINKED ISSUE #1489: Two hand-written copies of DOC_EXTENSIONS remain, with no drift guard
Three places in the codebase decide "is this file documentation". Two agree; the third is missing `.mdx`.

- `packages/core/src/repowise/core/generation/categories.py:23` — `{".md", ".mdx", ".rst", ".txt", ".adoc"}`
- `packages/core/src/repowise/core/ingestion/git_indexer/fix_shape.py:64` — `{".md", ".mdx", ".rst", ".txt", ".adoc"}`
- `packages/core/src/repowise/core/analysis/knowledge_graph.py:165` — `{".md", ".rst", ".txt", ".adoc"}` ← **missing `.mdx`**

The third one feeds `_classify_file_type()`, so in a repo that writes its docs as `.mdx` (Next.js and Docusaurus sites, mostly) those files are typed `"file"` in the knowledge graph instead of `"document"`, while every other part of the system treats them as docs.

## The fix

Add `".mdx"` to `_DOC_EXTENSIONS` in `packages/core/src/repowise/core/analysis/knowledge_graph.py`.

## Please also add a test

A test asserting the three sets agree, in the style of `tests/unit/test_no_test_path_copies.py`. The one-line fix is easy; the test is what stops the next copy drifting.

## Context, and what is not in scope

All three sets are hand-written, while the language registry already declares markdown as `{.md, .mdx, .markdown, .mdown}`. Deriving them from the registry is a larger change and **not** in scope here — please keep this PR to the one extension plus the test.

rishu685:
Closes #1489

### Context
Two hand-written literal copies of the documentation extension set (`{".md", ".mdx", ".rst", ".txt", ".adoc"}`) remained in `categories.py` and `fix_shape.py`, with no architectural test in place to prevent future drift.

### Solution
1. **Consolidated Copies**:
   - Pointed `_DOC_SUFFIXES` in `packages/core/src/repowise/core/generation/categories.py` to `support_paths.DOC_EXTENSIONS`.
   - Pointed `_DOC_EXT` in `packages/core/src/repowise/core/ingestion/git_indexer/fix_shape.py` to `support_paths.DOC_EXTENSIONS`.
2. **Added Drift Guard Test**:
   - Created `tests/unit/test_no_doc_extension_copies.py` (following `test_no_test_path_copies.py`) to inspect the AST across `packages/` and prevent new hand-written doc-extension sets or constants from being introduced outside `repowise.core.support_paths`.
   - Verified that `categories._DOC_SUFFIXES`, `fix_shape._DOC_EXT`, and `knowledge_graph._DOC_EXTENSIONS` all alias the exact same `DOC_EXTENSIONS` object.

### Verification
```bash
pytest tests/unit/test_no_doc_extension_copies.py tests/unit/test_support_paths.py tests/unit/generation/test_categories.py tests/unit/ingestion/test_fix_shape.py
# 58 passed in 6.02s

ruff check packages/core/src/repowise/core/generation/categories.py packages/core/src/repowise/core/ingestion/git_indexer/fix_shape.py tests/unit/test_no_doc_extension_copies.py
# All checks passed!

### DIFF
--- a/packages/core/src/repowise/core/generation/categories.py
+++ b/packages/core/src/repowise/core/generation/categories.py
@@ -13,6 +13,8 @@ from __future__ import annotations
 
 from pathlib import PurePosixPath
 
+from repowise.core.support_paths import DOC_EXTENSIONS
+
 # Categories the file_page prompt knows how to adapt to.
 CATEGORY_CODE = "code"
 CATEGORY_CONFIG = "config"
@@ -20,7 +22,7 @@ CATEGORY_DOC = "doc"
 CATEGORY_DATA = "data"
 CATEGORY_PIPELINE = "pipeline"
 
-_DOC_SUFFIXES = frozenset({".md", ".mdx", ".rst", ".txt", ".adoc"})
+_DOC_SUFFIXES = DOC_EXTENSIONS
 _CONFIG_LANGUAGES = frozenset({"yaml", "toml", "json", "ini", "properties", "hcl"})
 _DATA_DIR_TOKENS = frozenset({"migrations", "versions", "models", "schema", "schemas", "entities"})
 _DATA_SUFFIXES = frozenset({".sql", ".prisma", ".graphql", ".proto"})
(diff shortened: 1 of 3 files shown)


RaghavChamadiya (approved):
Both replaced sets were byte-identical to the canonical one, so this is a pure dedup with no behaviour change, and the guard is the half that makes it stick. test_known_copies_still_exist is the detail I liked: an allowlist that cannot silently rot into a list of files that stopped being copies years ago.

PR #2169, opened 2026-09-08
fix(postgres): widen git_function_blame.symbol_id to Text

### LINKED ISSUE #2162: [Bug] Postgres schema. 512 chars are not sufficient for git_function_blame.symbol_id column
## Describe the Bug
Postgres schema. 512 chars are not sufficient for git_function_blame.symbol_id column

## Steps to Reproduce

1. Setup postgres db and set REPOWISE_DB_URL="postgresql+asyncpg://repowise:repowise@localhost:5432/repowise"
2.  Run `repowise init --no-prose --yes`
3. See error on persistence stage : value too long for type character varying(512) SQL: INSERT INTO git_function_blame

Aman-goel-04:
## Summary

- Widens `git_function_blame.symbol_id` from `VARCHAR(512)` to `Text`, it mirrors `WikiSymbol.symbol_id`, which is already unbounded, but wasn't caught by #1565/`0062` since that migration targeted symbol *name* columns at 255 chars, not a symbol *id* at 512
- Adds migration `0064` (Postgres-only, same shape as `0062`)

## Related Issues

Fixes #2162

## Test Plan

- [x] Tests pass (`pytest`)
- [x] Web build passes (`npm run build`) *(no frontend changes)*

## What was done

- [x] Reproduced the original error locally against Postgres, confirmed the fix resolves it
- [x] `alembic upgrade head` / `downgrade -1` verified round-trip on a stamped pre-fix schema
- [x] Added an integration test (`tests/integration/test_postgres_git_function_blame_symbol_id.py`), confirmed it fails without the fix and passes with it

## Checklist

### DIFF
--- a/packages/core/alembic/versions/0064_widen_git_function_blame_symbol_id.py
+++ b/packages/core/alembic/versions/0064_widen_git_function_blame_symbol_id.py
@@ -1,0 +1,74 @@
+"""Widen git_function_blame.symbol_id to ``Text`` (PostgreSQL only).
+
+``symbol_id`` stores ``"{path}::{name}"`` and mirrors ``WikiSymbol.symbol_id``,
+which is unbounded ``Text`` — but this column was declared ``VARCHAR(512)``.
+A deeply-nested generated file (protobuf/OpenAPI bindings, minified bundles)
+routinely produces a ``path::name`` combination past 512 characters, and
+PostgreSQL enforces the declared length where SQLite ignores it. So the
+narrow type is invisible on the default backend and aborts the run on the one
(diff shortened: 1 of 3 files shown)


RaghavChamadiya (approved):
Guarding the migration on the dialect is the right call, and the reason you gave for it is the one that matters: a batch_alter_table on SQLite would rebuild the table to change nothing, since it never enforced the length in the first place. The downgrade refusing to truncate rather than papering over it with left(col, 512) is also right. Integration tests do not run in CI, so the regression test is documentation of the repro for now, which is still worth having next to the migration.

PR #2170, opened 2026-09-08
fix: raise tree-sitter floor to 0.25, test query compilation

### LINKED ISSUE #2165: [Bug] Raise the tree-sitter floor to 0.25 and add a test that compiles every language query
We declare `tree-sitter>=0.23,<1` in [pyproject.toml#L37](https://github.com/repowise-dev/repowise/blob/a92ea0b59652d8c32fed22d3d8e458238bd652ff/pyproject.toml#L37). That floor is wrong. With `tree-sitter` 0.23.x or 0.24.x installed, 13 of our 27 language queries fail to compile with `Impossible pattern`, and every file in those languages is indexed with zero symbols. Found while investigating #2163, where a VB.NET repo indexed as `Files indexed 10 · Symbols 0`.

Same repowise version, same grammar package versions, only the `tree-sitter` runtime changed:

| tree-sitter | queries that fail to compile |
|---|---|
| 0.23.2 | csharp, dart, fsharp, go, javascript, php, python, razor, rust, scala, shell, swift, vbnet |
| 0.24.0 | same 13 |
| 0.25.0, 0.25.2, 0.26.0 | none |

A fresh `pip install repowise` resolves to the newest `tree-sitter` and is fine, so this only bites when repowise is installed into an environment that already pinned an older one. Nothing in CI catches it, because CI also resolves to latest.

## What needs doing

1. Bump the floor to `tree-sitter>=0.25,<1`.
2. Add a test that compiles the query for every language in the registry and asserts none fail. There is nothing like this today: [`_load_compiled_query`](https://github.com/repowise-dev/repowise/blob/a92ea0b59652d8c32fed22d3d8e458238bd652ff/packages/core/src/repowise/core/ingestion/parser.py#L746) has a single caller, at [parser.py#L1135](https://github.com/repowise-dev/repowise/blob/a92ea0b59652d8c32fed22d3d8e458238bd652ff/packages/core/src/repowise/core/ingestion/parser.py#L1135), and no test exercises it. It swallows the failure and returns `None`, so assert on the return value rather than expecting an exception.
3. Add a CI job that installs `tree-sitter` at exactly the declared floor and runs that one test.

On point 3, please do not reach for a full `--resolution lowest-direct` job. This repo has a lot of dependencies with loose floors and that will fail for reasons unrelated to this bug. Pinning `tree-sitter` alone to the floor version and running the single query-compile test is enough and stays green.

## Reproducing

```bash
uv venv -p 3.12 .venv
VIRTUAL_ENV=.venv uv pip install repowise==0.49.0
VIRTUAL_ENV=.venv uv pip install "tree-sitter==0.24.0"
```

Then `repowise init --yes` against any repo with source in one of the 13 languages reports 0 symbols for them. The per-language detail comes from `_load_compiled_query` returning `None`.

Aman-goel-04:
## Summary

- Raise the `tree-sitter` floor from `>=0.23,<1` to `>=0.25,<1`. 0.23.x/0.24.x fail to compile 13 of 27 language queries (`Impossible pattern`), silently indexing those languages at zero symbols
- Add a test that compiles every language's query against its registered grammar and asserts the result isn't `None`, instead of expecting an exception
- Add a CI job that pins `tree-sitter` to the exact declared floor (0.25.0) and runs that one test, so a future floor regression is caught without a full `--resolution lowest-direct` job

## Related Issues

Fixes #2165

## Test Plan

- [x] Tests pass (`pytest`)
- [x] Web build passes (`npm run build`) *(if frontend changes)*: N/A, no frontend changes

## Checklist

- [x] I have updated documentation if needed: N/A, no docs affected

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -44,6 +44,41 @@ jobs:
         run: uv run pytest tests/unit/ -v --tb=short
 
   # ---------------------------------------------------------------------------
+  # tree-sitter floor — installs the declared minimum tree-sitter version and
+  # compiles every language's query against it. `uv sync` alone always
+  # resolves the newest tree-sitter, which is why #2165 (13 of 27 language
+  # queries failing to compile against tree-sitter 0.23.x/0.24.x, with each
+  # failure silently indexing the language at zero symbols) went uncaught.
+  # Deliberately narrow: only tree-sitter itself is pinned down to the floor,
+  # not `--resolution lowest-direct` for the whole dependency set, which
+  # would fail on unrelated loose floors elsewhere in this repo.
+  # ---------------------------------------------------------------------------
+  test-tree-sitter-floor:
+    name: tree-sitter floor (query compilation)
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Install uv
+        uses: astral-sh/setup-uv@v3
+        with:
+          version: "latest"
+          enable-cache: true
+
+      - name: Install dependencies
+        run: uv sync --all-packages
+
+      - name: Pin tree-sitter to the declared floor
+        run: uv pip install "tree-sitter==0.25.0"
+
+      - name: Run query-compilation test
+        # --no-sync: plain `uv run` re-syncs the venv from uv.lock before
+        # running, which would silently reinstall the newest tree-sitter and
+        # undo the pin above.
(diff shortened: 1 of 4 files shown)


RaghavChamadiya (approved):
The floor bump alone would have been the easy half. Pinning tree-sitter back down to exactly 0.25.0 in its own job, with --no-sync so the pin survives, is what actually stops this regressing, and asserting on the None return rather than an exception matches what _load_compiled_query really does in production.

PR #2173, opened 2026-09-08
fix(ui): raise floating layers above modals so selects open in front

### LINKED ISSUE #2172: [Bug] Wiki style dropdown in the Add Repository dialog renders behind the modal
## Describe the Bug

In the web UI, opening the **Add Repository** dialog and clicking the **Wiki style** dropdown opens the select listbox *behind* the dialog. The listbox is painted under the dialog's backdrop and panel, so it is neither visible nor clickable — there is no way to change the wiki style from this dialog. It always stays on the default, `comprehensive`.

The cause is a z-index token mismatch between two sibling portals:

| Element | Class | Token value |
|---|---|---|
| `DialogOverlay` — `packages/ui/src/ui/dialog.tsx:20` | `z-[var(--z-modal)]` | **40** |
| `DialogContent` — `packages/ui/src/ui/dialog.tsx:41` | `z-[var(--z-modal)]` | **40** |
| `SelectContent` — `packages/ui/src/ui/select.tsx:68` | `z-[var(--z-dropdown)]` | **20** |

Tokens are defined in `packages/ui/styles/globals.css:389-395`:

```css
--z-base: 0;
--z-elevated: 10;
--z-dropdown: 20;
--z-sidebar: 30;
--z-modal: 40;
--z-command: 50;
--z-toast: 60;
```

`SelectContent` wraps its content in `<SelectPrimitive.Portal>` (`select.tsx:64`), and `DialogContent` wraps its own in `<DialogPortal>` (`dialog.tsx:38`). Both portal to `document.body`, so they are siblings in the same stacking context and their raw z-index values compete directly — 20 loses to 40. The overlay's `backdrop-blur-sm` also creates its own containing block, which makes the occlusion total rather than just dimmed.

Nothing at the call site overrides it: `packages/ui/src/onboarding/add-repo-wizard.tsx:343` is a bare `<SelectContent>`.

Focus/pointer handling is fine — Radix `Dialog` and `Select` both default to `modal`, so the listbox does receive focus and arrow keys/typeahead do work. It is purely a paint-order problem, which arguably makes it worse: the value silently changes under an invisible listbox.

## Steps to Reproduce

1. `repowise serve`
2. Open the web UI and click **Add Repository** (sidebar, mobile nav, or the empty-repos state).
3. In the dialog, click the **Wiki style** select (showing "Comprehensive").
4. The listbox opens behind the dialog — nothing appears to happen.

## Expected Behavior

The Wiki style listbox renders above the Add Repository dialog, and the four options (Comprehensive / Caveman / Reference / Tutorial) are visible and clickable.

## Actual Behavior

The listbox renders underneath the dialog overlay and panel. No visible change on click; the option list cannot be seen or clicked.

## Scope

`add-repo-wizard.tsx:343` is currently the only `<SelectContent>` rendered inside a `<Dialog>` — every other usage is on a plain page (`settings/provider-settings.tsx`, `symbols/symbol-table.tsx`, `web/.../settings/*`), which is why this is the only place it shows up today. (The settings-page wiki-style picker in `settings/general-form.tsx` is a custom radio-card list, not a `Select`, so it is unaffected.)

The same latent mismatch applies to `Popover` (`packages/ui/src/ui/popover.tsx:21`) and `Tooltip` (`packages/ui/src/ui/tooltip.tsx:20`), both `z-[var(--z-dropdown)]`, if either is ever rendered inside a `Dialog`.

## Possible Fixes

1. **Point-fix the call site** — pass `className="z-[var(--z-modal)]"` (or higher) to the `<SelectContent>` in `add-repo-wizard.tsx`. Smallest diff, but leaves the trap in place for the next nested select.
2. **Raise the floating-layer tokens above `--z-modal`** so portaled popovers/selects/tooltips always sit above dialogs, since they are transient and always attached to something below them. This is the usual layering convention and fixes Popover/Tooltip at the same time, but it is a design-system change and would want a comment in `globals.css` documenting the intent.

Happy to open a PR for whichever you prefer — I'd lean towards (2) with a comment on the token block, but (1) if you'd rather keep the scale as-is.

## Environment

- OS: Windows 11 Pro (26200)
- Python version: 3.12.0
- Installation method: pipx
- Browser: Chrome
- Also verified present on current `main` (`d019d389`) by reading the source — the three z-index values above are unchanged there.

jgt87:
## Summary

- Move `--z-dropdown` above the surface tokens (20 → 55) so portalled floating layers paint in front of the surfaces they can be anchored inside. Fixes the Wiki style `Select` in the Add Repository dialog opening behind the dialog overlay.
- Document the scale in `globals.css`: which layers must outrank which, and why the portalling makes DOM nesting irrelevant here.
- Add `packages/ui/__tests__/z-layering.test.ts`, an ordering guard for the `--z-*` scale in the style of the existing `token-drift.test.ts`.

## Related Issues

Fixes #2172

## The bug

Radix portals `SelectContent` to `document.body`, so inside a `Dialog` it is a **sibling** of the dialog portal rather than a child of it. The DOM nesting that would normally settle paint order is gone, and the raw z-index values decide:

| Element | Class | Value |
|---|---|---|
| `DialogOverlay` — `packages/ui/src/ui/dialog.tsx:20` | `z-[var(--z-modal)]` | 40 |
| `DialogContent` — `packages/ui/src/ui/dialog.tsx:41` | `z-[var(--z-modal)]` | 40 |
| `SelectContent` — `packages/ui/src/ui/select.tsx:68` | `z-[var(--z-dropdown)]` | **20** |

The overlay's `backdrop-blur-sm` also establishes a containing block, which makes the occlusion total rather than merely dimmed.

It fails quietly in a way worth noting: Radix still gives the hidden listbox focus, so arrow keys and typeahead change the wiki style under a dropdown the user cannot see.

## Why the token change rather than a call-site override

Passing `className="z-[var(--z-modal)]"` to the one `<SelectContent>` in `add-repo-wizard.tsx` is a smaller diff and fixes the reported symptom. I went the other way because `Popover` (`popover.tsx:21`) and `Tooltip` (`tooltip.tsx:20`) carry the same `--z-dropdown` and would fail identically the first time either is rendered inside a `Dialog` — the point-fix leaves that armed.

**55 rather than 45**, because the command palette hardcodes `z-[calc(var(--z-modal) + 1)]` (`packages/web/src/components/search/command-palette.tsx:105`) instead of reading `--z-command`. A dropdown that only cleared `--z-modal` would still open behind the palette.

Happy to swap to the narrow fix if you'd prefer to keep the scale as-is.

## On the test

No render test can catch this: jsdom has no paint and no stacking contexts, and the wizard's existing test renders the `Select` happily today. So the guard asserts the ordering of the scale itself, and additionally pins the components to the tokens it sorts — an ordered scale means nothing once a portalled layer is switched to a bare `z-50`.

Verified it fails before the fix:

```
- []
+ [
+   "--z-dropdown (20) must be above --z-sidebar (30)",
+   "--z-dropdown (20) must be above --z-modal (40)",
+   "--z-dropdown (20) must be above --z-command (50)",
+ ]
AssertionError: expected 20 to be greater than 41
```

## Out of scope, but noticed

`--z-base`, `--z-elevated`, `--z-sidebar` and `--z-command` currently have no call sites — the palette offsets `--z-modal` directly instead of using `--z-command`. Left alone here; happy to fold a cleanup into a separate PR if useful.

## Test Plan

- [x] Web UI type check — `npm run type-check`, exit 0
- [x] Web UI lint — `npm run lint`, exit 0 (one pre-existing warning in `coverage/freshness-table-wrapper.tsx`, untouched here)
- [x] Web build passes — `npm run build`, exit 0 (Next.js build + VS Code extension bundle)
- [x] Lint passes — `uv run ruff check .`, "All checks passed!"
- [x] New test fails before the fix and passes after; `token-drift`, `brand` and `add-repo-wizard` tests pass alongside it
- [ ] `pytest` — not completed locally (still running when this was opened). This change touches no Python; deferring to CI's 3.11/3.12/3.13 jobs.

One note on the `packages/ui` vitest suite: it is flaky on my machine independently of this change — 14 failures on a clean `main` checkout, then 12 and 13 across two runs with this patch applied, with the failing *sets* differing between runs. Total test count moves 1671 → 1676, exactly the five added here. All tests this change could plausibly affect pass in isolation. Flagging in case that is known; I have not investigated it and it is not addressed by this PR.

## Checklist

- [x] All existing tests still pass *(subject to the pre-existing flakiness noted above)*
- [x] I have updated documentation if needed *(the token block is now commented; no other docs reference the z scale)*

### DIFF
--- a/packages/ui/__tests__/z-layering.test.ts
+++ b/packages/ui/__tests__/z-layering.test.ts
@@ -97,4 +97,36 @@ describe("z-index layering", () => {
     // Overlay and content both, or the listbox clears one and not the other.
     expect(dialog.match(/z-\[var\(--z-modal\)\]/g) ?? []).toHaveLength(2);
   });
+
+  /**
+   * The mirror of the test above, and the one the ordering test cannot do on
+   * its own: SURFACES lists the tokens, not the elements wearing them. A panel
+   * that reaches for a FLOATING token inherits its rank, and `--z-dropdown` at
+   * 55 puts that panel above the dialogs and the command palette it should sit
+   * under.
+   *
+   * `graph-canvas-shell.tsx` is why this exists. The rail was pinned to
+   * `--z-dropdown` so it would tie with the tooltips that portal past it —
+   * harmless while `--z-dropdown` was 20 and below `--z-modal`, and an open
+   * rail painting over every dialog under `lg` once it was 55.
+   */
+  it("keeps surfaces off the floating tokens", () => {
+    const surfaces: Array<[string, string]> = [
+      ["graph/graph-canvas-shell.tsx", "--z-sidebar"],
+      ["graph/graph-context-menu.tsx", "--z-modal"],
+    ];
+
+    for (const [file, token] of surfaces) {
+      const text = readFileSync(join(UI_SRC, file), "utf8");
+      expect(text, `${file} should position itself with ${token}`).toContain(
+        `z-[var(${token})]`,
+      );
+      for (const float of FLOATING) {
+        expect(
+          text,
+          `${file} is a surface and must not sit on ${float}, which outranks every surface`,
+        ).not.toContain(`z-[var(${float})]`);
+      }
+    }
+  });
 });
--- a/packages/ui/__tests__/z-layering.test.ts
+++ b/packages/ui/__tests__/z-layering.test.ts
@@ -1,0 +1,100 @@
+import { describe, it, expect } from "vitest";
+import { readFileSync } from "node:fs";
+import { join } from "node:path";
+
+/**
+ * Ordering guard for the `--z-*` scale.
+ *
+ * `token-drift.test.ts` pins that a token *exists*. This pins the one thing
+ * about the z scale that carries meaning: the order.
+ *
+ * The failure it exists to catch is silent. Radix portals Select, Popover and
+ * Tooltip content to `document.body`, so a Select opened inside a Dialog is a
+ * *sibling* of the dialog portal, not a child of it — the DOM nesting that
+ * would normally settle paint order is gone, and the raw z-index values decide.
+ * While `--z-dropdown` sat at 20 and `--z-modal` at 40, the wiki style Select in
+ * the Add Repository dialog opened behind the dialog overlay. Nothing threw,
+ * nothing logged, and the listbox still took focus, so arrow keys changed the
+ * value under a dropdown the user could not see.
+ *
+ * No render test can catch this: jsdom has no paint and no stacking contexts,
+ * and the wizard's own test renders the Select happily today. The assertion has
+ * to be on the scale itself.
+ */
+
+const CSS = join(__dirname, "../styles/globals.css");
+const UI_SRC = join(__dirname, "../src");
+
+/** Surfaces own a region of the page. */
+const SURFACES = ["--z-base", "--z-elevated", "--z-sidebar", "--z-modal", "--z-command"];
+/** Floating layers are transient and anchored to a control on a surface. */
+const FLOATING = ["--z-dropdown"];
+
+function zTokens(): Map<string, number> {
+  const css = readFileSync(CSS, "utf8");
+  const out = new Map<string, number>();
+  for (const m of css.matchAll(/(--z-[a-z0-9-]+)\s*:\s*(\d+)\s*;/g)) {
+    out.set(m[1]!, Number(m[2]!));
+  }
+  return out;
+}
+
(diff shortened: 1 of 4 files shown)


RaghavChamadiya (approved):
Thanks @jgt87. The revised stacking-context change keeps the graph rail and dropdowns in the shared sidebar layer, updates the rationale, and adds the regression guard requested in the earlier review. The change now matches the modal layering contract. Approved.

PR #2177, opened 2026-09-09
fix(ingestion): warn when a language query fails to compile (#2166)

### LINKED ISSUE #2166: [Bug] Tell the user when a language query fails to compile, instead of reporting zero symbols
Split out of #2163. Separate from the version floor itself (#2165), this failure mode is invisible, and that is what cost the reporter the time.

When a language's query does not compile, [parser.py#L784](https://github.com/repowise-dev/repowise/blob/a92ea0b59652d8c32fed22d3d8e458238bd652ff/packages/core/src/repowise/core/ingestion/parser.py#L784) logs at warning and returns `None`. The CLI discards that log. The index then completes successfully, reports `Symbols 0`, and the summary still says the language is 95% of the repo, because that line is a file census and not a count of what was parsed:

```
Files indexed   10
Symbols         0
Languages       vbnet 95%, yaml 2%, markdown 2%, json 0%
```

There is already the right channel for this. [`_report_missing_grammars`](https://github.com/repowise-dev/repowise/blob/a92ea0b59652d8c32fed22d3d8e458238bd652ff/packages/core/src/repowise/core/pipeline/phases/ingestion.py#L27) runs a preflight scoped to the languages traversal actually found and calls `emit_warning` with a clear message when a grammar is not installed. It is called once, at [ingestion.py#L470](https://github.com/repowise-dev/repowise/blob/a92ea0b59652d8c32fed22d3d8e458238bd652ff/packages/core/src/repowise/core/pipeline/phases/ingestion.py#L470). A query that will not compile deserves the same treatment. The docstring on [`missing_grammar_languages`](https://github.com/repowise-dev/repowise/blob/a92ea0b59652d8c32fed22d3d8e458238bd652ff/packages/core/src/repowise/core/ingestion/parser.py#L852) already names this gap as a known ceiling.

**One design point worth knowing before you start.** `missing_grammar_languages` deliberately uses `importlib.util.find_spec` rather than importing, to stay cheap enough to run on every index without pulling every grammar into the parent process. Compiling a query needs the grammar loaded, which is exactly the cost that function avoids. Two workable ways around it:

- Only check the languages the traversal actually found. That is a bounded set, and the parse pool is about to load those grammars anyway.
- Or do the check inside the worker, where the grammar is already loaded, and report it back.

Either is fine. Please say which you picked in the PR.

## Scope

- Warn, do not hard-fail. A repo with one incidental Go file should still index if the Go query is broken.
- Second half, optional and fine as a follow-up: make the `Languages` line stop claiming a language it extracted nothing from. The string is built at [reporting.py#L180](https://github.com/repowise-dev/repowise/blob/a92ea0b59652d8c32fed22d3d8e458238bd652ff/packages/cli/src/repowise/cli/commands/init_cmd/reporting.py#L180) and used at [#L193](https://github.com/repowise-dev/repowise/blob/a92ea0b59652d8c32fed22d3d8e458238bd652ff/packages/cli/src/repowise/cli/commands/init_cmd/reporting.py#L193). Something like `vbnet 95% (not parsed)` is honest and cheap.

To reproduce, install `tree-sitter==0.24.0` alongside repowise 0.49.0 and index a Python or Go repo.

rishu685:
Fixes #2166

### Summary of Changes
- **Parser (`packages/core/src/repowise/core/ingestion/parser.py`)**:
  - Extracted `_compile_query(lang, grammar_tag)` with `@cache` to compile queries and capture error messages on failure.
  - Implemented `failed_query_languages(language_tags)` bounded to the languages discovered during traversal.
- **Ingestion Pipeline (`packages/core/src/repowise/core/pipeline/phases/ingestion.py`)**:
  - Added `_report_query_compilation_failures(stats, progress)` alongside `_report_missing_grammars` to emit actionable warnings via `emit_warning` when query compilation fails.
  - Ensures other unaffected languages continue indexing normally without interruption.
- **Tests (`tests/unit/ingestion/test_query_preflight.py`)**:
  - Added unit test suite covering compilation failures, progress warning emissions, query cache handling, and resilience.

### Verification
- `pytest tests/unit/ingestion/test_query_preflight.py` (7 passed)
- `pytest tests/unit/ingestion/test_grammar_preflight.py` (8 passed)
- `pytest tests/unit/pipeline/` (160 passed)

### DIFF
--- a/packages/core/src/repowise/core/ingestion/parser.py
+++ b/packages/core/src/repowise/core/ingestion/parser.py
@@ -779,6 +779,7 @@ def _compile_query(lang: str, grammar_tag: str | None = None) -> tuple[object | 
         return None, str(exc)
 
 
+@cache
 def _load_compiled_query(lang: str, grammar_tag: str | None = None) -> object | None:
     """Process-wide cache of compiled tree-sitter Query objects.
 
--- a/packages/core/src/repowise/core/ingestion/parser.py
+++ b/packages/core/src/repowise/core/ingestion/parser.py
@@ -743,20 +743,15 @@ def _cpp_export_macro_parent(node: Node, parent_names: dict[int, str]) -> str | 
 
 
 @cache
-def _load_compiled_query(lang: str, grammar_tag: str | None = None) -> object | None:
-    """Process-wide cache of compiled tree-sitter Query objects.
+def _compile_query(lang: str, grammar_tag: str | None = None) -> tuple[object | None, str | None]:
+    """Compile and return (Query, None) or (None, error_str).
 
-    Compiling `.scm` queries is non-trivial; in process-pool parsing each worker
-    would otherwise recompile per file. ``grammar_tag`` may differ from
-    ``lang`` when a language reuses another's grammar at a different
-    variant — e.g. ``.tsx`` files reuse ``typescript.scm`` but must bind
-    to the JSX-aware ``tsx`` grammar so React components don't drown in
-    ERROR nodes.
+    Cached process-wide so preflight and parsing workers never recompile the same query.
     """
     grammar = grammar_tag or lang
     language = _get_language(grammar)
     if language is None:
-        return None
+        return None, None
 
     # The spec names the query file, so a language can reuse another's
     # queries wholesale (svelte -> typescript.scm). Every other spec declares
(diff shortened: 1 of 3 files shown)


RaghavChamadiya (approved):
Thanks @rishu685. Restoring the cache at the load boundary keeps a broken query warning to once per process while preserving the compiled-query cache, and the new tests cover warning deduplication and real progress behavior. The prior blocker is resolved. Approved.

PR #2187, opened 2026-09-10
fix(server): detect repo index presence from database in list_repos (#2139)

### LINKED ISSUE #2139: [Bug] GET /api/repos cannot detect registered-but-never-indexed repos under PostgreSQL (wiki.db gate)
## Describe the Bug

`GET /api/repos` still decides “registered but never indexed” by checking for `<local_path>/.repowise/wiki.db`. That signal is meaningless under `REPOWISE_DB_URL`: Postgres indexes never create that file, and in workspace mode a later pass often **overwrites** the flag anyway.

So on shared Postgres you lose the honest `needs_index` distinction the sidebar contract was written for — either false `needs_index` (when the overwrite does not run) or false `indexed` (registration-only rows stamped indexed just because a `Repository` row exists).

This is the `/api/repos` sibling of #1034. #1853 does not touch `repos.py`.

## Steps to Reproduce

**A — registration without index (workspace + Postgres)**

1. Workspace with `REPOWISE_DB_URL` → PostgreSQL.
2. Ensure a member has a `repositories` row (e.g. registration / failed first index) but **no** `graph_nodes` / wiki pages.
3. Confirm no `<repo>/.repowise/wiki.db`.
4. `GET /api/repos`: that member is still stamped `workspace_status="indexed"` by the workspace attach pass (see below), so the sidebar cannot show “needs index”.

**B — wiki.db flag sticks (no workspace overwrite)**

1. Same Postgres DB, serve **without** workspace config loaded (`workspace_config is None`).
2. Registered rows with no `wiki.db` get `workspace_status="needs_index"` from the filesystem check and return immediately — even when the shared DB holds a full index.

## Expected Behavior

On a configured shared database, “indexed vs needs_index” should follow index presence in that DB (e.g. file nodes / completed initial job), not the presence of a repo-local SQLite file. Registration alone must not imply indexed.

## Actual Behavior

The code assumes initial-index always creates `wiki.db`, then (in workspace mode) mostly ignores that check.

## Environment

- Repowise version: 0.48.0
- Installation method: Docker (`REPOWISE_DB_URL=postgresql+asyncpg://…`)

### Call sites (v0.48.0) — `packages/server/src/repowise/server/routers/repos.py` `list_repos`

**1. Filesystem “never indexed” probe** (~210–223):

```python
# Flag registered-but-never-indexed repos. … the initial-index path always
# establishes <repo>/.repowise/wiki.db, its absence means the first index
# hasn't run.
for resp in responses:
    if resp.workspace_status is None and resp.local_path:
        if not (_Path(resp.local_path) / ".repowise" / "wiki.db").is_file():
            resp.workspace_status = "needs_index"
```

**2. Workspace attach overwrites to indexed** (~241–250), for any response whose `local_path` resolves to a configured member:

```python
resp.workspace_alias = entry.alias
resp.is_primary = bool(entry.is_primary)
resp.workspace_status = "indexed"  # unconditional
indexed_aliases.add(entry.alias)
```

**3. Synthetic stubs** (~273–304) only for aliases **not** in `indexed_aliases` (no matching `Repository` row in the merged list). Those correctly get `needs_index` / `missing_dir` — but anything that *is* registered is forced to `indexed` in step 2, so Postgres cannot express “registered, not yet indexed” via the wiki.db absence the comment describes.

**4. Early return** (~230–231): if `workspace_config is None`, responses return after step 1 — so the wiki.db → `needs_index` assignment is what the client sees.

### Interaction with empty `workspace_sessions`

`list_repos` also merges rows from `app.state.workspace_sessions` (~177–192). On Postgres that map is empty for the same reason as the serve-registration bug (startup skips members without `wiki.db`). Listing still often looks fine because the primary session *is* the shared DB and already contains every `Repository` row — which is why #1034’s screenshots could show `/api/repos` populated while `/api/workspace` said `needs_index`. The remaining bug here is the **status semantics**, not “repos missing from the list”.

### Suggested fix shape

- If `get_configured_db_url()` is set, replace the wiki.db probe with a shared-DB signal (e.g. `graph_nodes` with `node_type='file'` for that `repository_id`, or a completed `initial_index` / sync job).
- In the workspace attach pass, do not unconditionally stamp `indexed`; reuse that signal.
- Keep synthesizing stubs only for members with no `Repository` row.

### Related

- #1034 / #1853 — `/api/workspace` listing (different endpoint; status already being fixed there)
- #2138 — serve `workspace_sessions` / search fan-out (empty secondary map); complementary, not a duplicate of this status bug
- #2140 — primary vector store selection under Postgres

rishu685:
Fixes #2139

### Summary of Changes
- **Router (`packages/server/src/repowise/server/routers/repos.py`)**:
  - Replaced the local filesystem probe `(Path(local_path) / ".repowise" / "wiki.db").is_file()` with an honest database query checking for file-typed `GraphNode` rows across active primary and secondary workspace sessions.
  - Correctly marks registered repositories with `0` file nodes as `needs_index` (or `missing_dir` if directory does not exist on disk) under both SQLite and PostgreSQL.
  - Removed the unconditional `resp.workspace_status = "indexed"` assignment in the workspace attach loop, allowing registered-but-unindexed workspace members to honestly retain `needs_index`.
- **Tests (`tests/unit/server/test_repos_status.py` & `test_workspace_sidebar.py`)**:
  - Added unit test suite covering registered-but-unindexed repos, PostgreSQL/shared DB environments without local `wiki.db`, workspace status persistence, and missing directories.

### Verification
- `pytest tests/unit/server/test_repos_status.py` (5 passed)
- `pytest tests/unit/server/test_workspace_sidebar.py` (6 passed)
- `pytest tests/unit/server/test_repos.py` (18 passed)
- `pytest tests/unit/server/test_workspace_router.py` (63 passed)
- `ruff check packages/server/ tests/unit/server/` (0 errors)

### DIFF
--- a/packages/cli/src/repowise/cli/commands/augment_cmd/search.py
+++ b/packages/cli/src/repowise/cli/commands/augment_cmd/search.py
@@ -412,7 +412,7 @@ def _wiki_db_exists(repo_path: Path) -> bool:
     return (repo_path / ".repowise" / "wiki.db").exists()
 
 
-def _fast_pagerank_file_order(repo_path: Path, paths: list[str]) -> list[str] | None | object:
+def _fast_pagerank_file_order(repo_path: Path, paths: list[str]) -> list[str] | object | None:
     """``_pagerank_file_order`` over stdlib sqlite3. ``_ORM`` to fall back."""
     if not _wiki_db_exists(repo_path):
         return None
@@ -438,7 +438,7 @@ def _fast_search_enrich(
     mode: str,
     result_count: int,
     matched: dict[str, int] | None,
-) -> str | None | object:
+) -> str | object | None:
     """Triage and the widened rescue without the ORM. ``_ORM`` to fall back.
 
     The zero-result rescue is not served here on purpose: 45% of its queries
--- a/packages/server/src/repowise/server/routers/repos.py
+++ b/packages/server/src/repowise/server/routers/repos.py
@@ -250,9 +250,12 @@ async def list_repos(
     # Reuses the workspace "needs_index" / "missing_dir" contract the sidebar renders.
     for resp in responses:
         if resp.workspace_status is None and resp.id not in indexed_repo_ids:
-            if resp.local_path and not Path(resp.local_path).is_dir():
-                resp.workspace_status = "missing_dir"
-            else:
+            try:
+                if resp.local_path and not Path(resp.local_path).is_dir():
+                    resp.workspace_status = "missing_dir"
+                else:
+                    resp.workspace_status = "needs_index"
+            except OSError:
                 resp.workspace_status = "needs_index"
 
     # Augment with workspace metadata. We do this in a second pass (rather
--- a/packages/server/src/repowise/server/routers/repos.py
+++ b/packages/server/src/repowise/server/routers/repos.py
@@ -170,8 +170,23 @@ async def list_repos(
     what powers the web UI sidebar — silently dropping unindexed repos
     used to cause the "I only see the primary" Discord report.
     """
-    result = await session.execute(select(Repository).order_by(Repository.updated_at.desc()))
-    repos = list(result.scalars().all())
+    has_files_subq = (
+        select(GraphNode.repository_id)
+        .where(GraphNode.node_type == "file")
+        .group_by(GraphNode.repository_id)
+        .subquery()
+    )
+    result = await session.execute(
+        select(Repository, has_files_subq.c.repository_id.is_not(None))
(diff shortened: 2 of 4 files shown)


RaghavChamadiya (approved):
Reading the index state out of the database instead of off the filesystem is the right move, and you took the half the report did not ask for: line 249 used to stamp "indexed" on any row whose path matched a workspace entry, so a member registered with index: false read as indexed on plain SQLite too. Gating that on indexed_repo_ids fixes both backends with one predicate. The subquery is index-only against ix_graph_nodes_repo_type, which covers both the filter and the grouping key, so the sidebar does not pay a table scan for this.

PR #2195, opened 2026-09-11
fix(server): register workspace members and fan-out search on shared database (#2138)

### LINKED ISSUE #2138: [Bug] Workspace serve only registers sessions/FTS when wiki.db exists (Postgres search fan-out skips members)
## Describe the Bug

In workspace mode under `REPOWISE_DB_URL` (shared PostgreSQL), per-repo indexes live in the configured DB and no `<repo>/.repowise/wiki.db` is created. Startup and workspace search still key off that file:

1. **Serve lifespan** only opens engines / populates `workspace_sessions` / `workspace_fts` when `wiki.db` exists.
2. **Search fan-out** then skips any member without `wiki.db` and reads `repo_id` via `sqlite3`, so semantic/FTS search never visits non-primary members even when they are fully indexed in Postgres.

Same root assumption as #1034 / #1853, but different call sites (those only touch `workspace.py`).

## Steps to Reproduce

1. Create a workspace with 2+ repos and set `REPOWISE_DB_URL` to PostgreSQL.
2. Index every member (confirm `repositories` + `graph_nodes` rows in Postgres; confirm no per-repo `.repowise/wiki.db`).
3. Start `repowise serve` against the workspace root.
4. Observe startup: no `workspace_repo_dbs_loaded` (or count 0); `app.state.workspace_fts` empty for member ids.
5. Run a workspace search (API or UI) whose hits only exist in a non-primary member → empty / primary-only results.

## Expected Behavior

- Shared-DB deployments register each workspace member against the configured session factory / FTS (keyed by `repository_id` / `local_path`) without requiring `wiki.db`.
- Search fan-out visits every indexed member via that registration (or an equivalent shared-DB lookup).

## Actual Behavior

Members without `wiki.db` are skipped at startup and again in search fan-out.

## Environment

- Repowise version: 0.48.0
- Installation method: Docker (`REPOWISE_DB_URL=postgresql+asyncpg://…`, workspace mode)

### Call sites (v0.48.0)

**1. Lifespan registration** — `packages/server/src/repowise/server/app.py`

```python
# ~289–317
for repo_entry in ws_config.repos:
    repo_path = (_Path(ws_root) / repo_entry.path).resolve()
    repo_db = repo_path / ".repowise" / "wiki.db"
    if not repo_db.exists():
        continue
    # … sqlite3.connect(repo_db) → SELECT id FROM repositories …
    db_url_posix = repo_db.as_posix()
    if db_url and db_url_posix in db_url.replace("\\", "/"):
        app.state.workspace_fts[repo_id] = fts
        continue
    repo_engine = create_engine(f"sqlite+aiosqlite:///{db_url_posix}")
    # … workspace_sessions[repo_id] = repo_sf …
    # … workspace_fts[repo_id] = FullTextSearch(repo_engine) …
```

On Postgres the `exists()` continue fires for every member, so both maps stay empty. The primary-FTS branch (`db_url_posix in db_url`) also never hits for `postgresql+asyncpg://…` (same substring trick as `build_primary_vector_store`).

**2. Search fan-out** — `packages/server/src/repowise/server/routers/search.py`

```python
# ~152–176, comment: "Fan-out: iterate over every workspace repo with an indexed wiki.db"
for entry in ws_config.repos:
    repo_path = (ws_root_path / entry.path).resolve()
    db_path = repo_path / ".repowise" / "wiki.db"
    if not db_path.exists():
        continue
    with _sql.connect(str(db_path)) as conn:
        row = conn.execute("SELECT id FROM repositories LIMIT 1").fetchone()
    rid = row[0] if row else None
    # … resolve_repo_vector_store(…, rid) …
    # … workspace_fts[rid].search(…)  # only if rid was registered at startup
```

Even if fan-out stopped requiring `wiki.db`, FTS fallback still needs `rid in workspace_fts`, which startup never populated — hence filing these together.

### Why routes look “half fine”

`resolve_session_factory` (`deps.py` ~42–55) falls back to `app.state.session_factory` when `workspace_sessions` has no entry. On shared Postgres that primary factory *is* the shared DB, so some `repo_id`-scoped ORM reads still work. Search fan-out does **not** use that fallback for FTS membership — it requires the maps filled above.

### Existing helpers to copy

- `get_configured_db_url()` / `resolve_db_url()` — `packages/core/.../persistence/database.py`
- `crud.get_repository_by_path(session, local_path)` — `packages/core/.../persistence/crud/repository.py` (exact `local_path` match; callers should pass `str(path.resolve())`)
- Pattern in #1853 for workspace listing: resolve identity from configured session by path, then filter aggregates by `repository_id`

### Suggested fix shape

When `get_configured_db_url()` is set:

1. For each workspace member, `get_repository_by_path` on the primary session.
2. Register `workspace_sessions[repo_id] = app.state.session_factory` (shared) and `workspace_fts[repo_id] = app.state.fts` (shared FTS already built on that engine).
3. In search fan-out, resolve `rid` the same way (or iterate `workspace_fts` keys) instead of `wiki.db` + `sqlite3`.

### Out of scope but related

`RepoRegistry` / `open_repo_db` (`packages/core/.../workspace/registry.py`) still opens per-repo SQLite for MCP/chat contexts (see `app.py` comment citing #970). Separate from `workspace_sessions`, but the same Postgres deploy will hit it for chat tools.

### Related

- #1034 / #1853 — workspace listing/sync router only
- #2140 — primary vector store `wiki.db`-in-URL matching (`search_helpers.build_primary_vector_store`, also called from this lifespan at `app.py` ~221)
- #2139 — `/api/repos` indexed vs `needs_index` semantics

rishu685:
### Summary

Fixes #2138.

In workspace mode under a configured database URL (e.g. `REPOWISE_DB_URL` pointing to shared PostgreSQL), per-repo indexes and metadata live directly inside the shared database, so individual `<repo>/.repowise/wiki.db` files are not created. 

Previously:
1. **Server lifespan startup (`app.py`)** skipped every workspace member repo because it checked `if not repo_db.exists(): continue`, leaving `app.state.workspace_sessions` and `app.state.workspace_fts` empty.
2. **Search fan-out (`search.py`)** skipped member repos without `wiki.db` and used `sqlite3.connect` to resolve repository IDs, preventing semantic search and FTS fan-out from discovering or querying non-primary workspace members.

### Changes Made

1. **Workspace Lifespan Registration (`packages/server/src/repowise/server/app.py`)**:
   - Detects when `get_configured_db_url()` is set.
   - Resolves member repositories in the shared database via `get_repository_by_path(session, str(repo_path))`.
   - Registers `app.state.workspace_sessions[repo.id]` to the shared session factory and `app.state.workspace_fts[repo.id]` to the shared FTS instance.
   - Populates `app.state.workspace_path_to_repo_id` for fast repository ID resolution.

2. **Database Helpers (`packages/server/src/repowise/server/repo_db.py`)**:
   - Updated `is_primary_db_for` to return `True` when `get_configured_db_url()` is set, ensuring configured databases are treated as canonical for all repos and avoiding local SQLite file expectations.

3. **Workspace Search Fan-Out (`packages/server/src/repowise/server/routers/search.py`)**:
   - Resolves workspace member repository IDs via `workspace_path_to_repo_id` without requiring `wiki.db`.
   - Queries member LanceDB vector stores when available and falls back to `workspace_fts.get(rid)`.
   - Deduplicates shared `FullTextSearch` execution across workspace members to prevent redundant queries on shared databases.
   - Deduplicates search results across workspace members by `page_id`.

4. **Unit Tests (`tests/unit/server/test_workspace_search_postgres.py`)**:
   - Added tests verifying workspace lifespan member registration, full-text search fan-out, and semantic search FTS fallback in shared database mode.

### Verification

- Ran unit tests:
  ```bash
  pytest tests/unit/server/test_workspace_search_postgres.py tests/unit/server/test_search.py tests/unit/server/test_workspace_sidebar.py tests/unit/server/test_repos.py


RaghavChamadiya (approved):
Thanks @rishu685. Repository scoping now reaches both SQLite and Postgres full-text search, including the widened retry path, and the two-repository fixtures prove scoped A/B results alongside unscoped behavior. This resolves the cross-repository leakage from the earlier review. Approved.

PR #2196, opened 2026-09-11
fix(workspace): stop dropping the repo you pointed at when it is a submodule

oldschoola:
## The bug

Running `repowise init` inside a git submodule that also contains ordinary nested clones builds a workspace out of the clones and silently drops the repo you were standing in. Reported from `seedpatch/silo`:

```
  #       Repository                      Path                                  Status
  [1]     Exorth-rsps                     server_bases/Exorth-rsps              new
  [2]     osrs-world-map-object-dumper    tools/osrs-world-map-object-dumper    new
```

`silo` is absent, unselectable, and nothing says why.

## Why

`silo/.git` is a file (`gitdir: ../.git/modules/silo`), so `_is_submodule(root)` is true. Its two nested repos are plain clones — `.git` directories, not submodules — so they are discovered regardless of the flag, which makes `found` non-empty and tips the run into workspace mode. `scanner.py` then reached:

```python
if not root_is_sub or include_submodules:
    found.insert(0, (root, root_is_sub))
```

`include_submodules` decides whether the walk descends into submodules it *discovers*. The root is not a discovery — it is the path the caller named — so gating it there removed the one repo the user explicitly asked for.

Confirmed directly against the reporter's checkout, old scanner vs new:

```
BEFORE (main): 2 repo(s) -> ['Exorth-rsps', 'osrs-world-map-object-dumper']   silo present: False
AFTER  (fix): 3 repo(s) -> ['.', 'Exorth-rsps', 'osrs-world-map-object-dumper']  silo present: True
```

## Scope: two branches, one rule

The single-repo branch carried the same confusion pointing the other way:

```python
if root_is_sub and not include_submodules:
    return ScanResult(repos=[], root=root)
```

A submodule with no nested repos scanned to **zero** repos rather than the one it was handed.

To be straight about this half: **no current caller is visibly broken by it.** `init` falls through to its single-repo path when the scan is empty and works fine, and `workspace scan` filters the root against `existing_paths` anyway. It is changed because a function asked "which git repos are here" should not answer "none" about a directory that is one, and because leaving it would apply `include_submodules` to the root in one branch and not the other — the exact confusion that produced this bug, preserved half-fixed for the next reader. If a reviewer would rather keep this PR to the one-line workspace fix, that half can be dropped without affecting the report.

Discovery itself is untouched: a submodule found *under* a plain root still requires `--include-submodules`.

## Workaround on affected versions

`--include-submodules` does make the root selectable, so it unblocks anyone on a released build. It is not equivalent to this fix — it also un-filters genuine submodules during discovery. On the reporter's checkout it yields four repos rather than three:

```
OLD code, --include-submodules=False -> 2 repo(s), silo present: False
OLD code, --include-submodules=True  -> 4 repo(s), silo present: True
```

## Tests

`test_submodule_at_root_excluded` asserted `len(result.repos) == 0`. It pinned exactly this bug, so it is replaced rather than re-pinned:

- `test_root_submodule_is_scannable_on_its_own` — a bare submodule root returns itself, flagged `is_submodule`.
- `test_root_submodule_leads_its_own_workspace` — the reporter's topology: a submodule holding two plain clones lists the root first, then the clones. This is the regression test for the report.
- `test_discovered_submodule_still_respects_the_flag` — guards the other side, so the root rule cannot leak into discovery.

`tests/unit/workspace` is green (1136 passed). Across `tests/unit/{cli,workspace,server}`: 6706 passed, 3 failed — `test_mascot.py::test_short_tagline_below_the_banner_width` and two in `test_update_provider_prompt.py`, all of which reproduce on unmodified `main` (this branch is `main` plus the scanner change only). `ruff check`, `ruff format --check` and `mypy` are clean on both touched files.

### DIFF
--- a/packages/core/src/repowise/core/workspace/scanner.py
+++ b/packages/core/src/repowise/core/workspace/scanner.py
@@ -88,7 +88,10 @@ def scan_for_repos(
     - If *root* itself is a git repo, returns a single-element result.
     - Stops descending once a ``.git`` boundary is found (no nested repos).
     - Skips common junk directories (``node_modules``, ``.venv``, etc.).
-    - Submodules (``.git`` is a file) are excluded unless *include_submodules* is True.
+    - Submodules (``.git`` is a file) found *during the walk* are excluded
+      unless *include_submodules* is True. The flag never applies to *root*
+      itself: that is the repo the caller explicitly pointed at, not something
+      discovery turned up, so it is always eligible.
     """
     root = Path(root).resolve()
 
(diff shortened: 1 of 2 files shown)


RaghavChamadiya (approved):
The distinction this turns on is the right one and the code now states it: include_submodules governs what the walk *discovers*, and the root is not a discovery, it is the path the caller named. I checked the two branches against their callers. init gates workspace mode on len(scan.repos) > 1 (command.py:933, :968), so the reporter's topology went from two repos to three with silo leading, and the bare-submodule case goes from zero to one without changing anything downstream, since resolve_repo_path already returns a resolved path and the len == 1 redirect at :1007 is a no-op. Taking the second half too: you are right that leaving the flag applied to the root in one branch and not the other preserves exactly the confusion that caused this.

PR #2197, opened 2026-09-11
fix(workspace): give the index phase the provider it was launched with

oldschoola:
## The bug

Running `repowise init --provider omp --model omp/ccs/claude-opus-5` across a workspace reports:

```
Not run: pull requests (No LLM provider is configured), git history (No LLM
provider is configured), comments (No LLM provider is configured)
```

The run was given a provider. It generated pages with it. The decision sources still reported that none was configured.

Not provider-specific — any `--provider` hits this in workspace mode. Single-repo init is unaffected.

## Why

The index phase mines decisions with a model. `pr`, `git_history` and `comments` each declare a model stage, and `analysis.py` decides whether those stages can run from the client it was handed:

```python
runtime = {rt.key: rt for rt in policy.runtime(provider_available=llm_client is not None)}
```

Single-repo init passes one (`init_cmd/command.py`):

```python
llm_client = None if dry_run else (provider if not index_only else decision_provider)
...
run_pipeline(..., generate_docs=False, llm_client=llm_client, ...)
```

The workspace path did not (`init_cmd/workspace.py`):

```python
run_pipeline(
    repo.path,
    ...
    generate_docs=False,
    mode=(...),
)
```

`llm_client` defaults to `None`, so `provider_available` was `False` for every repo in the workspace. `ctx.provider` was sitting right there — the very next block used it for generation. Nothing failed; the model-backed sources silently did not run, and the report honestly said why.

## The fix

Bind the provider to the repo once, before the pipeline call, and pass it:

```python
repo_provider = (
    None if ctx.dry_run else _workspace_generation_provider_for_repo(ctx.provider, repo.path)
)
```

Bound rather than reused verbatim because a provider that shells out with a working directory (`REPO_PATH_PROVIDERS`) has to point at the repo being indexed, not at whichever one resolved it first. Generation now reuses that same instance instead of re-deriving it a few lines later.

A dry run still passes `None`, matching single-repo init — decision extraction and KG enrichment both reach for the client before the dry-run return.

**Not addressed here:** single-repo init resolves a separate `decision_provider` so `--index-only` still mines decisions with a model. The workspace path has no equivalent, so an index-only workspace stays modelless. That is a distinct gap from the one reported and is left alone rather than widened into this PR.

## Tests

`tests/unit/cli/test_workspace_init_provider.py`, reusing the stub harness already established by `test_workspace_init_dry_run.py`. `run_mode="fast"` isolates the index phase — generation is skipped, so whatever reaches `run_pipeline` got there on the index phase's account.

- `test_the_index_phase_is_given_the_resolved_provider` — the reported bug.
- `test_a_repo_path_provider_is_rebound_to_the_repo_being_indexed` — a `codex_cli`-shaped provider arrives rebound.
- `test_a_dry_run_keeps_the_index_phase_modelless` — guards the single-repo parity so the dry-run contract is not lost to a later refactor.

Each was confirmed non-vacuous: deleting the `llm_client=repo_provider` line fails all three with `KeyError: 'llm_client'`, and they pass with it restored.

`test_workspace_init_provider.py` + `test_workspace_init_dry_run.py` + `test_editor_setup.py`: 59 passed. `ruff check` and `mypy` clean. The wider `tests/unit/cli` sweep was not run to completion on this branch.

`ruff format --check` flags `workspace.py`, but it does so on unmodified `main` too — the only reformat it wants is a pre-existing mis-indented `console.print` block that this change does not touch, so it is left alone rather than folded in as unrelated churn.

### DIFF
--- a/packages/cli/src/repowise/cli/commands/init_cmd/workspace.py
+++ b/packages/cli/src/repowise/cli/commands/init_cmd/workspace.py
@@ -314,6 +314,20 @@ def _ingest_and_generate_repo(repo: Any, idx: int, total: int, ctx: _WorkspaceCt
     )
     ensure_repowise_dir(repo.path)
 
+    # The index phase mines decisions with a model too -- pull requests, git
+    # history and code comments all have a model stage -- so it needs the
+    # provider as much as generation does. Leaving it out made every one of
+    # those sources report "No LLM provider is configured" on a workspace run
+    # that had been given a perfectly good one, while the single-repo path
+    # (which passes its client) mined them normally.
+    #
+    # Bound to this repo rather than reused verbatim: a provider that shells
+    # out with a working directory has to point at the repo being indexed, not
+    # at whichever one resolved the provider first.
+    repo_provider = (
+        None if ctx.dry_run else _workspace_generation_provider_for_repo(ctx.provider, repo.path)
+    )
+
     try:
         with Progress(
             SpinnerColumn(spinner_name=OWL_SPINNER, style=BRAND_STYLE),
@@ -338,6 +352,7 @@ def _ingest_and_generate_repo(repo: Any, idx: int, total: int, ctx: _WorkspaceCt
                     exclude_patterns=ctx.exclude_patterns if ctx.exclude_patterns else None,
                     include_submodules=ctx.include_submodules,
                     generate_docs=False,
+                    llm_client=repo_provider,
                     mode=(
                         OrchestratorMode.FAST
                         if ctx.run_mode == "fast"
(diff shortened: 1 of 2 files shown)


RaghavChamadiya (approved):
This is a real parity gap and the diagnosis is right: single-repo init passes llm_client into the index-phase run_pipeline with generate_docs=False (command.py:1431), _run_decision_extraction consumes it (orchestrator.py:522-524), and the workspace path resolved a provider for generation only. So decision mining was modelless on exactly the runs that had been given a provider. Binding it per repo rather than reusing the first-resolved instance is the right call for the same reason generation already does it, and hoisting that one call so both phases share it keeps the two from drifting apart. Keeping a dry run modelless matches single-repo and is the conservative side to err on.

PR #2199, opened 2026-09-11
docs(quickstart): --no-editor-setup covers the repo-root .mcp.json too

oldschoola:
The Claude Code section said `repowise init` writes a repo-root `.mcp.json` "unconditionally", and that `--no-editor-setup` / `REPOWISE_SKIP_EDITOR_SETUP=1` gates only the `~/.claude/settings.json` registration.

That describes the behaviour before the flag was widened. `write_editor_project_files` now covers both halves — its own docstring records why ("a user who read the flag as 'leave my repo alone' still got `.mcp.json`, `.claude/CLAUDE.md` and two `.vscode` files written into their working tree") — with `.repowise/mcp.json` as the one deliberate exception, written either way because it is what `repowise mcp .` prints and is how you opt back in by hand.

So a reader following this page expected a file that opting out no longer writes. Found while wiring a non-Claude MCP client to a repo whose `.mcp.json` was missing.

Text only; one hunk.

### DIFF
--- a/docs/start/QUICKSTART.md
+++ b/docs/start/QUICKSTART.md
@@ -86,12 +86,16 @@ This is the payoff: your agent reads the index instead of your codebase.
 
 <details open><summary><b>Claude Code</b></summary>
 
-**`repowise init` already did this.** It writes a repo-root `.mcp.json`
-unconditionally, and unless you passed `--no-editor-setup` or set
-`REPOWISE_SKIP_EDITOR_SETUP=1`, it also registers repowise with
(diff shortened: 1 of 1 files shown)


RaghavChamadiya (approved):
Correct, and I checked it against the code rather than the changelog: write_editor_project_files calls save_mcp_config(repo_path) before the gate and then returns [] under is_editor_setup_disabled (editor_setup.py:316-323), so .repowise/mcp.json really is written either way and everything else really is covered by the one flag. The docs were describing the behaviour the flag used to have. Docs-only, and Python 3.12 is already green on the matrix.

PR #2204, opened 2026-09-11
fix(cli): warn Windows users before upgrading

### LINKED ISSUE #2149: [Bug] - upgrading Repowise required manually killing the `repowise` task
## Describe the Bug

---
Every time I upgrade Repowise I get this error message and must kill the "Repowise" process manually (on Windows OS)

I also have to close the IDE before upgrading Repowise because I want to be sure it will use the latest Repowise version after upgrade.

## Steps to Reproduce

1. Run `uv tool upgrade repowise` 
2. Get error at the end: "the process cannot access the file because it is being used by another process"
3. Kill "Repowise process" in *Windows* task manager
4. Re-run succeed

## Expected Behavior

Command to work at fist-shot

## Environment

- OS: Windows 11
- Python version: 3.12.1
- Repowise version: Any

roian6:
## Summary

- Warn Windows users to stop running Repowise server or MCP processes before an upgrade command is shown.
- Apply the guidance to both `doctor` and the shared update/what's-new advisory while preserving non-Windows output.
- Clarify that `Repowise: Stop Server` only stops a server started by the VS Code extension; otherwise the MCP client must be closed or a manually started server stopped in its terminal.

## Related Issues

Fixes #2149

## Test Plan

- [x] Focused update-advisory suite: 45 passed in an isolated Python 3.11 dependency slice
- [x] Windows and non-Windows guidance regressions cover both advisory surfaces, including the reported `uv tool upgrade repowise` path
- [x] `ruff check` passes on all five changed files
- [x] `ruff format --check` passes on all five changed files
- [ ] Full project `pytest`: dependency installation is blocked on Linux/aarch64 because `tree-sitter-gdscript==6.1.0` cannot find `tree_sitter/parser.h`; the same failure reproduces on unmodified `main`

## Checklist

- [x] All focused existing tests pass
- [x] Documentation is not separately needed; this change corrects the user-facing guidance itself

### DIFF
--- a/packages/cli/src/repowise/cli/commands/doctor_cmd/advisories.py
+++ b/packages/cli/src/repowise/cli/commands/doctor_cmd/advisories.py
@@ -70,7 +70,7 @@ def _print_cli_version_status() -> None:
     Runs once per invocation (the CLI version is global, not per-repo).
     """
     try:
-        from repowise.cli.update_check import get_cli_update_check
+        from repowise.cli.update_check import get_cli_update_check, windows_upgrade_caveat
 
         check = get_cli_update_check()
     except Exception:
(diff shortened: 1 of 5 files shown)


RaghavChamadiya (approved):
Thanks @roian6. The shared Windows upgrade caveat is emitted before the suggested update command in both Doctor and What's New, without changing the command resolution contract. The focused tests cover both surfaces. Approved.

PR #2224, opened 2026-09-11
fix(ui): clamp generation progress percentages

### LINKED ISSUE #2176: [Bug] Generation progress percentage is not clamped, producing 4820% and a malformed CSS transform
## Describe the Bug

`GenerationProgress` divides `completed_pages` by `total_pages` without clamping the result, so an out-of-range job record renders a percentage above 100. Mine showed:

```
241 / 5 pages          4820%
```

The label is the visible part, but the unclamped value also breaks the progress bar's rendering and its accessibility.

### The missing guard

Two sibling components compute the same percentage. Only one clamps it.

`packages/ui/src/dashboard/active-job-banner.tsx:38-40` — correct:

```tsx
const progress =
  job.total_pages > 0
    ? Math.min(100, Math.round((job.completed_pages / job.total_pages) * 100))
    : 0;
```

`packages/ui/src/jobs/generation-progress.tsx:69-73` — unclamped:

```tsx
const progress = job
  ? job.total_pages > 0
    ? Math.round((job.completed_pages / job.total_pages) * 100)
    : 0
  : 0;
```

That value is used both as the label (`generation-progress.tsx:152`) and as the bar's value (`:138`).

### What the out-of-range value does to `Progress`

`packages/ui/src/ui/progress.tsx:27` positions the indicator by string interpolation:

```tsx
style={{ transform: `translateX(-${100 - (value ?? 0)}%)` }}
```

With `value = 4820`, `100 - 4820` is `-4720`, so the template produces `translateX(--4720%)` — a malformed CSS value. The browser drops the declaration, the indicator keeps its default `w-full`, and **the bar renders completely full** regardless of actual progress.

Radix separately rejects the out-of-range value. Rendering `<Progress value={4820} />` against `value={48}` for comparison:

```
value=4820  aria-valuenow=null  data-state=indeterminate  style=transform: translateX(--4720%);
value=48    aria-valuenow=48    data-state=loading        style=transform: translateX(-52%);
```

and logs, once per render:

```
Invalid prop `value` of value `4820` supplied to `Progress`. The `value` prop must be:
  - a positive number
  - less than the value passed to `max` (or 100 if no `max` prop is set)
  - `null` or `undefined` if the progress is indeterminate.

Defaulting to `null`.
```

So a single missing clamp produces four distinct problems: a nonsensical label, a bar pinned full, `aria-valuenow` dropped so assistive technology gets no progress at all, and console error spam for the duration of the job.

## Steps to Reproduce

1. `repowise serve`, trigger a full resync, and let it reach a phase whose total is unknown (this is the job-record bug filed alongside this one).
2. Once `completed_pages` exceeds `total_pages`, watch the generation progress panel.
3. The percentage exceeds 100, the bar is full, and the browser console fills with the Radix warning above.

## Expected Behavior

The displayed percentage is clamped to 0-100, and the bar receives a value within range even when the job record is inconsistent.

## Actual Behavior

`4820%`, a permanently full bar, no `aria-valuenow`, and a console error per render.

## Possible Fixes

1. **Clamp at the call site** — mirror `active-job-banner.tsx:40` in `generation-progress.tsx:71`. Smallest change, restores parity between the two components.
2. **Clamp inside the primitive** — have `Progress` clamp `value` into `[0, max]` before both the Radix prop and the transform. Defends every current and future caller, and removes the malformed-CSS failure mode entirely. My preference, ideally alongside (1).

The string interpolation at `progress.tsx:27` is worth hardening either way: any negative input silently yields `--N%` rather than an error, so the failure is invisible until someone inspects the element.

## Related

The out-of-range job record that triggers this is filed separately. These are independent defects — the clamp is worth having regardless of whether the record is fixed, since it is already present in the sibling component.

## Environment

- OS: Windows 11 Pro (26200)
- Installation method: pipx
- Browser: Chrome

akkupratap323:
## What changed

`GenerationProgress` could display percentages above 100% when a job reported more completed pages than its stale total. For `241 / 5`, the UI rendered `4820%`, passed an invalid value to Radix, dropped `aria-valuenow`, and generated malformed transform CSS.

This change clamps the displayed generation percentage to `0-100`. It also hardens the shared `Progress` primitive so all finite inputs are bounded before reaching Radix or the indicator transform; non-finite inputs retain the indeterminate state.

## Validation

- Focused generation and progress tests: 21 passed
- Shared UI TypeScript check: passed
- Shared-package import boundary check: passed
- Full shared UI suite: 1,658 passed; two unrelated system-map suites could not load a user-level PostCSS plugin, and two existing heavy UI tests timed out on this Windows machine
- Regression tests were run before implementation: 9 failed on overflow, negative, and non-finite inputs

Closes #2176

### DIFF
--- a/packages/ui/__tests__/jobs/generation-progress.test.tsx
+++ b/packages/ui/__tests__/jobs/generation-progress.test.tsx
@@ -3,6 +3,39 @@ import { render, screen, fireEvent } from "@testing-library/react";
 import { GenerationProgress } from "../../src/jobs/generation-progress.js";
 
 describe("GenerationProgress", () => {
+  it.each([
+    { completed: 241, total: 5, percentage: 100 },
+    { completed: -1, total: 5, percentage: 0 },
+    { completed: 3, total: 10, percentage: 30 },
+    { completed: 5, total: 5, percentage: 100 },
+    { completed: 241, total: 0, percentage: 0 },
+  ])("shows $percentage% for $completed / $total pages", ({ completed, total, percentage }) => {
+    render(
+      <GenerationProgress
+        job={{
+          id: "j1",
+          status: "running",
+          total_pages: total,
+          completed_pages: completed,
+        }}
+        log={[]}
+        elapsed={1000}
+        actualCost={null}
+        stuckPending={false}
+        cancelling={false}
+        onCancel={vi.fn()}
+      />,
+    );
+
+    expect(screen.getByText(`${percentage}%`)).toBeInTheDocument();
+    expect(screen.getByText(`${completed} / ${total} pages`)).toBeInTheDocument();
+    const bar = screen.getByRole("progressbar");
+    expect(bar).toHaveAttribute("aria-valuenow", String(percentage));
+    expect(bar.firstElementChild).toHaveStyle({
+      transform: `translateX(${percentage - 100}%)`,
+    });
+  });
+
   it("renders the queued state for pending jobs", () => {
     render(
       <GenerationProgress
--- a/packages/ui/__tests__/ui/progress.test.tsx
+++ b/packages/ui/__tests__/ui/progress.test.tsx
@@ -1,0 +1,36 @@
+import { afterEach, describe, expect, it, vi } from "vitest";
+import { render, screen } from "@testing-library/react";
+import { Progress } from "../../src/ui/progress.js";
+
+afterEach(() => vi.restoreAllMocks());
+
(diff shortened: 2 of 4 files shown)


RaghavChamadiya (approved):
Thanks @akkupratap323. Clamping at both the generation boundary and the shared progress component covers raw percentages as well as finite derived values, while preserving indeterminate handling for missing and non-finite input. The regression coverage matches those cases. Approved.

PR #2226, opened 2026-09-11
fix(cli): honor editor_files opt-outs during init and update

### LINKED ISSUE #2137: update/init rewrite .vscode/mcp.json + .vscode/extensions.json unconditionally; editor_files: {vscode_mcp: false} is never read
### Summary

Setting the documented project-file opt-out for VS Code (`editor_files: {vscode_mcp: false}` in `.repowise/config.yaml`) has no effect: every `repowise update` — even `--index-only` — and every `init` rewrites `.vscode/mcp.json` and `.vscode/extensions.json` into the repo.

### Environment

- repowise **0.44.0**, Windows 11 (the write itself is platform-independent)
- Repo config carries: `editor_files: {claude_md: false, agents_md: false, vscode_mcp: false, cursor_rules: false}`

### What I found reading the source

- `editor_files.claude_md` / `editor_files.agents_md` **are** consulted by the claude/codex integrations (`cli/editor_integrations/claude.py`, `codex.py` read the config via the `editor_files` mapping).
- The VS Code integration **never reads config**: `VSCodeSetup.write_project_files` → `_write_vscode_files` (`cli/editor_integrations/vscode.py`) writes unconditionally; the only gate is `options.disabled_project_files`.
- `update_cmd/command.py::_refresh_editor_stamp` builds `EditorSetupOptions` exclusively from the `--agents-md` CLI flag — it never loads `editor_files` from config, so even the generic `disabled_project_files` gate cannot see config opt-outs on the update path.
- `PROJECT_FILE_ID = "vscode_mcp"` is defined (`cli/agent_targets/targets/vscode.py:46`), so the config key exists in the schema — but no code path consults it (grepped the whole package: the only `vscode_mcp` occurrences are the ID constant and the save functions).

### Repro

1. `repowise init` any repo (VS Code files get written).
2. Add `editor_files: {vscode_mcp: false}` to `.repowise/config.yaml`.
3. Delete `.vscode/`.
4. Run `repowise update --index-only`.

Output: `✓ VS Code MCP configured (.vscode/mcp.json)` and `✓ VS Code extension recommended (...)` — both files are recreated, contradicting the config.

### Expected

The `editor_files` opt-outs honored on **all** write paths, matching the claude_md/agents_md behavior — ideally by loading the config into `EditorSetupOptions.disabled_project_files` inside `_refresh_editor_stamp` (and the init path), rather than relying on CLI flags alone.

### Workaround we currently ship

Repowise leaves invalid-JSON (JSONC) files untouched (`left unchanged (not valid JSON...)`), so an inert comment-only stub at `.vscode/mcp.json` + `.vscode/extensions.json` blocks the rewrite. That is what we now ship across ~20 repos — it works, but it is a hack: the files look like broken JSON to every editor, and it relies on an error-path escape hatch rather than a supported opt-out.

### Secondary note

`repowise agents remove --target=all --scope=both` unwires registrations but does not delete the project files the integrations created (`.vscode/`, `.mcp.json`, `.repowise/mcp.json`) — we had to remove them manually afterwards. With the config opt-out landing this matters less, but having `remove` also delete (or offer to delete, e.g. `--purge-project-files`) the files it originally created would round the lifecycle out.

Thanks!

z2Ace0107:
## Summary

Fix `editor_files.<project_file_id>: false` being ignored during editor project-file setup.

- Resolve persisted opt-outs once in the setup layer using registered agent target descriptors.
- Honor `vscode_mcp: false` during both `init` and `update`.
- Keep explicit per-run project-file overrides stronger than persisted opt-outs.
- Keep Codex MCP and hooks registration independent from the `agents_md` opt-out.

The separate project-file purge behavior mentioned in the issue remains out of scope.

## Validation

- `uv run pytest tests/unit/cli/test_editor_setup.py tests/unit/cli/test_vscode_setup.py tests/unit/cli/test_agent_target_baseline.py -q` (`78 passed`)
- `uv run ruff check packages/cli/src/repowise/cli/editor_setup.py packages/cli/src/repowise/cli/editor_integrations/codex.py tests/unit/cli/test_editor_setup.py` (passed)
- `uv run ruff format --check packages/cli/src/repowise/cli/editor_setup.py packages/cli/src/repowise/cli/editor_integrations/codex.py tests/unit/cli/test_editor_setup.py` (passed)
- `uv run mypy packages/cli/src/repowise/cli/editor_setup.py packages/cli/src/repowise/cli/editor_integrations/codex.py` (passed)
- `npm run lint` (passed; existing warning in an untouched UI file)
- `npm run type-check` (passed)

Fixes #2137

### DIFF
--- a/packages/cli/src/repowise/cli/editor_integrations/codex.py
+++ b/packages/cli/src/repowise/cli/editor_integrations/codex.py
@@ -37,15 +37,16 @@ class CodexSetup:
         setup_override = options.integration_overrides.get(self.integration_id)
         agents_override = options.project_file_overrides.get(self.project_file_id)
         agents_override_present = self.project_file_id in options.project_file_overrides
+        agents_disabled = self.project_file_id in options.disabled_project_files
         # Codex is opt-in, and stays that way. It is the one integration that
         # already got this right — nothing is written unless --codex or the
         # checklist asked for it — which is why #1499 names the other two.
         if setup_override is None or setup_override is False:
-            if agents_override_present:
-                written = maybe_generate_agents_md(
+            if agents_override_present and not agents_disabled:
+                agents_path = maybe_generate_agents_md(
                     console_obj, repo_path, agents_md=agents_override
                 )
-                return [written] if written is not None else []
+                return [agents_path] if agents_path is not None else []
             return []
 
         installed = is_codex_cli_installed()
@@ -56,13 +57,14 @@ class CodexSetup:
         hooks_path = save_codex_hooks_config(repo_path)
         console_obj.print(f"  [{OK}]✓[/] Codex hooks registered ({hooks_path})")
         written = [Path(config_path), Path(hooks_path)]
-        agents_path = maybe_generate_agents_md(
-            console_obj,
-            repo_path,
-            agents_md=True if agents_override is None else agents_override,
-        )
-        if agents_path is not None:
-            written.append(agents_path)
+        if not agents_disabled:
+            agents_path = maybe_generate_agents_md(
+                console_obj,
+                repo_path,
+                agents_md=True if agents_override is None else agents_override,
+            )
+            if agents_path is not None:
+                written.append(agents_path)
 
         if not installed:
             console_obj.print(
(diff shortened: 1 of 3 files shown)


RaghavChamadiya (approved):
Thanks @z2Ace0107. Resolving explicit editor-file opt-outs into `disabled_project_files` preserves default behavior and explicit overrides while keeping the Codex MCP and hook toggles separate from AGENTS.md generation. The regression tests cover those contracts. Approved.

PR #2233, opened 2026-09-12
fix(health): count Pascal for-in loops, add a performance dialect

Laex:
## Summary

- Pascal's `for x in collection do` parses as its own `foreach` grammar node, distinct from the counted `for` -- and it was missing from `loop_kinds`, so the complexity walker skipped it entirely. Any Pascal function using a for-in loop understated both CCN and nesting depth. Confirmed against a real ~150-file Delphi codebase: 70 occurrences in one module alone.
- Registers a Pascal `PerfDialect` (`io_in_loop` / `hot_path_sync_io`), closing the other visible gap: the performance pillar was completely silent for Pascal (0.8% LOC coverage on a real Delphi repo's health report, 100% after this change). Scoped to filesystem and subprocess sinks matched by distinctive RTL/VCL/FPC call names (`CopyFile`, `TFileStream.SaveToFile`, `ShellExecute`, `CreateProcess`, ...). DB and network sinks are deliberately left out: Pascal's `uses` clause has no import-classification table behind it yet, so there is no evidence to disambiguate `Open`/`Get`/`Post` from an unrelated method of the same name -- precision over recall.
- Bumps `HEALTH_ANALYZER_VERSION` to 12 (documented stamp convention) so an existing index re-scores these findings on its next update.
- Updates `docs/CHANGELOG.md` (+ the bundled mirror) and `docs/layers/LANGUAGE_SUPPORT.md`.

## Test plan

- [x] New regression test for the `foreach` nesting/CCN undercount (`test_pascal_foreach_counts_as_a_loop`)
- [x] New Pascal cases in the perf pass corpus (`test_perf_io_in_loop.py`): bare filesystem call in a loop, attribute-call stream I/O in a loop, process spawn in a loop, and a negative case (`Copy` the built-in substring function vs. `TFile.Copy`)
- [x] `test_model_version_coupling.py` pin updated to the new stamp
- [x] Full `tests/unit/health/` suite passes (1627 passed, 18 skipped)
- [x] Validated live against a real ~150-file Delphi codebase: perf coverage 0.8% -> 100%, and genuine `io_in_loop` hits surfaced (e.g. a retry-delete loop, recursive tree-copy/delete/list operations) with no false positives on the `Copy`-substring-builtin case


RaghavChamadiya:
Thanks @Laex for the measurement behind this one. "70 occurrences in one module alone" is the kind of number that turns a plausible gap into a confirmed one, and scoping the new `PerfDialect` to filesystem and subprocess sinks while deliberately leaving DB and network out is the right call. Pascal's `uses` clause has no import-classification table behind it, so there is genuinely no evidence to tell `Open`/`Get`/`Post` apart from an unrelated method of the same name. Saying that in the commit message rather than quietly guessing is what makes this reviewable.

One blocker, and it is about authorship, not about the work. Commit `e25a144e` ("chore: retrigger CI") carries:

```
Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>
```

We do not merge commits with a coding agent as author or co-author. Every other commit on the branch is authored and committed by you with no trailer, so the ask is small: that commit is an empty CI retrigger, so the cleanest fix is to drop it entirely with an interactive rebase, or amend it and delete the trailer line, then `git push --force-with-lease`. Nothing else about the branch needs to change.

Worth noting so it does not bite you on the rebase: you have already carried the stamp forward to `HEALTH_ANALYZER_VERSION = 22` with `_STAMPS = (22, 2, 2)`, and main is currently on 21, so that part is correct as it stands right now. If anything else lands on main first you will need to bump past it again.

Ping me once the trailer is gone and CI is green and I will take another look.

Laex:
Done — rebuilt the branch as a single commit rebased directly onto current upstream/main (no merge commits, no empty CI-retrigger commit), so the trailer is gone from the whole branch, not just amended away. Force-pushed with `--force-with-lease`.

Stamp is confirmed correct against current main: `HEALTH_ANALYZER_VERSION = 22`, `_STAMPS = (22, 2, 2)`. CI is running on the new push.

RaghavChamadiya (approved):
Thanks @Laex for rebuilding the branch cleanly and for keeping up with the stamp while main kept moving. It lands at HEALTH_ANALYZER_VERSION 31 against main's 27, so no collision. Gating every new branch behind an empty-by-default field or an explicit pascal check means no other language changes behaviour, and classifying each unit in a uses clause separately, without seeding the bare System prefix, avoids the obvious over-match.

PR #2237, opened 2026-09-13
Fix API client handling of 204 No Content responses

Selene29:
## Summary

- Treat successful HTTP 204 responses as empty results instead of parsing them as JSON.
- Add regression tests for saving and deleting provider API keys.

## Problem

The provider key endpoints return `204 No Content` after completing their mutation. The shared API client attempted to parse every successful response with `response.json()`.

Saving or deleting a provider key therefore succeeded on the server, but the client threw `Unexpected end of JSON input`. The UI displayed a failure toast and skipped its normal refresh.

## Testing

- `npm test` in `packages/api-client`: 65 tests passed
- `npm run type-check` in `packages/api-client`


RaghavChamadiya (approved):
Thanks @Selene29. Handling `204 No Content` before JSON parsing fixes the API-client contract without weakening handling for other successful responses, and the provider-key deletion tests exercise the real callers. Approved.

PR #2238, opened 2026-09-13
Match Kimi K2 and K3 reasoning contracts

Selene29:
## Summary

- Restrict K2 explicit reasoning control to disabling thinking.
- Expose K3 `low`, `high`, and `max` effort choices and send them through the top-level `reasoning_effort` field.
- Leave reasoning fields unset for `auto` so Kimi keeps its provider default.

## Problem

K2 advertised graded effort choices even though every enabled choice became the same thinking toggle. K3 did not expose or forward its documented reasoning-effort levels.

This made the catalog promise controls that K2 could not distinguish while omitting controls K3 supports.

## Testing

- 22 Kimi provider tests passed
- Whitespace checks passed


RaghavChamadiya (approved):
Thanks @Selene29. The K2 and K3 paths now match their distinct contracts: K2 uses the explicit thinking switch, while K3 sends only its supported top-level effort values and omits automatic mode. The targeted request-shape tests cover the boundary. Approved.

PR #2239, opened 2026-09-13
Fix DeepSeek thinking controls and effort mapping

Selene29:
## Summary

- Send `reasoning_effort` at the documented top level and map portable effort choices correctly.
- Restrict advertised reasoning controls to supported DeepSeek model IDs.
- Disable thinking for streaming chat because Repowise does not retain `reasoning_content` for tool-call follow-ups.

## Problem

DeepSeek generation nested `reasoning_effort` inside the thinking payload and mapped low effort to high. Model discovery also granted reasoning support to unknown V4-looking names.

Streaming chat could enable thinking even though the shared conversation history cannot replay the required reasoning content on later tool calls.

## Testing

- 22 DeepSeek provider tests passed
- Whitespace checks passed


RaghavChamadiya (approved):
Thanks @Selene29. The explicit reasoning-model allowlist, supported effort mapping, and `thinking` request shape match the DeepSeek contract. Disabling thinking for the current streaming path also avoids losing required `reasoning_content` across tool-call turns. Approved.

PR #2240, opened 2026-09-13
Derive LiteLLM reasoning capabilities from local metadata

Selene29:
## Summary

- Build LiteLLM model options from the loaded local model metadata.
- Advertise exact reasoning efforts from explicit levels and capability flags.
- Use the same metadata interpretation for discovery and runtime validation.

## Problem

LiteLLM discovery offered a fixed `low`, `medium`, and `high` set for reasoning models, while execution used a separate capability lookup. The two paths could disagree about which effort values a model accepts.

Catalog construction also consulted per-model helpers and included non-text model categories. This change keeps enumeration local, filters those categories, and treats explicit negative or always-on metadata as authoritative.

## Testing

- 6 LiteLLM provider tests passed
- Whitespace checks passed


Selene29:
Confirmed. I reproduced the split on the locked LiteLLM 1.90.2 catalog: both `vercel_ai_gateway/deepseek/deepseek-r1` and `vercel_ai_gateway/xai/grok-3-mini` exposed only `auto` in discovery while runtime accepted `low`, `medium`, and `high`.

Fixed in `01f5ce36` after rebasing the PR onto current `main`.

- Discovery and runtime now share the same capability resolution.
- Decisive direct metadata still wins.
- Provider-prefixed rows use LiteLLM's case-insensitive bare-model fallback.
- Models missing from `model_cost` use `get_model_info()` and `supports_reasoning()` with the provider resolved from the model ID.
- The loaded catalog is indexed once, so discovery remains local and bounded. The installed 2,830-model catalog completed in 22 ms in the live reproduction.

Added regressions for both a missing catalog row and an indecisive provider-prefixed row. Validation passed: 8 LiteLLM provider tests, Ruff lint, Ruff format, and `git diff --check`. The two reported models now expose the same `auto/low/medium/high` modes in discovery and execution.

RaghavChamadiya (approved):
Thanks @Selene29 for reproducing both named cases on the locked catalog before changing anything. Routing discovery and runtime through one shared function closes the split I was worried about, and the regression tests fail loudly if either path escapes to get_model_info, which is the right way to pin it.

PR #2241, opened 2026-09-13
Fix MCP tool toggles for workspace repositories

Selene29:
## Summary

- Route MCP tool updates to the workspace database selected by the request body's repo ID.
- Cover the HTTP route with separate primary and workspace databases.

## Problem

`PATCH /api/mcp/tools` carries `repo_id` in its JSON body. The handler used the request URL resolver, which saw no path or query repo ID and fell back to the primary database.

Valid non-primary workspace repositories therefore returned `404 Repo not found`, even though the same IDs worked through `/api/repos/{repo_id}`. The GET tool endpoint masked the lookup failure by showing the default tool list, so the switches appeared usable but could not save.

## Testing

- 8 MCP and workspace-routing tests passed
- Whitespace checks passed
- Confirmed MCP tool toggles save in the workspace dashboard


RaghavChamadiya (approved):
Thanks @Selene29. Resolving the session factory from the repository ID fixes workspace-backed PATCH requests where no query parameter is available, and the separate-database test proves the write reaches the selected repository. Approved.

PR #2242, opened 2026-09-13
Fix repository paths for server-side CLI providers

Selene29:
## Summary

- Build server-selected providers through the shared provider argument mapping.
- Pass the target repository path to Codex CLI and OpenCode.
- Add regression coverage for both repository-aware providers.

## Problem

Server jobs correctly resolved the repository being indexed, but `get_chat_provider_instance()` dropped that path while constructing the provider.

Codex CLI then used the server process directory. In workspace mode this was the workspace root rather than a Git repository, so decision extraction failed with `Not inside a trusted directory`. OpenCode had the same missing path even though both providers are registered as repository-aware.

## Testing

- 62 provider and job-executor tests passed
- Confirmed the generated Codex command uses the CalculationsArena repository path
- Whitespace checks passed


RaghavChamadiya (approved):
Thanks @Selene29. Passing the resolved repository path through the shared provider configuration path restores repository-local CLI settings while preserving the existing repository config and environment overrides. The focused tests cover that precedence. Approved.

PR #2245, opened 2026-09-14
feat: add object pascal assertion and performance markers

krishna231122-del:
## Summary

- Implements Object Pascal assertion and performance markers, fulfilling the planned roadmap item.
- Adds `PascalPerfDialect` with custom callee extraction logic to handle bare parenless statements and assignments unique to the Pascal AST.
- Registers the new dialect in the performance metrics engine and adds comprehensive unit test coverage.

## Test Plan

- [x] Tests pass (`pytest`) - Added dialect-specific callee extraction tests for Pascal AST nodes.
- [ ] Web build passes (`npm run build`) *(if frontend changes)* - N/A

## Checklist


sloemo01:
Two things are red here and both are quick to clear.

**The three test failures are your own new tests**, and the cause is subtle enough to be worth writing down. All three come back `None`:

```
FAILED test_pascal_dialect.py::test_callee_root_name_assignment - assert None == 'GetProfile'
FAILED test_pascal_dialect.py::test_callee_root_name_exprCall   - assert None == 'MyCall'
FAILED test_pascal_dialect.py::test_callee_root_name_statement  - assert None == 'Assert'
```

Your override unwraps the shape and then hands the unwrapped node to `super().callee_root_name(...)`, but the base method expects the *call*, not the callee. It opens with:

```python
fn = call_node.child_by_field_name("function")
if fn is None:
    named = [c for c in call_node.children if c.is_named]
    fn = named[0] if named else None
```

For `assignment` you pass the `rhs`, which is a bare identifier. It has no `function` field, and a leaf identifier has no named children, so `fn` stays `None` and the method returns before the identifier walk ever runs. The `statement` branch has the same problem when `named[0]` is the identifier itself.

So the unwrapping is right and the delegation is the part that does not fit. Either pull the root name out of the shape you just unwrapped, or keep the base method and give it something call-shaped. The loop inside it, descending `object` then `value` while looking for `identifier` / `property_identifier` / `field_identifier`, is the part worth reusing.

**The ruff leg is three errors, all auto-fixable**, in `tests/unit/health/test_pascal_dialect.py`:

```
I001  Import block is un-sorted or un-formatted
F401  `pytest` imported but unused
F401  `tree_sitter.Node` imported but unused
```

`ruff check --fix` clears all three. Worth running the two legs locally before the next push:

```
uv run pytest tests/unit/health/test_pascal_dialect.py -q
uv run ruff check packages/core/src/repowise/core/analysis/health/perf/dialects/pascal.py tests/unit/health/test_pascal_dialect.py
```

RaghavChamadiya:
Thanks @krishna231122-del for this. Pascal assertions and parenless calls were real gaps, and you were right to go after both.

I am closing it because #2233, opened two days before this one, covered the same ground (parenless zero-argument calls in the performance pass, DUnit-style assertions, plus Pascal `for-in` loops and uses-clause sinks) and has now merged. The three new tests here were also still failing on the callee extraction @sloemo01 described, so there was no quick path to landing both.

If you want to keep working in this area, the Pascal extractors are one of the easiest places in the repo to have an impact, and I am happy to point you at something specific. Just ask.

PR #2253, opened 2026-09-15
fix(workspace): allow removing missing workspace repos from UI (#2234)

### LINKED ISSUE #2234: [Bug] A workspace entry whose directory is gone cannot be removed from the UI — and the sidebar tells you to do exactly that
## Describe the Bug

Rename or delete a repo folder that is listed in `.repowise-workspace.yaml` and the
workspace entry becomes unremovable from the web UI. There is no API surface for it at
all, so the stale row is permanent until the user finds out that the fix is a CLI command
and a server restart.

Three things compound:

1. **`DELETE /api/repos/{repo_id}` cannot delete a synthetic entry.** `list_repos`
   synthesizes rows for workspace members that have no database row, with a prefixed id
   (`routers/repos.py:307`):

   ```python
   status = "needs_index" if abs_path.is_dir() else "missing_dir"
   synthetic_id = f"ws:{entry.alias}"
   ```

   `delete_repo` (`routers/repos.py:564`) opens with a database lookup:

   ```python
   repo = await crud.get_repository(session, repo_id)
   if repo is None:
       raise HTTPException(status_code=404, detail="Repository not found")
   ```

   A `ws:` id has no row by construction, so every delete of a synthetic entry 404s.
   And a delete that *did* somehow succeed would not help: the row is re-synthesized
   from the YAML on the next listing.

2. **No workspace endpoint can edit the config.** `routers/workspace.py` is read-only
   apart from `POST /api/workspace/sync`; nothing exposes the `workspace remove` that
   the CLI has. So this is not "the UI forgot to call it" — there is nothing to call.

3. **The UI promises the affordance anyway.** The sidebar renders a `missing` pill whose
   title is (`components/layout/sidebar.tsx:287`):

   > "Directory missing — open Workspace to remove or fix"

   and links to `/workspace`. The Workspace page then renders **no actions at all** for
   that exact status (`app/workspace/page.tsx:179`):

   ```tsx
   actionsFor={(repo) =>
     repo.status === "missing_dir" ? null : (
       <SyncButton alias={repo.id} label={repo.status === "indexed" ? "Sync" : "Index now"} />
     )
   }
   ```

   Meanwhile `attentionSentence` nags about it on every visit — "1 directory is missing on
   disk (click-to-posh). Each is marked in the list below." — pointing at a list that
   offers nothing to do about it. The user is sent to a dead end by name.

Secondary finding: `workspace remove` from the CLI does not take effect while `serve` is
running. `list_repos` reads `request.app.state.workspace_config` (`repos.py:247`), which
is loaded once at startup, so the entry survives in the UI until the server is restarted
— with no indication that a restart is what is missing.

## Steps to Reproduce

1. Have a workspace at a root with `.repowise-workspace.yaml` listing several repos.
2. Rename one of the member directories on disk (here: `Click-to-PoSh` → `Push to PoSh`).
   The old alias stays in the YAML; the sync script adds the renamed folder as a *new*
   entry, so the workspace now has both.
3. `repowise serve` from the workspace root, open the UI.
4. Sidebar shows `click-to-posh` with a `missing` pill; its tooltip says to open Workspace
   to remove it. The Workspace page lists it with no actions.
5. Try the API directly:

   ```console
   $ curl -X DELETE http://127.0.0.1:7337/api/repos/ws:click-to-posh
   {"detail":"Repository not found"}
   ```

## Expected Behavior

A workspace member whose directory no longer exists can be removed from the UI, in the
place the UI already points at. Concretely, any of:

- `DELETE /api/repos/{repo_id}` accepts a `ws:<alias>` id, and drops the entry from
  `.repowise-workspace.yaml` rather than looking for a database row; or
- a dedicated endpoint (`DELETE /api/workspace/repos/{alias}`) that wraps the same code
  path as `repowise workspace remove`, with a Remove action wired to it in
  `actionsFor` for `missing_dir`.

Either way the config change should be visible without restarting `serve` — reload
`workspace_config` after the mutation, or re-read it per request.

If removal from the UI is deliberately out of scope, the honest minimum is to stop
advertising it: change the sidebar tooltip and have `attentionSentence` name the command
that actually works (`repowise workspace remove <alias>`, then restart `serve`).

## Actual Behavior

The entry cannot be removed from the UI by any route. `DELETE /api/repos/ws:<alias>`
returns `404 {"detail":"Repository not found"}`. The only working fix is out-of-band:

```console
$ repowise workspace remove click-to-posh
Writing .repowise-workspace.yaml with 21 repo(s), down from 22 — dropping: click-to-posh
✓ Removed repo 'click-to-posh' from workspace.
```

followed by restarting `serve`, because the running server holds the old config in memory.

## Environment

- OS: Windows 11 Pro 26200
- Python version: 3.12.0
- Repowise version: 0.50.0 (`repowise --version`); code paths above verified identical on
  upstream `main` @ `aef2d429`
- Installation method: pipx

## Additional Context

Related but distinct: #2139 covers how `workspace_status` is *computed* for
registered-but-never-indexed repos under Postgres. This one is about the absence of any
removal path for the synthetic entries that status produces — same listing code, opposite
end of the lifecycle.

`missing_dir` is reachable by ordinary use: renaming a project folder is not an error
state, and a workspace that auto-registers new subdirectories will happily add the new
name alongside the orphaned old one. Every rename leaves a permanent duplicate.

I am happy to open a PR for whichever shape you prefer. The two designs differ in ways
worth settling before code: whether `/api/repos` should mutate workspace config at all,
and whether a `missing_dir` entry whose index data still exists on disk should be removed
silently or warn first.

rishu685:
## Summary
Closes #2234

When a workspace member directory is removed or renamed on disk:
1. The server creates a synthetic repo entry with `status: "missing_dir"` and ID `ws:<alias>`.
2. The UI sidebar tells users to open the Workspace page to remove or fix it, but the Workspace page rendered no action buttons for `missing_dir` rows.
3. No API endpoint existed to remove workspace configuration entries, and calling `DELETE /api/repos/ws:<alias>` returned a 404 database error.
4. Removing via the CLI (`repowise workspace remove <alias>`) was not reflected in running `repowise serve` instances because in-memory `workspace_config` was only loaded at startup.

### Changes Made

- **Server Endpoints & Live State Sync**:
  - Added `DELETE /api/workspace/repos/{alias}` in `routers/workspace.py` to drop the entry from `.repowise-workspace.yaml` and save the file to disk.
  - Live reload `request.app.state.workspace_config` and clean up `workspace_sessions`, `workspace_fts`, `workspace_path_to_repo_id`, and `workspace_vector_stores` caches so the running server immediately reflects the removal without requiring a restart.
  - Added synthetic `ws:<alias>` ID handling in `DELETE /api/repos/{repo_id}` (`routers/repos.py`) to delegate cleanly to workspace configuration removal.
  - Added and re-exported `WorkspaceRepoRemovedResponse` schema.

- **API Client**:
  - Added `removeWorkspaceRepo(alias)` to `packages/api-client/src/workspace.ts` and exported `WorkspaceRepoRemovedResponse` interface.

- **Web UI**:
  - Added `RemoveWorkspaceRepoButton` component in `packages/web/src/app/workspace/sync-buttons.tsx`.
  - Wired `actionsFor` in `packages/web/src/app/workspace/page.tsx` to display the "Remove" button when `repo.status === "missing_dir"`.

- **Unit Tests**:
  - Added `tests/unit/server/test_workspace_remove_repo.py` covering:
    - Successful removal and persistence to `.repowise-workspace.yaml`
    - Live in-memory state and cache cleanup
    - 404 response for unknown alias
    - 404 response outside workspace mode
    - Synthetic `DELETE /api/repos/ws:<alias>` deletion

### Verification
- Ran full workspace test suite (74 tests passed):
  `pytest tests/unit/server/test_workspace_router.py tests/unit/server/test_workspace_remove_repo.py tests/unit/server/test_workspace_sidebar.py`
- Verified TypeScript builds without errors.


rishu685:
@RaghavChamadiya Hi sir, can you please review this PR once as i can't exactly find the reason where it is failing actually and getting confused

RaghavChamadiya:
Thanks @rishu685, and sorry this sat. Assigned #2234 to you, since you claimed it and then delivered, which is the right order.

The CI failure here is real and it is a small one. All three Python jobs fail on the same assertion:

```
AssertionError: packages/types/src/generated/http.ts is stale. Run: python scripts/generate_http_types.py
```

`packages/types/src/generated/http.ts` is generated from the live FastAPI schema and a test compares the committed copy against it, so any change to a response model is also a change to that file. Your diff touches `packages/server/src/repowise/server/schemas/workspace.py`, so the contract moved and the committed TypeScript did not follow. Regenerate and commit the result in this PR.

Two traps in that step, both of which produce a plausible-looking but wrong file, so they are worth knowing before you run it:

Run the generator with the source paths ahead of the installed package, otherwise an installed `repowise.server` shadows the tree and roughly forty schemas vanish, which shows up as a large deletion:

```
PYTHONPATH="packages/core/src;packages/server/src;packages/cli/src" python scripts/generate_http_types.py
```

And the generator writes CRLF, so normalise to LF before committing or git reports the whole file as changed.

Sanity check before you push: the diff on `http.ts` should contain your workspace fields and nothing else. If it also removes unrelated lines, your local FastAPI or pydantic version differs from CI's and those removals should not be committed.

The rest of the change I am happy with, and once this is green I will merge it without another review round.

rishu685:
@RaghavChamadiya done! looks good to go :)

RaghavChamadiya (approved):
Thanks @rishu685 for turning the regenerated `http.ts` around quickly. That was the whole of the red, and CI is green on the head that carries it.

For anyone reading this later: the rule is that a change to `schemas/workspace.py` has to ship its `packages/types/src/generated/http.ts` regeneration in the same PR, because the drift guard compares them and will fail any branch that moves one without the other.

The `repo_id.startswith("ws:")` shortcut in the generic delete path is the part I looked at hardest, and routing workspace aliases away from the repo deletion path rather than teaching that path about workspaces is the right split.

PR #2255, opened 2026-09-15
fix(health): count Go select arms and TS/JS default arms in complexity

JikSau-D:
## Summary

`LanguageNodeMap` is a set of tree-sitter node-type strings, so a missing entry
fails silently: the arm is never counted and the file is scored as if that
control flow does not exist. Two grammars emit a dedicated arm node that the
maps never named. Both misses were reproduced against the real walker
(`walk_file`), not read off the grammar.

### 1. Go: `select` was invisible

tree-sitter-go spells a channel `select` as `select_statement` with
`communication_case` arms. Neither type appeared in the map:

| Case | Before | After |
|---|---|---|
| 3-arm `select` | **CCN 1** (straight line) | CCN 2 |
| equivalent 3-case `switch` | CCN 2 | CCN 2 |
| `select` whose arm carries an `if` | CCN 3 | CCN 4 (== equivalent `switch`) |
| `select` with a `default` arm | CCN 2 | CCN 2 (stays flat) |

Only `default` registered before the fix, because `default_case` is shared with
`switch` -- which is what shows this is an omission, not a deliberate silence.
Every Go repo using channels under-reported `max_ccn`, `complex_method`,
`nested_complexity` and the maintainability index for those functions.

### 2. TypeScript/JavaScript: a complex `default:` arm added nothing

The JS/TS grammar emits `switch_default` as its own node type; `case_kinds`
named only `switch_case`. Measured on the same dispatch:

| Language | Before | After |
|---|---|---|
| TypeScript / JavaScript | **CCN 3** | CCN 4 |
| Java / C# / Objective-C / C | CCN 4 | CCN 4 |

`_JS = _TS`, so `jsx` / `tsx` / `svelte` / `vue` inherit the fix.

## Changes

- `_GO`: `switch_kinds += "select_statement"`, `case_kinds += "communication_case"`.
- `_TS` (`_JS`): `case_kinds += "switch_default"`.
- `HEALTH_ANALYZER_VERSION` 11 -> 12 and `test_model_version_coupling._STAMPS`
  updated in the same commit. Same reasoning as #2225: `HealthWalkCache` keys
  on the stamp and the file bytes, and a node-map change moves neither, so
  without the bump an existing index keeps serving the old results.

## Steps -- 7/7 completed

- [x] 1. Reproduced both undercounts with `walk_file` on minimal snippets.
- [x] 2. Confirmed the node-type names against the real grammars
      (`select_statement`, `communication_case`, `switch_default`).
- [x] 3. Patched the two language maps in `complexity/languages.py`.
- [x] 4. Added table cases (`_GO_CASES`, `_TS_CASES`) plus two cross-language
      parity tests: `test_go_select_agrees_with_an_equivalent_switch`,
      `test_typescript_default_arm_agrees_with_java`.
- [x] 5. Bumped `HEALTH_ANALYZER_VERSION` 11 -> 12 and updated the coupling pin.
- [x] 6. Ran the affected suites: `tests/unit/health`,
      `tests/unit/change_health`, `tests/unit/cli/test_health_rescore_gate.py`
      -- 1671 passed, 1 skipped.
- [x] 7. `ruff check` clean on all four changed files.

## Test plan

```bash
uv run pytest tests/unit/health/ tests/unit/change_health/ tests/unit/cli/test_health_rescore_gate.py
uv run ruff check packages/ tests/
```

## Notes

- Flat-dispatch policy is unchanged: a flat Go `select` (with or without
  `default`) scores 2, exactly like a flat Go `switch`; a flat TS switch still
  scores 2.
- `STORE_FORMAT_VERSION` and `PARSER_SCHEMA_VERSION` are untouched; existing
  indexes re-score on their next update, no re-index required.
- Found while checking whether the same class of omission exists elsewhere
  after #2225.


JikSau-D:
Friendly ping for review when you have a moment. CI is green (Python 3.11/3.12/3.13 tests, ruff, Web UI lint + type check, tree-sitter query compilation, VS Code extension).

Short recap of the change: Go `select` arms (`select_statement` / `communication_case`) and the JS/TS `switch_default` node were missing from the complexity maps, so a three-arm `select` scored CCN 1 (a straight line) and a complex `default:` arm added nothing. Same class of fix as #2225, with `HEALTH_ANALYZER_VERSION` bumped to 12 in the same commit.

Happy to adjust scope or split anything out if that makes review easier.

RaghavChamadiya:
Thanks @JikSau-D, and sorry for the silence. You filed this on the 15th and nobody replied, which is on me, not on you.

The diagnosis is right and I checked all of it against current main.

`_GO.switch_kinds` is `{"expression_switch_statement", "type_switch_statement"}` and `case_kinds` is `{"expression_case", "type_case", "default_case"}` (`complexity/languages.py:302`). No `select_statement`, no `communication_case`. And `switch_default` appears nowhere in the health tree at all; I grepped the whole of `analysis/health/` for both and got nothing, so neither miss has been fixed by anything else in the meantime.

The argument I found most convincing is the one you put in a single line: `default_case` is already in the Go map, and it is there only because `switch` shares it. So the arm node for the same construct was being counted while the construct itself was not, which is what makes this an omission and not a deliberate decision to treat `select` as straight-line code. A three-arm `select` scoring CCN 1 while the equivalent `switch` scores 2 is not a defensible pair of numbers.

Bumping `HEALTH_ANALYZER_VERSION` in the same commit is correct and I am glad you did it without being asked. `HealthWalkCache` keys on the stamp and the file bytes, and a node-map change moves neither, so without the bump every existing index keeps serving the old scores and the fix reaches nobody who already ran an index.

**Two things to fix, both mechanical.**

Main has moved since you opened this. `HEALTH_ANALYZER_VERSION` is now `13` (`health/engine.py:129`) and `_STAMPS` is `(13, 2, 2)` (`tests/unit/health/test_model_version_coupling.py:25`). Your branch bumps 11 to 12, so after a rebase it needs to go to **14**, with `_STAMPS` as `(14, 2, 2)`. That version line is also the reason GitHub is showing this as conflicting; two other analyzer changes landed in between.

```
git fetch origin && git rebase origin/main
```

The conflict should be confined to those two numbers. Nothing else you touched has moved.

Your CI was green on your current head, and I have no doubt it will be again. Push the rebase and I will merge it on green with no further review round.

One thing worth saying since you offered to split the scope: do not. Two node-map omissions of the same class, in one commit, with one version bump, is the right size. Splitting them would mean two stamp bumps and two cache invalidations for what is one correction.

JikSau-D:
Rebased onto ca5bafa6. One note on the stamp: main had already moved past the 13 you saw -- the assertion-tier bump (#2371) took it to 14 -- so this lands as 14 -> 15 with `_STAMPS = (15, 2, 2)`, one above main, as you intended.

Conflicts were exactly the two numbers you predicted; `languages.py` re-applied cleanly. Tests re-run locally: `tests/unit/health`, `tests/unit/change_health`, `tests/unit/cli/test_health_rescore_gate.py` -- 1938 passed, 1 skipped; ruff clean.

CI is waiting for your approval to run on the new head and thx～

JikSau-D:
Main moved its analyzer stamp twice under this PR (#2393 took it to 15; this now lands as 15 -> 16), and every health PR that lands bumps it again, so each round is a fresh rebase + conflict. If you can approve and merge on green ASAP, this stops being a moving target more quickly. Same four files, tests green (1965 passed, 1 skipped).

Issue #1694, opened 2026-08-18
[Bug] build_jvm_gradle_index crashes on a relative repo path


RaghavChamadiya:
**Summary**

`build_jvm_gradle_index` subtracts one path from another where only one of the
two has been resolved, so they have no common base and `relative_to` raises.
Any multi module Gradle repository indexed through a relative path dies with a
`ValueError`. Passing an absolute path works, which is why this went unnoticed.

**Where**

`packages/core/src/repowise/core/ingestion/resolvers/jvm_gradle.py:348`

```python
root_dir=mod_dir.relative_to(resolved_repo).as_posix() if module_name else "",
```

`mod_dir` has not been resolved, `resolved_repo` has. Lines 361 and 372 do the
same subtraction but against paths taken from the already resolved side, so they
are fine. Line 348 is the only bad one.

**Repro**

Clone any multi module Gradle project (caffeine works) and index it through a
relative path:

```
repowise init ../../../test-repos/caffeine
```

```
ValueError: '..\..\..\test-repos\caffeine\caffeine' is not in the subpath of 'C:\...\test-repos\caffeine'
```

The same command with an absolute path succeeds.

**Expected**

Relative and absolute paths both index without error.

**Impact**

Affects every multi module Gradle repository. It is a hard crash, not a
degradation, so nothing gets indexed at all.

**Done looks like**

`mod_dir` is resolved before the subtraction, and a test indexes a small fixture
with a multi module Gradle layout through a relative path and asserts the index
builds. Absolute paths keep behaving exactly as they do now.

Roughly one line plus a test.

nightcityblade:
Hi, I'd like to work on this. I'll submit a PR shortly.

ASVLCII:
I see this is assigned to @nightcityblade, so I will not open a competing PR under the contribution policy. I have independently reproduced the relative-path crash and have a minimal patch plus regression test ready locally; the targeted Gradle resolver suite passes (9 tests) and Ruff passes on both changed files. If the current assignee is no longer working on it and a maintainer would like to reassign it, I can submit promptly.

RaghavChamadiya:
Thanks @ASVLCII for checking the assignment before opening anything, and for saying what you had actually run instead of just offering. That is the right way to handle an overlap.

[...]

PR #2257, opened 2026-09-15
fix(server): count stale pages without loading content

### LINKED ISSUE #2029: [Bug] check_staleness logs a count but never queues regeneration
## What

The server's `check_staleness` scheduler job says it queues regeneration for stale wiki pages. It counts them and logs the number. Nothing is ever enqueued, so a hosted or self-hosted server detects staleness on a 15-minute cadence and takes no action on it.

## Evidence

`packages/server/src/repowise/server/scheduler.py`, module docstring:

> Staleness checker — finds stale wiki pages and queues regeneration.

The job body runs a `SELECT` for `Page.freshness_status in ('stale', 'expired')` per repo, then calls `logger.info("staleness_check", ...)`. There is no enqueue call anywhere in the function.

The sibling `polling_fallback` job does enqueue, but only `mode: "sync"`, which per `docs/scale/AUTO_SYNC.md` is the index-only path: *"Server sync does not regenerate wiki pages (no LLM cost)."* So neither scheduled job reaches wiki prose.

## Why it matters beyond the docstring

Nothing else closes this loop. `repowise update` sets `file_pages_only=True` and stops the cascade before the model-written pages, and the function that marks a concept page stale (`mark_page_ids_stale`) is only reached from `generate`'s own cascade, never from `update`. So on a server whose repos sync via hook or webhook, the model-written pages never regenerate and are never marked stale — `doctor` reports a clean stale count while the prose drifts from HEAD. There is no signal anywhere until someone reads a page and finds it describes code that moved.

Context for where I hit this: a repo of ~6,500 files whose module pages were written once and then silently diverged over weeks of commits, with `doctor` reporting no staleness throughout.

## Which way should it go?

Both directions are defensible and the choice is yours, which is why this is an issue and not a PR:

1. **Make the job enqueue.** Matches the documented intent, but it means scheduled LLM spend for every server operator, and `AUTO_SYNC.md` currently promises the opposite ("no LLM cost"). That is a product decision about billing, not a bug fix.
2. **Correct the docstring**, and treat regenerating prose as explicitly operator-triggered. Cheap and honest, but leaves the underlying gap: no shipped mechanism keeps concept pages fresh, and `generate --stale` cannot find them because nothing marks them.

A third option, if the concern is cost: have the job mark the pages stale without enqueueing, so `generate --stale` and `doctor` both start reporting the truth and the operator decides when to spend.

Happy to open a PR for whichever you prefer. Related but distinct: #1744 (stale `file_page` rows surviving `update`) and #1845 (prioritising oldest stale pages under budget, which assumes queuing already happens).

ASVLCII:
## Summary

- Count stale and expired pages in SQL for each repository, so the scheduled check no longer loads rendered markdown just to log a total.
- Correct the scheduler module description: this job reports stale pages and does not queue regeneration.

## Related Issues

Fixes #2029

## Test Plan

- [x] `pytest tests/unit/server/test_scheduler.py -q` (3 passed, Python 3.12)
- [x] `ruff check` and `ruff format --check` on both changed files
- [x] `git diff --check`

## Checklist

- [x] The change stays within the assigned scheduler scope.
- [x] A regression test checks per-repository counts and confirms the query does not select page content.


ASVLCII:
Quick status: the scheduler fix and regression test are up. The focused Python 3.12 scheduler tests pass (3/3), Ruff passes, and the Repowise code-health check is green. The PR is ready for review whenever you have a moment. I kept this scoped to the stale-page count and left the assigned #1694 work alone.

PR #2275, opened 2026-09-16
fix(types): regenerate the HTTP contract for the test-impact fields

RaghavChamadiya:
`tests/unit/server/test_generated_http_types.py` compares the committed TypeScript contract against the live FastAPI schema. It has been failing on `main` since #2268, so every branch cut from `main` inherits a red suite.

The blast-radius response gained a `deleted` test-impact status, per-file `change_status` and `head_present`, and a `deleted_files` list. The generated contract was never regenerated alongside them.

Generated output only. No schema, endpoint or behaviour change — the four added lines are exactly what the generator emits for the fields already being returned.


RaghavChamadiya:
Already fixed by the v0.51.0 release (#2269), which regenerated the same contract. The committed http.ts on main is now byte-identical to this branch's.

PR #2308, opened 2026-09-17
fix(server): zero total_pages when a phase reports no total

### LINKED ISSUE #2175: [Bug] Job progress denominator goes stale when a phase reports an unknown total (241 / 5 pages)
## Describe the Bug

While a sync / re-index runs, the progress readout pairs a numerator from the *current* phase with a denominator left over from an *earlier* one. My running `full_resync` displayed:

```
241 / 5 pages
```

`completed_pages` is phase-scoped and resets each phase; `total_pages` only updates when a phase happens to know its item count. When a phase reports an unknown total, the previous phase's total stays on the record while the numerator climbs from zero.

### Where it happens

`JobProgressCallback.on_phase_start` (`packages/server/src/repowise/server/job_executor.py:170-180`) resets the numerator and assigns an optional total:

```python
def on_phase_start(self, phase: str, total: int | None) -> None:
    self._phase = phase
    ...
    # Reset per-phase counters so the bar shows progress within the current phase
    self._completed = 0
    self._total = total          # <- may be None
```

That is written out in `_async_update` (`job_executor.py:248-255`):

```python
await update_job_status(
    session, self._job_id, "running",
    completed_pages=self._completed,
    total_pages=self._total,      # <- None
    current_level=_PHASE_LEVELS.get(self._phase, 0),
)
```

and the writer treats `None` as "leave unchanged" (`packages/core/src/repowise/core/persistence/crud/repository.py:291-292`):

```python
if total_pages is not None:
    job.total_pages = total_pages
```

So `completed_pages` is overwritten every flush while `total_pages` silently retains a value that describes different work.

**34 `on_phase_start(...)` call sites pass `None`**, including both phases visible in my run:

- `packages/core/src/repowise/core/pipeline/orchestrator.py:593` — `on_phase_start("knowledge_graph.skeleton", None)`
- `packages/core/src/repowise/core/pipeline/phases/generation.py:161` — `on_phase_start("generation", None)`

The job is created with the default `total_pages=0` (`packages/server/src/repowise/server/routers/repos.py:737-742` passes no total), so the stale `5` came from an earlier phase that did know its count.

## Steps to Reproduce

1. `repowise serve` on a workspace.
2. Trigger a full resync on a repository.
3. Watch the progress readout as it moves into a phase that reports no total — `knowledge_graph.skeleton` or `generation`.
4. The numerator restarts and climbs past the denominator, which is left over from the previous phase.

## Expected Behavior

The two numbers describe the same unit of work. When a phase's total is unknown, the UI should show an honest indeterminate state or a bare processed count — not a count paired with an unrelated denominator.

## Actual Behavior

`241 / 5 pages`. Verified directly against the job record:

```json
"total_pages": 5, "completed_pages": 241, "current_level": 2, "status": "running",
"config": {"mode": "full_resync"}
```

## Possible Fixes

There is a contract decision here, which is why I am opening this as an issue rather than a PR:

1. **Clear the denominator when a phase reports no total** — have `on_phase_start` set `total_pages = 0` for unknown-total phases, so `total_pages > 0` (already the frontend's guard for whether to show a fraction at all) correctly reports "no meaningful denominator". Needs `update_job_status` to distinguish "leave unchanged" from "set to zero", since `None` currently means the former.
2. **Make the pair atomic** — never write `completed_pages` without writing the `total_pages` belonging to the same phase, so the two cannot describe different work.
3. **Carry the phase on the record** so the UI can label what is being counted, rather than calling every unit a "page" (`241 / 5 pages` is also mislabelled — the units in these phases are graph nodes and edges, not pages).

Happy to open a PR once you have picked a direction.

## Related

The out-of-range fraction also produces a nonsensical percentage in the UI; filed separately as the frontend does not clamp it.

## Environment

- OS: Windows 11 Pro (26200)
- Python version: 3.12.0
- Installation method: pipx

Aman-goel-04:
## Summary

- Fix JobProgressCallback._async_update() so a phase starting with an unknown total (total=None) zeroes total_pages instead of leaving a previous phase's total on the record
- Add a regression test proving the zeroing, plus a test proving callers that omit total_pages still preserve the existing value

## Related Issues

Fixes #2175

## Test Plan

- [x] Tests pass (pytest tests/unit/server/test_job_executor.py)

## Checklist


RaghavChamadiya (approved):
This is the change I asked for on #2175 and nothing else, which is exactly right. `update_job_status` keeps `None` meaning "leave unchanged", so the three callers that depend on that are untouched, and the fix sits at the one place that knows a phase has started with no count.

The test is the part that makes it convincing: it seeds a stale `241 / 5`, drives the real `JobProgressCallback` through `on_phase_start` with `None`, and asserts both halves are cleared. That drives the actual path instead of the helper in isolation.

Thanks for waiting on this, and sorry it took a few days to get to.

PR #2309, opened 2026-09-17
fix(csharp): align builtin_parents with shared .NET BCL set

### LINKED ISSUE #2304: [Bug] C# builtin_parents is missing ten BCL types that VB.NET's list already carries
**Summary**

The two .NET language specs disagree about which base types are framework builtins.

`packages/core/src/repowise/core/ingestion/languages/specs/csharp.py:83-97` lists twelve: `Object`, `ValueType`, `Enum`, `Exception`, `SystemException`, `ApplicationException`, `IDisposable`, `IEnumerable`, `IEnumerator`, `IComparable`, `ICloneable`, `IEquatable`.

`languages/specs/vbnet.py:60-92` lists those **plus** `Form`, `Control`, `UserControl`, `Component`, `Page`, `Window`, `Attribute`, `EventArgs`, `MarshalByRefObject`, `INotifyPropertyChanged`, `IComparer`, `ICollection`, `IList`, `IDictionary`.

So `class Foo : EventArgs` in C# stays a real heritage relation, while the identical `Inherits EventArgs` in VB.NET is correctly filtered.

**Mechanism**

An unfiltered builtin parent can reach the heritage resolver's global-unique tier. If any class in the repository is literally named `EventArgs` or `Attribute`, the relation binds to it.

**Impact**

**Latent rather than firing today**, and worth stating plainly: neither Ocelot nor eShopOnWeb declares a colliding class, so no wrong edge is produced in either. The exposure is any C# repository that declares a type sharing one of these BCL names, which is not unusual for `Attribute` or `EventArgs` subclasses named after them.

`test-repos/Ocelot/src/LoadBalancer/LeaseEventArgs.cs:5` and `test-repos/eShopOnWeb/src/BlazorShared/Attributes/EndpointAttribute.cs:5` exercise the unfiltered path today.

**Done looks like**

The C# list carries the same BCL names the VB.NET list does. The two specs share a runtime, so a name that is a builtin parent in one is a builtin parent in the other.

Worth deciding while fixing: whether these belong in a shared .NET constant rather than two hand-kept frozensets. The drift is the actual defect; copying the names fixes this instance and leaves the mechanism in place.

Tests: the heritage type-name tests — a C# `class Foo : EventArgs` and `class Bar : Attribute` asserting no heritage relation, mirroring whatever assertion already covers VB.NET's `Form`.

Voyagerroc-Lab:
Fixes #2304

### Summary of Changes
- Aligned C# `builtin_parents` in `packages/core/src/repowise/core/ingestion/languages/specs/csharp.py` with the shared .NET BCL types already present in `vbnet.py`:
  - Added `Form`, `Control`, `UserControl`, `Component`, `Page`, `Window`, `Attribute`, `EventArgs`, `MarshalByRefObject`, `INotifyPropertyChanged`, `IComparer`, `ICollection`, `IList`, `IDictionary`.
  - Prevents C# inheritance relations like `class Foo : EventArgs` or `class Bar : Attribute` from leaking into the heritage graph and binding to matching repository types.
- Added unit tests in `tests/unit/ingestion/test_csharp_extractors.py` verifying that all these BCL base types and interfaces are correctly filtered out from the heritage graph in C# code.
- Tested against pytest suite with all 13 tests passing.


RaghavChamadiya (approved):
Thanks @Voyagerroc-Lab for finding the parity rather than the symptom. The interesting part of #2304 is that `vbnet.py`'s own comment says it inherits the C# set verbatim, so the two drifting apart is a bug in the claim the code already makes about itself. Copying the missing 14 across in one go closes it rather than patching the one type someone happened to hit.

Verified on current main before merging: `csharp.py` carried 12 entries and `vbnet.py` carried those same 12 plus exactly the 14 this adds.

`test_builtin_bcl_parents_filtered` asserting `result.heritage` is empty across five classes, including the multi-interface `Entity : MarshalByRefObject, IDisposable, INotifyPropertyChanged` case, is real negative coverage rather than a single happy path.

This is your first merged change here. Thanks for picking up something unglamorous.

One follow-up for whoever wants it, not for this PR: `fsharp.py` has its own `builtin_parents` with `Attribute`, `EventArgs` and `MarshalByRefObject` but still missing `Form`, `Control`, `UserControl`, `Component`, `Page`, `Window`, `INotifyPropertyChanged`, `IComparer`, `ICollection`, `IList` and `IDictionary`. Same gap, one file over.

PR #2316, opened 2026-09-17
fix(godot): resolve scene connections to C# script methods

### LINKED ISSUE #2303: [Bug] A Godot [connection] to a C# method produces no edge
**Summary**

`_add_connection_edges` builds its handler lookup only from GDScript files.

`packages/core/src/repowise/core/ingestion/framework_edges/godot.py:308-318`

The `functions` map is populated where `parsed.file_info.language == "gdscript"` (line 310). A scene whose `script` ext_resource points at a `.cs` file resolves the path correctly at `:350` and then finds nothing in `functions` at `:353`, so no edge is created.

**Mechanism**

The lookup is keyed by resolved script path and method name, which is language-neutral. Only the population step is not. C# method symbols are already extracted by the normal C# pipeline, so the data exists; it just is not offered to this map.

**Repro**

`test-repos/godot-demo-projects/mono/` has twelve `.tscn` files with `.cs` scripts and `[connection]` blocks, across `dodge_the_creeps`, `squash_the_creeps`, `multiplayer_pong`, `pong` and `android_iap`. None produces a connection edge.

**Impact**

Every Godot project using C# rather than GDScript, which is a first-class Godot option, not a niche. Signal handlers reached only through a scene connection have no incoming edge and can read as dead.

Silent under-reporting.

**Done looks like**

The `functions` map also accepts `csharp` symbols of function or method kind, so the existing `[connection ... method="..."]` lookup works regardless of the script's language.

Worth noting what stays out of scope: `connect(..., "method_name")` string dispatch is a different problem and should not be folded in here. The corpus has zero occurrences of that legacy form, and resolving it by name matching risks binding to the wrong same-named handler.

Tests: `tests/unit/ingestion/test_godot_project_awareness.py` — a `.tscn` plus `.cs` fixture asserting a `[connection ... method="OnPressed"]` resolves to the C# method symbol.

Aman-goel-04:
## Summary

- Widen _add_connection_edges' handler lookup in godot.py to include C# symbols, not just GDScript, so a scene [connection] pointing at a .cs script's method resolves to an edge
- Widen _GodotSceneConnectionHandler.detect() the same way, since a pure C# ("Mono") Godot project with zero .gd files would otherwise never trigger the handler at all

## Related Issues

Fixes #2303

## Test Plan

- [x] Tests pass (pytest tests/unit/ingestion/test_godot_project_awareness.py)

## Checklist

- [x] I have updated documentation if needed; not applicable, no user-facing docs affected


RaghavChamadiya (approved):
Thanks @Aman-goel-04 for leaving the third gate alone. `_add_class_name_edges` at `godot.py:140` also tests `!= "gdscript"`, and widening that one too would have been the obvious over-reach, but `class_name` is GDScript syntax with no C# equivalent, so it is correctly GDScript-only. Changing exactly the two gates that needed it is what makes this easy to review.

Verified on current main that `godot.py:310` still skips any parsed file whose language is not `gdscript` when building the handler lookup, and that `_GodotSceneConnectionHandler.detect` is still gated on `_has_gdscript` alone, so a pure C# Godot project never triggers the handler at all.

PR #2318, opened 2026-09-17
fix(swift): stop a qualified extension naming its qualifier as a parent

### LINKED ISSUE #2297: [Bug] extension Ns.Type: P records the namespace as conforming to the protocol
**Summary**

`extension Ns.Type: P {}` records **two** heritage relations where there is one: the correct `Type extends P`, and a spurious `Ns extends P`, where `Ns` is only the qualifier.

Reproduced at HEAD:

```python
src = b"""
enum Ns { struct Type {} }
protocol P {}
extension Ns.Type: P {}
"""
# result.heritage -> [('Ns', 'P'), ('Type', 'P')]
```

**Mechanism**

tree-sitter-swift parses `Ns.Type` as a **flat** `user_type` with two `type_identifier` children rather than a nested qualifier chain:

```
class_declaration "extension Ns.Type: P {}"
  user_type "Ns.Type"
    type_identifier "Ns"
    type_identifier "Type"
  inheritance_specifier "P"
```

The capture is unanchored, so it matches both children.

`packages/core/src/repowise/core/ingestion/queries/swift.scm:24-29`

```scheme
; extension Foo: Protocol — name is nested under user_type
(class_declaration
  (user_type
    (type_identifier) @symbol.name
  )
) @symbol.def
```

`extract_heritage` then dedupes on `(line, name)`, which does not collapse two different names on one line, so both survive.

**Impact**

A wrong edge, not a missing one, which is the worse direction. It reaches every consumer of heritage: the `extends`/`implements` reachability used by dead code, the architecture graph, and `get_context`.

Swift sits near the bottom of the hand-graded precision table, so this is worth removing on its own terms.

Unsized across the corpus. Only qualified extensions are affected; plain `extension Foo: P` has a single-identifier `user_type` and is correct today.

**Done looks like**

Only `('Type', 'P')` is recorded. Either anchor the capture to the last `type_identifier` in the `user_type`, or take the last matching name per `def_node` in `extract_heritage`.

Confirm `class Foo: Bar` and plain `extension Foo: Protocol` are unchanged — both are single-identifier `user_type`s and already correct.

Tests: `tests/unit/ingestion/test_swift_extractors.py`, alongside `test_extension_protocol_conformance`. Assert the relation set **equals** `{("Type", "P")}` rather than merely contains it. Asserting containment would pass against the bug.

sloemo01:
## What

Fixes #2297. `extension Ns.Type: P {}` recorded two conformance relations where there is one: the correct `Type -> P`, and a spurious `Ns -> P` that named the qualifier as conforming on its own.

## Root cause

`queries/swift.scm:24-29` captured the name with an unanchored `(type_identifier) @symbol.name` inside `user_type`. tree-sitter-swift parses `Ns.Type` as one flat `user_type` holding two `type_identifier` children, so the capture matched both. The qualifier came through as a symbol name, and `_extract_swift_heritage` records a relation for every `user_type` name it finds, which is how `Ns` ended up as a parent of `P`.

Reproduced at HEAD before the fix:

```
heritage: [('Ns', 'P'), ('Type', 'P')]
symbols:  [('Ns', ...), ('Type', ...), ...]
```

## Changes

- Anchored the capture to the last child with `.` in `packages/core/src/repowise/core/ingestion/queries/swift.scm`. A qualified name now resolves to its trailing identifier.

Single-identifier `user_type`s are untouched: plain `extension Foo: P`, `class Foo: Bar`, and `protocol Pet: Animal` capture exactly what they did before.

## Tests

- `test_qualified_extension_records_only_the_type` and `test_deeply_qualified_extension_takes_the_trailing_identifier` assert the relation set equals `{("Type", "P")}` and `{("C", "P")}` rather than containing those pairs. Containment passes against the bug.
- `test_plain_extension_is_unaffected_by_the_anchor` pins the single-identifier case.
- Sabotage-verified: with the `.scm` change stashed, both qualified-extension tests fail with the exact spurious edges (`('A', 'P')`, `('B', 'P')`); both pass with it.

Local runs: `test_swift_extractors.py` 8 passed, `test_swift_resolver.py` 7 passed, `test_swift_same_module.py` 10 passed, and `tests/unit/ingestion/parser/test_query_compilation.py` 27 passed (the gate that compiles every language query). Full `tests/unit/ingestion/` came back 3285 passed with one unrelated failure, `test_repo_totals_churn.py::test_a_rebased_branch_in_the_history_falls_back`, which fails on unmodified main too (it shells out to `git rebase master` and this clone's default branch is `main`).


RaghavChamadiya (approved):
Anchoring the capture at the trailing identifier is the correct fix: `user_type` is a flat node here, so an unanchored capture was always going to take the qualifier as well as the type. What makes this easy to approve is that the tests assert relation-set equality and not containment, and they pin the plain `extension Foo: P` and deeply qualified `A.B.C` cases alongside the reported one, so a regression in either direction fails loudly.

PR #2323, opened 2026-09-17
fix(graph): pair Objective-C .h and .m header/implementation files

### LINKED ISSUE #2300: [Bug] Objective-C .h/.m files never pair, so header consumers never reach the implementation
**Summary**

`_resolve_cpp_header_pairs` gates on language before it gates on extension.

`packages/core/src/repowise/core/ingestion/graph/_resolvers.py:310`

```python
if pf.file_info.language in ("c", "cpp"):
```

Objective-C files carry `language == "objectivec"` (`languages/specs/objectivec.py:25`), so `.m` and `.mm` files are never collected as candidate sources. `Foo.h` and `Foo.m` never receive the `header_source_pair` edge that lets a header's consumers reach the implementation.

The `source_exts` tuple at `:298-306` also lists no `.m`/`.mm`, so both gates need widening.

**Mechanism**

Nothing downstream constrains this to C and C++. `header_source_pair` is already unconditional in `COHESION_HINTS` (`cohesion.py:69`), so the restriction is in the producer alone.

**Repro**

Index `test-repos/AFNetworking` or `test-repos/SDWebImage` and look for any edge with `hint_source="header_source_pair"` between a `.h` and its `.m`. There are none.

**Impact**

Every Objective-C project. The `.h` and `.m` halves of a class stay unconnected, so a consumer that imports the header does not reach the implementation file.

**Done looks like**

`"objectivec"` is accepted at the language gate and `.m`/`.mm` are in `source_exts`.

Note this is genuinely two edits, not one: widening either gate alone changes nothing.

Worth checking in the same pass: `.mm` is Objective-C++ and may also be reported as `cpp` depending on how the file is routed. Confirm which tag it actually carries before assuming both need adding.

Tests: `tests/unit/ingestion/test_cpp_header_pairs.py` — an Objective-C case asserting `Foo.h` pairs with `Foo.m`.

sloemo01:
## What

Fixes #2300. `_resolve_cpp_header_pairs` gated on language before it gated on extension, so Objective-C `.h`/`.m` files were never paired.

## Root cause

Two gates, both closed to Objective-C:

```python
if pf.file_info.language in ("c", "cpp"):   # _resolvers.py:310
    ...
source_exts = (".c", ".cpp", ".cxx", ".c++", ...)   # :298-306
```

Objective-C files carry `language == "objectivec"` (`specs/objectivec.py` claims `.m` and `.mm`), so they never entered the candidate set. `Foo.h` and `Foo.m` got no `header_source_pair` edge, and a consumer importing the header could not reach the implementation.

Nothing downstream constrains the pairing to C and C++. `header_source_pair` is unconditional in `COHESION_HINTS` (`cohesion.py:69`), and the pairing loop only compares same-directory lowercase stems.

## Changes

Widened both gates: the language check now admits `"objectivec"`, and `source_exts` includes `.m` and `.mm`. The two are inseparable, since widening either alone pairs nothing.

I checked the `.mm` routing the issue asked about before assuming. The `.mm` extension is Objective-C++, so it could plausibly carry `"cpp"`. It does not: the objectivec spec claims both `.m` and `.mm`, and the cpp spec claims neither, so both extensions belong in the list.

## Tests

`tests/unit/ingestion/test_cpp_header_pairs.py` gains a `TestObjectiveCHeaderSourcePairing` class covering the `.m` and the `.mm` case. Both fail with the source change reverted and pass with it. The neighbouring cpp header-pairing and cohesion tests still pass, and `ruff check .` is clean.


RaghavChamadiya (approved):
Both gates needed widening together and this does both, which is the part that would have been easy to half-fix. Worth recording why the header side works: `traverser.py` content-sniffs a `.h` into `objectivec`, so the header genuinely arrives at this producer with that language once the gate admits it, and the pairing is not relying on the extension list alone.

PR #2324, opened 2026-09-17
fix(ingestion): capture C# null-conditional call sites

### LINKED ISSUE #2296: [Bug] A C# null-conditional call (a?.Foo()) produces no call site at all
**Summary**

`queries/csharp.scm` has four invocation patterns: a bare call, a member or self call, a chained call, and a fluent construction. All four anchor on `member_access_expression`.

`packages/core/src/repowise/core/ingestion/queries/csharp.scm:185-249`

tree-sitter-c-sharp represents `a?.Foo()` differently: a `conditional_access_expression` wrapping a `member_binding_expression`. A repo-wide grep finds **zero** patterns for either node type anywhere in `packages/core/src/repowise/core/ingestion`.

So the call is not resolved-and-wrong, and not unresolved. It is never captured, which means nothing downstream knows it existed.

**Mechanism**

A shape no query matches is absent rather than low-confidence. It does not appear in the unresolved bucket either, so it is invisible to the recall measurements that would otherwise flag it.

**Repro**

```csharp
public void Notify(ILogger? logger)
{
    logger?.LogInformation("done");
}
```

Index the file and look for a call site targeting `LogInformation`. There is none.

**Impact**

`a?.Foo()` is ordinary modern C#. Previously sized at **1,650 of Ocelot's 22,582 invocation nodes**, roughly 7%. `test-repos/Ocelot` is available for re-sizing.

Under-reporting only. No wrong edge is produced.

**Done looks like**

A new pattern capturing an `invocation_expression` whose `function` is a `member_binding_expression` reached through a `conditional_access_expression`, with the receiver as `@call.receiver` and the binding's name as `@call.target`, mirroring the existing member-call pattern at `:197-206`.

**Keep this scoped to the plain identifier receiver.** The chained and fluent conditional-access shapes each raise their own receiver-typing question, and bundling them turns an additive capture fix into a resolution change.

Tests: the C# call-site extraction tests — a `logger?.LogInformation(...)` fixture asserting a call site with the correct target and receiver, and that it resolves through the existing typed-receiver tiers with no resolver change.

sloemo01:
## What

Fixes #2296. `a?.Foo()` produced no call site at all.

## Root cause

`csharp.scm` has four invocation patterns and all four anchor on `member_access_expression`. tree-sitter-c-sharp spells the null-conditional form differently:

```
invocation_expression
  function: conditional_access_expression
    condition: identifier
    member_binding_expression
```

Nothing in `packages/core/src/repowise/core/ingestion` matched either node type, so the call was not resolved-and-wrong and not unresolved either. It was never captured, which is why it stayed invisible to the recall measurements that would otherwise have flagged the gap.

## Changes

A fifth invocation pattern anchored on the conditional access form, with the receiver as `@call.receiver` and the binding's name as `@call.target`, mirroring the existing member-call pattern so the call lands on the receiver-typing tiers the resolver already has. No resolver change was needed.

Scope is the plain identifier receiver only. A chained or constructed receiver inside the conditional access raises its own receiver-typing question, so bundling those shapes would turn an additive capture fix into a resolution change.

## Tests

9 new tests in `tests/unit/ingestion/test_csharp_null_conditional_calls.py`. Five fail without the pattern and pass with it; the other four are controls that do not depend on it. The wider `tests/unit/ingestion` suite is 3291 passed, and `ruff check .` is clean.

## One pre-existing limitation, left alone

The receiver-type scan does not read a C# nullable annotation, so `ILogger? logger` types nothing. That is equally true of a plain `logger.LogInformation(...)` call on unmodified `main`, which makes it a typing gap rather than a capture gap, and it is out of scope here.


RaghavChamadiya (approved):
Anchoring on `conditional_access_expression` is right: that is genuinely how `a?.Foo()` parses, and since the four existing patterns all anchor on `member_access_expression`, the two cannot both match one invocation. Scoping the receiver to a plain identifier keeps the first version of this narrow, which I prefer here.

PR #2325, opened 2026-09-17
fix(ingestion): resolve a nested sub-package from its own manifest

### LINKED ISSUE #2301: [Bug] A workspace sub-package with its own package.json resolves to nothing
**Summary**

Some packages ship sub-packages nested inside their own tree, each with its own `package.json` and entry point, without being separate workspace members. solid ships `packages/solid/web`, `packages/solid/store` and others this way.

When something imports `solid-js/web`, the resolver matches the outer package and then tries to resolve `web` as a path under the outer package's own layout. It never considers that `web` is itself a package.

`packages/core/src/repowise/core/ingestion/resolvers/ts_workspace.py:573-584`

```python
direct = _probe_path(f"{dir_posix}/{sub}", ctx.path_set)
if direct is not None:
    return direct
for src_root in ("src", "lib", "dist"):
    cand = _probe_path(f"{dir_posix}/{src_root}/{sub}", ctx.path_set)
    if cand is not None:
        return cand
return spare_export_target()
```

`_probe_path` (`:470-489`) additionally tries `<base>/index.*`. None of these reads `<dir>/<sub>/package.json`.

**Mechanism**

The probes nearly work, which is why this is easy to miss. `_probe_path` would find `packages/solid/web/index.ts` if it existed. It does not. Verified on disk:

```
test-repos/solid/packages/solid/web/package.json
test-repos/solid/packages/solid/web/src/index.ts
```

The entry point is `web/src/index.ts`, named by `web`'s **own** manifest. The outer package's `src` root is `packages/solid/src`, so the `src` fallback probes `packages/solid/src/web`, which is a different place entirely.

**Repro**

Index `test-repos/solid` and resolve an import of `solid-js/web`. It resolves to nothing and becomes an external node; every call through it is absent from the call graph.

**Impact**

Any package publishing nested sub-packages this way. solid is the confirmed instance in the corpus.

Under-reporting only.

**Done looks like**

When the four existing probes fail, check for `<dir>/<sub>/package.json` and resolve the subpath as its own package, reading its `main`/`module`/`exports` and probing relative to `<dir>/<sub>`.

Branch 2 of the same function (`:552-571`) already does bare-package resolution and is the logic to mirror, rooted one level deeper.

This can only replace an unresolved external node with a real file. It cannot retarget an import that already resolves, because it runs only after every existing probe has failed.

Tests: the `ts_workspace` resolver tests — a fixture with a nested `package.json` under a workspace member, asserting the subpath specifier resolves to the nested entry file rather than `None`.

sloemo01:
## What

Fixes #2301. A workspace sub-package nested inside a member, with its own `package.json`, resolved to nothing.

## Root cause

Some packages publish sub-packages inside their own tree, each with its own manifest and entry point, without declaring them as workspace members. solid ships `packages/solid/web`, `packages/solid/store` and others this way.

When something imports `solid-js/web`, the resolver matched the outer package and then probed `web` as a path under the outer package's own layout (`_probe_path` at `ts_workspace.py:470-489`, the subpath branch at `:573-584`). None of those probes read `<member>/<sub>/package.json`, which is where the entry point is actually named:

```
test-repos/solid/packages/solid/web/package.json
test-repos/solid/packages/solid/web/src/index.ts
```

The member's own `src` root is `packages/solid/src`, so the `src` fallback probed `packages/solid/src/web`, a different place entirely.

## Changes

The subpath fallback now checks for `<member>/<sub>/package.json` after every existing probe has failed, and resolves the subpath as its own package: its `exports["."]` map, then `index.*`, then `module`/`main`, then spare export targets and conventional source roots, all probed relative to `<member>/<sub>`. This mirrors the bare-package branch rooted one level deeper. A declaration file is refused as a fallback entry, as it already is higher up.

The manifest read goes to disk rather than to the in-memory path set, because a manifest is not an indexed source file, so the path set can never say whether one exists. It is memoized on the resolver context, following the existing `godot_project_root` precedent for the same reason.

Running last keeps the change additive: it replaces an unresolved external node with a real file, and cannot retarget an import that already resolves.

## Tests

`tests/unit/ingestion/test_typescript_resolver.py` passes 78 tests, 4 of them new. With the source change reverted those 4 fail with `assert None == 'packages/solid/web/src/index.ts'` and `assert 'external:solid-js/web' == ...web/src/index.ts'`. The 4 no-movement guards in the same class pass either way.

`ruff check .` is clean, the adjacent ts_workspace and ts dead-code files pass 52 tests, and a scratch copy of solid's real on-disk layout resolves `solid-js/web` to `packages/solid/web/src/index.ts`.


RaghavChamadiya:
Holding this one. @Aman-goel-04 claimed #2301 at 17:16 on the 17th, shortly before this went up, so the issue stays with them. Longer note on #2345.

sloemo01:
Closing this, since #2301 was already claimed by @Aman-goel-04 before this branch went up. The issue stays with them, so leaving this open only parks a review that is not going to happen.

PR #2326, opened 2026-09-17
fix(mcp): name the capability, not one host's tool names, in no-index guidance

### LINKED ISSUE #2305: [Bug] The no-index guidance tells every MCP client to use Read/Grep/Glob, including clients that have neither
**Summary**

When no index exists, every repowise tool returns a shaped notice whose guidance names three Claude Code tools.

`packages/server/src/repowise/server/mcp_server/_failure_shield.py:57-61`

```python
"guidance": (
    "Until an index exists, every repowise tool will return this "
    "notice. Answer questions about this repo with your built-in "
    "tools (Read/Grep/Glob) for the rest of the session."
),
```

The same text appears again at `:135` for the schema-repair case.

The codebase documents elsewhere that this is not universally true.

`packages/cli/src/repowise/cli/editor_integrations/claude_config.py:227`

```python
# Codex has no Read/Grep/Glob tools and no SessionStart block, so the shell is
```

**Mechanism**

The server cannot know which built-in tools the calling client has. Naming a specific host's tools is a guess that happens to be right for one client.

Naming *something* is deliberate and correct: when there is no index, every repowise tool is dead for the session, so telling the agent to fall back is the right call. It is the specific names that do not travel.

**Repro**

Call any repowise tool from a Codex-based client against an un-indexed repository. The guidance names three tools that client does not have.

**Impact**

Every non-Claude-Code MCP client, on the one code path that fires when nothing else works. The agent is most dependent on this instruction exactly when it is least able to follow it.

**Done looks like**

The guidance names the capability rather than the tools: the agent's own file-reading and searching tools, whatever they are called.

The same phrasing appears in three skill files and should move together, one of which is the codex plugin's own and therefore carries the same problem:

* `plugins/claude-code/skills/codebase-exploration/SKILL.md:41`
* `plugins/codex/.../SKILL.md:36`
* `plugins/shared/skills/codebase-exploration.md:49`

Tests: `tests/unit/server/mcp/test_failure_shield.py:47` currently pins the literal string `"Read/Grep/Glob"`, so it fails by design until updated. Keep it a literal pin on the new wording; that pin is what surfaced this.

sloemo01:
## What

Fixes #2305. The no-index guidance named three Claude Code tools, so it was unactionable for every other MCP client.

## Root cause

When no index exists, the failure shield returns a shaped notice whose guidance reads:

```
Answer questions about this repo with your built-in tools
(Read/Grep/Glob) for the rest of the session.
```

That phrasing appears on two shapes in `_failure_shield.py` (the not-indexed notice at `:57-61` and the schema-repair notice around `:135`), and a third variant lives in the shared codebase-exploration skill.

The server cannot know which built-in tools the calling client has. Those three names are Claude Code's, and a Codex client has none of them (`claude_config.py:227` documents this). The agent is most dependent on this instruction exactly when it is least able to follow it.

Naming something is deliberate and stays. When there is no index, every repowise tool is dead for the session, so telling the agent to fall back is correct. Only the host-specific names are wrong.

## Changes

The guidance now names the capability instead of the tools: the agent's own file-reading and searching tools, whatever they are called. The session-scoped framing and the fallback instruction are unchanged.

The same sentence lives in `plugins/shared/skills/codebase-exploration.md`, from which the claude-code and codex copies are rendered, so it moves with the shield and the wording stays one.

## Tests

`tests/unit/server/mcp/test_failure_shield.py` now pins the new wording on both shapes and asserts the old names are gone. With the source change reverted both tests fail on the old text; with it, 15 passed. `ruff check .` is clean, and 148 tests pass across the shield, plugin-content, tool-bridge and tool-adapter files.


RaghavChamadiya (approved):
Naming the capability instead of one host's tool names is the right call. The guidance is served to every MCP client, and three of them do not have a Read, a Grep or a Glob, so the old wording was unactionable for exactly the clients that most needed it. Covering the three skill copies in the same pass keeps them from drifting apart again.

PR #2327, opened 2026-09-17
fix(rust): widen type_use captures to the wrapped type shapes

### LINKED ISSUE #2298: [Bug] Rust type_use misses &Foo, Box<Foo> and std::io::Error, though the extractor already handles them
**Summary**

`queries/rust.scm` feeds only a bare `type_identifier` into `@param.type`, with one narrow special case for `&dyn Trait`. A parameter or return type written `&Foo`, `Box<Foo>` or `std::io::Error` produces no `type_use` edge.

`packages/core/src/repowise/core/ingestion/queries/rust.scm:329-373`

The head extractor already handles all three.

**Mechanism**

`_rust_head_type_identifier` (`packages/core/src/repowise/core/ingestion/parser_helpers.py:1341-1403`) unwraps `reference_type`/`pointer_type`, `generic_type`, `scoped_type_identifier` (taking the rightmost segment), and `dynamic_type`/`abstract_type`. It is registered in `TYPE_HEAD_EXTRACTORS["rust"]` and unit-tested on exactly these shapes in `tests/unit/ingestion/test_rust_type_use.py` (`test_reference_and_mut_are_unwrapped`, `test_scoped_type_takes_rightmost_component`, `test_generic_head_is_the_constructor`).

So the machinery exists and is proven. Only the query is narrow, which makes this a query-only change.

**Repro**

```rust
fn takes(x: &MyType) {}
fn boxed(y: Box<MyType>) {}
fn io() -> std::io::Error { unimplemented!() }
```

None of the three produces a `type_use` edge to `MyType`/`Error`.

**Impact**

Under-reporting on a common Rust shape: a borrowed or boxed parameter is the normal way to pass a type. `type_use` is in `REACHABILITY_USE_EDGE_TYPES`, so a type used only this way can read as unreachable.

Unsized. The affected shapes are not counted today.

**Done looks like**

The `@param.type` positions accept the wrapper shapes the extractor already unwraps, rather than only a bare `type_identifier`.

This cannot mint a wrong edge: the extractor returns the head type or nothing, and a type name that resolves to no repository symbol is dropped. The change is additive by construction.

Worth checking in the same pass whether the return-type position at `:357-359` needs the same widening; it is bare-only today for the same reason.

Tests: extend `tests/unit/ingestion/test_rust_type_use.py` from the extractor level to the query level, asserting `&Foo`, `Box<Foo>` and `std::io::Error` in parameter and return position each produce a `type_use` edge.

sloemo01:
## What

Fixes #2298. `type_use` missed `&Foo`, `Box<Foo>` and `std::io::Error`, so most Rust type uses went unrecorded.

## Root cause

`rust.scm` captured the bare `type_identifier` at each type position. A wrapped type is not a `type_identifier`, so it never matched the query and its head never reached the Rust head extractor. Since `type_use` is in `REACHABILITY_USE_EDGE_TYPES`, a type used only through a wrapper could read as unreachable.

The extractor already handles all three shapes. `_rust_head_type_identifier` (`parser_helpers.py:1341-1403`) unwraps `reference_type`, `pointer_type`, `generic_type`, `scoped_type_identifier` (taking the rightmost segment), and `dynamic_type`/`abstract_type`, and is registered in `TYPE_HEAD_EXTRACTORS["rust"]` with unit tests on exactly these shapes. Only the query was narrow, which makes this a query-only change.

## Changes

The parameter, return and trait-bound patterns now capture the enclosing type node, and the type-argument pattern accepts the shapes the extractor unwraps. That alternation deliberately excludes `type_binding`, whose head name is the binding (`Item` in `Iterator<Item = Foo>`), not a type. The turbofish-only `generic_function` pattern is subsumed by the anchored `type_arguments` one.

The change cannot mint a wrong edge: the extractor returns the head name or nothing, and a head naming no repository symbol is dropped by the resolver.

## Tests

New query-level and graph-level tests in `tests/unit/ingestion/test_rust_type_use.py` cover `&Foo`, `Box<Foo>` and `std::io::Error` in both parameter and return position, each producing a `type_use` edge.

With the query change reverted, 20 of the file's 47 tests fail. With it, 47 pass and the full `tests/unit/ingestion` suite (3293 tests) stays green.


RaghavChamadiya:
Holding this one. @Aman-goel-04 claimed #2298 at 10:13 on the 17th with a full implementation plan, about nine hours before this PR, so the issue stays with them. Longer note on #2345.

sloemo01:
Closing this, since #2298 was already claimed by @Aman-goel-04 before this branch went up. The issue stays with them, so leaving this open only parks a review that is not going to happen.

PR #2330, opened 2026-09-17
fix(objectivec): blank call-shaped availability macros trailing a declaration

### LINKED ISSUE #2299: [Bug] A trailing availability macro cascades an Objective-C file into 819 error nodes
**Summary**

`prepare_objectivec_source` blanks a fixed set of **whole-line** macros before parsing: `NS_ASSUME_NONNULL_BEGIN`/`END`, `NS_HEADER_AUDIT_*`, `CF_ASSUME_NONNULL_*`, `NS_REFINED_FOR_SWIFT`, `NS_SWIFT_UI_ACTOR`.

`packages/core/src/repowise/core/ingestion/parser_helpers.py:817-835`

A **call-shaped** macro trailing a declaration is not covered, and one is enough to derail the rest of the file.

Measured on `test-repos/AFNetworking/AFNetworking/AFURLSessionManager.m`:

| sanitiser | ERROR nodes |
|---|---|
| today | **819** |
| with `AF_API_AVAILABLE(...)` spans blanked | **99** |

**Mechanism**

The trigger is a declaration like `AFURLSessionManager.m:98`:

```objc
@property (readwrite, nonatomic, copy) ... taskDidFinishCollectingMetrics
    AF_API_AVAILABLE(ios(10), macosx(10.12), watchos(3), tvos(10));
```

The grammar has no rule for the macro, error recovery fails to resync, and the damage runs past the declaration into the rest of the file. Calls inside affected method bodies then have no enclosing function to attribute to, so they are credited to `__module__` instead of their real caller.

That last part is the reason this matters beyond a parse statistic: the caller end of those edges is wrong, not missing.

**Repro**

Parse `AFURLSessionManager.m` through `prepare_objectivec_source` and count `ERROR` nodes in the resulting tree. 819 today.

**Impact**

Any Objective-C project using availability macros, which is most of them that support more than one OS version. AFNetworking and SDWebImage both do.

**Done looks like**

Trailing call-shaped macros are blanked the same way the bare-line ones already are, preserving byte offsets so every recorded position stays correct. `_blank_matches` already does exactly this.

Enumerate the macro names rather than matching any `IDENT(...)` before `;` or `{`. A general rule would blank real code; an enumerated list (`AF_API_AVAILABLE`, `API_AVAILABLE`, `API_DEPRECATED`, `NS_SWIFT_NAME`, `NS_AVAILABLE`, `NS_DEPRECATED`) is what the existing `_OBJC_BARE_MACRO_LINE_RE` does and is the safe shape.

Tests: `tests/unit/ingestion/parser/test_objectivec.py` — a property and a method carrying a trailing `AF_API_AVAILABLE(...)`, asserting the file parses without the cascade and that a call inside the method is attributed to the method rather than `__module__`.

sloemo01:
## What

Fixes #2299. A call-shaped availability macro trailing a declaration cascaded an Objective-C file into hundreds of ERROR nodes, and the calls after it lost their real caller.

## Root cause

`prepare_objectivec_source` blanks a fixed set of **whole-line** macros (`NS_ASSUME_NONNULL_BEGIN`/`END`, `NS_HEADER_AUDIT_*`, and friends). A call-shaped attribute macro is not a whole line. It is the last token of a declaration the grammar otherwise reads fine:

```objc
@property (readwrite, nonatomic, copy) ... taskDidFinishCollectingMetrics
    AF_API_AVAILABLE(ios(10), macosx(10.12), watchos(3), tvos(10));
```

The grammar has no rule for it, so it reads as an unclosable declarator or attribute list. Error recovery then fails to resync at the terminating semicolon and the damage runs past that declaration into the rest of the file.

That last part is why this matters beyond a parse statistic. Every call inside the affected method bodies has no enclosing function to attribute to, so call resolution credits them to the per-file `__module__` node instead of the method that really makes the call. The caller end of those edges was wrong, not missing.

## Changes

A new `_OBJC_TRAILING_MACRO_RE` blanking the attribute macros (`AF_API_AVAILABLE`, `API_AVAILABLE`, `API_DEPRECATED`, `NS_SWIFT_NAME`, `NS_AVAILABLE`, `NS_DEPRECATED`) through the existing `_blank_matches` helper, so every byte offset outside a match survives and line/column numbers stay valid.

It is enumerated by name rather than a general `IDENT(...)` rule because the grammar cannot tell an attribute from a call: `NS_ENUM(NSInteger, Kind)` after a `typedef` and `NSLog(@"%@", x)` both have that shape, and blanking either would corrupt real code. A match must also *terminate* a declaration (followed by the closing `;`, the `{` opening a method body, or end of line, since a definition may put its brace on the next line), which leaves the same name inside an argument list alone. The terminator itself is outside the match, so the declaration still ends where it did.

## Verification

Measured on the real `AFNetworking/AFURLSessionManager.m`, counting ERROR and missing nodes:

| | ERROR nodes |
|---|---|
| `main` | **857** |
| this PR | **4** |

I ran that measurement myself against the actual file rather than relying on the issue's estimate.

- `tests/unit/ingestion/parser/test_objectivec.py`: 46 passed, 7 of them new.
- Sabotage-verified: with `parser_helpers.py` reverted to `main`, 4 of the new tests fail (both cascade tests, the offset/blanking test, and the next-line-brace test). Restored, all 46 pass.
- `tests/unit/ingestion/`: 3289 passed, 5 skipped, 2 xfailed with the known-unrelated rebase test deselected.
- `ruff check .`: clean.


RaghavChamadiya (approved):
Thanks for keeping this byte-preserving. Blanking through `_blank_matches` holds the length exactly, and the zero-width lookahead on the terminator means the `;` or `{` is never consumed, so every other symbol range in an affected file keeps its offsets. The test asserting equal length and equal newline count is the one I would have asked for.

Keeping the macro list an enumerated allowlist instead of a general call-shaped rule is also the right instinct, since a blanket pattern would have taken `NS_ENUM` with it.

PR #2331, opened 2026-09-17
fix(ingestion): index TypeScript class fields holding a function

### LINKED ISSUE #2302: [Bug] static create = (...) => {} is invisible: class fields are captured for their type only
**Summary**

`queries/typescript.scm` has exactly one `public_field_definition` pattern, and it captures the field's **type annotation**, never the field itself.

`packages/core/src/repowise/core/ingestion/queries/typescript.scm:276-278`

```scheme
; Class field types: class C { f: Field }
(public_field_definition
  type: (type_annotation (_) @param.type))
```

So a class property holding a function, the common `static create = (...) => {}` factory shape, produces no symbol. Nothing can call it, and it cannot be reported as unused.

**Mechanism**

Worth ruling out the obvious suspect: this is **not** the callable-ancestor filter that deliberately drops functions nested inside other functions (`parser.py:1360-1374`). Class bodies are exempt from that filter, which is why ordinary methods work. There is simply no pattern for the shape.

**Repro**

```ts
class Thing {
  static create = (name: string) => new Thing(name);
}
```

`Thing.create` appears in no symbol table.

**Impact**

A common TypeScript idiom, particularly for static factories and for class properties holding arrow-bound handlers.

Under-reporting. Previously observed as part of a wider extraction gap where two competing tools held a symbol and we held none.

**Done looks like**

A `public_field_definition` pattern binding `@symbol.def` when the value is an `arrow_function` or `function_expression`, leaving the existing type-annotation capture intact.

Gate on the value being a function. A pattern matching every class field would mint symbols for ordinary data properties, which is a different and much larger change.

This changes no resolution tier: it adds a callee identity that calls can already reach by name once it exists.

Tests: the TypeScript extractor tests — a `static create = (...) => {}` and a non-static `handler = () => {}` asserting both become symbols, plus a plain `count = 0` asserting it does not.

sloemo01:
## What

Fixes #2302. A class field holding a function produced no symbol, so the common static-factory idiom was invisible to every consumer.

## Root cause

`typescript.scm` had exactly one `public_field_definition` pattern, and it captured the field's **type annotation**, never the field itself:

```scheme
; Class field types: class C { f: Field }
(public_field_definition
  type: (type_annotation (_) @param.type))
```

So `static create = (name: string) => new Thing(name)` produced nothing. Nothing could call it, and it could not be reported as unused.

Worth ruling out the obvious suspect, as the issue asked: this was not the callable-ancestor filter at `parser.py:1360-1374`. `_has_callable_ancestor` returns False for the field node (its ancestors are `class_body` and `class_declaration`, neither of which maps to a callable kind), which is why ordinary methods work. There was simply no pattern for the shape.

## Changes

A `public_field_definition` pattern binding `@symbol.def` and `@symbol.name`, gated on the value being an `arrow_function` or `function_expression`, with `public_field_definition` mapped to the `"method"` kind.

The gate is the point. A pattern matching every class field would mint symbols for plain data properties, which is a larger and different change. I verified the gate independently on the current tree:

| source | symbols produced |
|---|---|
| `static create = (name) => new Thing(name)` | `create` (method) |
| `handler = () => {}` | `handler` (method) |
| `run = function () {...}` | `run` (method) |
| `count = 0` | **none** |
| `label: string = 'x'` | **none** |

The existing type-annotation pattern is untouched, so `f: Field` still emits its `field_type` reference.

The pattern accepts both arrow parameter shapes, function expressions, async and generic arrows, and an optional accessibility modifier, so `private h = () => {}` does not read as public. Because the node type is now a callable kind, it also joins the callable-ancestor filter, which means a method inside a class expression assigned to a field no longer lands as a member of the outer class. That is pinned as a test, and it matches what the module-level form (`const K = class Foo { m() {} }`) has always done.

## Tests

`tests/unit/ingestion/parser/test_typescript.py`: 55 passed, 10 of them new.

Sabotage-verified: with both source files reverted to `main`, 8 of the 10 new tests fail. The 2 that still pass are negative assertions that hold either way (`count = 0` producing no symbol), which is the correct behavior for a control.


RaghavChamadiya:
Holding this one, and the reason is not the code. @rehan6025 claimed #2302 at 11:05 on the 17th, before this went up, and #2302 is labelled `good first issue`, which I keep for people making an early contribution. I have assigned it to them. Longer note on #2345.

Leaving this open, not closing it. If they go quiet I will come back to you.

sloemo01:
Closing this, since #2302 was already claimed by @rehan6025 before this branch went up. The issue stays with them, so leaving this open only parks a review that is not going to happen.

PR #2332, opened 2026-09-17
fix(server): clear the job total when a phase reports no count

### LINKED ISSUE #2175: [Bug] Job progress denominator goes stale when a phase reports an unknown total (241 / 5 pages)

sloemo01:
## What

Fixes #2175. A phase that reports no item count kept the previous phase's denominator, so the UI showed readouts like `241 / 5 pages`.

## Root cause

`JobProgressCallback.on_phase_start` resets the phase-scoped numerator and assigns the phase's optional total:

```python
self._completed = 0
self._total = total          # may be None
```

`_async_update` then passed both fields straight through, and `update_job_status` treats `None` as "leave unchanged". A phase reporting no total therefore kept the previous phase's denominator on the row while the numerator climbed from zero. The two numbers described different units of work.

## Changes

Two commits, because the bug had a data half and a visible half.

**Server:** `_async_update` now writes an explicit `0` for an unknown total instead of passing `None`. An explicit `0` was already distinguishable from "leave unchanged" in `update_job_status` (only `None` skips the assignment), so no signature changed. The contract is now documented at the writer.

**Web:** `LiveJobProgress` interpolated both fields unconditionally, so a cleared denominator still rendered as `241/0 pages`. The text is now guarded on `total > 0`, the same way the percentage beside it already was, and an unknown total reads as a running count with no fraction.

I added the web half because without it the fix would have swapped one bad readout for another. The issue scoped the frontend out, so if you would rather keep this to the server change, the second commit drops off cleanly on its own.

## Tests

**Server** (`tests/unit/server/test_job_executor.py`, `tests/unit/persistence/test_crud.py`): 60 passed. A known-total phase still writes its denominator; an unknown-total phase leaves `total_pages` at 0 with `completed_pages` counting the phase; one CRUD test pins the `None`-vs-`0` distinction.

Sabotage-verified: with `job_executor.py` reverted to `main`, `test_unknown_total_phase_clears_the_stale_denominator` fails with `assert 5 == 0`.

**Web** (`packages/web/src/components/jobs/live-job-progress.test.tsx`): 2 passed. Sabotage-verified by reverting the component to `main`, which renders `241/0 pages` and fails the assertion.

`ruff check .` is clean and the wider server suite (`tests/unit/server/` plus `tests/unit/cli/test_progress_non_tty.py`) is 3345 passed.


RaghavChamadiya:
Holding this one, and of the set it is the clearest case. On #2175 @jgt87 filed the issue and explicitly handed it to @Aman-goel-04, I assigned it to them on the 14th, and their PR #2308 was already open when this went up. Longer note on #2345.

I am reviewing #2308 in this session. Leaving this open until that resolves, not closing it.

RaghavChamadiya:
Following up as promised: @Aman-goel-04's #2308 is merged, so #2175 is fixed on main and this one is now redundant. Their version is the one I asked for on the issue, a single line at `job_executor.py:253` that leaves `update_job_status`'s `None` semantics alone, so the three callers that depend on "leave unchanged" are untouched.

Closing this is your call, not mine. Sorry it cost you the work.

sloemo01:
Main already carries this fix, so I am closing rather than rebasing a duplicate.

`job_executor.py` now writes `total_pages=self._total if self._total is not None else 0`, which is exactly what this branch did, and `tests/unit/server/test_job_executor.py` has its own case for #2175 (`test_phase_start_with_unknown_total_clears_stale_denominator`, asserting `job.total_pages == 0` after a phase starts with an unknown total). The UI side is there too: `live-job-progress.tsx` reads `total > 0` before computing a percentage, so a zero denominator no longer renders.

That leaves nothing unique here. The only piece this branch had that main does not is a three-line comment on the None-versus-zero contract in `update_job_status`, and main's code already reads that way on its own.

The conflict is what prompted the check, and it was the useful signal: the same lines were changed on both sides.

PR #2333, opened 2026-09-17
fix(esm): read destructured dynamic imports as named bindings

### LINKED ISSUE #2230: Bug: Dynamic ESM Destructuring Emits Wildcard Instead of Named Bindings
## Describe the Bug

When a TypeScript or JavaScript file uses a dynamic ESM import with destructuring:

```typescript
const { fn, calculate } = await import('./mod');
```

the parser emits `imported_names = ["*"]` (namespace wildcard) and `bindings = []`
instead of the correct `imported_names = ["fn", "calculate"]` with matching `NamedBinding`
objects.

Because `"*"` tells the dead-code analyzer to treat **every exported symbol** in the
target file as live, any exports in `./mod` that are genuinely unused will never be
flagged. This creates **false negatives** in `unused_export` dead-code findings for
every module that is dynamically imported with destructuring.

---

## Steps to Reproduce

1. Add the following test to `tests/unit/ingestion/parser/test_typescript.py`:

```python
def test_dynamic_import_destructuring_currently_broken(parser: ASTParser) -> None:
    src = b"""
    async function load() {
        const { fn, calculate } = await import('./mod');
    }
    """
    fi = _make_file_info("src/app.ts", "typescript")
    imp = next(i for i in result.imports if i.module_path == "./mod")

    assert imp.imported_names == ["fn", "calculate"]
    assert {b.exported_name for b in imp.bindings} == {"fn", "calculate"}
```

2. Run:
```bash
pytest tests/unit/ingestion/parser/test_typescript.py::test_dynamic_import_destructuring_currently_broken -v
```

3. See assertion failure.

---

## Expected Behavior

For `const { fn, calculate } = await import('./mod')`, the parser should emit:

```python
imported_names = ["fn", "calculate"]
bindings = [
    NamedBinding(local_name="fn",       exported_name="fn"),
    NamedBinding(local_name="calculate", exported_name="calculate"),
]
```

This is identical to the behaviour already produced for the static equivalent:

```typescript
import { fn, calculate } from './mod';
// → imported_names = ['fn', 'calculate']  ✅
```

and for CommonJS destructuring:

```javascript
const { fn, calculate } = require('./mod');
// → imported_names = ['fn', 'calculate']  ✅
```

---

## Actual Behavior

```
FAILED — AssertionError: Expected ['fn', 'calculate'] but got ['*'].
```

The parser hits the hardcoded wildcard branch in `parser.py:L1918–1935`:

```python
# parser.py:L1928 — always emits wildcard, never reads the destructuring LHS
imported_names=["*"],
bindings=[],
```

The tree-sitter query in `queries/typescript.scm` captures the inner `call_expression`
(`import('./mod')`) as `@import.statement`. The destructuring LHS `{ fn, calculate }`
lives in the **parent** `variable_declarator` node, which the parser never inspects.

### Downstream effect in the UI

On the File Detail page's **"Depends on"** tab, the edge to `./mod` shows `*` under
the file name instead of listing specific imported symbols. More critically, no
`unused_export` findings are raised for genuinely unused exports in `./mod`, even
when only one symbol out of many was destructured.

---

## Environment

- **OS:** Windows 11
- **Python version:** 3.12
- **Repowise version:** 0.1.2 (`repowise --version`)
- **Installation method:** pip (monorepo dev install via `.venv`)

---
### Fix location & Scope

The fix is a **parent-node walk** inside `parser.py:L1918–1935`. After detecting the
dynamic `import()` call, walk up through any `await_expression` to find a
`variable_declarator`; if its `name` field is an `object_pattern`, extract the
property names exactly as `_extract_require_bindings()` already does in
`extractors/bindings/ts_js.py:L11–67`. This must also correctly handle aliased destructuring (e.g., `const { fn: myFn } = await import(...)`).

#### ⚠️ Intentional Wildcard Exclusions

While destructuring should extract named bindings, the following forms **MUST intentionally remain as wildcards (`["*"]`)**:
1. **Bare module assignments** (`const mod = await import('./mod')`)
2. **Lazy route callbacks** (`() => import('./views/Profile')`)

In these cases, the developer is deliberately capturing or forwarding the entire module namespace object at runtime. Emitting a wildcard `["*"]` is strictly required here to prevent the dead-code analyzer from aggressively generating false positives.

#### Out of Scope
Forms **not** covered by the simple fix (out of scope for this issue):
- `import('./mod').then(({ fn }) => ...)` — binding is in a callback parameter
- `const { fn } = await import('./mod').then(x => x)` — chained call in between

sloemo01:
## What

Fixes #2230. A destructured dynamic ESM import was recorded as a namespace wildcard, so genuinely unused exports in the target module were never flagged.

## Root cause

`const { fn, calculate } = await import('./mod')` produced `imported_names = ["*"]` with an empty bindings list.

The `typescript.scm` dynamic-import pattern captures the inner `import()` call, while the destructuring target lives on the parent `variable_declarator`, one `await_expression` up. Nothing inspected that parent, so every dynamic import fell through to the wildcard.

The `"*"` sentinel tells the dead-code analyzer that every export of the target file is live. Any genuinely unused export in a module reached this way could never be reported: a false negative in `unused_export` findings for every module dynamically imported with destructuring.

## Changes

`dynamic_import_bindings()` walks one `await_expression` up from the captured call to the enclosing `variable_declarator` and, when its `name` field is an `object_pattern`, extracts the property keys as `imported_names` with matching `NamedBinding` objects (local alias on `local_name`, source-side name on `exported_name`), exactly as the CommonJS `require()` path already did.

The `object_pattern` reader is factored into `_bindings_from_object_pattern()` so both paths spell a destructuring clause identically rather than growing two copies.

## Behaviour, checked on the current tree

| source | result |
|---|---|
| `const { fn, calculate } = await import('./mod')` | `["fn", "calculate"]` |
| `const { a, b: c } = await import('./mod')` | `["a", "b"]` with `c -> b` |
| `const mod = await import('./mod')` | `["*"]` (unchanged) |
| `() => import('./views/Profile')` | `["*"]` (unchanged) |

The two wildcard cases are deliberate. A bare binding captures the whole namespace object, and a lazy route callback forwards it to the runtime router, so in both the developer does consume everything the module exports and the wildcard is the honest record. Emitting named bindings there would make the analyzer report exports the file really uses.

Ceiling: one hop through `await_expression` only. A chained call in between (`await import('./m').then(x => x)`) and a binding inside a callback parameter (`import('./m').then(({ a }) => ...)`) stay on the wildcard, matching the shapes the issue scopes out.

## Tests

`tests/unit/ingestion/parser/test_typescript.py` plus a new `tests/unit/ingestion/test_esm_dynamic_import_destructuring.py`: 53 passed. The new file proves the downstream `unused_export` effect the issue describes, not just the parse shape.

Sabotage-verified with a file swap of both source files back to `main`: 4 tests fail (`assert ['*'] == ['fn', 'calculate']`), 49 pass. The two wildcard-exclusion tests pass either way, which is correct for controls that pin behaviour which must not change.

`ruff check .` is clean, and the wider `tests/unit/ingestion/` (3290 passed) plus `tests/unit/dead_code/` and `tests/integration/test_ts_dead_code.py` (410 passed) are green.


RaghavChamadiya:
Holding this one. @akshatmalik-bruh claimed #2230 on 12 September and I confirmed the finding to them on the 14th, so that one has been spoken for over a week. Longer note on #2345.

sloemo01:
Closing this, since #2230 was already claimed by @akshatmalik-bruh before this branch went up. The issue stays with them, so leaving this open only parks a review that is not going to happen.

PR #2334, opened 2026-09-17
fix(providers): make codex_cli subprocess concurrency configurable

### LINKED ISSUE #2266: [Feature] codex_cli: configurable subprocess concurrency (hard Semaphore(1) serializes generate)
## Problem

`core/providers/llm/codex_cli.py` hard-codes `asyncio.Semaphore(1)` for the codex subprocess, so `repowise generate --concurrency 4` with `codex_cli` still writes pages strictly one at a time (`ps` shows a single `codex exec` at any moment). A 104-page `generate` on our repo took ~95 min serial vs ~18 min after raising the semaphore to 4. `claude_cli.py` already has `_DEFAULT_CONCURRENCY = 4` with a `REPOWISE_CLAUDE_CLI_CONCURRENCY` override, with a comment explaining exactly this trade-off.

## Proposed Solution

Mirror `claude_cli`: `REPOWISE_CODEX_CLI_CONCURRENCY` env, default 4:

```python
self._subprocess_semaphore = asyncio.Semaphore(_resolve_concurrency())  # was Semaphore(1)
```

We ran this patched locally for ~500 pages across 9 repos with no rate-limit errors from Codex.

## Alternatives Considered

Respect `--concurrency` directly for CLI providers — but a separate knob is better since the seat limit is per account, not per run.

## Additional Context

Repowise 0.50.0, Codex CLI 0.154.0, macOS.

sloemo01:
## What

Fixes #2266. `codex_cli` hard-coded `asyncio.Semaphore(1)`, so `repowise generate --concurrency 4` still wrote pages strictly one at a time.

## Root cause

`CodexCliProvider._get_semaphore` built its semaphore from the literal `1`, so only one `codex exec` was ever in flight. Each page is a full CLI process, which is why a 104-page generate took roughly 95 minutes serial.

`claude_cli` had already solved the same trade-off. This mirrors it.

## Changes

`_resolve_concurrency()` reads `REPOWISE_CODEX_CLI_CONCURRENCY`, defaults to 4, warns and falls back to the default when the value is not an integer, and clamps to a minimum of 1. `_get_semaphore()` builds from that instead of the literal.

As in `claude_cli`, the variable is a true override rather than a clamp: it can raise the fan-out above 4 as well as lower it, since the seat limit is per account and only the operator knows which plan they are on. The implementation is structurally identical to `claude_cli._resolve_concurrency`, including the comment discipline, so the two providers read the same way.

One stale docstring line claiming the provider "serializes subprocess calls by default" was directly contradicted by this change and is corrected. The knob is documented in `docs/agent/CODEX.md` and `website/codex.md`, next to the existing `claude_cli` concurrency notes.

## Tests

`tests/unit/test_providers/test_codex_cli_provider.py`: 23 passed.

The old `test_generate_serializes_subprocess_calls` asserted `max_active == 1`, which pinned the bug rather than the contract, so it is rewritten as `test_generate_respects_configured_concurrency` (env=2, three calls, expects `max_active == 2`). Three `_resolve_concurrency()` unit tests cover blank, invalid, and below-one values.

Sabotage-verified by reverting only the semaphore line: `test_generate_respects_configured_concurrency` fails with `assert 1 == 2` and `test_generate_fans_out_to_the_default_concurrency` with `assert 1 == 4`.

`ruff check .` is clean, and `tests/unit/test_providers/` plus `tests/providers/` are 369 passed.

## Not verified

No `codex` CLI is installed here, so the fix is verified against mocked subprocesses. I did not observe four live `codex exec` processes running concurrently, and the issue's claim that ~500 pages ran without rate-limit errors at fan-out 4 is the reporter's observation, not reproduced here.


sloemo01:
One red leg, and it looks like a flake rather than this change.

`tests/unit/distill/test_rewrite_perf.py::test_a_ledgered_invocation_stays_under_budget` asserts a wall-clock median under 200 ms and measured 208.8 ms on that runner. The same file passes locally in 2.59s (7 passed), the 3.11 and 3.12 legs are green, and the only source file this PR touches is `codex_cli.py`, which is not on that hook's import path.

I cannot retrigger the job from here (needs admin on this repo), so a re-run would settle it.

RaghavChamadiya (approved):
Thanks @sloemo01. Clamping to a minimum of 1 and warning on an invalid value rather than falling over is the right handling, since the realistic bad input here is someone exporting `REPOWISE_CODEX_CLI_CONCURRENCY=0` and expecting it to mean "unlimited".

The red leg you flagged on 17 September was `test_rewrite_perf.py`, not your code. Your #2385 just merged and fixes that class of wall-clock flake at the source, so it should stop landing on unrelated PRs.

Non-blocking follow-up, already filed as #2266: `opencode.py:320` has the identical hardcoded `Semaphore(1)`. Same fix, different provider, and still unclaimed if you want it.

PR #2336, opened 2026-09-18
fix(godot): resolve uid:// references to scripts through their .uid sidecars

### LINKED ISSUE #2294: [Bug] GDScript uid:// references never resolve, so most Godot 4 scene dependencies are invisible
**Summary**

`resolve_gdscript_import` refuses every `uid://` reference outright.

`packages/core/src/repowise/core/ingestion/resolvers/gdscript.py:147-148`

```python
if raw.startswith(_UID_PREFIX) or raw.startswith(_USER_PREFIX):
    return _miss(raw, ctx)
```

Godot 4.4 addresses `[ext_resource]` entries by UID rather than by path, so this is not a corner case. Across `godot-demo-projects` there are **1,763** `uid://` references in `.tscn` files alone. Every one of them is a scene dependency the graph does not have.

The mapping it needs is sitting next to each file. Godot writes a `<file>.uid` sidecar containing exactly the uid and nothing else:

```
$ cat 2d/bullet_shower/bullets.gd.uid
uid://e3fsq0i746o1
```

**Mechanism**

Two things line up, and both are deliberate in isolation.

1. The resolver treats `uid://` the same as `user://`, a runtime-only scheme that genuinely cannot be resolved statically. `uid://` can be, because the sidecar is in the repository.
2. The sidecars are excluded from indexing, so nothing ever reads them.

`packages/core/src/repowise/core/ingestion/languages/specs/godot_resource.py:87`

```python
blocked_extensions=(".import", ".uid"),
```

That exclusion is right for what it was for, keeping one-line metadata files out of the wiki and the dead-code pass. It also means the uid map has no reader.

There is already a precedent for the scan this needs. `godot_project_root` (`resolvers/gdscript.py:97-110`) globs `project.godot` from the filesystem and caches it on the context, for exactly the reason that applies here, in its own words: the file "is not guaranteed to be among the indexed source files".

**Repro**

```
$ grep -rho 'uid://' test-repos/godot-demo-projects --include=*.tscn | wc -l
1763
```

Index `godot-demo-projects` and look for edges from any `.tscn` whose `[ext_resource]` uses a `uid://` id. There are none. A scene that references its script by uid has no edge to that script, so the script reads as unreferenced.

**Impact**

Every Godot 4.4 project, which is the version that made uid addressing the default. The effect is silent under-reporting: a missing edge, not a wrong one, so nothing downstream looks broken. It costs scene-to-script reachability, and a script reached only by uid can surface as dead code.

Unsized beyond the 1,763 references above. Nothing counts refused `uid://` references today, so there was no signal that this was happening.

**Done looks like**

`uid://abc123` resolves to the file whose sidecar contains that uid, and still misses cleanly when no sidecar matches.

The shape to follow is `godot_project_root`: glob `*.uid` once per build, cache the `uid -> repo-relative path` map on the context, and consult it in `resolve_gdscript_import` ahead of the `_miss` at line 148. A sidecar is `<file>.uid`, so the mapped path is the sidecar's own path with `.uid` stripped.

Two things worth deciding while fixing rather than after:

* Whether `.uid` should stay in `blocked_extensions`. It should. The files need reading, not indexing, and the glob reads them directly the same way `project.godot` is read.
* `user://` must keep refusing. It is a runtime path with no repository counterpart, and it shares the branch being changed.

Wrong-edge risk here is close to zero: uids are unique by construction, so a match is an identity, not a guess. If two sidecars somehow carry one uid, refuse rather than pick.

Tests: `tests/unit/ingestion/test_gdscript_resolver.py` — a fixture with a `.uid` sidecar asserting the reference resolves to the source file, one asserting a `uid://` with no matching sidecar still misses, and one asserting `user://` is unaffected.

sloemo01:
Closes #2294

## What was wrong

Godot 4.4 addresses resources by uid, so a script's `preload("uid://...")` or `extends` can name one instead of a `res://` path. Those references reached the resolver and fell through to `add_external_node`: the edge showed up in the graph with no target.

The uid to file mapping lives in the `<file>.uid` sidecar Godot writes beside each script. Those sidecars are one-line build artifacts, so the spec keeps them out of the index, and the resolver had nothing to read.

## What this does

Builds the uid map off the filesystem, the same way `project.godot` is discovered and for the same reason: the file carrying the answer is not an indexed source. One walk per build, cached on the resolver context.

Only scripts carry a sidecar. Scenes and `.tres` files keep their uid in their own `[gd_scene format=4 uid=...]` header, because Godot's text loaders declare custom uid support and `should_create_uid_file` is false for them. A scene uid therefore misses here and still reaches the graph through the `path=` Godot writes beside it on the `[ext_resource]` header.

When two sidecars claim the same uid, that uid is left out of the map entirely. This happens in real repos: `godot-demo-projects` ships two `main.gd` under different demos that share a uid. Both sidecars are equally its owner, so either choice is a guess, and a wrong edge is worse than a missing one.

## Verification

`test_gdscript_resolver.py` and `test_godot_project_awareness.py`: 87 passed.

Reverting `resolvers/gdscript.py` to main fails 5 of them. Restoring the file leaves the tree clean.

## Scope

The uid references this resolves come from script call sites, meaning `load`, `preload`, and `extends`. That is where a `uid://` string reaches the resolver at all. A `.tscn` or `.tres` names its dependencies with `path=` on its `[ext_resource]` headers, and the extractor reads that field, so scene dependencies already reach the graph without a uid lookup.

So this closes the script-side gap. The scene-side path was not broken, and the module docstring in `lightweight_imports/godot.py` carries the measurement behind that. This change updates that note rather than leaving the old one standing.


RaghavChamadiya (approved):
Refusing an ambiguous uid instead of guessing is what makes this safe to take. Two sidecars claiming one uid is a real shape in `godot-demo-projects`, and dropping the claim there means a wrong edge never gets minted; a uid pointing at an unindexed file missing cleanly has the same property. Leaving scene uids alone because Godot writes no sidecar for them, and `[ext_resource]` already carries `path=`, is the correct scope.

PR #2337, opened 2026-09-18
fix(claude_cli): count the whole prompt in token totals and record the run's cost

### LINKED ISSUE #2267: [Bug] claude_cli records input_tokens=2 per page; status/costs show zero tokens
## Describe the Bug

Pages written via `claude_cli` record `input_tokens=2` (output looks right, e.g. 2465), so `repowise status` shows `Total tokens 0` / `repowise costs` shows "No cost records found" for a wiki of 100+ Opus pages. `codex_cli` pages record plausible input counts (~22k) on the same repo.

`claude_cli.py` ~line 431 reads `usage.input_tokens` from the `claude -p --output-format json` payload. Claude Code's JSON puts most of the prompt under `cache_creation_input_tokens` / `cache_read_input_tokens` and only the uncached remainder in `input_tokens`; the code reads the cache fields (lines 435-436) but they don't appear to reach the persisted `input_tokens` / cost record.

## Steps to Reproduce

1. `repowise generate --provider claude_cli --model claude_cli/claude-opus-5 --page module_page:<x> --cascade none`
2. `sqlite3 .repowise/wiki.db "select model_name,input_tokens,output_tokens from wiki_pages where id='module_page:<x>'"`

## Expected Behavior

`input_tokens` ≈ prompt size (~20k for a module page), `repowise costs` lists the run ($0 for a seat, but token volume visible).

## Actual Behavior

`claude_cli/claude-opus-5|2|2465`; `costs` empty.

## Environment
- OS: macOS 26 (arm64)
- Python: 3.13
- Repowise: 0.50.0 (`uv tool install`)
- Claude Code CLI / Codex CLI 0.154.0 as subprocess providers

sloemo01:
Closes #2267

Two holes in the same provider.

## Token totals summed to zero

Claude Code reports only the uncached remainder of the prompt as `input_tokens`. A wiki page prompt is large and its stable prefix is cached, so the bulk arrives as `cache_creation_input_tokens` on the first page of a type and `cache_read_input_tokens` on every later page. Persisting `input_tokens` alone recorded 2 tokens against a roughly 20k token page, which is what made `repowise status` sum to zero for a whole wiki.

The three counters are disjoint: a prompt token is sent uncached, read from the cache, or written to it. Their sum is the prompt total the rest of the codebase means by `input_tokens`, so that is what is stored now. `cached_tokens` stays the read half on its own, per the provider docs. `uncached_input_tokens` is kept as its own field and added to the debug log, so the split stays visible.

## No cost rows

The provider never wrote an `llm_costs` row, so `repowise costs` printed "No cost records found" for a complete run.

The record is booked under the `claude_cli/` prefix, which the cost table prices at $0.00 (a seat is not API spend), while `repowise costs` still shows the run's token volume. It is awaited inline instead of fired off detached, because a detached write can still be in flight when the event loop is torn down after generation. `record()` swallows its own persistence errors, so generation is unaffected either way.

## Verification

`test_claude_cli_provider.py`: 39 passed. Full `tests/unit/test_providers/` sweep: 316 passed.

Reverting `claude_cli.py` to main fails 7 of the provider tests. Restoring it leaves the tree clean.


RaghavChamadiya (approved):
Summing the three dispositions into the persisted total is right, and keeping the uncached figure separately means nothing is lost by doing it. Booking the ledger row at $0.00 is the part I would have forgotten: a seat-billed run costs nothing and still needs to show its token volume, or `repowise costs` reports silence for work that happened.

One follow-up that is not yours and not a reason to hold this. The same undercount is live in the direct API provider: `anthropic.py:234` sets `input_tokens=response.usage.input_tokens` on its own while the two cache fields are collected separately at `:242-246`. That path is billed for real money, so the under-total there costs more than this one did. I am filing it separately.

PR #2338, opened 2026-09-18
fix(delete): delete a repository the configured database knows about

### LINKED ISSUE #2320: [Bug] repowise delete requires local wiki.db when using PostgreSQL
`repowise delete` cannot delete a repository index when Repowise is configured to use PostgreSQL and the repository does not have a local `.repowise/wiki.db`.

The command checks for the repository-local SQLite database before resolving the configured database URL, so it exits without looking up the repository in PostgreSQL.

## Steps to Reproduce

1. Configure Repowise with `REPOWISE_DB_URL` pointing to PostgreSQL.
2. Use a repository whose index is stored in the shared PostgreSQL database and which has no local `.repowise/wiki.db`.
3. Run:

   ```sh
   repowise delete --path /workspace/example-repository
   ```

4. Observe:

   ```text
   Database not found.
   ```

If the repository has no `.repowise` directory, the command instead reports:

```text
No .repowise/ directory found. Run 'repowise init' first.
```

## Expected Behavior

With `REPOWISE_DB_URL` configured for PostgreSQL, `repowise delete --path` should locate the repository by path in PostgreSQL, prompt for confirmation, and delete that repository's indexed data without requiring a local SQLite database.

## Actual Behavior

The command requires a repository-local `.repowise/wiki.db` before resolving or using the configured database URL. PostgreSQL-backed repositories therefore cannot be deleted through the CLI command.

## Environment

- OS: Linux
- Repowise version: 0.50.0 (`25c89ab307a02e2a89d548c0a1b46aae8d6d20fe`)
- Installation method: Repowise installation or official container image
- Database: PostgreSQL via `REPOWISE_DB_URL`

## Additional Context

The failure is specific to the delete command's local SQLite existence checks; the same PostgreSQL database can contain a valid repository registration and indexed data.

Verified against upstream `v0.50.0` at commit `25c89ab307a02e2a89d548c0a1b46aae8d6d20fe`. The guard appears in `packages/cli/src/repowise/cli/commands/delete_cmd.py` before the command calls `get_db_url_for_repo()`.

There is a second usability issue in the same command: `--path` selects the database URL, but when no repository ID is supplied it still displays every repository and prompts for a numbered selection. The documented `--path` behavior should select the repository whose `local_path` matches the supplied path.

The v0.50.0 integration tests already cover the analogous configured-database behavior for `status` and `doctor`, but do not cover `delete`. A regression test should exercise the PostgreSQL configuration directly, then assert that delete reaches the configured database and removes the selected repository.

## Suggested Fix Shape

Resolve the database URL before requiring `.repowise/wiki.db`. For a shared PostgreSQL database, look up the repository by its normalized local path and delete its indexed records through the configured session/database. Preserve the existing local SQLite behavior when no shared database is configured.

When `--path` is provided, use the normalized path to select the matching `Repository` row rather than treating it only as a database-location hint.

sloemo01:
Closes #2320

## What was wrong

`repowise delete` opened with two checks written for the repo-local SQLite default: that a `.repowise/` directory exists, and that `wiki.db` lives inside it. With a shared store configured (`REPOWISE_DB_URL` pointing at PostgreSQL) neither is true, so the command printed "No .repowise/ directory found" and refused to delete a repository the configured database had a row for.

`--path` had the mirror problem. It was documented and treated as a hint for locating the database rather than as a way to name the repository to delete, so passing a checkout path did not select its row.

## What this does

The two `.repowise/` checks now run only when no database URL is configured, so behaviour and messages are unchanged for the default case.

`--path`, or the current directory, names the repository: the row whose `local_path` matches is selected. Both sides are resolved before comparing, so a symlink or a relative argument still matches, and a stored path that no longer exists compares as a string instead of raising. When nothing matches and no path was given, every known repository is listed for numbered selection.

## Verification

`tests/integration/test_cli.py -k delete`: 4 of the new `TestDeleteWithConfiguredDb` cases fail with `delete_cmd.py` reverted to main. Restoring it leaves the tree clean.

One limitation worth stating plainly: the new `tests/integration/test_delete_cmd_postgres.py` skips here because no PostgreSQL server is available, so the PostgreSQL path itself is not exercised in this environment. The configured-database logic is covered by `TestDeleteWithConfiguredDb` against a fake database. The Postgres file only becomes meaningful with a real server behind it, and it is worth running that way before merge.


RaghavChamadiya (approved):
Gating the local-file checks behind `db_configured()` is exactly the right shape, and reusing the existing helper instead of writing a third way to ask the same question is why this reads as a small change.

The `--path` half is the part I looked at hardest, since a selection bug here deletes the wrong repository. It holds up: `_same_local_path` requires resolved-path equality, a non-match prints and returns without touching anything, an explicit positional `repo_id` still wins, and the numbered prompt is unchanged when neither is given.

Noting your own caveat plainly: the PostgreSQL dialect itself is not exercised here, because the integration job never runs on a pull request in this repo. The configured-database logic is covered against a shared SQLite store, which is the most CI can do. Taking it on that basis.

PR #2339, opened 2026-09-18
fix(init): stop --force in a worktree from becoming a paid upgrade

### LINKED ISSUE #1482: [Bug] `init --force` on a worktree silently becomes a paid `update --full` (model regeneration)
## Describe the Bug

`--force` is documented as "Regenerate all pages, ignoring existing" and is the free, no-model re-index escape hatch. When init runs in a **seedable worktree** (a detected linked worktree, or `--seed-from`), it delegates to `run_update(... full=force ...)` instead of doing a free re-index.

## Steps to Reproduce

1. In a git repo, create a linked worktree: `git worktree add <path>`
2. Run `repowise init --force` inside the worktree **without** a provider configured.
3. Trace: `detect_worktree_base` finds the base (init_cmd/command.py:802-809), `seed_index_from_base` succeeds, then `run_update(full=force=True)` is called (command.py:841).
4. In `run_update`, the `if full:` branch (update_cmd/command.py:666) calls `upgrade_to_full` (upgrade_flow.py:413-537), which calls `resolve_provider` (upgrade_flow.py:445) → **"No provider configured" error** (or, with a key set, spends money regenerating the *entire* wiki with the model).

## Expected Behavior

`init --force` should re-index/regenerate pages in the mode the user invoked (free template wiki or existing mode), regardless of worktree seeding. Only `update --full` should trigger model regeneration.

## Actual Behavior

On a seedable worktree, `init --force` routes into a paid, whole-repo LLM regeneration. With no provider configured the command aborts outright. `--index-only --force` is swallowed the same way.

## Environment

- OS: macOS 14
- Python version: 3.13
- Repowise version: 0.41.0

## Additional Context

Code path: `init_cmd/command.py:841` (`full=force,`) → `update_cmd/command.py:666-692` (full branch before lock) → `upgrade_flow.py:445` (`resolve_provider` raises without a key) and `upgrade_flow.py:506-527` (full regeneration + `save_state(... full_index=True)`).

The lock comment at update_cmd/command.py:896-906 also confirms the `full` branch never acquires the update lock (see related issue for `--full` lock bypass).

No existing test covers the worktree `--force` combination.

sloemo01:
Closes #1482

## What was wrong

Two mechanisms collided in a linked worktree.

`init` seeds a worktree's index from its main checkout, because indexing from scratch is the wasteful path. `--force` is the request to do that re-index anyway, so the two should not combine. The seed block then delegated to `run_update` and passed `full=force`, and `full` is not "re-index": it is `update --full`, the upgrade that resolves a provider and regenerates the whole wiki with a model.

So `init --force` in a seedable worktree either died with "No provider configured" or spent money the user never asked for. `--index-only --force` was swallowed the same way.

## What this does

When `--force` is passed in a seedable worktree, the seed is dropped and the ordinary init path below runs. That is what `init --force` already does outside a worktree: re-index and regenerate every page in the mode the user invoked, free when that mode is structural. A line is printed saying the seed was not used, so the behaviour change is visible rather than silent.

`update --full` stays the only paid path, and the delegate no longer receives `full=force` at all.

## Verification

`tests/integration/test_cli.py -k "force or worktree"`: 8 passed. Reverting `init_cmd/command.py` to main fails 3 of the new `TestWorktreeInitForce` cases.

The wider `tests/unit/cli` suite is 2566 passed, 1 failed. That failure is `test_agent_target_baseline.py::test_integration_writes_match_baseline`, and it is pre-existing and platform-specific rather than related to this change. It fails identically on an unrelated branch:

```
'Library/Application Support/Claude/claude_desktop_config.json'
  != 'AppData/Roaming/Claude/claude_desktop_config.json'
```

The baseline was recorded on Windows, and macOS writes a different path.


RaghavChamadiya (approved):
Thanks @sloemo01 for the spy test asserting the paid path is never reached. That is the test that matters here: the failure mode is not a crash, it is quietly spending money, and an assertion that `upgrade_to_full` is never called is the only thing that actually pins it.

Verified the cause on current main. `init_cmd/command.py:993` passes `full=force` into `run_update`, so `init --force` in a worktree becomes `update --full`, which resolves a provider and does a paid regeneration. In a seedable worktree that either died with "No provider configured" or silently billed, neither of which is what `--force` means on `init`.

Also good that the no-provider case is covered, since that was the visible half of the bug and the one people actually reported.

PR #2340, opened 2026-09-18
fix(coverage): report LF:0 as not applicable instead of 0% covered

### LINKED ISSUE #2193: [Bug] A file with zero coverable lines (LF:0) is ingested as 0% coverage, so coverage_gradient imputes what it documents it never will
## Describe the Bug

A file with **zero coverable lines** (`LF:0`) is ingested as `line_coverage_pct = 0.0` instead of "not applicable". Downstream that is indistinguishable from a file that has coverable lines of which none were hit, so `coverage_gradient` and `untested_hotspot` both fire on a file that contains no executable code at all.

This bypasses the guarantee `coverage_gradient` documents for itself:

> **Absent coverage ≠ zero coverage.** When no coverage report was ingested (`line_coverage_pct is None`) the biomarker is silent - it never imputes uncovered for missing data.

The guard is `if cov is None: return []`. It holds when no report was ingested, but a `0/0` record hands the detector a hard `0.0`, so the biomarker imputes exactly what the docstring promises it never will.

The same function already handles the analogous case correctly one branch over — `branch_coverage_pct` becomes `None` when there are no branches, while `line_coverage_pct` becomes `0.0` when there are no lines:

```python
# core/analysis/health/coverage/lcov.py:46-51
line_pct = (hit / total * 100.0) if total else 0.0     # 0 lines  -> "0% covered"
branch_pct: float | None
if has_branches and branches_found:
    branch_pct = branches_hit / branches_found * 100.0
else:
    branch_pct = None                                   # 0 branches -> "not applicable"
```

## Steps to Reproduce

Minimal lcov with one type-only file (`LF:0`) and one genuinely uncovered file:

```
TN:
SF:src/types.ts
FNF:0
FNH:0
LF:0
LH:0
BRF:0
BRH:0
end_of_record
TN:
SF:src/real.ts
FNF:2
FNH:0
DA:1,0
DA:2,0
LF:2
LH:0
BRF:0
BRH:0
end_of_record
```

```python
from repowise.core.analysis.health.coverage.lcov import parse_lcov

for f in parse_lcov(open("repro.lcov").read()).files:
    print(f"{f.file_path:16} coverable_lines={f.total_coverable_lines}  "
          f"line_pct={f.line_coverage_pct!r:8} branch_pct={f.branch_coverage_pct!r}")
```

## Expected Behavior

The two records are not the same fact and should not collapse to the same value. A file with nothing to cover should be reported as unknown/not-applicable (`None`), the way `branch_coverage_pct` already is, so the coverage biomarkers stay silent on it.

## Actual Behavior

```
src/types.ts     coverable_lines=0  line_pct=0.0      branch_pct=None
src/real.ts      coverable_lines=2  line_pct=0.0      branch_pct=None
```

Both report `0.0`. `total_coverable_lines` does carry the distinction, but nothing downstream consults it before applying the coverage deductions.

**On a real repo.** A TypeScript barrel of 28 `export interface` / `export type` declarations and no executable statement (`LF:0 FNF:0 BRF:0`, emitted by `@vitest/coverage-v8`) scored:

| | |
|---|---|
| `coverage_gradient` | −2.00 · "100% of lines uncovered (0% line coverage)" |
| `untested_hotspot` | −2.00 · critical · "Hotspot with 0% line coverage and 21 dependents" |
| score | 2.5 |
| `weighted_deficit` | 2079 — **15.8% of the whole repo gap**, ranked #2 in `high_leverage_files` |

So `fix_first`'s runner-up was a file with no code to fix. Excluding the module from the coverage report moved the repo from 2 alert files to 1 and the average from 8.04 to 8.12, which is the same bug seen from the other side.

## Root cause (and why it is not a one-line fix)

`FileCoverage` cannot express the distinction for lines, but can for branches:

```python
# core/analysis/health/coverage/model.py:19-21
file_path: str
line_coverage_pct: float          # no None
branch_coverage_pct: float | None  # has None
```

The three `if total else 0.0` sites are a consequence of that asymmetry, and all four parsers share it:

- `coverage/lcov.py:46` — `(hit / total * 100.0) if total else 0.0`
- `coverage/cobertura.py:96` — same expression
- `coverage/clover.py:77` — `(len(covered) / total_n * 100.0) if total_n else 0.0`
- `coverage/repowise_json.py:111` — clamps `pct` with no zero-coverable guard

A fix presumably widens the field to `float | None` and returns `None` when there is nothing coverable, then lets the existing `cov is None` guards in `coverage_gradient` / `coverage_gap` / `untested_hotspot` do their job unchanged.

Worth deciding explicitly whether "no coverable lines" should be `None` (unknown) or `100.0` (vacuously covered). `None` seems right and matches the branch behaviour, but it is a product call — an lcov record whose producer genuinely instrumented the file and found nothing to instrument is arguably a different statement from no record at all.

## Environment

- OS: macOS 27.0.0 (Darwin, arm64)
- Python version: 3.14.6
- Repowise version: 0.49.0 (also present in 0.48.0)
- Installation method: pipx
- Coverage producer: `@vitest/coverage-v8` 4.x, `reporter: ['text', 'lcov']`

## Additional Context

The distinction is already available at the point of the bug — `total_coverable_lines` is set in the same `flush()` that computes `line_pct` — so the parsers have everything they need to answer "not applicable" without consulting the source tree.

sloemo01:
Closes #2193

## What was wrong

A coverage report that instruments a file and finds no executable statement emits `LF:0`. That answers a different question from `LH:0 LF:2`, which is a file with coverable lines and none of them exercised. Both were stored as `0.0`, so every consumer reading a percentage treated "nothing to cover" as "nothing covered".

That also broke a documented promise. `coverage_gradient`, `coverage_gap`, and `untested_hotspot` carry `cov is None` guards so absent coverage is never imputed as uncovered, and with `0.0` those guards could not tell the two cases apart.

## What this does

`line_coverage_pct` is `float | None` now, and a file with nothing coverable stores `None`. This mirrors `branch_coverage_pct`, which has always been nullable and has always reported NULL when there are no branches.

The change runs through each layer that carries the value:

- **Parsers** (`clover`, `cobertura`, `lcov`, `repowise_json`, `discovery`) emit `None` for a zero-`LF` record instead of `0.0`.
- **Persistence**: migration `0066` makes the column nullable, with the model and CRUD following.
- **Server**: the health route skips `None` when accumulating a coverage ratio, and sorts with a `None`-aware key so those rows land last. One latent bug came with that: sorting on `line_coverage_pct` directly raises `TypeError` once `None` is a possible value.
- **UI**: the figure and the bar render an em dash for the not-applicable case, matching the other coverage surfaces, rather than a red 0.0% band.

## Verification

`test_coverage_zero_coverable_lines.py` (new, 726 lines): 18 passed. Reverting `analysis/health/coverage/` to main and re-running that same file fails 4 of them. Restoring leaves the tree clean at the same commit.

The migration uses `batch_alter_table`, which SQLite needs for a column alteration, and ships a matching `downgrade`.

The wider persistence, health, pipeline, server, and CLI suites ran at 8524 passed, 2 failed. Both failures are pre-existing and unrelated to this change. One is `test_agent_target_baseline.py::test_integration_writes_match_baseline`, whose baseline was recorded on Windows, and it fails identically on an unrelated branch. The other is `test_doctor_mcp_smoke.py::test_the_registered_command_is_the_one_launched`, which passes on its own and only fails under full-suite ordering. I confirmed the isolation and confirmed this change touches nothing on that path, but I did not pin down the ordering interaction itself.


RaghavChamadiya:
Holding this one. @modelpath-dev claimed #2193 on 10 September and confirmed on the 12th that they were proceeding with the `None` approach, so that one has been spoken for a week. Longer note on #2345.

sloemo01:
Closing this, since #2193 was already claimed by @modelpath-dev before this branch went up. The issue stays with them, so leaving this open only parks a review that is not going to happen.

PR #2342, opened 2026-09-18
docs(reference): add the API facts OpenAPI cannot carry

sloemo01:
Adds `docs/reference/API_REFERENCE.md`, cross-linked from `docs/README.md`.

The maintainer's comment on the issue narrowed this hard, and this follows his scope rather than the issue body. The endpoint reference is already generated and already served: `repowise serve` publishes `/openapi.json` and a Swagger UI at `/docs`, both derived from the route signatures and the Pydantic models. A hand-maintained table covering the same 33 router modules drifts the first time someone adds a query parameter, and it drifts silently because nothing tests prose against routes.

So the document carries only what OpenAPI cannot, in the order he put them:

- **Which endpoints spend money.** `POST /api/repos/{id}/full-resync`, the chat endpoint, and `refactoring/{id}/generate-code` make model calls. Nothing in the schema says so.
- **Authentication.** The `REPOWISE_API_KEY` bearer requirement, which endpoints enforce it, and what happens when the header is absent.
- **Error payload shape** and the response envelope conventions, stated once instead of inferred per route.
- **SSE streaming semantics** for `GET /api/jobs/{id}/stream` and the chat endpoint: event names, ordering, and how a client knows the stream ended.
- **Repeated conventions**, such as `fields=summary` versus `fields=full` on `/api/pages`, documented once.
- **A short orientation table**, resource area to router, so a reader knows which part of `/docs` to open.

Every statement is read out of the code rather than inferred, and the two corrections from his comment are applied: `MCP_TOOLS.md` lives at `docs/agent/`, and `docs/reference/` holds `CLI_REFERENCE.md`, `COMPUTED_GLOSSARY.md`, `CONFIG.md`, `TELEMETRY.md`, and `UPGRADING.md`.

Docs only. No code changes.


sloemo01:
Overlap worth knowing about before you spend time here: #2094 opened the same file, `docs/reference/API_REFERENCE.md`, for this issue on 3 September. You pushed back on its shape that morning, and it has been idle since, so this is not meant to race it. But you should have both in front of you when you pick one to read.

If #2094 is the one you would rather land, say so and I will close this.

PR #2344, opened 2026-09-18
feat(glossary): treat a user-authored glossary as the authority on naming

sloemo01:
Part of #2262.

Treats a glossary the team authored as the authority on naming, and exposes it to agents.

Repowise already mines a repository's house vocabulary deterministically and renders it as the onboarding Glossary page. That is the descriptive half: what the repo says its words mean. This adds the prescriptive half, so a glossary a person wrote down is the one canonical term and the rest of the pipeline follows it.

## What is in this PR

- A declared glossary is parsed and loaded, with the parser in `declared_glossary.py`.
- Declared terms enter the vocabulary path with `status="declared"`, and a declared term is never dropped for lack of corroboration in the code, which is the point of a person having written it down.
- An avoid index carries the phrases the declared glossary rules out, so a mined term that conflicts with a declared one is flagged rather than silently competing.
- An MCP tool exposes the glossary to agents.

## Where this sits in #2262

The issue describes four increments. This PR is the first two:

| increment | here |
|---|---|
| 1. Ingest a declared glossary as the authoritative term source | yes |
| 2. Expose terms as data to agents | yes |
| 3. Flag naming drift (health detector and PR mode) | not here |
| 4. Bootstrap a `CONTEXT.md` draft from what is mined | not here |

## What is not

This is the first reviewable slice. The drift-flagging half, meaning reporting where the code diverges from the declared glossary, is deliberately not here and remains open work.

## Verification

Rebased onto `main` on 2026-09-21, which had moved 113 commits ahead and left the branch with a real conflict in `tests/fixtures/mcp/tool_response_shapes.json`. The conflict was that #2452 deleted the `change_shape` block from that fixture when it restructured change-risk reporting, while this branch had reordered keys inside that block. The deletion is the newer intent, so the block is gone and this branch's own addition to the same fixture, the `get_glossary` entry, is kept. The fixture parses and holds 18 top-level keys.

```
tests/unit/generation/test_declared_glossary_authority.py   55 passed
tests/unit/server/mcp/test_get_glossary.py                   9 passed
tests/unit/server/mcp/test_response_budget_contracts.py     55 passed
tests/unit/server/mcp/test_wire_representation_contract.py   4 passed
                                                           123 passed
```


sloemo01:
Rebased onto `main` and pushed. The branch was 113 commits behind and carried a real conflict, which is what the `dirty` state on this PR was.

The conflict was in `tests/fixtures/mcp/tool_response_shapes.json`, and resolving it needed a decision rather than a pick:

- `#2452` deleted the `change_shape` block from that fixture when it restructured change-risk reporting. Main now has zero occurrences of it.
- A commit on this branch had touched the same block, reordering two of its keys.

The deletion is the newer intent, so the block stays gone. This branch's own addition to the same fixture, the `get_glossary` entry, is kept. Both sides survive in the sense that matters: main keeps its restructure, this branch keeps its new tool.

```
before:  mergeable=false, mergeable_state=dirty, 113 behind main
after:   mergeable=true,  0 behind, 4 ahead

tests/unit/generation/test_declared_glossary_authority.py   55 passed
tests/unit/server/mcp/test_get_glossary.py                   9 passed
tests/unit/server/mcp/test_response_budget_contracts.py     55 passed
tests/unit/server/mcp/test_wire_representation_contract.py   4 passed
                                                           123 passed
```

The fixture parses, with 18 top-level keys against main's 17, the difference being the tool this PR adds.

I also linked the body to `Part of #2262`, which is the issue this implements the first two increments of. The PR did not reference it before, which was an omission on my side.

PR #2345, opened 2026-09-18
feat(store): opt-in global store mode keeps the repo untouched

sloemo01:
Adds an opt-in global store mode: a repository's index lives under `$HOME/.repowise` instead of inside the checkout, so a repository can be indexed without changing anything in it.

This is for the case the issue describes, where the person indexing a repository is not in a position to make repository-wide changes. With the mode on there is no `.repowise/` directory, nothing new to gitignore, and nothing new in `git status`.

## Design

- One resolution point, `store_location.py`, decides where the index lives, so every layer reads the same answer instead of each deriving its own.
- The entry is keyed on the checkout's absolute path, so two checkouts of one repository get two entries and the same checkout always resolves to the same entry whichever directory the command runs from.
- The switch is per run: `--global-store`, or `REPOWISE_GLOBAL_STORE=1` for CI and multi-repo use, with an explicit off.
- It is deliberately not a `config.yaml` key, since that file lives in the store this module is choosing and cannot also choose it. What carries the decision past the run that made it is the store itself: a checkout whose global entry already exists keeps resolving to it, so a plain `repowise update` after a global `init` finds the index rather than creating the `.repowise/` the user asked not to have.
- `REPOWISE_GLOBAL_STORE_ROOT` relocates the root, which is what keeps tests out of a real home directory.

## Not in this PR

The issue also asks for querying insights for past commits from a central index, and for a migration path for existing per-repo `.repowise` directories. Both are remaining work.

## Verification

`tests/unit/core/test_store_location.py` and `tests/unit/cli/test_global_store_mode.py` cover the resolution and the CLI surface. The broader `tests/unit/pipeline` and `tests/unit/ingestion` run was 3445 passed with one failure, `test_repo_totals_churn.py::test_a_rebased_branch_in_the_history_falls_back`, which is pre-existing and environment-specific: it shells out to `git rebase master` and git exits 128 here.

One note on the branch: it also carries a small test-quality commit to `tests/unit/cli/test_augment_read_skeleton.py` that is unrelated to the store work. Flagging it rather than leaving it hidden; happy to drop it if you would rather keep the PR single-purpose.

## Verification

Past the unit tests, the promise was checked end to end on a throwaway checkout:

- `init --global-store` leaves `git status --porcelain` empty and creates the entry under the store root.
- A plain `update` and `doctor` afterwards, with no flag, still leave the tree untouched and read the store.
- Two checkouts of one repository produce two entries.
- `--local-store` forces the repo-local store and leaves the global entry alone.

That check is why the routing reaches further than the four files it started in. Around ten writers bypassed the resolver and still wrote into the repository, the pipeline parse, centrality and duplication caches and the knowledge graph among them, and a later plain `update` re-planted `.mcp.json`, `.claude/` and `.vscode/` through the editor refresh. Those call sites now go through the single resolver, and the editor refresh is gated on the store actually being repo-local.

Sabotage, both restored byte-identical: forcing the resolver back to the repo-local path fails 7 of the new tests, and removing the editor-refresh guard fails the test asserting the tree stays untouched.

The regression sweep covered `tests/unit/cli` (2576 passed), `tests/unit/server` (3333), `tests/unit/health` and `tests/unit/precedent` (1958), `tests/unit/pipeline` and `tests/unit/ingestion` (3445, plus the known pre-existing churn failure), and `tests/unit/distill` (811).


RaghavChamadiya:
Thanks @sloemo01. First, the run you just put up is good work, and several of them are on things nobody else was going to pick up. #2338 is a fair example: it reuses `db_configured()` instead of reinventing the check, and it fixes the `--path` selection bug the issue only mentions in passing. That is a PR that saves me a review cycle rather than costing me one.

You have 46 PRs merged here, which is among the highest of anyone working on this, and a real amount of what works in this project works because you fixed it. I would like to keep that going for a long time, and that is exactly why I am writing the rest of this out properly instead of quietly letting the duplicates pile up.

A few of the recent PRs landed on issues someone else had already claimed, in most cases a few hours after their comment:

- #2331 on #2302, claimed by @rehan6025 at 11:05
- #2327 on #2298, claimed by @Aman-goel-04 at 10:13 with a full implementation plan
- #2325 on #2301, claimed by @Aman-goel-04 at 17:16
- #2333 on #2230, claimed by @akshatmalik-bruh on 12 September and confirmed by me on the 14th
- #2340 on #2193, claimed by @modelpath-dev on 10 September
- #2332 on #2175, where @jgt87 handed the issue to @Aman-goel-04, I assigned it, and their PR #2308 was already open

I am holding those six rather than merging them, and those issues stay with the people who claimed them. I know that is an unwelcome thing to hear about finished work, and I am not reading it as bad faith. You claim properly when you think of it: #2297 and #2289 both have your comment in the thread before the PR. The gap is checking whether somebody else got there first, not the claiming itself.

The reason I weigh this above the duplicated review time is simple. Someone who writes out a detailed plan and then watches a finished PR appear the same evening usually does not come back. I would rather lose a few duplicate fixes than lose the contributors.

Two things that would help.

**Please leave `good first issue` alone.** That label is how new people get a first contribution, and it is the only thing I have that brings them in. `help wanted` is the opposite: I put it on work that needs someone who already knows the codebase, which is you. Take as many of those as you want. #2326, #2330, #2334, #2336, #2337 and #2338 are all in that group and none of them are a problem.

**Say you are taking it in the thread before you start.** One line is enough. It means the next person reads the thread and picks something else instead of duplicating you.

One honest word on volume. You have 38 open PRs, about 30% of the queue, and 22 of those went up in the last two days. My review time is the constraint here, not your output, so a deep queue mostly means your own work waits longer. If you kept roughly six open at a time I think more of it would land, and sooner.

If you want somewhere to put all of this, the Postgres surfaces are the best unclaimed work in the repo and nobody is on them. Reviewing #2338 turned up a fourth one that is not filed yet: `workspace_cmd.py:150` decides "not indexed" from `repowise_dir.exists()` alone, and `_query_repo_counts` returns `(0, 0)` whenever the local `wiki.db` is missing (`:196-197`). Neither consults `get_configured_db_url()`, and that file has no reference to it at all. So under a configured Postgres database every repo in a workspace reads as "not indexed" with zero counts under `repowise workspace list`, however well indexed it is. Same bug class as #1034, #2148 and #2320, in a surface none of them touch. It is yours if you want it.

None of this changes that I am glad you are here. Keep sending them, just give the thread a look first.

RaghavChamadiya:
Filed the `workspace list` one as #2368, with the mechanism written up and the `db_configured()` precedent from your own #2338 named as the fix. Unassigned and yours if you want it.

Also merged today from your queue: #2318, #2323, #2324, #2326, #2330, #2336, #2337 and #2338. Reviewing #2337 turned up the same token undercount in the direct Anthropic API provider, where the bill is real money instead of a seat, so that is filed as #2367.

rishu685:
bro has created its life time contribution to this repo at once!

sloemo01:
You are right about the volume, and I want to be concrete about what changes rather than just agree.

I had not read this comment until today. I was watching CI and merge state, and a paragraph addressed to me on a PR I thought of as "one of the open ones" did not surface. That is on me: review feedback was the thing I was not reading.

What I am changing:

- **Claiming first.** I will read the issue thread and post one line before starting. The six you listed are closed. Two of them (#2340, #2332) I would have caught by reading the thread; #2332 especially, since @jgt87 had handed the issue to @Aman-goel-04 and you had assigned it.
- **`good first issue`: left alone from here.** That lane is not mine.
- **No more opening until the queue is down.** 26 open now, and 18 of those have never been looked at. Adding to that is what I was doing and it is the wrong direction.

Where I would use the time, if it is useful to you: the Postgres surfaces. I have not touched them. If there is one you would rather I start on than another, say which and I will take that one and leave the rest.

Thank you for the specific counts. "My review time is the bottleneck" is not something I can see from my side, and it is the useful half of the feedback.

sloemo01:
Rebased onto `main` and pushed. The branch was 113 commits behind and carried three real conflicts, which is what the `dirty` state was.

## The conflicts, and why they were decisions

All three were the same collision. Main added `has_db_store()` and `resolve_db_url()` for the configured-database case; this branch adds `resolve_store_dir()` for global store mode. Each side was fixing a different instance of one bug: a guard that answers "is there a store here" by testing the repo-local path alone.

| file | resolution |
|---|---|
| `core/persistence/database.py` | `has_db_store` keeps main's configured-database branch and resolves the file path through `resolve_store_dir` instead of hardcoding `.repowise` |
| `core/workspace/update.py` | main's `has_db_store(repo_path)` guard, because it now covers both cases; the raw `.is_file()` check is gone |
| `cli/worktree.py` | union of both imports: main's `warn` and this branch's `resolve_store_dir` |
| `cli/commands/update_cmd/persistence.py` | same as `workspace/update.py` |

Neither side was dropped. The rule I applied: main's newer guard names the question better, so it survives, and it gets taught the second location.

## The bug the rebase surfaced

Once rebased, seven tests failed, and they were pointing at something I had not noticed: `init` still wrote `.repowise/omissions/` into the checkout after the index itself had moved. Three modules spell that path themselves.

- `savings/transcript.py` joined `(".repowise", "omissions")` at three call sites.
- `savings/recorder.py` joined the parts in `sidecar_path`.
- `savings/pricing.py` built the pricing cache path the same way.

All of them are the exact failure this mode exists to prevent, and the sidecar landed after my branch was written, which is why the original test suite did not see it. Each asks `resolve_store_dir` now.

## Verification

```
before:  mergeable=false, dirty, 113 behind main
after:   mergeable=true, 0 behind, 4 ahead

tests/unit/cli/test_global_store_mode.py    14 passed  (7 were failing before the sidecar fix)
tests/unit/core/test_store_location.py      34 passed
tests/unit/savings/                        141 passed
                                           189 passed
ruff check                                 clean
```

Three new arms pin the sidecar, and they were sabotage-checked rather than trusted:

```
recorder.sidecar_path hardcoded again  -> 1 failed  (the sidecar arm)
transcript._sidecar_dir hardcoded again -> 10 failed (the original seven, plus the three new)
```

Issue #2352, opened 2026-09-18
[Docs] GRAPH.md quotes benchmark figures that BENCHMARKS.md already corrected


RaghavChamadiya:
**Summary**

`docs/layers/GRAPH.md` restates a benchmark result that no longer matches the source of truth in `docs/BENCHMARKS.md`.

**Mechanism**

`docs/BENCHMARKS.md` widened one language cell from 30 to 40 rows across two repositories, which changed both the counts and the denominator. That file was corrected in commit `d416aca99`. `GRAPH.md` was never updated to match.

**Repro**

`docs/layers/GRAPH.md:165` reads:

> **229 of 270 correct for us, 154 of 270 for CodeGraph 1.5.0**, intervals disjoint.

`docs/BENCHMARKS.md:211-212` reads:

> | **repowise** | **240/280 = 85.7%** | [81.1, 89.3] |
> | **CodeGraph 1.5.0** | **164/280 = 58.6%** | [52.7, 64.2] |

Both the counts and the denominator differ.

**Impact**

A reader of GRAPH.md gets a stale figure that understates repowise's result and overstates the comparison, and the two documents contradict each other when read together.

**Done looks like**

GRAPH.md quotes the same figures as BENCHMARKS.md, or better, drops the restated numbers and links to BENCHMARKS.md as the single source for this benchmark.

Tests: a plain text fix. Worth asking in the pull request whether the repository's documentation drift check can pin GRAPH.md's quote against BENCHMARKS.md, so the next update cannot silently leave it behind again.

PR #2359, opened 2026-09-18
docs: point GRAPH.md graded figures at BENCHMARKS.md (#2352)

dyk1454683243-sudo:
## Summary
Closing this PR — the branch currently replaces the full GRAPH.md design page with a short stub (large deletion). A proper follow-up should keep the page and only sync the hand-graded figures (`240/280`, `164/280`) with BENCHMARKS.md §7 (issue #2352).

Sorry for the noise.


sloemo01:
Closing this yourself, with the reason written out, was the right call and it is worth saying so plainly.

A branch that replaces the whole design page with a stub is exactly the kind of thing that costs a maintainer a review cycle to discover. Catching it before anyone had to is the judgement the `good first issue` label exists to find.

If you want to retry, the fix is much narrower than that branch was: keep the page, and sync only the hand-graded figures so they agree with `docs/BENCHMARKS.md`, which is the source of truth here. The issue is #2352 and it is still open.

Two things that will save you a pass:

- the table at `BENCHMARKS.md:213-224` is what the prose has to match, and java's row is `n=40` rather than 30, which is where the denominator comes from
- the same percentage is quoted a second time further down GRAPH.md, so it is worth grepping rather than fixing the one occurrence

PR #2360, opened 2026-09-18
fix(mcp): record unrecognised mode in ignored_arguments for search_codebase

### LINKED ISSUE #2347: [Bug] search_codebase silently coerces an unrecognised mode to auto
**Summary**

`_VALID_MODES` is `{"auto", "concept", "symbol", "path", "hybrid"}`, and `_resolve_mode` lowercases the incoming value and quietly falls back when it is not in that set.

`packages/server/src/repowise/server/mcp_server/tool_search.py:925`, `:928-949`

```python
m = (mode or "auto").lower()
if m not in _VALID_MODES:
    m = "auto"
```

**Mechanism**

The repository already has the mechanism for this. `resolve_enum_argument` returns a valid value, or drops it and records it in an `ignored` list that `attach_ignored_arguments` surfaces to the caller.

`packages/server/src/repowise/server/mcp_server/_helpers.py:254-283`

`search_codebase` already uses it for `kind`:

`packages/server/src/repowise/server/mcp_server/tool_search.py:1155`

```python
kind = resolve_enum_argument(kind, _VALID_KINDS, argument="kind", ignored=ignored)
```

`mode` never goes through it, so a typo is absorbed instead of reported.

**Repro**

Call `search_codebase` with `mode="symbl"`. The search runs the `auto` heuristics and the response carries no ignored-argument entry for `mode`, which is what the same call with a misspelled `kind` would produce today.

**Impact**

The caller gets results from a different search strategy than the one asked for, with nothing in the response saying so.

This is the same defect that #1496 reported against `get_dead_code`'s `kind` argument, and `resolve_enum_argument` is what closed it. Here it is arguably worse: the call does not go empty, it silently answers a different question.

**Done looks like**

Route `mode` through `resolve_enum_argument` exactly as `kind` is at `:1155`, so an invalid value is dropped and reported rather than remapped inside `_resolve_mode`. Leave `_resolve_mode`'s internal fallback in place as a defensive no-op.

MCP tool docstrings in this file sit under a size budget, so do not expand the `mode` description at `:1146` beyond its current line.

Tests: extend the argument-validation tests that already cover an invalid `kind` with an invalid `mode`, asserting it is reported in the ignored arguments and that the search still runs under `auto`.

Voyagerroc-Lab:
Closes #2347

### Summary
When an unrecognized `mode` was passed to `search_codebase`, it was previously silently coerced to `auto` inside `_resolve_mode` without recording the invalid argument in `ignored_arguments`. As with other closed-vocabulary filters (e.g. `kind` or `include`), a typo should be transparently recorded in `ignored_arguments` so the caller knows the argument was dropped while still recovering and executing under `auto`.

### Changes
- In `packages/server/src/repowise/server/mcp_server/tool_search.py`, validate `mode` via `resolve_enum_argument(mode, _VALID_MODES, argument="mode", ignored=ignored)` before resolving concrete branch heuristics.
- In `tests/unit/server/mcp/test_argument_validation.py`, added `TestSearchMode` covering unknown mode reporting and verified that known modes do not pollute `ignored_arguments`.


8rulerstar:
Thanks for picking this up. One thing I noticed: `_resolve_mode` lowercases `mode` before checking it, so on current main `mode="Symbol"` (or `"SYMBOL"`) runs a symbol search. `resolve_enum_argument` compares case-sensitively, so with this change `"Symbol"` is reported as unknown and the call falls back to `auto`. I ran both against "authentication service": main returns `mode: symbol`, this branch returns a concept search with `ignored_arguments` naming `"Symbol"`.

If case-insensitive modes should keep working, lowercasing before validation would preserve that, and a `mode="Symbol"` test would guard it.

RaghavChamadiya:
Thanks @Voyagerroc-Lab for moving the `ignored` list above the resolve call. That ordering is genuinely required, since a mode rejected after the list is built has nowhere to be recorded, and you got that part right. Leaving `kind` alone was also correct.

There is one blocking problem, and @8rulerstar already called it out in the comment above with a working repro. I have verified it against current main rather than taking it on trust:

- `_VALID_MODES = {"auto", "concept", "symbol", "path", "hybrid"}` at `tool_search.py:925`, all lowercase.
- `resolve_enum_argument` compares with `if value is None or value in valid` at `mcp_server/_helpers.py:314`. Plain membership, case-sensitive.
- `_resolve_mode` lowercases its own input at `tool_search.py:938` (`m = (mode or "auto").lower()`).

So on main, `mode="Symbol"` works: `_resolve_mode` folds the case itself. On this branch the value goes to `resolve_enum_argument` as typed, misses the set, gets dropped into `ignored_arguments` and silently falls back to `auto`. That turns a working call into a concept search, which is a worse outcome than the bug this PR set out to fix.

The fix is to lowercase before validating and keep the original spelling for the report, so `mode="Symbol"` still resolves and only a genuine typo like `"Symbl"` is reported, echoed back as the caller typed it.

I should be straight with you about what happens next, though. #2347 was assigned to @8rulerstar after they claimed it on the issue, and their #2392 implements this with that case handling and a test for it. Your PR here is earlier by timestamp, opened about half an hour after the issue was filed, but it never carried a claim comment, and in this repo the claim process on the issue is what binds. So #2392 is the one that will merge.

That is not a comment on your work, and I would rather not lose you over a race nobody told you was running. Your #2309 merged a few minutes ago and was a clean fix. If you want something unclaimed right now, `javascript.scm` has no `public_field_definition` pattern at all, so plain JS class fields holding a function are never indexed, the same gap #2302 just fixed for TypeScript. Nobody has that one and I would review it quickly.

PR #2363, opened 2026-09-18
fix(ingestion): mint reference edges for Go functions in composite literals (#2351)

### LINKED ISSUE #2351: [Bug] A Go function used only as a composite-literal value mints no reference edge
**Summary**

A Go function referenced only as a value inside a composite literal, such as `template.FuncMap{"greet": greet}`, produces no call or reference edge, so it reads as fully dead even though it is registered and called at runtime.

**Mechanism**

The one pattern that rescues a function used as a first-class value matches only a qualified `selector_expression` directly inside a call's argument list.

`packages/core/src/repowise/core/ingestion/queries/go.scm:94-105`

```
; Qualified function value passed as call argument: f(pkg.Handler, ...)
(call_expression
  arguments: (argument_list
    (selector_expression
      operand: (_) @reference.receiver
      field: (field_identifier) @reference.name
    )
  )
)
```

The only pattern touching composite literals captures the literal's *type*, never a value inside it.

`packages/core/src/repowise/core/ingestion/queries/go.scm:132-135`

```
(composite_literal
  type: (_) @param.type)
```

Nothing captures a value inside a `composite_literal`.

**Repro**

A Go file where `greet` and `handler` are referenced only through `template.FuncMap{"greet": greet, "handler": handler}`, plus a `doSomething(handler)` call. Parsed at the current commit, `calls` holds only the `doSomething` call site and `references` is empty. Both functions read as dead.

**Impact**

Under-reporting only. No wrong edge is created. Unsized across the corpus; `test-repos/hugo` and `test-repos/cobra` are available for sizing how often the shape occurs.

**Done looks like**

A pattern capturing a function name used as a value inside a `composite_literal`'s element list and minting a reference edge, mirroring the qualified-argument pattern at `:94-105`.

**Keep this scoped to the composite-literal value.** A bare identifier passed as a call argument, as in `doSomething(handler)`, is a broader shape with its own false-positive risk and deserves its own issue.

Tests: a fixture with a function referenced only through a `template.FuncMap`-style composite literal value, asserting a reference edge is minted; and a fixture confirming the composite-literal type capture at `:132-135` is unaffected.

rishu685:
Fixes #2351.

In Go codebases, functions referenced as values inside composite literals (such as registration tables `template.FuncMap{"greet": greet, "handler": pkg.Handler}`, struct initializers `Config{OnEvent: greet}`, or slice literals `[]func(){greet}`) previously minted neither call nor reference edges. As a result, functions used only in these positions were incorrectly flagged as dead code.

This PR adds tree-sitter query patterns to capture bare function identifiers and package-qualified selector expressions within:
1. `keyed_element` value positions (map entries and struct field initializers)
2. `composite_literal` body elements (slice/array literal element lists)

## Changes
- **`packages/core/src/repowise/core/ingestion/queries/go.scm`**:
  - Added query patterns for `keyed_element` values (`@reference.name` and optional `@reference.receiver`).
  - Added query patterns for direct elements within `composite_literal` bodies.
  - Kept existing `(composite_literal type: (_) @param.type)` query intact.
- **`tests/unit/ingestion/test_value_reference_edges.py`**:
  - Added unit test cases for map, struct, and slice composite literal function value references asserting incoming `references` edges without creating false `calls` edges.

## Verification
- Added test coverage in `TestGoFunctionValues`:
  - `test_func_value_in_keyed_map_composite_literal`
  - `test_func_value_in_struct_composite_literal`
  - `test_func_value_in_slice_composite_literal`


RaghavChamadiya (approved):
Thanks @rishu685 for asserting `references` without `calls` in the tests. A function named in a composite literal is a value, not an invocation, and a patch like this is easy to get wrong in the direction of minting call edges that never happen at runtime. Pinning both sides is what makes it reviewable.

Verified on current main that `go.scm` has no `keyed_element` or `composite_literal` patterns, so map values, struct field values and slice elements all went unrecorded.

The patterns match any identifier in those positions rather than only functions, which matches how `references` edges already work elsewhere in this file. A local variable there simply resolves to no node, so this does not open a new false-positive class.

PR #2364, opened 2026-09-18
fix(ingestion): retire inert reads membership from SYMBOL_USE_EDGE_TYPES (#2349)

### LINKED ISSUE #2349: [Bug] SYMBOL_USE_EDGE_TYPES still carries the inert reads membership
**Summary**

`SYMBOL_USE_EDGE_TYPES` still lists `"reads"` as a symbol-to-symbol use edge, but nothing can produce a `reads` edge whose source is a symbol node.

`packages/core/src/repowise/core/ingestion/models.py:554-573`, the entry at `:567`

**Mechanism**

The file's own comment at `:549-553` explains why: the symbol-level producer moved to `framework_binds`, leaving `csharp_member_reads` as the only remaining producer.

`packages/core/src/repowise/core/ingestion/languages/csharp_member_reads.py:143-163`

`_add_reads_edge` adds the edge with `edge_type="reads"` at `:159`, and its call sites pass file paths on both ends. A file node can never be a symbol node's predecessor, so the membership can never match. It is inert rather than wrong.

Two things already lean on this. `EXECUTION_EDGE_TYPES` (`:584`) explicitly subtracts `"reads"`, which only makes sense if the membership is understood to be dead weight. `REACHABILITY_USE_EDGE_TYPES` (`:591`) is built on top of `SYMBOL_USE_EDGE_TYPES` and inherits the same inert member.

**Repro**

Read `models.py:546-573` beside `csharp_member_reads.py:130-163`. The file-to-file shape of every `reads` edge the codebase emits is visible directly in `_add_reads_edge`'s call sites, so no script is needed.

**Impact**

None on current behaviour. This is dead vocabulary, not a wrong edge. The cost is to a reader who sees `"reads"` in a symbol-use set and assumes it does something.

**Done looks like**

The proof first, then the removal. A guard test that a `reads` edge's source is never a symbol node, so removing the membership provably cannot change any existing classification. Only once that test passes, drop the `"reads"` entry and correct the comment at `:549-553`.

**Keep this scoped to the one membership.** `FILE_DEPENDENCY_EDGE_TYPES` at `:540-543` also contains `reads`, where it is legitimate and load-bearing for file-level dependencies. Do not touch it, and do not change anything `csharp_member_reads` emits.

Tests: the guard test above, added before the removal. No existing test should change behaviour.

rishu685:
Fixes #2349.

`SYMBOL_USE_EDGE_TYPES` listed `"reads"` as a symbol-to-symbol use edge, but its symbol-level producer was previously moved to `framework_binds`. The only remaining producer of `"reads"` edges is `csharp_member_reads.py`, which strictly emits file-to-file dependency edges (`Caller.cs -> Order.cs`).

Because a file node is never a symbol node, `"reads"` in `SYMBOL_USE_EDGE_TYPES` was completely inert. This PR retires `"reads"` from `SYMBOL_USE_EDGE_TYPES` and synchronizes all downstream vocabulary collections across the engine, server, and UI packages.

## Changes
- **`packages/core/src/repowise/core/ingestion/models.py`**:
  - Removed `"reads"` from `SYMBOL_USE_EDGE_TYPES`.
  - Simplified `EXECUTION_EDGE_TYPES` to `SYMBOL_USE_EDGE_TYPES - {"references"}`.
  - Corrected comments on `FILE_DEPENDENCY_EDGE_TYPES` and `SYMBOL_USE_EDGE_TYPES`.
  - Kept `"reads"` in `FILE_DEPENDENCY_EDGE_TYPES` intact.
- **`packages/server/src/repowise/server/schemas/intelligence.py`**:
  - Updated `SYMBOL_RELATION_GROUPS["reference"]` to `frozenset({"references"})` to keep the partition exact.
- **`packages/ui/src/graph/symbol-relations.ts`**:
  - Removed `reads` from `RELATION_LABELS`.
- **`tests/unit/ingestion/test_csharp_member_reads.py`**:
  - Added `test_reads_edges_are_strictly_file_to_file` guard test verifying all `reads` edges connect file paths and never symbol IDs (`::`).
- **`tests/unit/ingestion/test_edge_type_vocabulary.py`**:
  - Added `test_reads_edge_producer_is_strictly_file_level` verifying that only `csharp_member_reads.py` produces `reads` edges.

## Verification
- Added guard tests verifying `reads` edges are strictly file-to-file.
- Synchronized tests in `test_symbol_relation_vocabulary.py` and `test_edge_type_vocabulary.py`.


sloemo01:
Three legs are red on a single test, and the cause sits two files away from your diff:

```
FAILED test_onboarding_key_concepts.py::test_relation_verb_covers_the_edge_types
  not (dict_keys([... 'reads' ...]) - frozenset([... no 'reads' ...]))
```

The link is `key_concepts.py:81`:

```python
_CONCEPT_EDGE_TYPES = SYMBOL_USE_EDGE_TYPES
```

Dropping `reads` from `models.py` therefore drops it here as well, while `_RELATION_VERB` in the same file still carries a `reads` verb. The test is a two-way check on purpose, so it fails on the pair rather than on either side alone:

```python
assert not (_CONCEPT_EDGE_TYPES - _RELATION_VERB.keys())
assert not (_RELATION_VERB.keys() - _CONCEPT_EDGE_TYPES)
```

Nothing in your diff shows this because `key_concepts.py` is not among your five files.

One thing worth knowing before you delete the verb. The comment directly above `_CONCEPT_EDGE_TYPES` records a deliberate decision to keep `reads` in this shared view rather than trim it, on the grounds that a future cross-file `reads` "should start counting here without anyone remembering to add it". So this is a design question rather than leftover cleanup: either retire the `reads` verb too and rewrite that comment to match, or leave `reads` in the shared set and take the retirement up in #2349 on its own. Both are defensible. I would only make the choice explicitly, because that comment is the single place the reasoning is written down, and it currently says the opposite of what the change does.

Local check:

```
uv run pytest tests/unit/generation/test_onboarding_key_concepts.py -q
```

RaghavChamadiya (approved):
Thanks @rishu685, and thanks @sloemo01 for catching the `_CONCEPT_EDGE_TYPES` knock-on. That was the right thing to flag: it aliases `SYMBOL_USE_EDGE_TYPES` directly, so trimming the shared set silently changes what onboarding prose can talk about, and the vocabulary-parity test would have gone red for a reason nobody reading this diff would have predicted.

What makes the follow-up commit better than a minimal fix is that it corrected the stale comment instead of just deleting the entry. The old text said the symbol-layer producer was `framework_edges/express.py`. I checked: `express.py` emits no `reads` edges at all any more, and `csharp_member_reads.py:159` is the only producer of them anywhere in the repo. Both its endpoints are file paths, `_add_reads_edge(graph, path, target)` with `target` coming out of `build_csharp_type_to_file`, so `reads` really is file to file and really was inert in a symbol-level set. Leaving the old explanation in place would have left the next person defending a producer that no longer exists.

Verified the rest on current main too: `reads` was still listed in `SYMBOL_USE_EDGE_TYPES`, and after the change the only places retaining it are the file-level list and `c4_builder/labels.py`, which is the expected pair.

PR #2365, opened 2026-09-18
fix(ingestion): align C# builtin_parents with VB.NET BCL list (#2304)

### LINKED ISSUE #2304: [Bug] C# builtin_parents is missing ten BCL types that VB.NET's list already carries

rishu685:
Fixes #2304.

C# and VB.NET share the same .NET Base Class Library (BCL) runtime. However, `packages/core/src/repowise/core/ingestion/languages/specs/csharp.py` only listed 12 builtin parent types in `builtin_parents`, while `vbnet.py` listed those plus 14 additional BCL base classes and interfaces.

Because types like `EventArgs`, `Attribute`, UI forms (`Form`, `Control`, etc.), and collection interfaces were omitted from C#'s `builtin_parents`, classes inheriting from these BCL types were eligible to match arbitrary repo-local symbols with colliding names in the heritage resolver's global-unique tier.

This PR adds the missing 14 BCL types to C#'s `builtin_parents` and adds parity and extraction unit tests.

## Changes
- **`packages/core/src/repowise/core/ingestion/languages/specs/csharp.py`**:
  - Added `Form`, `Control`, `UserControl`, `Component`, `Page`, `Window`, `Attribute`, `EventArgs`, `MarshalByRefObject`, `INotifyPropertyChanged`, `IComparer`, `ICollection`, `IList`, and `IDictionary` to `builtin_parents`.
- **`tests/unit/ingestion/test_csharp_extractors.py`**:
  - Added `test_bcl_builtin_parents_are_filtered` verifying that inheriting from BCL base types yields no heritage records.
- **`tests/unit/ingestion/test_language_capabilities.py`**:
  - Added `test_csharp_and_vbnet_builtin_parents_parity` asserting `csharp.builtin_parents == vbnet.builtin_parents`.

## Verification
- Verified C# heritage extraction filters all BCL base types.
- Verified spec-level parity test between C# and VB.NET.


RaghavChamadiya:
@rishu685 #2304 just landed via #2309, so this one is superseded and I am closing it. The reasoning is worth spelling out because the two PRs were the same fix.

Neither of you claimed #2304 on the issue before writing code. @Voyagerroc-Lab opened #2309 about two hours after the issue was filed on 17 September; this PR opened about twenty hours later, and your comment saying a PR was up came after that. With no claim comment on either side the tiebreak falls to who got there first, which was them. It is also their first merged change here, and a straightforward parity fix is the kind of thing a newcomer should get to land.

Your version had something theirs did not, and I want it: `test_csharp_and_vbnet_builtin_parents_parity`. #2309 fixes today's drift but nothing stops the two sets separating again, and a parity test is the thing that makes this the last time. Please open a small follow-up with just that test against current main, and I will merge it.

While you are in there, `fsharp.py` has the same shape of gap: its `builtin_parents` already has `Attribute`, `EventArgs` and `MarshalByRefObject`, but is missing `Form`, `Control`, `UserControl`, `Component`, `Page`, `Window`, `INotifyPropertyChanged`, `IComparer`, `ICollection`, `IList` and `IDictionary`. I verified that on current main. If you want to fold F# into the same parity guard rather than leaving a third set to drift, that would be better than what either PR did.

PR #2366, opened 2026-09-18
fix(ingestion): refine Python symbol visibility using module __all__ (#2350)

### LINKED ISSUE #2350: [Bug] Python visibility ignores __all__, so an excluded name still reads as exported
**Summary**

`py_visibility()` decides Python export visibility purely from the name: a leading underscore is private, everything else is public. It never consults the module's `__all__`.

`packages/core/src/repowise/core/ingestion/extractors/visibility.py:25-30`

**Mechanism**

Given:

```python
__all__ = ["foo"]
def foo(): ...
def _bar(): ...
def baz(): ...
```

Parsing this through the parser at the current commit gives `foo` public, `_bar` private, and `baz` public. `baz` is not underscore-prefixed, so it reads as exported even though the module deliberately excludes it. The reverse also holds: an underscore-prefixed name listed in `__all__` still reads private.

**Repro**

Parse the snippet above and inspect each symbol's visibility. `baz` comes back public.

**Impact**

Unsized. Names shaped like `baz`, public-looking but deliberately left out of `__all__`, read as part of the module's API when they are not, which reaches anything treating public as exported, including unused-export detection. No count of how often this occurs in real code has been run.

**Done looks like**

A bounded signal, not a rescue. `__all__` is advisory and frequently stale in real code, so treating it as authoritative would hide genuinely dead exports. Presence in `__all__` should raise confidence that a symbol is public, capped so it can never alone mark something reachable.

Propose a concrete cap in the pull request so it is easy to argue with, rather than leaving it implicit.

**Keep this scoped to a literal `__all__`.** Re-export chains, and an `__all__` built through a comprehension or `+=`, should be treated as absent rather than half-parsed.

Tests: a name present in `__all__`, a name absent from it, an underscore-prefixed name present in it, and a dynamically built `__all__` which must be treated as absent.

rishu685:
## Summary

Fixes #2350.

Refines Python symbol visibility extraction by inspecting literal module-level `__all__` definitions:
- Top-level symbols included in static/literal `__all__` definitions (lists, tuples, sets) become `public` (even if underscore-prefixed).
- Top-level symbols omitted from `__all__` are demoted to `private`.
- Class and inner function members remain unaffected and continue using scope/name-based visibility.
- Dynamic `__all__` constructions (comprehensions, concatenations, mutations via `+=`, or non-literal items) are treated as absent, falling back cleanly to standard name-based heuristics without error.

## Changes
- **`visibility.py`**: Added `py_module_all_names` to safely extract literal `__all__` symbols using AST and `refine_py_visibility` to adjust top-level declaration visibility.
- **`parser.py`**: Extracted module `__all__` once per Python file and refined visibility of top-level declarations.
- **`test_py_visibility_refinement.py`**: Added unit tests covering literal lists/tuples, omitted symbols, underscore exports, dynamic `__all__` fallbacks, class methods, and modules without `__all__`.


rishu685:
@RaghavChamadiya Sir, can you please review this as well :)

PR #2369, opened 2026-09-18
fix(mcp): remove dead context_hint and clean up get_context (#2348)

### LINKED ISSUE #2348: [Bug] context_hint always returns None, so every call site gets nothing
**Summary**

Every branch of `context_hint` returns `None`.

`packages/server/src/repowise/server/mcp_server/_meta.py:602-613`

```python
def context_hint(targets: list[str], compact: bool, include: set[str] | None = None) -> str | None:
    """Hint for `get_context` callers.

    Conservative: only fires when the call shape suggests the agent could
    have used a cheaper tool, AND the suggestion is unambiguously safe.
    """
    if not targets:
        return None
    # If caller requested source and got a large symbol, nudge toward Read with offset
    if include and "source" in include and len(targets) == 1:
        return None  # source mode provides its own truncation info
    return None
```

**Mechanism**

The middle branch describes a nudge in its comment and then returns the same value as the fallthrough below it, so the condition decides nothing and the function is a no-op for all inputs.

**Repro**

Call it with any arguments, including the case its own comment names (`include={"source"}` with one target). The result is `None` every time.

**Impact**

The `_meta` hint field it feeds can never carry a value. It is called once:

`packages/server/src/repowise/server/mcp_server/tool_context/context.py:232`

and the docstring at `context.py:61` points at `_meta.context_hint` as the place where the `source` case is tested and left out deliberately, which reads as though a decision is encoded here when none is.

**Done looks like**

A direction chosen and then applied, not a mechanical edit. Either implement the case the comment names, or remove the function together with its call site at `context.py:232` and correct the comment at `context.py:61`.

Say which you are taking, and why, before writing code. That judgement is the substance of this one.

Tests: if implemented, a case asserting a non-None hint for the single-target source-mode call and `None` otherwise. If removed, that `get_context`'s `_meta` block and its existing tests still pass without the hint sourced from here.

rishu685:
Fixes #2348.

Removes the dead `context_hint` function in `_meta.py`, eliminates its invocation in `get_context`, and cleans up circular/confusing documentation in `context.py`:
- `get_context` is explicitly designed as a triage / structural tool, not a raw source retrieval tool (`"source"` is not an allowed `include` block; source retrieval is handled via `include=["skeleton"]` or the `Read` tool).
- `context_hint` returned `None` across every branch and served no active purpose.
- Removing `context_hint` simplifies `get_context`'s `_build_meta` call, eliminates dead code, and clarifies the `_INCLUDE_BLOCKS` comment.

## Changes
- **`_meta.py`**: Removed dead `context_hint` function.
- **`context.py`**: Removed `_context_hint` import, removed `hint=` from `_build_meta` in `get_context`, and updated comments describing why `source` is omitted from `_INCLUDE_BLOCKS`.
- **`test_context.py`**: Added unit test `test_get_context_meta_envelope` verifying `get_context` returns a well-formed `_meta` envelope without dead/spurious `hint` fields.


RaghavChamadiya (approved):
Thanks @rishu685 for checking that every branch in `context_hint` returns `None` before deleting it, instead of just noting that it looked unused. That is the difference between removing dead code and removing code that happens to be quiet.

Confirmed it had exactly one caller, and that `build_meta`'s `hint` parameter already defaults to `None`, so dropping the keyword is behaviour-preserving. The test asserting the `_meta` envelope carries no `hint` key is the right guard against it creeping back.

PR #2373, opened 2026-09-18
fix(ingestion): extract named bindings for destructured dynamic ESM imports 

### LINKED ISSUE #2230: Bug: Dynamic ESM Destructuring Emits Wildcard Instead of Named Bindings

akshatmalik-bruh:
## Summary

- Fixed dynamic ESM import parsing (`const { fn, calc } = await import('./mod')`) to emit explicit named bindings (`["fn", "calc"]`) instead of namespace wildcard `["*"]`, preventing false negatives in dead-code analysis.
- Preserved `*` wildcard fallback behavior for bare module imports (`const mod = await import()`), lazy-route callbacks (`() => import()`), and rest-element patterns (`{ fn, ...rest }`).
- Encapsulated tree-sitter AST object pattern extraction inside `extract_dynamic_import_bindings` in `ts_js.py` to maintain language neutrality in `parser.py`.

## Related Issues

Fixes #2230

## Test Plan

- [x] Tests pass (`pytest tests/unit/ingestion/`) — 3245 passed
- [x] Lint passes (`ruff check .` & `ruff format --check .`)

## Checklist


sloemo01:
Verified independently, in case it saves a review cycle.

I ran the new tests with the change, then ran them again with only the two source files reverted. One note on method: reverting to *current* main is not usable here, because the branch is 97 commits behind it and main's `parser.py` now imports names your `visibility.py` predates. So I reverted to this branch's merge base (`8eafa7efe`) instead, which removes your change and nothing else.

| state | result |
|---|---|
| with the change | 51 passed |
| source reverted to merge base | **3 failed, 48 passed** |

```
FAILED TestDynamicESMImportDestructuring::test_await_import_destructuring_extracts_named_bindings
FAILED TestDynamicESMImportDestructuring::test_await_import_aliased_destructuring
FAILED TestDynamicESMImportDestructuring::test_dynamic_import_destructuring_javascript
```

and the assertions are the bug from #2230 in one line each:

```
assert ['*'] == ['fn', 'calculate']
assert ['*'] == ['helper']
```

The half I checked hardest is what passes in both states, because a binding extractor's real risk is over-capturing: `test_bare_await_import_namespace_stays_wildcard`, `test_lazy_route_callback_stays_wildcard` and `test_rest_element_in_object_pattern_stays_wildcard` all stay wildcards with and without your change. The wildcard is still correct for a namespace and for a rest element, and the change does not touch either.

One practical thing: at 97 commits behind main this will likely need a rebase before it merges, and the `parser.py` import difference above is the first place that will bite.

To reproduce:

```
git checkout 8eafa7efe -- packages/core/src/repowise/core/ingestion/extractors/bindings/ts_js.py packages/core/src/repowise/core/ingestion/parser.py
uv run pytest tests/unit/ingestion/parser/test_typescript.py -q
```

PR #2375, opened 2026-09-18
fix(security): mask secret values in security finding snippets (#2119)

### LINKED ISSUE #2119: Mask the secret value in persisted and served security finding snippets

akshatmalik-bruh:
## Summary

- Redacts captured credential values (`val[:4] + "****"`) in security scan snippets before findings are persisted or served.
- Adds Alembic migration `0067` to clear existing plaintext credential snippets so subsequent scans repopulate them in masked form.
- Covers per-line scans, spanning patterns, and history scanning via a single `SECRET_KINDS` post-pass before `scan_file` returns.

## Related Issues

Fixes #2119

## Test Plan

- [x] Tests pass (`pytest`)

## Checklist


RaghavChamadiya (approved):
Thanks @akshatmalik-bruh for putting the masking at one choke point instead of at each append site. `scan_file` is the only thing that mints a finding, so running the mask once on its way out covers the per-line loop, the spanning-pattern loop and `HistorySecurityScanner` without any of them needing to know about it. That is the difference between one fix and three that drift apart later.

Confirmed the bug on current main: `security_scan.py:446` stores `snippet = line.strip()[:120]` verbatim for every `SECRET_KINDS` hit, and there is no masking anywhere in the file.

The migration to scrub existing rows is the right call, and so is refusing to make the downgrade reversible. Re-exposing plaintext to undo a redaction would defeat the point of the change.

Non-blocking nit, take it or leave it: the PR body still carries the template's `Fixes #123` / `Closes #456` placeholder lines above the real `Fixes #2119`.

PR #2380, opened 2026-09-18
Fix #2222: [Bug] Telemetry flusher causes console window to flash on Windows during VS Code index updates

### LINKED ISSUE #2222: [Bug] Telemetry flusher causes console window to flash on Windows during VS Code index updates
## Describe the Bug

On Windows, Repowise causes a console window to briefly appear when telemetry is enabled.

This is especially noticeable during automatic index updates triggered by the VS Code extension, where a console window flashes even though the update is supposed to run in the background.

Disabling telemetry fixes the issue completely:

```powershell
repowise telemetry disable
```

After disabling telemetry and restarting VS Code, automatic index updates no longer cause a console window to flash.

Re-enabling telemetry also reproduces the problem directly:

```powershell
repowise telemetry enable
```

Running that command causes a console window to briefly appear as well.

## Steps to Reproduce

1. Install Repowise using `uv tool`.

2. Install the Repowise VS Code extension.

3. Ensure telemetry is enabled:

   ```powershell
   repowise telemetry enable
   ```

4. Observe that a console window briefly appears.

5. Open an indexed repository in VS Code.

6. Allow Repowise to perform an automatic index update.

7. Observe another console window briefly appear during or immediately after the update.

8. Disable telemetry:

   ```powershell
   repowise telemetry disable
   ```

9. Restart VS Code and allow another automatic index update.

10. Observe that the console window no longer appears.

A process trace taken while reproducing the issue showed:

```text
Code.exe
└─ repowise.exe update --progress json
   └─ python.exe
      └─ python.exe
         ├─ git.exe
         └─ python.exe -m repowise.cli.platform.telemetry.flusher
```

No `cmd.exe` process was observed in the process trace, despite a console window visibly flashing.

## Expected Behavior

Repowise telemetry operations and automatic index updates should run without opening or flashing a console window.

The telemetry flusher is intended to run detached in the background on Windows, so enabling telemetry or flushing telemetry should not create any visible console window.

## Actual Behavior

A console window briefly flashes when telemetry is enabled.

It also flashes during automatic Repowise index updates while telemetry is enabled.

Disabling telemetry with:

```powershell
repowise telemetry disable
```

eliminates the issue completely.

Running:

```powershell
repowise telemetry enable
```

causes the console window to appear again.

## Environment

* OS: Windows 11
* Python version: uv-managed Python
* Repowise version: `0.49.0`
* Installation method: `uv tool`
* Repowise VS Code extension: `0.10.0`
* uv version: `0.12.13`

## Additional Context

The VS Code extension appears to already launch the Repowise CLI with:

```typescript
spawn(executable, args, {
    shell: false,
    windowsHide: true,
});
```

The Repowise telemetry flusher also appears intended to avoid creating a console window on Windows by using `DETACHED_PROCESS | CREATE_NO_WINDOW`.

Since:

* telemetry enabled → console flash occurs
* `repowise telemetry enable` → console flash occurs
* automatic index update with telemetry enabled → console flash occurs
* telemetry disabled → console flash disappears

the issue appears specifically related to the telemetry process/flusher on Windows rather than the main VS Code index-update process.

pamod-madubashana:
Fixes #2222

## Summary

Fixes #2222 — console window flashing on Windows during VS Code index updates
when telemetry is enabled.

**Root cause:** `_spawn_flusher()` in
`packages/cli/src/repowise/cli/platform/telemetry/emitter.py` launched the
detached telemetry flusher with `sys.executable` — the console-subsystem
`python.exe` — relying only on `DETACHED_PROCESS | CREATE_NO_WINDOW`. A
console-subsystem binary spawned from a GUI/console-less parent (VS Code
extension launches `repowise` with `windowsHide: true`) can still briefly
allocate/flash a console, and no `STARTUPINFO` `SW_HIDE` was passed. Every CLI
exit (including `repowise telemetry enable` and background `repowise update`
runs) spawns the flusher, so the flash appeared on each of those paths and
disappeared when telemetry was disabled.

**What changed** (`emitter.py` only):
- Added `_flusher_executable()`: on Windows prefers a `pythonw.exe` sibling
  next to `sys.executable` (windowless GUI subsystem — Windows never allocates
  a console for it) when that file exists, otherwise falls back to
  `sys.executable`. POSIX behavior unchanged.
- `_spawn_flusher()` now launches that executable, and on Windows additionally
  passes `STARTUPINFO` with `STARTF_USESHOWWINDOW` / `SW_HIDE` (guarded by
  `hasattr`, fail-silent) alongside the existing
  `DETACHED_PROCESS | CREATE_NO_WINDOW` flags.
- Wrapped `subprocess.Popen` in try/except returning `False` on failure, so the
  spawn itself is fail-silent (telemetry must never break a command); the
  existing `atexit _flush` suppression is unchanged.
- POSIX path (`start_new_session=True`, DEVNULL stdio) unchanged.

**Why it addresses the issue:** the flusher no longer uses a
console-subsystem executable on Windows when `pythonw.exe` is available, and
even in fallback it requests window hiding at both the process-creation and
window-show layers — matching the documented recipe for invisible background
processes. `telemetry enable`, normal command exits, and VS Code-triggered
`update` runs therefore spawn no visible console.

## Verification

- `python -m py_compile packages/cli/src/repowise/cli/platform/telemetry/emitter.py` — passed (no output).
- `python -c "import sys; sys.path.insert(...); from repowise.cli.platform.telemetry import emitter; ..."` — import OK; `_flusher_executable` present; POSIX spawn path confirmed (`start_new_session=True`, no `creationflags`).
- Full unit suite could NOT be run: neither `uv` nor `pytest` is installed in this environment (`uv ...` → spawn failure, `python -m pytest` → `No module named pytest`). Per the Makefile the relevant command would be `uv run pytest tests/unit/cli/test_telemetry.py -v`, plus `uv run ruff check packages/ tests/` — neither was runnable here. No test files were modified.


RaghavChamadiya:
Thanks @pamod-madubashana. The diagnosis here is right and I confirmed it on current main: `_spawn_flusher` in `platform/telemetry/emitter.py` launches the console-subsystem `sys.executable` with only `DETACHED_PROCESS|CREATE_NO_WINDOW` and no `STARTUPINFO`, which is exactly the combination that still flashes a console on Windows.

Preferring a `pythonw.exe` sibling *and* setting `STARTF_USESHOWWINDOW`/`SW_HIDE` is the correct pairing rather than either alone, and wrapping the `Popen` in a try/except so a spawn failure degrades to "telemetry did not flush" instead of breaking the user's command is the right posture for something this peripheral.

Two things before it can merge.

The two empty `fork-error.txt` and `fork-view-error.txt` files at the repo root need to come out. Same as on your #2381.

The other is tests. I know this is a Windows-only visual symptom that CI on Linux cannot observe, and I am not asking you to prove the window stopped flashing. But the new logic is branching code and can be tested as such: patch `os.name` and `sys.executable` and assert which executable `_flusher_executable()` picks, that it falls back cleanly when no `pythonw.exe` sibling exists, and that `startupinfo` carries the hide flags when `os.name == "nt"`. Right now the existing telemetry tests monkeypatch `_spawn_flusher` out entirely, so nothing exercises any of this.

Worth knowing: CI has never actually run on this PR. It sits at `action_required` because workflow runs on fork PRs need a maintainer to approve them, so nothing here is CI-verified yet, not even lint. I will get that approved.

RaghavChamadiya:
@pamod-madubashana thanks for turning both asks around. The junk files are gone and the tests cover exactly the branches I named. Two things before this can move, and the first is a correction of my own comment above.

**I got the flags wrong on the 20th.** I called `pythonw.exe` plus `SW_HIDE` "the correct pairing", but the flag combination this PR keeps is the actual bug, and I had already said so on #2222 six days earlier. `emitter.py` still sets:

```python
kwargs["creationflags"] = 0x00000008 | 0x08000000  # DETACHED_PROCESS | CREATE_NO_WINDOW
```

Microsoft documents that `CREATE_NO_WINDOW` is ignored when combined with `DETACHED_PROCESS`. So the flusher starts with no console at all, and when `sys.executable` is a venv or `uv tool` trampoline `python.exe`, the trampoline launches the real interpreter, a console app with no console to inherit, and Windows gives it a new visible one. That is the flash. With `CREATE_NO_WINDOW` alone the flusher gets a hidden console that its child inherits. Preferring `pythonw.exe` does sidestep it when the sibling exists, but the fallback path, which is exactly the install types where it does not, still flashes. `DETACHED_PROCESS` is not what keeps the flusher alive after the parent exits either; a Windows child outlives its parent without it.

**Claim order.** @codessensei claimed #2222 on 13 September with this diagnosis and a ready commit, and I confirmed the claim on the 14th. They have not opened the PR since. I missed that when I reviewed yours, which is on me, not you. I have pinged them on #2222. If they do not pick it up in the next few days, this PR goes in once it drops `DETACHED_PROCESS` (keep your `pythonw.exe` preference, it does no harm) and a test asserts the flags are `CREATE_NO_WINDOW` alone.

PR #2381, opened 2026-09-18
Fix #2379: Classify `.env` dotfiles as config paths

### LINKED ISSUE #2379: Classify `.env` dotfiles as config paths
## Summary

`CONFIG_EXTENSIONS` in `packages/core/src/repowise/core/support_paths.py` explicitly lists `".env"` as a config extension, but `is_doc_or_config_path()` matches on `PurePosixPath(path).suffix` only — and `pathlib` reports an empty suffix for dotfiles (`PurePosixPath('.env').suffix == ''`). As a result the file literally named `.env` (at any depth) is never recognised as config, while `foo.env` is.

The same suffix-only idiom is duplicated in the knowledge-graph node classifier (`_classify_file_type` in `packages/core/src/repowise/core/analysis/knowledge_graph.py`), which therefore labels `.env` a generic `"file"` instead of `"config"`.

## Steps to reproduce

```bash
PYTHONPATH=packages/core/src python3 -c 'from repowise.core.support_paths import is_doc_or_config_path as f, file_population; print(is_doc_or_config_path(".env"), is_doc_or_config_path("proj/.env"), is_doc_or_config_path("foo.env")); print(file_population(".env", is_test=False))'
```

## Expected vs actual

- Expected: `is_doc_or_config_path('.env')` → `True` (it is the canonical file the `".env"` entry names), `is_doc_or_config_path('proj/.env')` → `True`, and `file_population('.env', is_test=False)` → `"doc"`.
- Actual: `False`, `False`, `True` (for `foo.env`), and `file_population('.env', …)` → `"production"` — so `.env` files are treated as production code by every surface that hides non-production (entry points, tours, module grouping).

## Evidence

- `packages/core/src/repowise/core/support_paths.py:75-80` — `CONFIG_EXTENSIONS` contains `".env"`.
- `packages/core/src/repowise/core/support_paths.py:84-87` — `is_doc_or_config_path` checks only `.suffix`, which is `''` for `.env`.
- `packages/core/src/repowise/core/analysis/knowledge_graph.py:167-172` — `_classify_file_type` repeats the suffix-only check (`ext in _CONFIG_EXTENSIONS`), so `.env` falls through to `"file"`.
- Observed output: `[('.env', False), ('proj/.env', False), ('foo.env', True), ('README.md', True)]` and `file_population('.env') == 'production'` vs `file_population('foo.env') == 'doc'`.
- Dedup: searched existing issues and PRs (open and closed) for `is_doc_or_config_path`, `CONFIG_EXTENSIONS`, `.env config classification`, `dotfile suffix`, `file_population` — no plausibly related report (the only `dotfile` hits are unrelated feature/semantic matches).

pamod-madubashana:
Fixes #2379

# PR Summary — Issue #2379: Classify `.env` dotfiles as config paths

## What changed
- `packages/core/src/repowise/core/support_paths.py` — `is_doc_or_config_path()` now falls back to matching the lowercased basename (`PurePosixPath(path).name`) against `CONFIG_EXTENSIONS`, in addition to the existing suffix check. `pathlib` reports an empty suffix for dotfiles, so the literal file `.env` (at any depth) previously never matched even though `".env"` is listed in `CONFIG_EXTENSIONS`.
- `packages/core/src/repowise/core/analysis/knowledge_graph.py` — `_classify_file_type()` gets the same basename fallback (`p.name.lower() in _CONFIG_EXTENSIONS`), so `.env` classifies as `"config"` instead of falling through to `"file"`.

No other behavior changes: `foo.env` still matches via suffix, `README.md`/doc handling is untouched, and the fallback only fires when the suffix check misses (it generalizes to any future dotfile-style entry added to `CONFIG_EXTENSIONS`).

## Why it addresses the issue
- `is_doc_or_config_path(".env")` and `is_doc_or_config_path("proj/.env")` now return `True` (were `False`).
- Consequently `file_population(".env", is_test=False)` returns `"doc"` instead of `"production"`, so `.env` is hidden by every surface that hides non-production (entry points, tours, module grouping).
- `_classify_file_type(".env", "", False)` now returns `"config"` instead of `"file"`.

## Verification
- Project test command per `Makefile`: `uv run pytest tests/ -v` (unit subset: `uv run pytest tests/unit/ -v`). `uv` and `pytest` are not installed in this unattended environment (`uv: not found`, system `python3 -m pytest`: `No module named pytest`, HostedToolCache Pythons likewise have no pytest), so the pytest suite could not be executed here — stated plainly rather than implying a test run.
- Direct reproduction (the issue's repro, run with the repo's own code on `PYTHONPATH`) was executed and passes:
  - Command: `PYTHONPATH=packages/core/src python3 -c 'from repowise.core.support_paths import is_doc_or_config_path as f, file_population; print(f(".env"), f("proj/.env"), f("foo.env"), f("README.md")); print(file_population(".env", is_test=False))'`
    Result: `True True True True` / `doc` (before fix: `False False True True` / `production`).
  - Command: `PYTHONPATH=packages/core/src python3 -c 'from repowise.core.analysis.knowledge_graph import _classify_file_type as c; print(c(".env", "", False), c("proj/.env", "", False), c("foo.env", "", False), c("README.md", "", False), c("src/main.py", "python", False))'`
    Result: `config config config document file` (before fix `.env` → `file`).
- Relevant existing tests identified (not runnable here): `tests/unit/test_support_paths.py`, `tests/unit/analysis/test_knowledge_graph.py` (`TestFileTypeClassification`), `tests/unit/generation/test_categories.py`.


RaghavChamadiya:
Thanks @pamod-madubashana for filing #2379 with the mechanism already worked out rather than just the symptom. The diagnosis is right, and I confirmed it on current main: `is_doc_or_config_path` tests `PurePosixPath(path).suffix.lower()`, and a dotfile like `.env` has no suffix at all, so the extension check never fires and it falls through as code.

Three things before I can merge this.

The blocking one is that the PR adds two empty files at the repo root, `fork-error.txt` and `fork-view-error.txt`. They look like leftovers from whatever tooling produced the branch. Same two files are in your #2380. They need to come out of both.

Second, there are no tests. This is a classification change with an obvious over-reach failure mode, where a fix that matches any dotfile turns `.gitignore` into config too, so I want the negative cases pinned: `.gitignore` stays production, and `foo.env` still classifies by extension rather than by name.

Third, non-blocking but worth doing while you are here: the patch repeats the same name check inline in both `support_paths.py` and `knowledge_graph.py`. Those two have already drifted once, which is the whole reason this bug exists in two places, so a single shared helper that both call would stop it happening a third time.

I should be open with you about the situation around this one. @zhongxiao-chang independently opened #2449 for the same issue two days later, with the shared helper and the regression tests this one is missing. You filed the issue and opened a PR four minutes later, which is about as clear a claim as it gets, and in this repo the claim binds, so this is yours to finish rather than something I am going to merge out from under you. Push those changes and it goes in.

PR #2385, opened 2026-09-18
fix(test): stop two wall-clock assertions flaking on shared runners

sloemo01:
Two wall-clock assertions have been flaking on shared runners, each for a different reason. Both are fixed at the cause rather than by loosening a threshold.

## 1. `_meta.timing_ms` was rounded by a different rule from the rest of the payload

`tests/unit/server/mcp/test_number_precision.py::test_real_tool_payload_rounds_a_seeded_raw_double` failed on CI with:

```
raw doubles in get_risk: [('$get_risk._meta.timing_ms', 1077.85)]
```

The envelope wrote `round(float(timing_ms), 2)`. Every other float on that wire goes through `_rounding.round_float`, which rounds to *significant digits*. For a duration in the thousands of milliseconds the two rules disagree: `round(1077.85, 2)` keeps a fractional part, while `round_float(1077.85)` is `1078.0`.

So the field tripped the no-raw-doubles guard whenever a tool call landed off an integer millisecond, which is most of the time under load and never on a fast machine. That is why it looked like a flake and why it kept coming back. `timing_ms` now goes through `round_float` like everything else. That helper returns `None` for a non-finite measurement, and since `null` is not a valid duration on this wire the key is omitted in that case rather than emitted.

Demonstrated on the exact value CI caught:

```
old round(1077.85, 2)    = 1077.85   guard accepts: False
new round_float(1077.85) = 1078.0    guard accepts: True
```

## 2. The ledgered rewrite budget judged a single run of twelve subprocess spawns

`tests/unit/distill/test_rewrite_perf.py::test_a_ledgered_invocation_stays_under_budget` asserts a 200 ms median and measured 208.8 ms. The test already takes a median precisely so one slow spawn cannot decide the run, but on a loaded runner the whole batch of twelve is slow together, which the median cannot absorb.

The budget stays at 200 ms. When the first run's median is already over, it is measured once more and the better of the two is judged. A real regression repeats across both runs; a scheduling blip does not. The failure message still prints the full distribution from the run being judged, so an outlier stays visible rather than hidden behind one number.

## Verification

- `tests/unit/server/mcp/test_number_precision.py`: 33 passed. `ruff check` clean.
- `tests/unit/distill/test_rewrite_perf.py`: run and result below.
- The timing fix is verified by direct comparison rather than by sabotage, since a sabotage of a timing-dependent guard can pass by luck. The two lines in section 1 are that comparison.

Both failures were seen on PRs whose diffs touch neither file, which is what pointed at these assertions rather than at the code under review.


RaghavChamadiya (approved):
Thanks @sloemo01 for fixing the rounding at the source rather than loosening the assertion. `build_meta` was doing a bare `round(float(timing_ms), 2)` at `_meta.py:406` while `round_float` already exists at `mcp_server/_rounding.py:51` and does significant-digit quantizing, so the test was not flaky so much as asserting against the wrong rounding. Routing it through the shared quantizer means the next timing field added gets the right behaviour without anyone remembering.

Leaving the budget unchanged is the right restraint. A flake fix that quietly raises the ceiling stops being a guard.

PR #2386, opened 2026-09-18
fix(rust): widen type_use captures to wrapper shapes 

### LINKED ISSUE #2298: [Bug] Rust type_use misses &Foo, Box<Foo> and std::io::Error, though the extractor already handles them

Aman-goel-04:
## Summary

- Widen the @param.type captures in rust.scm for parameter and return-type positions to accept the wrapper node itself instead of requiring a bare type_identifier, so &Foo, impl Foo and std::io::Error reach the existing head extractor
- Add a capture for a bare type argument inside type_arguments, covering Box<Foo>'s inner type (the outer Box head is filtered as a builtin, so the inner type needs its own capture)
- Remove the now-redundant &dyn Trait / impl Trait special-case query patterns, since the broadened parameter capture subsumes them

## Related Issues

Fixes #2298

## Test Plan

- [x] Tests pass (pytest tests/unit/ingestion/test_rust_type_use.py, pytest tests/unit/ingestion -k rust, 135 passed)

Measured on goose (576 Rust files) to confirm the scope-widening guard holds: 
- type_use edges: 199 -> 487 (+288)
- calls edges: 19219 -> 19219 (unchanged)

## Checklist


sloemo01:
Verified independently, in case it saves a review cycle.

I ran the new tests with the change, then ran them again with only `queries/rust.scm` reverted to this branch's merge base. That keeps your tests and removes your query change, so anything that fails is genuinely pinned by the change rather than by the diff as a whole.

| state | result |
|---|---|
| with the query change | 27 passed |
| `rust.scm` reverted | **4 failed, 23 passed** |

The four that fail without it are precisely the wrapper shapes:

```
FAILED TestRustWrapperShapesReachTheResolver::test_reference_parameter
FAILED TestRustWrapperShapesReachTheResolver::test_scoped_parameter
FAILED TestRustWrapperShapesReachTheResolver::test_reference_return_type
FAILED TestRustWrapperShapesReachTheResolver::test_scoped_return_type
```

The part I checked hardest is the one that passes in both states: `test_the_generic_constructor_itself_is_not_captured`. Widening a capture query is the usual way to start resolving the wrapper instead of the wrapped type, and that does not happen here.

Branch is one commit behind main, so it should still apply cleanly. To reproduce:

```
git checkout origin/main -- packages/core/src/repowise/core/ingestion/queries/rust.scm
uv run pytest tests/unit/ingestion/test_rust_type_use.py -q
```

RaghavChamadiya (approved):
Thanks @Aman-goel-04 for supplying the numbers rather than asserting the widening was safe. `type_use` edges 199 -> 487 with `calls` edges unchanged at 19219 across goose's 576 Rust files is exactly the pair I asked for, and the second half is the one that matters: the scope-widening failure mode shows up as new call edges, and there are none.

Checked the two hazard sites on current main. The fold into `import_targets` is still at `graph/builder.py:482-485` guarded on `not data.get("no_scope_widening")`, and the guard itself is still set at `type_ref_resolution.py:872`. This PR touches only `queries/rust.scm` and its test file, so the newly captured wrapper shapes reach the resolver through the same gate every other `type_ref` already goes through. There is no path in the diff that gets around it.

`test_the_generic_constructor_itself_is_not_captured` is the right negative case, since the obvious wrong version of this change captures `Box` alongside `MyType`.

PR #2391, opened 2026-09-18
fix(mcp): keep pathless symbol rows out of get_answer's homonym union

### LINKED ISSUE #2346: [Bug] get_answer can serve a symbol row with an empty file path
**Summary**

`by_name` is built straight from the query results with no path guard.

`packages/server/src/repowise/server/mcp_server/tool_answer/symbols.py:429-431`

```python
by_name: dict[str, list] = {}
for row in res.scalars().all():
    by_name.setdefault(row.name, []).append(row)
```

Both `homonyms["union"]` branches, at `:481` and `:487`, then call `_verified_dict(c)` on every candidate with no `file_path` check.

**Mechanism**

`_verified_dict` (`:441-449`) does `_text_cache[row.file_path] = _read_repo_text(repo_root, row.file_path)`. For a row whose `file_path` is empty that reads with an empty path, and the reply can name an empty string as a file to open.

The same file already guards this case on the other path. The anchor loop at `:497-500` reads:

```python
for sym in chosen:
    fp = sym.file_path
    if not fp:
```

with a comment saying why a pathless row must not win a rank. The union branches have no equivalent.

**Repro**

Compare the two paths in the same file. A `WikiSymbol` row with `file_path == ""` that matches a homonym query is dropped at `:498` before it can become an anchor, and kept at `:481` and `:487` where it reaches `_verified_dict`. Nothing between `:429` and `:487` filters it.

**Impact**

Under-specified rather than wrong: the reply names a path that cannot be opened. How often a pathless row exists is index-dependent and **unsized** here. It has been seen on some snapshots and not on a fresh build, so treat the frequency as unknown and the mechanism as certain.

**Done looks like**

Skip rows with an empty `file_path` where `by_name` is built at `:429-431`, so no pathless row reaches `_verified_dict` by any route.

Worth checking in the same pass: with the shared guard in place the `if not fp` check at `:498-500` becomes redundant, and it is the clearest signal of whether any other consumer of a `WikiSymbol` row is missing the same check.

Tests: a case in the `get_answer` symbol-resolution suite seeding a homonym set that includes a row with `file_path=""`, asserting the union branch never reads with an empty path and the row does not appear in the reply.

8rulerstar:
## Summary

- Skip `WikiSymbol` rows with an empty `file_path` when `by_name` is built, so a pathless row can't reach `_verified_dict` through either union branch.
- Kept the `if not fp` guard in the anchor loop, as you suggested — thank you for explaining why it's worth keeping.
- One side effect worth flagging: when a homonym's only twin is pathless, the real definition now anchors on its own instead of forming a union.

## Related Issues

Fixes #2346

## Test Plan

- [x] `uv run pytest tests/providers/ tests/unit/`: no new failures compared to `main`
- [x] `ruff check .`
- New tests cover both union branches (narrowed and bare) and fail without the fix.

## Checklist

- [ ] I have updated documentation if needed (not needed)

Thanks again for the detailed guidance. I'm happy to change anything you'd like.


RaghavChamadiya (approved):
Thanks @8rulerstar. Dropping pathless rows once at `by_name` construction is the narrowest place that covers the anchor and both union branches, and the tests drive the narrowed branch and the bare branch separately, which is exactly the gap I asked about. Spying on `_read_repo_text` to assert an empty path is never read proves the mechanism is gone, not just that the output looks right.

PR #2392, opened 2026-09-18
fix(mcp): report an unknown search_codebase mode instead of coercing it

### LINKED ISSUE #2347: [Bug] search_codebase silently coerces an unrecognised mode to auto

8rulerstar:
## Summary

- Route `mode` through `resolve_enum_argument` right after `kind`, so an unknown mode is reported in `ignored_arguments` and the search still runs under `auto`.
- Moved the `ignored` list above the mode handling and lowercase a valid mode before validating, so `mode="Symbol"` keeps working. `kind` and `_resolve_mode`'s fallback are unchanged.
- Note: an empty string or a mode with surrounding spaces is now reported too, where it used to fall back silently.

## Related Issues

Fixes #2347

## Test Plan

- [x] `uv run pytest tests/providers/ tests/unit/`: no new failures compared to `main`
- [x] `ruff check .`
- New `TestSearchMode` cases cover a typo (reported as spelled), the `auto` fallback (including a path-shaped query), a valid mode, and case-insensitivity.

## Checklist

- [ ] I have updated documentation if needed (not needed)

Thank you for pointing out the ordering issue. Please let me know if you'd like anything changed.


RaghavChamadiya:
Thanks @8rulerstar, this does all three of the things I asked for on #2347 and I checked each one against the diff rather than taking the description's word for it.

Lowercasing before validating is the one that mattered. `_VALID_MODES` is all lowercase at `tool_search.py:925` and `resolve_enum_argument` compares with a plain `value in valid` at `mcp_server/_helpers.py:314`, so handing `mode` straight over would have dropped `mode="Symbol"` into `ignored_arguments` and silently fallen back to `auto`. Your tests pin both halves of that: `"Symbol"` still resolves, and `"Symbl"` is reported back in the spelling the caller typed rather than folded. The `ignored` list is above the resolve call, and `kind` is untouched.

CI had never actually run on this, because fork PRs need a maintainer to approve the workflow. That was on me and I have approved it now, so the matrix is building. It takes about 25 minutes. Assuming it comes back green I will merge it without another review pass.

RaghavChamadiya (approved):
Thanks for this, @8rulerstar. Reporting the unknown mode instead of quietly switching strategies is exactly right, and the tests pin both the "reported as spelled" case and the real `auto` fallback, which is the part that's easy to get wrong. Approving and merging. One heads-up: a follow-up will move the mode routing into a small `_query_shape` module, so `_VALID_MODES` will be imported from there. Nothing for you to change.

PR #2396, opened 2026-09-18
fix(generation): only prescribe ruff format when the repo declares it

### LINKED ISSUE #2384: [Bug] A comment mentioning "formatter" makes generated agent files prescribe `ruff format .`
## Summary

`detect_build_commands` decides a repository uses the ruff formatter if the strings `"ruff"` and `"format"` both appear anywhere in `pyproject.toml`. Because that is a substring test over the whole file, the word "formatter" in a comment is enough.

`packages/core/src/repowise/core/generation/editor_files/tech_stack.py:458-459`

```python
if "format" not in commands and "ruff" in text and "format" in text:
    commands["format"] = "ruff format ."
```

The result reaches `CLAUDE.md` and the other generated agent instruction files through `editor_files/fetcher.py:91`, where it is rendered as a `Format:` command. So this is not a suggestion a person skims, it is an instruction an agent reads and acts on.

## Demonstrated on this repository

This repo does not configure the ruff formatter. `pyproject.toml` has `[tool.ruff]`, `[tool.ruff.lint]`, `[tool.ruff.lint.isort]`, `[tool.ruff.lint.flake8-bugbear]` and `[tool.ruff.lint.per-file-ignores]`, and **no `[tool.ruff.format]` section at all**. Ruff is configured here as a linter.

The only two matches for `format` in the entire file are inside comments:

```
:195  # Ruff — linter + formatter
:217  "E501",    # line too long (handled by formatter)
```

Both describe ruff, neither declares a formatter. That is enough to switch the command on, and our own generated `CLAUDE.md` currently carries:

```
- Format: `ruff format .`
```

CI never runs `ruff format` (`.github/workflows/ci.yml` runs `uv run ruff check .` and nothing else), and the dependency range is wide enough that a fresh environment can resolve a newer ruff whose formatter rewrites hundreds of files. So the command we hand to every agent working on this repository produces a large unreviewable diff that nothing asked for and nothing checks.

## The codebase already contains the argument against it

This is the part that makes it a gap and not a disagreement about intent. A stricter detector already exists for the precedent layer, and its docstring names this function as the loose one:

`packages/core/src/repowise/core/precedent/structural.py:439-447`

```
Deliberately stricter than :func:`detect_build_commands`, whose format
inference fires on a ``pyproject.toml`` containing the words "ruff" and
"format" anywhere. That is fine for a suggested command and wrong as the
premise of a stored fact: ruff-as-linter beside black-as-formatter is a
common pairing, and the two disagree, so the inference would tell a
black-clean repo it is not formatter-clean by a formatter it never chose.
```

The reasoning is right and the conclusion drawn from it was too narrow. "Fine for a suggested command" holds when a person reads the suggestion and ignores it. It does not hold once the command is written into the file that instructs coding agents, because there the suggestion is the action.

`_declares_ruff_format` also refuses when a competing formatter is declared, via `_COMPETING_FORMATTERS = ("[tool.black]", "[tool.blue]", "[tool.yapf]")`. `detect_build_commands` has no equivalent check; grepping `tech_stack.py` for `tool.black` returns nothing. So a project running black beside ruff is currently told to format with ruff, which actively fights the formatter it did choose.

## Scope beyond this repo

Any Python project that uses ruff as a linter and happens to write "formatter" in a comment, a tool description or a disabled-rule note gets the same wrong command. "Linter + formatter" is a natural way to describe ruff in a comment, so the false-positive shape is a common one and not an exotic one.

## Not sized

I have not measured how many repositories in the corpus this fires on incorrectly. The mechanism is certain and demonstrated here; treat the frequency as unknown.

## Done looks like

The format command is emitted only when the repository actually declares the ruff formatter, using the test that already exists instead of a second one. `_declares_ruff_format` checks for a real `[tool.ruff.format]` section, falls back to an explicit `ruff format` invocation in `Makefile`, `package.json`, `.pre-commit-config.yaml` or `justfile`, and refuses outright when a competing formatter is declared. That is the behaviour wanted here.

Reuse it instead of reimplementing, so the two paths cannot drift into disagreeing about the same repository. It currently lives in `precedent/structural.py`; moving it somewhere both callers can reach, or having `tech_stack.py` call it, are both fine and the choice is an implementation detail.

Worth a look in the same pass: the sibling inferences directly above are substring tests of the same shape (`"pytest" in text` for the test command, `"mypy" in text` for typecheck). They are less damaging, since running the wrong test command fails loudly instead of rewriting the tree, but if one of them is wrong for the same reason it is cheap to fix while the file is open.

## Related

Same defect class as #2142, where the heading harvest mined `# Install dependencies` out of a fenced bash block as house vocabulary: a prose-blind substring match standing in for a structural check.

JikSau-D:
Closes #2384.

`detect_build_commands` turned the `Format: ruff format .` command on for any `pyproject.toml` containing both `ruff` and `format` as substrings, so a comment like "Ruff - linter + formatter" was enough. Generated agent files render that line as an instruction, and a repo that runs black beside ruff was told to format with a tool it never chose.

This reuses the stricter declaration test that already existed for the formatter-drift episode instead of keeping a second, looser one:

- moves `_declares_ruff_format` out of `precedent/structural.py` into `core/repo_formatters.py` as `declares_ruff_format`, so the suggested command and the stored fact read the same declaration and cannot drift apart
- `tech_stack.py` now calls it instead of the substring check: the command is emitted only for a real `[tool.ruff.format]` section or an explicit `ruff format` / `ruff-format` invocation in the Makefile, package.json, `.pre-commit-config.yaml` or justfile, and a declared competing formatter (black, blue, yapf) suppresses it
- regression tests cover the comment-only false positive, the declared section, the competing formatter, and an explicit pre-commit hook

Verification:

- `uv run pytest tests/unit/generation tests/unit/precedent tests/unit/core` - 1762 passed
- `uv run ruff check packages/ tests/` - clean

I left the sibling `pytest`/`mypy` substring inferences alone, per the issue's "worth a look": they fail loudly rather than rewriting the tree, and can be a separate change.


RaghavChamadiya:
@JikSau-D this one is going to close without merging, and I want to explain why in full, because it is not about the quality of the change.

Your extraction is the cleaner of the two. Pulling the check out into a neutral `core/repo_formatters.py` and having both `precedent/structural.py` and `generation/editor_files/tech_stack.py` import from there avoids creating a `generation` -> `precedent` dependency, and it matches the absolute-import convention the sibling files in that directory already use. You also had one more regression case than the version that landed. I read both diffs side by side and that is a fair reading of them.

The reason it is not merging is claim order. @Aman-goel-04 commented on #2384 at 10:14 on 18 September naming `tech_stack.py:458-459`, quoting the exact `"ruff" in text and "format" in text` mechanism, and saying he intended to reuse `_declares_ruff_format` rather than reimplement it. That is a substantive claim, and he was the assignee. Your PR opened at 13:13 the same day, about three hours later, and the only comment on the issue from you is the one announcing the PR. In this repo the first substantive claim binds, and quality only breaks ties between claims, not between a claim and a later PR. If I merged the better patch here I would be telling everyone that writing up a plan and waiting is worth less than going straight to code, and the people who lose that race are exactly the ones I cannot afford to lose.

So this is on me for not getting to #2384 faster, not on you.

Two things I would genuinely like. The shared-module placement is better than what we merged, and I would take a follow-up PR that moves the helper out of `precedent/structural.py` into a neutral module and updates both importers, as a standalone refactor. Your extra pre-commit-hook case is worth folding in at the same time. Second, if you want something unclaimed right now, #2350 is open and is a real design question rather than a mechanical fix: it asks for `__all__` membership to be a *bounded* signal on Python symbol visibility, with the cap argued in the PR rather than assumed, and nobody has taken that argument on yet.

Sorry for the outcome on this one.

RaghavChamadiya:
Superseded by #2444, which merged just now. The reasoning is in the comment above, along with the follow-up I would like from this branch.

PR #2402, opened 2026-09-18
fix(ingestion): resolve a nested sub-package from its own manifest 

### LINKED ISSUE #2301: [Bug] A workspace sub-package with its own package.json resolves to nothing

Aman-goel-04:
## Summary

- Add _resolve_nested_subpackage as a new fallback in resolve_via_workspaces, checked only after every existing probe on the outer package has failed
- Reads <dir>/<sub>/package.json directly and resolves the subpath as its own package (exports field, then index, then main/module, then a source-root index), rooted one level deeper than the outer package, mirroring the existing bare-package resolution
- _probe_path itself is untouched, so every other caller keeps its current behavior with no added file-read cost

## Related Issues

Fixes #2301

## Test Plan

- [x] Tests pass (pytest tests/unit/ingestion/test_typescript_resolver.py, pytest tests/unit/ingestion -k "typescript or workspace or ts_", 655 passed)

Added TestNestedSubPackage (5 tests) modeled on the real solid-js/packages/solid/web shape from the issue: own src entry, main field, exports field, plus two negative tests confirming an already-resolving subpath is never retargeted and a missing manifest still resolves to None.

## Checklist


sloemo01:
Nice work landing this.

The ordering is the right call: checking the nested manifest only after every probe on the outer package has failed means the new fallback cannot change a resolution that already worked. That is what makes it safe to slot into a path other people depend on, and it is the part a reviewer would otherwise have to check line by line.

One thing that may be useful. I had a branch on this same issue before it was closed for the claim (`fix/tsws-nested-package`, still on my fork at sloemo01/repowise). It is larger than this one, so it may carry a case yours does not. Take whatever is useful from it and ignore it otherwise, no expectation either way.

RaghavChamadiya (approved):
Thanks @Aman-goel-04 for ordering this defensively. Slotting `_resolve_nested_subpackage` in only after `spare_export_target()` has already returned `None` means the new fallback can never override a resolution the existing probes already made, which is what makes it safe to drop into a path other people depend on. The negative test that pins exactly that (`test_a_subpath_that_already_resolves_is_not_retargeted`) is the one I would have asked for.

Verified on current main that `_resolve_nested_subpackage` does not exist and `resolve_via_workspaces` has no fallback past the outer package's own `src`/`lib`/`dist` probes, so this is the gap the issue described. `_probe_path` is untouched, so no other caller changes behaviour.

PR #2403, opened 2026-09-18
fix(cli): read workspace list from the configured database

### LINKED ISSUE #2368: [Bug] workspace list reads every repo as "not indexed" under a configured shared database
## Summary

`repowise workspace list` decides whether a repository is indexed from the presence of local files, so under a configured shared database every member reads as "not indexed" with zero counts, however well indexed it actually is.

`packages/cli/src/repowise/cli/commands/workspace_cmd.py:150`

```python
if not repowise_dir.exists():
    table.add_row(label, rel_path, "-", "-", "-", "[yellow]not indexed[/yellow]")
    continue
```

and `_query_repo_counts` at `:196-197`

```python
db_path = get_repowise_dir(repo_path) / "wiki.db"
if not db_path.exists():
    return 0, 0
```

Neither consults `get_configured_db_url()`. Grepping the whole file for `get_configured_db_url` or `db_configured` returns nothing, so there is no shared-database branch anywhere in it.

## Consequence

With `REPOWISE_DB_URL` set and no local `.repowise/wiki.db`, every row prints `-` for files, symbols and indexed time, with the status `not indexed`. The summary line beneath the table is built from the same counter, so it reports `0/N repos indexed` for a fully indexed workspace.

That also makes the output actively misleading and not merely empty: the recommended next step a user draws from it is to reindex, which is unnecessary work against a database that already holds the pages.

This is the fourth instance of the same bug class, and the others are already known: #1034 (workspace repos not reconciled with the configured database), #2148 (workspace watch under PostgreSQL), and #2320 (`repowise delete` requiring a local `wiki.db`). In each case a local `.repowise/` directory or a local `wiki.db` is used as a proxy for "this repository is indexed", which stops being true the moment a shared database is configured.

`repowise delete` had exactly this shape and the fix there was to gate the local-file checks behind the existing `db_configured()` helper at `packages/cli/src/repowise/cli/helpers.py:306-314`. The same helper applies here.

## Done looks like

`workspace list` asks where the database is before deciding what a missing local file means. When a shared database is configured, the indexed check and the counts come from it; when it is not, the current local-file behaviour is unchanged.

Worth handling in the same pass: `indexed_count` feeds the summary line, so it needs to move with the per-row decision, and `check_repo_staleness` is called only on the indexed branch, so rows that currently short-circuit to "not indexed" have never had their staleness evaluated either.

## Scope note

I have verified the reading path only. I have not checked whether other `workspace_cmd` subcommands share the assumption, and that is worth a look while the file is open.

sloemo01:
Closes #2368

## What was wrong

`repowise workspace list` decided whether a repository was indexed from `repowise_dir.exists()` alone, and `_query_repo_counts` opened a local `wiki.db` whatever store the repository had actually been indexed into. Neither consulted `get_configured_db_url()`, so under a configured shared database every member read as "not indexed" with zero counts, however well indexed it was.

## The fix

The directory check is gated on `db_configured()`, the precedent #2338 set, rather than a second convention for the same problem:

```python
counts = _query_repo_counts(abs_path)
indexed = counts is not None if configured_db else repowise_dir.exists()
```

With a shared database configured the store answers: a repository row for this path is the index, and its absence is the not-indexed verdict. With no shared database configured the directory check and the counts are unchanged.

`_query_repo_counts` now returns `None` for "no store could answer", which keeps apart two cases that both used to return `(0, 0)`:

- a shared store holding nothing for a repository still means **indexed**
- an existing local `.repowise/` with no readable store still reports **zero counts**

## Tests

Six cases in `tests/integration/test_cli.py`:

| test | asserts |
|---|---|
| `test_shared_db_repos_read_as_indexed` | a member with no local directory reads as indexed under a configured DB |
| `test_shared_db_repos_counts_are_scoped_per_repository` | counts come from that repository's rows, not the workspace total |
| `test_repo_absent_from_the_shared_db_reads_not_indexed` | no row is still the not-indexed verdict |
| `test_a_local_directory_is_not_the_verdict_when_a_db_is_configured` | a stray local `.repowise/` does not rescue an unindexed repository |
| `test_local_repos_without_a_configured_db_are_unchanged` | the default path is untouched |
| `test_local_store_keeps_its_empty_verdict` | an existing store with nothing in it still shows zeros |

Checked by reverting the source change and keeping the tests: the three shared-database cases fail reading "not indexed", and the three local-behaviour cases still pass.

`ruff check` and `ruff format --check` clean on both files, and all 58 tests in the file pass.


sloemo01:
Read the three checks; none of them changes anything here, and one is worth answering explicitly rather than leaving silent.

**1. The three files that "usually change" with these.** `init_cmd/command.py`, `job_executor.py` and `helpers.py` do not belong in this PR. The change is a test port: the two assertions from #2490 that #2403 did not have, moved into `tests/unit/cli/test_workspace_list_store_states.py`, plus the `test_cli.py` coverage that came with this branch. `workspace_cmd.py` carries the fix; nothing in the init or job-executor path is touched by it, and adding files to match a coupling heuristic is how a diff stops being reviewable.

**2. The cohesion finding in `TestWorktreeInitForce`.** That class is not introduced by this PR; `f8a264d` is the commit the bot cites, and this branch does not touch it. Nothing to do here either.

**3. The suggested order to run tests.** The unit tier is what gates this PR, and it is green: `test_workspace_list_store_states.py` 3 passed. The integration tier does not run on pull requests in this repo, which is why the same two assertions exist at the unit tier as well; that duplication is deliberate and was the point of the port.

Health moved 2.7 to 3.0 in the bot's own readout, which is the direction that matters here.

Issue #2148, opened 2026-09-06
[Bug] Workspace watch under PostgreSQL full-reindexes every change and leaks Postgres connections


T012m3n7oR:
## Describe the Bug

Two tightly coupled defects show up together under `REPOWISE_DB_URL` (shared PostgreSQL) with `repowise watch --workspace`:

### 1. Connection leak / pool exhaustion (symptom that surfaces first)

Long-running watch (alongside `serve` / `mcp` on the same DB) eventually fails with:

```text
asyncpg.exceptions.TooManyConnectionsError:
remaining connection slots are reserved for roles with the SUPERUSER attribute
```

That is a **server-side** exhaustion of non-superuser slots (`max_connections − superuser_reserved_connections`). It is not a client QueuePool timeout.

**Why connections accumulate**

Postgres engines are created via `create_engine()` with **no** `poolclass` / `pool_size` override:

```python
# packages/core/src/repowise/core/persistence/database.py
else:
    # PostgreSQL — asyncpg handles its own connection pool   # misleading
    kwargs["pool_pre_ping"] = True

engine = create_async_engine(db_url, **kwargs)  # → AsyncAdaptedQueuePool defaults
```

SQLite short-lived engines correctly use `NullPool`. Postgres short-lived engines get SQLAlchemy’s default pool (**`pool_size=5`, `max_overflow=10` → up to 15 server connections per engine**). asyncpg does **not** own a separate pool here; SQLAlchemy does.

Every workspace update opens one or more of these engines. If an engine is not `dispose()`d in the same event loop that created it — or is abandoned across `asyncio.run()` boundaries — pooled asyncpg connections stay open on Postgres until the backend times them out. That is a **connection leak** (client-side ownership lost; server-side slots remain occupied). The same lifecycle class is documented in #2062 for the cost-tracker path; watch/update creates many more engines under the gate below, so the leak is amplified and eventually hits `TooManyConnectionsError`.

### 2. Incremental updates never run (root amplifier)

Under Postgres, per-repo indexes live in the shared DB and **no** `<repo>/.repowise/wiki.db` is created. `update_single_repo_index` still requires that file before taking the incremental path:

```python
# packages/core/src/repowise/core/workspace/update.py — update_single_repo_index
if (
    not config_changed
    and base_ref
    and (repo_path / ".repowise" / "wiki.db").is_file()  # always false on Postgres
    and commit_exists(repo_path, str(base_ref))
):
[...]

PR #2404, opened 2026-09-18
fix(update): treat a configured database as a store

sloemo01:
This one is unfiled as far as I can tell, so the write-up lives here.

## The bug

Both head-commit writers guarded on the repo-local `wiki.db` file:

| where | guard |
|---|---|
| `core/workspace/update.py` `reconcile_repo_head_commit` | `not (repo_path / ".repowise" / "wiki.db").is_file()` |
| `cli/update_cmd/persistence.py` `heal_commit_offsets` | `not (root / ".repowise" / "wiki.db").is_file()` |

Both then build their engine from `resolve_db_url()`, whose first resolution step is `REPOWISE_DB_URL`. Under a shared database the repo-local file is absent by design, so the guard returns before the engine is ever built.

## What it costs

**The visible half.** `reconcile_repo_head_commit` is the single head-commit stamper for both update paths. `/api/repos`, MCP `_meta` and the health overview's `last_indexed_at` fallback all read `head_commit` and `updated_at` from that row. Under Postgres, `repowise update` succeeds and the row keeps its old values, so the "index behind checkout" signal stays stuck and the freshness timestamp freezes at the last full index.

**The quieter half.** `heal_commit_offsets` never runs at all, so `git_commits.committed_offset_minutes` is never backfilled on a Postgres-backed repo. The punch card stays UTC-only until someone re-indexes.

## The fix

`has_db_store()` sits next to `resolve_db_url` and asks the question that function already answers:

```python
if get_configured_db_url() is not None:
    return True
if repo_path is None:
    return False
return (Path(repo_path) / ".repowise" / "wiki.db").is_file()
```

The guarantee the old guard was protecting is that a stamp must never conjure an empty database. That still holds, because a filesystem default counts only when the file is really there. A configured database counts because it exists independently of this repo, and conjuring is not the risk.

## Tests

`tests/unit/cli/test_update_stamp_configured_db.py`, eight cases:

- `has_db_store` tells apart the three states: configured, local file, neither
- the stamp lands in a configured database with no repo-local file
- the heal reaches a configured database instead of returning early
- the repo-local path still stamps, unchanged
- no store at all still no-ops, so nothing is conjured
- an absent or empty `head` is still ignored

Verified by reverting only the two guards while keeping the helper and the tests: exactly the two configured-database cases fail, and the six behaviour-preservation cases still pass.

`ruff check` clean, and 1250 tests across the update, workspace and MCP freshness suites pass.


RaghavChamadiya (approved):
Thanks @sloemo01. The negative test refusing to conjure an empty database when nothing is configured and no local file exists is the right guard, because the lazy version of `has_db_store()` returns True for a URL that was set but never pointed at anything real.

Verified on current main that `has_db_store` did not exist and that both call sites tested `(repo_path / ".repowise" / "wiki.db").is_file()` directly. That is the same local-file-as-indexed-signal shape behind #1034, #2148, #2320 and #2368, so folding it into one helper is the right direction rather than another per-site patch.

Worth saying out loud, non-blocking: this fixes two call sites, and the same `wiki.db` `is_file()` test is still open in `pipeline/incremental.py` at four places and `workspace/update.py` at four more. Not this PR's job, and I would rather have the helper landed and reused than have it grow to cover everything at once, but that is the remaining surface and it is worth its own issue so it does not get rediscovered a sixth time.

Issue #2389, opened 2026-09-18
[Bug] A Rust impl block reads as private although rust.scm says it inherits its target's visibility


RaghavChamadiya:
**Summary**

`rust.scm` states that an `impl` block inherits its visibility from the type or trait it implements. It does not. No visibility is captured, the modifier text is empty, and every `impl` block in every Rust repository reads `private`, including the roughly 6,600 that implement a `pub` type.

**Mechanism**

The query captures the target type's name and nothing else, and says so in a comment that describes an inheritance that is never performed.

`packages/core/src/repowise/core/ingestion/queries/rust.scm:71-76`

```
; impl block — the "type" field identifies what is being implemented
; impl does NOT need visibility capture (inherits from type/trait)
(impl_item
  type: (type_identifier) @symbol.name
) @symbol.def
```

The other three `impl_item` patterns at `:77-95`, covering generic, scoped and reference target types, capture no `@symbol.modifiers` either. With no modifier text, `rust_visibility` returns `private`, so the "inherits from type/trait" claim resolves to a constant.

This is the same shape as the trait-item defect fixed in #2376, where Rust forbids a visibility modifier on a trait item and `rust_visibility` therefore read `private` for every method of a `pub trait`. That fix added `refine_rust_visibility`, which reads the enclosing `trait_item`'s own modifier and is wired into the existing cpp / csharp / ts refinement chain. It matches an exact node shape rather than walking ancestors, and it is the template here.

**Evidence**

Counted over the Rust corpus at the current commit:

| repo | `impl` symbols | targeting a `pub` type |
|---|---|---|
| serde | 367 | 213 |
| ripgrep | 442 | 276 |
| rtk | 68 | 43 |
| goose | 811 | 622 |
| bevy | 6,468 | 5,493 |

All 8,156 read `private`. On the figures above, 6,647 should read `public`.

**Impact**

Visibility is wrong on a large symbol population, which matters anywhere visibility is read: the public-surface view, `unused_export` eligibility, and any published-API work.

[...]

PR #2418, opened 2026-09-19
fix(dead-code): treat Rust `impl` blocks as uncallable and non-importable (#2388)

### LINKED ISSUE #2388: [Bug] An impl block is a callable symbol to both dead-code passes, and nothing can ever call one
**Summary**

A Rust `impl` block is emitted as a symbol named after its target type, and both dead-code passes treat it as a callable, deletable unit. Nothing calls an `impl` block by name, so it can never satisfy the call-edge test either pass applies. It is held out of both passes today only by accident, and each accident is one change away from ending.

**Mechanism**

`rust.scm` captures `impl_item` as a symbol, keyed on the target type's name. This is deliberate: the node is the parent container for the block's methods and a `heritage_node_types` member.

`packages/core/src/repowise/core/ingestion/queries/rust.scm:71-76`

```
; impl block — the "type" field identifies what is being implemented
; impl does NOT need visibility capture (inherits from type/trait)
(impl_item
  type: (type_identifier) @symbol.name
) @symbol.def
```

The kind that node carries, `impl`, is in neither exclusion set the dead-code analyzer applies:

* `_UNIVERSAL_NON_IMPORTABLE` (`analyzer.py:257-269`) holds `method`, `variable`, `field`, `property`, `enum_member`, `constant`, `type_alias`, `namespace`, `module`.
* `_UNCALLABLE_TYPE_KINDS` (`analyzer.py:351-358`) holds `struct`, `interface`, `enum`, `type_alias`.

So each pass excludes `impl` only incidentally:

* `_detect_unused_exports` skips a symbol whose visibility is not `public` (`analyzer.py:1201`). An `impl` block reads `private`, because Rust forbids a visibility modifier there and the query captures none, so it never reaches the gate that would flag it.
* `_detect_unused_internals` skips `language == "rust"` outright (`analyzer.py:1565-1569`), so the `private` reading that saves it from the first pass is exactly what would condemn it in the second.

**Evidence**

Removing the Rust language skip from `_detect_unused_internals` and diffing findings against baseline, both arms in one process, five repos:

| repo | gained `unused_internal` | of which kind `impl` |
|---|---|---|
| serde | 9 | 7 |
| ripgrep | 16 | 13 |
| rtk | 21 | 16 |
| goose | 56 | 38 |
| bevy | 114 | 101 |

Every one is a false positive. Graded from source: `impl std::fmt::Display for Error` (`crates/globset/src/lib.rs:243`) is dispatched by the compiler and deleting it breaks every `write!("{}", err)`; `impl GooseAcpAgent { ... }` (`crates/goose/src/acp/server/tools.rs:8`) is an inherent block holding live `pub(super)` methods. On goose, `GooseAcpAgent` alone fires 22 times, once per file that adds methods to that type.

**Impact**

No user-visible findings today: `unused_internal` is skipped for Rust, and `unused_export` is blocked by the `private` reading. `impl` is a Rust-only kind, so no other language is affected.

The reason to fix it is that it is load-bearing for anything that touches Rust visibility. The population is large:

| repo | `impl` symbols | targeting a `pub` type |
|---|---|---|
| serde | 367 | 213 |
| ripgrep | 442 | 276 |
| rtk | 68 | 43 |
| goose | 811 | 622 |
| bevy | 6,468 | 5,493 |

**This issue is a prerequisite for the visibility fix, and the order matters.** If an `impl` block is made to inherit its target's visibility, as `rust.scm:72` already claims it does, roughly 6,600 of the 8,156 symbols above become `public` and flow straight into `_detect_unused_exports`, where nothing can ever call them. Fixing visibility first converts a dormant defect into a live false-positive class across every Rust repository. Fix this one first and the visibility change is safe.

**Done looks like**

`impl` excluded from the callable population both dead-code passes consider, so that neither a visibility change nor lifting a language skip can expose it. `_UNCALLABLE_TYPE_KINDS` is the natural home: its docstring is precisely this argument, that a CALL-edge check is meaningless for the kind.

**Keep this scoped to the exclusion.** Whether `impl` blocks should be symbols at all is a separate and much larger question: the node is a `heritage_node_types` member and the parent of the block's methods, so removing it would orphan them. Do not touch the query.

Tests: a fixture asserting an `impl` block on a private type yields no `unused_internal` finding when the language skip is absent; and one asserting an `impl` block on a `pub` type yields no `unused_export` finding when its visibility reads `public`, which is the state the follow-up creates.

Note for whoever picks this up: `rust.scm` has two open PRs against it (#2327, #2386), both scoped to `type_use` captures. They do not overlap with this, but a rebase may be needed if you touch that file.

rishu685:
Fixes #2388

### Overview
In Rust, Tree-sitter captures `impl` blocks as AST symbol definitions with kind `"impl"`. These blocks represent structural containers holding method definitions and trait implementations for types; they cannot be invoked as call targets or imported directly by name (e.g. `use foo::impl;` is invalid).

Previously, `_UNCALLABLE_TYPE_KINDS` and `_UNIVERSAL_NON_IMPORTABLE` in `DeadCodeAnalyzer` omitted `"impl"`. As a result:
- In `_detect_unused_internals`, private `impl` blocks with no incoming `call` edges were treated as callable symbols and falsely flagged as dead internal code.
- In `_detect_unused_exports`, public `impl` blocks with no incoming named `imports` edges were susceptible to false positive unused-export findings.

### Changes
- **`packages/core/src/repowise/core/analysis/dead_code/analyzer.py`**:
  - Added `"impl"` to `_UNCALLABLE_TYPE_KINDS` so `_detect_unused_internals` excludes `impl` blocks from callable call-edge requirements.
  - Added `"impl"` to `_UNIVERSAL_NON_IMPORTABLE` so `_detect_unused_exports` excludes `impl` blocks from named export import requirements.
- **`tests/unit/dead_code/test_unused_internal.py`**:
  - Added `test_unused_internal_rust_impl_uncallable` to verify that a private `impl` block with no call edges is not flagged as `UNUSED_INTERNAL`.
- **`tests/unit/dead_code/test_unused_exports.py`**:
  - Added `test_unused_export_rust_impl_non_importable` to verify that a public `impl` block with no import edges is not flagged as `UNUSED_EXPORT`.

### Validation
- Validated that `rust.scm` remains untouched to preserve parent-child AST containment for methods and heritage types.
- Added comprehensive unit test coverage for both dead-code detection passes.


RaghavChamadiya (approved):
Thanks @rishu685 for setting visibility explicitly in the tests rather than letting them lean on the visibility fix in #2389. That keeps this PR independently landable, which matters here because #2389 says outright that landing it first without this exclusion would mint thousands of false `unused_export` findings. This is the half that has to go in first, and it does.

Verified on current main that neither `_UNCALLABLE_TYPE_KINDS` nor `_UNIVERSAL_NON_IMPORTABLE` carries `"impl"`.

One thing worth knowing rather than fixing: the `_UNCALLABLE_TYPE_KINDS` half is inert today, because `_detect_unused_internals` skips `language == "rust"` before it ever reaches the kind check. The `_UNIVERSAL_NON_IMPORTABLE` half is the one doing the work right now. Adding both is still right, it just means the new internals test would pass with or without that line. Not a blocker.

PR #2422, opened 2026-09-19
feat(decisions): flag a bundled claim on every capture path, not one lane

RaghavChamadiya:
Based on #2421.

## What this fixes

`needs_split` says a candidate joins two independent choices, so a reviewer
separates them rather than accepting one rule that is really two. Measured on
a byte copy of the dogfood store: **it is `False` for all 357 records**,
including the ones that visibly bundle.

Nothing was wrong with the detector. It is one line in
`discovery/grounding.py`, a pure function of the decision text, and it only
ever ran inside the `session_discovery` lane — which is off. Nothing asked it.

## The change

`SPLIT_MARKERS` and `bundles_decisions` move into
`analysis/decisions/lifecycle.py`, the pure vocabulary module both lanes
already resolve their words in, and `promotion_decisions` asks it.

It is asked **at promotion, not at staging**, for the same reason scope
binding and kind classification already are, and the file says so in the two
comments directly above: a row staged before the flag existed is judged on
its way out rather than staying unflagged forever. On this store that is the
difference between the flag applying to future rows only and applying to the
whole backlog.

| | n |
|---|---|
| staged rows (406) that flag | **62** |
| records (357) that flag | **29** |
| records flagged today | 0 |

The discovery lane's answer is kept and **only ever raised**. Both lanes ask
the same question of the same text, and which one ran first must not decide
the answer. The persistence layer already treats the flag the same way
(`crud/decisions.py:1320`, `needs_split=True if needs_split else None`), so
nothing walks back a split a reviewer asked for.

## The markers are not widened, and that is a measurement

| marker set | records held out of merging | duplicate pairs still collapsed | known bundles caught |
|---|---|---|---|
| none | 0 | 120 | 0 of 3 |
| **the three current markers** | **28** | **100** | **1 of 3** |
| widened with `" and "` | 157 | 44 | 2 of 3 |

Re-verified against the live store while writing this: the three markers flag
29 of 357 on decision text and 0 on titles; adding `" and "` flags 128 on
decision text and 66 on titles, 194 in all. That is 44% of the store held out
to catch one more bundle, and it takes 63% of the merge value with it.

The third known bundle (`Use explicit uninstall scope instead of a universal
confirmation flag`, which swallows `Add scoped repowise uninstall` and
`Reject a generic --yes uninstall flag`) joins its clauses with no
conjunction at all. **No marker set reaches it.** The guard is partial by
construction and that is accepted: a marker it misses costs a reviewer one
merge to undo, and one it invents costs a decision that never merges at all.

Porting turned out to be small — one shared predicate, one call, one dead
local removed — so the guard ships rather than being dropped.

## Why it ships before the identity change

Evidence-keyed identity folds records that share files and evidence. Three of
its 42 merge classes fold a bundled title together with the separate
decisions it bundles. This flag is the only signal that can hold those apart,
and it has to be non-zero before the merge exists.

## Tests

Five new cases in `tests/unit/analysis/test_decision_authority_containment.py`,
including the bare-conjunction case that pins the decision *not* to widen.
Mutation-tested, both caught:

- dropping the call in `promotion_decisions`: 1 failure
- widening `SPLIT_MARKERS` with `" and "`: 1 failure

`tests/unit/analysis` 2110 passed; `tests/unit/sessions` 206 passed; the four
decision suites in `tests/unit/persistence` 59 passed. `ruff check` clean.


RaghavChamadiya:
Amended after review. One real finding, fixed in the same commit: `bulk_upsert_decisions` read `needs_split` off the title group's *headline*, which is the highest-ranked source in the group. The lane that notices a bundle is usually not the highest-ranked one, so a session-mined candidate that flagged itself lost the flag the moment a CLI-authored record shared its title. It now reads the flag from every contributor. Mutation-tested: restoring the headline-only read fails the new test.

swati510 (approved):

PR #2426, opened 2026-09-19
feat(decisions): promote a candidate that states a reason, not one seen twice

RaghavChamadiya:
Based on #2425.

## Half one: the promotion bar

`SessionStagingStore.promotable()` gated on
`observations >= 2 or kind == "user_correction"`.

**The recurrence half has never fired.** Measured on a byte copy of the live
`sessions.db`, the observation distribution is `{1: 406}` — no staged row has
ever been seen twice. So `user_correction` was the entire promotion path, and
recurrence was rejecting rows on a condition nothing in this store could
satisfy. Waiting for a second sighting is not a quality bar when a second
sighting never comes.

It becomes: the row states a reason.

| | n |
|---|---|
| staged rows | 406 |
| **admitted by the new bar** | **224** |
| admitted by the old bar | 150 |
| admitted by both | 26 |
| **newly admitted** | **198** |
| old-bar rows the new bar would not admit | 124 |

### It is not a superset, and the plan for this did not say so

Those 124 are `user_correction` rows carrying **no rationale at all** — the
kind alone was carrying them. So the bar gates the **first** promotion only:

```python
first_promotion = r[7] is None
if first_promotion and not str(structured.get("rationale") or "").strip():
    continue
```

They are in the store either way. Applying the bar to re-emission would not
withdraw one of them; it would only stop them accreting the evidence of a
later sighting, which is information about a record that exists regardless. A
bar admits a record, it does not retract one already admitted.

On this store the practical difference is zero, because nothing has ever been
observed twice and so nothing ever re-emits. The choice matters on a store
with recurrence, which is exactly where the wrong reading would be invisible
here and wrong there.

### Why widening is safe now and was not before

What this promotes reaches an agent in the labelled `candidates` lane under
its own cap, not as rules to follow. Nothing here creates authority; a person
still accepts. That lane is the thing #2420 built, and it is the reason this
is a supply change rather than a quality one.

## Half two: the wiki, which this change points 198 more records at

`build_decision_maps` applied **no decision filter at all**
(`page_generator/helpers.py`). Every proposed record, and every dismissed
one, was indexed beside accepted rules and rendered into published pages as
settled fact. The resume path fed it the same way.

Two changes, and one deliberate non-change.

**Dismissed records never reach a page.** A tombstone is what stops
re-extraction re-proposing something already rejected; publishing it as a
decision says the opposite of what the dismissal meant.

**Accepted decisions outrank more confident proposals.** The two measure
different things: confidence says how well a candidate is grounded in what it
was mined from, an acceptance says a person read it and agreed it binds. A
page has room for three to eight of these, and this change offers several
times as many proposals to fill them with. Ranking is what makes a review
actually change the pages.

`rehydrate_decision_report` carries `status` through so a resumed run ranks
and filters identically to a fresh one. Without it every rehydrated record
read as a proposal, the accepted ones included.

**Filtering to accepted alone is not shippable, and that is measured, not
assumed.** This store holds 355 `proposed` records, 285 of them linked to
files, and **zero acceptances**. Phase 6's accepted-only predicate applied
here empties the decisions section of every page in the wiki. That is not a
more honest page, it is an absent one.

### What is left, and why it is not here

The honest full fix is to **label** each entry's authority where it is
rendered, so a proposal reads as a proposal rather than as a rule. That means
a new label in the i18n catalogue plus an edit to six templates
(`file_page.j2`, `module_page.j2`, `repo_overview.j2` and their `stub/`
variants), and it changes generated prose across every page in every
repository. It belongs in its own PR, not bolted onto a promotion-bar change.
This PR does the part that governs *which* records reach a page and in what
order; that one does the part that governs how they read.

## Tests

`tests/unit/sessions/test_staging.py`: a single choice with a reason now
promotes; one without does not; whitespace is not a reason; a correction
without a reason does not ride its kind; and a row already promoted keeps
accreting without one.

`tests/unit/generation/test_decision_scoping.py`: an accepted decision
outranks a more confident proposal, a payload with no status still ranks by
confidence, a dismissed record never reaches a page, the status reaches the
payload, and a record carrying no status reads as a proposal rather than as a
rule.

Three fixtures in `test_decision_discovery.py` seeded structured rows with no
rationale; they are about lane separation, so they now state one. One
assertion there (`dis_key not in rows  # one discovery observation does not
promote`) was incidentally pinning the recurrence bar; it now asserts what the
test is named for — the other lane kept its own row, kind and text.

Mutation-tested, all caught:

| mutation | result |
|---|---|
| restore the recurrence bar | 2 failures |
| apply the rationale bar to re-emission too | 1 failure |
| rank by confidence alone | 1 failure |
| publish dismissed records | 1 failure |

`tests/unit/persistence tests/unit/analysis tests/unit/sessions
tests/unit/generation` — **4812 passed**. `ruff check` clean.


RaghavChamadiya:
Rebased onto `main` after #2421 and #2422 merged, and amended with review fixes.

Two test comments were describing the old rule while still passing, so they now say what actually happens: the discovery lane promotes on the **first** sighting once a rationale is present (verified directly — `first.decisions` is 1 where it was 0), and a second sighting accretes evidence rather than being what admits the record. `key_concepts.py` also sliced `decisions_all` unranked, which this PR makes worse by admitting more proposals into that list; it now ranks.

`tests/unit/persistence tests/unit/analysis tests/unit/sessions tests/unit/generation` — 4813 passed on the rebased tree. `ruff check` clean.

swati510 (approved):

PR #2427, opened 2026-09-19
fix(rust): inherit `impl` block visibility from target type/trait declaration (#2389)

### LINKED ISSUE #2389: [Bug] A Rust impl block reads as private although rust.scm says it inherits its target's visibility

rishu685:
Fixes #2389

### Overview
In Rust syntax, `impl` blocks cannot write a visibility modifier of their own (e.g. `pub impl Foo` is invalid). `rust.scm` comments state that an `impl` block inherits its visibility from the type or trait it implements. However, because no modifier was captured by the query, `rust_visibility` previously fell back to `private` for every `impl` block across all Rust repositories.

Following the pattern established in #2376 for trait items, this PR extends `refine_rust_visibility` to resolve the target type/trait in the enclosing scope and inherit its declared visibility modifier.

### Changes
- **`packages/core/src/repowise/core/ingestion/extractors/visibility.py`**:
  - Extended `refine_rust_visibility` to handle `impl_item` nodes.
  - Implemented target type extraction (`_rust_extract_type_name`) supporting all 4 AST target shapes from `rust.scm`:
    1. Plain type: `(impl_item type: (type_identifier))`
    2. Generic type: `(impl_item type: (generic_type type: (type_identifier)))`
    3. Scoped type: `(impl_item type: (scoped_type_identifier name: (type_identifier)))`
    4. Reference type: `(impl_item type: (reference_type type: (type_identifier)))`
  - Resolved target names against sibling type declarations (`struct_item`, `enum_item`, `union_item`, `trait_item`, `type_item`) in the enclosing scope to inherit the corresponding visibility (`pub` -> `public`, `pub(crate)` -> `internal`, `pub(super)` -> `protected`, none -> `private`).
- **`tests/unit/ingestion/test_rust_impl_visibility.py`**:
  - Added test suite verifying `impl` blocks on `pub` vs private structs across plain, generic, scoped, and reference target types.
  - Added test coverage for `pub(crate)`, `pub(super)`, `enum`, and `union` targets.

### Validation
- Preserved AST query integrity (`rust.scm` remains untouched).
- All unit tests added and passing in `test_rust_impl_visibility.py`.


RaghavChamadiya:
@rishu685 #2418 is merged, so the dependency this one had is cleared. Pulling the dead-code exclusion into this branch yourself was the right instinct: #2389 says landing the visibility fix without it would mint roughly 6,600 false `unused_export` findings, and a PR that is only correct when another one lands first is a trap for whoever merges it.

Now that #2418 is in, that hunk is redundant here. `"impl"` is on main at `analyzer.py:268` and `:358`. GitHub still shows this as mergeable because both sides add byte-identical lines and git folds them, but the CI you have green ran against the old base, and I would rather not merge a set literal on the strength of a three-way merge doing the sensible thing with a duplicate entry.

`git fetch origin && git rebase origin/main`, drop the `analyzer.py` hunk and the two dead-code tests that came with it, and keep the visibility work plus `test_rust_impl_visibility.py`. Ping me when it is green and I will merge it, no further review needed.

Separately, the `"impl"` entry in `_UNCALLABLE_TYPE_KINDS` is inert today. `_detect_unused_internals` skips `language == "rust"` before it reaches the kind check, so the internals test passes either way. Worth knowing, nothing to change.

rishu685:
@RaghavChamadiya! done sir! its good to go now :)

RaghavChamadiya (approved):
Thanks @rishu685. The rebase dropped the duplicated analyzer.py hunk as asked, so the diff is now just the visibility rule and its tests. Unwrapping generic, scoped and reference targets before looking up the sibling declaration covers every shape rust.scm produces.

PR #2428, opened 2026-09-19
fix(ingestion): index TypeScript class fields holding a function (#2302)

### LINKED ISSUE #2302: [Bug] static create = (...) => {} is invisible: class fields are captured for their type only

rehan6025:
## Summary

- Adds a Tree-sitter query pattern in `typescript.scm` to capture `public_field_definition` nodes as symbols (`@symbol.def`) when the value is an `arrow_function` or `function_expression`.
- Maps `"public_field_definition": "method"` in TypeScript's `symbol_node_types` in `language_configs.py`.
- Gated strictly on function values so ordinary data properties (`count = 0`, `label: string = "x"`) never mint symbols, while preserving existing type-annotation extraction (`f: Field`).
- Adds comprehensive unit tests covering static factory fields, arrow handlers, function expressions, parameter shapes, visibility modifiers, and negative data-field assertions.

## Related Issues

Fixes #2302

## Test Plan

- [x] Tests pass (`pytest tests/unit/ingestion/parser/test_typescript.py` — 55 passed)
- [x] Query compilation passes across all 27 languages (`pytest tests/unit/ingestion/parser/test_query_compilation.py`)
- [x] All 550 parser tests pass (`pytest tests/unit/ingestion/parser/`)
- [x] Lint passes (`ruff check packages/core/`)
- [x] TypeScript type-check passes (`npm run type-check`)

## Checklist


rehan6025:
Hi @RaghavChamadiya,

The PR is open and ready for review! 

Note on CI: 20,496 tests passed, including all Tree-sitter query compilation and TypeScript parser tests. The only failing test across Python 3.11/3.12/3.13 is `tests/unit/cli/test_doctor_vector_decisions.py::test_decision_vector_ids_returns_namespaced_ids` (`UNIQUE constraint failed: decision_records.id`), which appears to be an unrelated upstream in-memory DB isolation issue on `main` when running the entire test suite.

Looking forward to your feedback!

RaghavChamadiya (approved):
Thanks @rehan6025 for gating the new pattern on the field's *value* rather than just capturing every `public_field_definition`. `count = 0` is a class field too, and the obvious version of this change mints a bogus symbol for it. The negative test pinning that is the one I went looking for.

Verified on current main that `typescript.scm` has a single `public_field_definition` pattern gated on `type: (type_annotation ...)`, so it only ever captured the annotation and never the field, which is exactly what #2302 described.

The nested class-expression case, where a method inside an inner class expression stopped being misattributed to the outer class, is a nice catch that the issue did not ask for.

One thing worth knowing rather than fixing here: `javascript.scm` has no `public_field_definition` pattern at all, not even the annotation one, so plain JS class fields have the same gap for a different reason. #2302 was scoped to TypeScript so this is not something you missed, but it is worth its own issue.

PR #2434, opened 2026-09-19
feat(decisions): a commit's file list stops answering per-file questions

RaghavChamadiya:
## The problem

`affected_files_json` was doing two jobs at once.

For a record mined from a commit it is **provenance**: these are the files the change touched. That is true, and worth keeping.

It was also the **governance binding** every path-scoped surface reads to answer "what governs this file". There it is mostly false. The `pr` and `git_archaeology` miners read one decision out of one commit body and then assign that commit's entire file list, so a decision mined from a 42-file refactor claimed all 42.

The canonical case, from the dev store: `packages/cli/src/repowise/cli/commands/augment_cmd/search_digest.py`, whose whole job is compacting Grep output, carried **five** decisions about *collapsing repeated file reads* — a mechanism in the sibling `read_state.py`. All five came from one 38-file commit that happened to touch it. One record, five wrong answers on five unrelated files.

## What was measured

41 labelled file/decision pairs across 10 real source files, stored under `local-stash/decision-layer-research/phase45-scope-2026-09-19/`.

| what an agent got back from `get_why` on a file | share |
|---|---|
| actually governs the file | 54% |
| tangential | 24% |
| outright noise, a different feature | 22% |

It tracks scope breadth and nothing else:

| the decision's scope | on topic |
|---|---|
| 1-10 files | 100% (n=7) |
| 11-31 files | 43% (n=7) |
| 32+ files | 44% (n=27) |

**80% of the 3,048 file→decision bindings in the store are commit footprints.**

## What this changes

A `scope_basis` column recording how a record's file list was arrived at, and one predicate, `binds_to_paths`, that every path-scoped surface asks:

| value | meaning |
|---|---|
| `""` | default. Binds. The backfill may repair it. |
| `"stated"` | a person said so. Binds. The backfill never touches it. |
| `"commit_footprint"` | a commit's file list. **Binds to nothing**, and keeps its files for provenance, staleness, search, the overview and lookup by id. |

Set at capture by `commit_scope_basis` (breadth > `MAX_GOVERNING_FILES` = 10, judged on the commit's *whole* list rather than the stored one), repaired on every index by `decision_migration.backfill_scope_basis`, and shipped with Alembic `0070` for the hosted Postgres path.

### Result

| | governs | tangential | noise |
|---|---|---|---|
| before | 54% | 24% | 22% |
| **after** | **88%** | **0%** | **12%** |

End to end on a byte copy of the real 575MB `wiki.db`: the schema reconciler adds the column cleanly, 110 of 367 records are marked, file links go 3,048 → 603, module links 1,492 → 434, 2,445 file entries survive as provenance, and a second run marks 0. `search_digest.py` now returns no decisions instead of five, and its sibling `read_state.py` keeps its one genuine record.

## Two of the three planned changes were already true

The plan named three changes. Verified against the bytes, two were no-ops, and they are recorded here so nobody re-derives them:

- **"Cap session-mined breadth via `SessionCandidate.edited`."** No session record is wide. They are 0 files (70) or 1-10 (67); `bind_scope_files` already caps at 20 and orders edited-first. Every wide record in the store is `pr` or `git_archaeology`.
- **"Exclude unscoped records from path queries."** They already cannot reach path mode: `query in json.loads('[]')` is `False`.

## Why this is a breadth rule and not a relevance rule

Two deterministic content signals were measured against the same labels and **both failed**. The probes are stored beside the labels so this is not re-litigated from intuition:

- **A file's share of the commit's diff** (`probe_diff_share.py`) does not separate. `search_digest.py` ranks *higher* than several genuinely-governing pairs.
- **Decision text vs the file's own diff hunk** (`probe_diff_overlap.py`) is worse than useless: it scores the canonical *wrong* answer highest of all noise pairs, because lexical similarity tracks the subsystem a file sits in rather than whether the decision governs it.

One signal did work and is **deliberately not shipped**. `probe_path_token_rescue.py` keeps a footprint's binding for the files whose *path tokens* its own text names, and measures 91% governing / 9% noise. It rests on 3 rescued pairs and a hand-tuned stopword list, which is too thin a basis for tuned machinery. It is documented as the way to buy recall back if the gate proves too aggressive in use.

## The cost, stated plainly

The gate drops 15 genuinely-governing pairs against 8 noise. A record that really is cross-cutting loses its file pages; it stays reachable by search, by the repo overview and by id. That is the trade the phase asked for, and it is why the rescue probe is stored rather than discarded.

## The judgement call worth flagging

`MAX_GOVERNING_FILES = 10` rests on an asymmetric evidence base, and a reader of the constant should know it. The **refusing** half is well evidenced (n=7 and n=27 in the two wide buckets). The **binding** half — that a 10-file commit-mined record binds all ten of them — rests on n=7. If a later measurement moves the bar, it should move down rather than up.

## Where the gate had to go, and why eight places

Gating `sync_decision_node_links` at the write path fixes only the two readers that use the graph (session injection, and the `get_risk` directives). Everything else reads `affected_files_json` directly, so each needed its own gate:

- `_why_path` and `_build_target_context` (`tool_why.py`) — one target routes to the first and two to the second, so they have to agree
- `build_decision_maps` (wiki file and module pages)
- `get_context` (`tool_context/targets.py`) — the per-file decisions lane agents hit most
- `get_answer` grounding (`_answer_context.py`) — ranked by overlap count, so a footprint outranked a record naming one file exactly, and the result renders into the prompt as grounded rationale
- per-file `stale_governance` findings (`health/governance.py`) — one HIGH finding per file
- the `has_decision` badge (`node_signals.py`)
- `governed_files`, the *denominator* for ungoverned hotspots (`crud/decisions.py`)
- `_sibling_coverage` (`_helpers.py`)

Modules are gated with files rather than kept. A commit-wide record does sound like a claim about the areas it touched, but its module list is `resolve_module_nodes` over the same unjustified file list, so a 42-file commit yields a dozen directories on exactly the evidence that made the file list wrong.

Deliberately left ungated: repo-wide surfaces (overview, search, reference lookup), staleness and regeneration, and the `+5.0` target tie-break in `_score_keyword_matches`, which only applies to records that already cleared the relevance floor.

## Two fixes outside the phase's scope, found on the way

1. **A scope a person stated was being stored and then ignored.** Six write paths wrote `affected_files_json` without touching the basis, so a footprint whose scope somebody narrowed kept refusing to bind — and the backfill would re-mark it on the next index. All six now write `"stated"`: `accept_decision` (which is what `decision confirm --scope` actually runs, not `update_decision_metadata`), the manifest importer's create and update paths, `update_decision_metadata`, `add_decision`'s `_restate`, and `update_decision_by_id`'s generic setattr map.

2. **The dedupe fold silently undid the backfill.** `apply_dedupe` unions the merged records' file lists and re-syncs node links unconditionally. A footprint folding into a narrow canonical handed it 40 files under an empty basis — and the canonical's `source` is usually not one the backfill repairs, so nothing would ever catch it again. The fold now recomputes the basis from the union's breadth, preserves a `stated` basis, and gates its own sync. Only reachable from `repowise decision review --dedupe`, so exposure was bounded.

## Testing

44 new tests across four files. 16 mutations run against them: 15 caught, and the 16th turned out to be an **equivalent** mutant (judging breadth before or after `commit_scope_files` cannot differ while `MAX_GOVERNING_FILES` stays below that function's own cap) — the test's docstring claimed a distinction the code did not rest on, so the claim was corrected rather than the code.

Two tests were vacuous when first written and only the mutation showed it: one read `governing` where the payload says `governing_decisions`, and one passed a query that suppressed the record for a reason unrelated to scope. Both now carry explicit negative controls, as does every gate test here.

Suites run: persistence, generation, analysis, health and pipeline — **6,678 passed, 1 skipped**. `tests/unit/server/mcp` and `tests/unit/cli` were run separately and are reported in a comment. `ruff check` clean across `packages/` and `tests/`.


RaghavChamadiya:
## Second commit: session scope

`24a5df012` extends this PR from the commit miners to the session miner, after measuring the half the first commit did not cover.

### Why it was needed

The first commit's gate keys on `scope_basis`, which only the two commit-derived miners set. Session records carried the empty default, so they bound to every file they named. Measuring that half directly, over 32 labelled file/decision pairs:

| session-sourced per-file | share |
|---|---|
| governs | **28%** |
| tangential | 53% |
| noise | **19%** |

Worse than the 54% baseline this PR set out to fix, on 30% of the per-file surface. My earlier "88% governing" figure covered only the part the first commit gated, and is corrected by this.

### What the measurement found

Session scope is **proximity**, not a footprint — the last few paths touched plus what the next few events touch. So breadth in files, the signal that worked for commits, does not separate at all here: 33–38% governing flat across scope sizes 1 through 4.

What separates is whether the files sit **together**:

| the record's files span | governs |
|---|---|
| one directory | **67%** (n=6) |
| two or more | 19% (n=16) |
| three or more top-level packages | **0%** (n=7) |

The dominant cause is a category error rather than a retrieval error: a repo-wide working rule ("avoid duplicated code", "keep pull requests OSS-safe", "do not invoke bare `python`") restated while editing several areas, then bound to each of them. All six noise pairs came from such rules, and all six were classified `architectural` rather than `agreement` — so gating on `kind` alone fixes almost nothing.

Keyword-groundedness was tested as an alternative and is 100% precise but keeps 2 of 32, because 20 of the 32 records contain no distinctive token at all. That is the diagnosis restated, not a usable gate: these records are not about code.

### The three changes

- **An agreement is scoped to the repository.** It governs how the work is conducted, so no file is what it is about. `AGREEMENT_SCOPE` already asserted this on the acceptance row; the scope column now agrees.
- **A session record spanning more than one directory binds to no path.** The same distinction `derive_decision_scope` already draws, rather than a second threshold.
- **Scope entries naming a file the index does not hold are pruned.** #2395 closed this at the write path on 2026-09-18, and every leaking record predates it — 50 records created 08-22, 08-23 and 09-04, none after — so this repairs rows already written rather than re-fixing a live path. Validated against the graph's own file nodes, the set `decision_node_links` points into, and skipped when the graph looks unbuilt so a failed build cannot empty every scope.

`binds_to_paths` now reads a `NON_BINDING_SCOPE_BASES` set, so a new basis is added in one place and all eight path-scoped surfaces honour it at once.

### Measured, on a byte copy of the dev store

| | |
|---|---|
| records pruned | 61 |
| scoped to the repository (agreements) | 30 |
| marked proximity | 8 |
| stale entries naming plan docs, sibling checkouts, worktrees | **114 → 0** |
| session bindings | **178 → 18** |
| total file links | 3,048 → 443 |

Session per-file precision on the labelled sample: **28% → 57% governing, 19% → 14% noise.** All three repairs are idempotent; a second run changes nothing.

### What I am not claiming

The one-directory rule rests on 32 pairs with cells of 6 to 16, and it drops 5 of the 9 governing pairs along with 20 of the 23 bad ones. It is the least-evidenced part of this PR. It is one predicate and should be re-measured on a larger sample before anything is built on top of it — in particular before Phase 5, which would multiply session records by 6x to 22x and make them 90% of the per-file surface.

Isolating the three: pruning alone moves precision 28% → 32%, the agreement rule shows no effect on this sample (both agreement pairs were already pruned) though it touches 30 records store-wide, and the one-directory rule does the rest.

### Testing

24 new tests in `test_decision_scope_basis.py` plus 7 in `test_decision_scope.py`. 5 further mutations run against them, all caught (21 total across the PR: 20 caught, 1 equivalent). Suites: persistence, analysis, sessions and generation — **4,856 passed**.

swati510 (approved):

PR #2433, opened 2026-09-19
fix: account for Anthropic cached input tokens

Hitesh-XS:
## Summary

Fix Anthropic input token accounting when prompt caching is used.

Previously, `input_tokens` only included the uncached input tokens reported by Anthropic, causing cache read and cache creation tokens to be excluded from the reported total.

## Changes

- Include `cache_read_input_tokens` in the total `input_tokens`.
- Include `cache_creation_input_tokens` in the total `input_tokens`.
- Keep `cached_tokens` as the cache-read token count.
- Keep cache creation and cache read counts in the `usage` breakdown.
- Add regression coverage for cached input token accounting.

### Unit tests

```text
pytest -q tests/unit/test_providers/test_anthropic_provider.py
14 passed in 4.60s

PR #2435, opened 2026-09-19
fix: resolve #2367 - [Bug] The Anthropic API provider counts only the uncached pr

### LINKED ISSUE #2367: [Bug] The Anthropic API provider counts only the uncached prompt remainder as input_tokens
## Summary

`claude_cli` counted only the uncached remainder of a prompt as `input_tokens`, which #2267 reported and #2337 fixed. The direct Anthropic API provider has the same undercount, and it has not been fixed.

`packages/core/src/repowise/core/providers/llm/anthropic.py:234`

```python
input_tokens=response.usage.input_tokens,
```

The two cache figures are collected four lines later and kept only inside the `usage` dict:

`:242-246`

```python
"cache_creation_input_tokens": getattr(
    response.usage, "cache_creation_input_tokens", 0
),
"cache_read_input_tokens": cached,
```

So the persisted `input_tokens` is the uncached remainder alone, and the two cached dispositions never reach the total.

## Why this one matters more than #2267

`claude_cli` is billed against a subscription seat, so its ledger rows price at $0.00 and the undercount distorted token volume only. This provider is billed per token against an API key. Prompt caching is on for wiki page generation, where the prompt is large and stable and therefore mostly cache reads after the first call, which is exactly the shape that makes the missing portion the majority of the prompt.

The consequence is that `repowise costs` and the cost ledger under-report both input tokens and the dollar figure derived from them, on the provider where the dollar figure is real.

## Not yet sized

I have not measured how far off the totals are on a real run, and the ratio depends on cache hit rate, so treat the mechanism as certain and the magnitude as unknown.

## Done looks like

`input_tokens` carries the whole prompt, matching what #2337 established for `claude_cli`: the uncached remainder plus cache reads plus cache creation, with the uncached figure kept separately so nothing is lost.

Two things worth doing in the same pass. Check `:251` and `:267`, which pass `result.input_tokens` onward, so they inherit whatever the fix decides. And pin the result with a recorded usage payload in a test, since the disjointness of the three fields is a property of the API contract and not something the repository can prove on its own.

fathirramadhan-web:
Closes #2367

### Overview
Addresses the problem reported in #2367 by resolving the underlying issue and ensuring test coverage passes.

### Verification
- Tested locally
- Linters/tests pass


RaghavChamadiya:
Thanks @fathirramadhan-web. The core change is right: summing the uncached remainder with `cache_read_input_tokens` and `cache_creation_input_tokens` is the fix #2367 describes.

I am holding this one, for two reasons. @Hitesh-XS asked for #2367 on the 18th and opened #2433 with the same fix the next day, a few minutes before this PR, so the issue is theirs under the repo's first-claim rule. And the diff adds a block to `.gitignore` headed `# Auto-added by github-engineer` (`*.pyc`, `*.whl`, `nem_stats.c`, `*.lib`...) that has nothing to do with the fix and names files this repo does not have. That looks like a tool writing into your branch without you noticing, and is worth checking on your side before your next PR.

If you are looking for something to pick up, commenting on an open issue first to claim it is the surest way to avoid this kind of overlap.

PR #2441, opened 2026-09-19
feat(decisions): ask the miner which files a commit's decision is about

RaghavChamadiya:
The `pr` and `git_archaeology` miners read one decision out of one commit body and then took that commit's **whole file list**, because the model was never asked which files the decision was about. A commit that bundles three unrelated changes produced three decisions each claiming all of its files, so `get_why` on any of them answered with the other two.

## What changes

- Both commit prompts now ask for `affected_files`, under a shared instruction that makes `[]` an explicitly correct answer.
- `mine_pr_bodies` shows the model the commit's file list. It never did — it asked for a decision from a subject and a body alone, and scored 20% on topic against git archaeology's 45% on the same store.
- The answer is intersected with the commit's own list before it is stored, so a path the model invented cannot bind, and the result is marked `commit_selected` rather than left under the breadth rule.
- An **empty** answer stays empty; a **missing** one falls back to the old behaviour, so a provider that has not seen the new prompt is unaffected.
- `backfill_scope_basis` repairs every legacy commit-derived row rather than only the wide ones, and leaves a selected scope alone.
- A dedupe fold keeps `commit_selected` only while every member of the fold chose its own files.

## Measured

Out of sample: 29 records and 124 blind-labelled pairs drawn from the dev store, three raters, pairs shuffled, "prefer the lower label when torn". The pass/fail bar was written to a file before any label existed.

| arm | n | governs | noise | 95% CI |
|---|---|---|---|---|
| bind the whole commit list (today) | 124 | 27% | 18% | 20–35% |
| the model's selection | 43 | **93%** | **0%** | 81–98% |

Arm B is a complete census of the selector's choices, not a sample of them.

It is not a precision/recall trade. The selection keeps 13% of the surface, and **all 20** of the records that had a genuinely governing file kept one. The 5 it emptied had none to keep — the selector drops exactly the records whose subject was never in the candidate list.

## Why the breadth gate goes with it

`MAX_GOVERNING_FILES` was the best signal available while nothing asked the model directly, but it never measured commit breadth. The miner's file list is an inversion of per-file git metadata, not the commit's diff (median 0.89 coverage of the real list, p25 0.45), so at the five-file cutoff **82% of the records that still bound came from commits that had touched more than five files**. The gate was selecting commits the indexer thinned.

The prize is the stratum `commit_footprint` blanks entirely today: 72 of 117 commit-derived records and 1,596 claimed file entries, which the selector recovers at 95% governs.

## Scope limit

This measures the selection stage on records that already exist. Whether asking for the new key also changes *which* decisions get extracted is unmeasured — worth watching on the first index after this lands.

## Tests

`tests/unit/analysis/test_decision_commit_selection.py` is new; the dedupe and scope-basis suites gained the fold and backfill cases. Nine mutants, one per behaviour a new test claims to pin, all killed. `tests/unit/persistence`, `tests/unit/analysis`, `tests/unit/generation/test_decision_scoping.py`, `tests/unit/server/mcp/test_why_scope_basis.py` and `tests/unit/health` pass locally.

Evidence and method in `local-stash/decision-layer-research/capture-2026-09-19/`.


swati510 (approved):

RaghavChamadiya:
Two review passes on the diff found two real bugs, both introduced by this branch. Both are fixed in `92268af75`, with tests.

**An accepted record was demoted.** `backfill_scope_basis` now repairs every legacy commit-derived row rather than only the wide ones, and it selected on `scope_basis == ""` alone. `accept_decision` writes `stated` only when the accepter passes an explicit scope, and **neither `repowise decision confirm` nor the web accept route does** — only `decision accept --scope ...`. So a decision a reviewer had confirmed as-is would have lost its file links on the next index, silently. Worse than silent, actually: `governed_files` is built from accepted path-binding records and `analysis/health/governance.py` emits an `ungoverned_hotspot` finding for every hotspot outside it, so the demotion would have *manufactured* health findings saying nobody had decided anything about those files. The repair now exempts a record by its acceptance row rather than by its basis.

**A narrow footprint came back binding through a dedupe fold.** The fold recompute calls `commit_scope_basis(union)`, which returns the binding empty basis for any union at or under the cutoff. That was safe while a footprint could only exist *above* the cutoff — and this branch made narrow footprints the normal case. A two-file footprint folding into a one-file canonical republished exactly the files the repair had just taken away. A member that cannot bind on its own now makes the union non-binding. That also fixes a pre-existing case the same lines owned: a `repository` (working agreement) or `session_proximity` canonical was being promoted to binding by any fold, because only `stated` was preserved.

Smaller, also from the reviews:

- A malformed `affected_files` (a list of dicts, of nulls) read as an empty selection, which under the new semantics binds the record to nothing *permanently*. It now falls back like any other unusable answer. `[]` still means "about none of these files".
- `proposed_files` is cleared whether or not the decision resolved to a commit. Only the resolved branch cleared it, so an unattributed decision carried the model's raw paths into `dataclasses.asdict`. Inert — every reader in `bulk_upsert_decisions` fetches keys by name, there is no `DecisionRecord(**headline)` and no strict model on that path — but false against the field's own docstring.
- `mine_git_archaeology` now uses `_MAX_PROMPT_FILES` and sorts before truncating, as `mine_pr_bodies` does. It inlined an unsorted `[:20]`, so which files the model could choose from depended on `_git_meta_map` iteration order rather than on the commit.
- `_scope_from_selection` is annotated `Sequence`, matching the helpers it calls.
- The `MAX_GOVERNING_FILES` docstring says what it still decides — the scope of a response that did not answer at all — rather than describing the retired rule as live.

Eight new tests. Sixteen mutants across both rounds, one per behaviour a test claims to pin, all killed.

## Two things the reviews raised that I have deliberately not changed

**Legacy `pr` records become non-binding and mostly stay that way.** `purge_proposed_decisions_by_source` is called for `git_archaeology` but never for `pr`, and the miner visits at most `_MAX_PR_BODIES = 25` commits per run, so most existing `pr` rows will be stamped `commit_footprint` and never re-mined. That is the intended outcome rather than a regression: those rows measured **20% governs and 23% noise**, so a record that keeps its text, its commit and its staleness signal while no longer answering "what governs this file" is strictly better than one that answers wrongly one time in four. Worth stating plainly because the word "repair" implies re-mining follows, and for `pr` it does not.

**`max_tokens` was not revisited.** The output now carries an `affected_files` array per decision, against budgets of 2500 (`pr`) and 2000 (`git_archaeology`) set before it existed. On truncation `_parse_decisions_json` fails `json.loads`, the `\[.*\]` salvage also fails on an unterminated array, and the whole batch of five commits is lost silently. The prompt pushes hard toward short lists so I have not seen it, but I have no measurement either way, and picking a new number blind is a guess. Flagging it rather than tweaking it; the durable fix is to make that truncation loud, which is a separate change.

PR #2444, opened 2026-09-19
fix(generation): stop a formatter comment from prescribing ruff format

### LINKED ISSUE #2384: [Bug] A comment mentioning "formatter" makes generated agent files prescribe `ruff format .`

Aman-goel-04:
## Summary

- Rename _declares_ruff_format to declares_ruff_format in precedent/structural.py and reuse it from detect_build_commands, replacing the substring test ("ruff" in text and "format" in text) that fired on a comment mentioning "formatter" with no formatter actually configured
- The reused check refuses when a competing formatter (black, blue, yapf) is declared, and requires either a real [tool.ruff.format] section or an explicit ruff format / ruff-format invocation in Makefile, package.json, .pre-commit-config.yaml or justfile

## Related Issues

Fixes #2384 

## Test Plan

- [x] Tests pass (pytest tests/unit/generation/test_tech_stack.py tests/unit/precedent, 208 passed)

Added 3 regression tests: a comment-only "formatter" mention with no [tool.ruff.format] (no command emitted), a real [tool.ruff.format] section (command present), and a competing [tool.black] declaration (no command even with both substrings present). Confirmed the exact reported bug reproduces red without the fix.

## Checklist

- [x] I have updated documentation if needed; not applicable, no user-facing docs affected


RaghavChamadiya (approved):
Thanks @Aman-goel-04 for reusing the detector that already exists instead of writing a second one. `_declares_ruff_format` was written for the formatter-drift episode and its own docstring names `detect_build_commands` as the loose one, so folding the two together closes the gap and removes the chance of them drifting apart later.

Confirmed the bug on this repo before merging: there is no `[tool.ruff.format]` section in our `pyproject.toml`, the only two matches for the old `"ruff" in text and "format" in text` test are comments, and our own generated `CLAUDE.md` was carrying `Format: ruff format .` because of it. That is the one command this repo must not run, so it was worth fixing properly.

The regression test for the comment-only case is the one that keeps this fixed.

PR #2449, opened 2026-09-20
fix: classify dotfiles named after a config extension as config

### LINKED ISSUE #2379: Classify `.env` dotfiles as config paths

zhongxiao-chang:
## Summary

`.env` — the file `CONFIG_EXTENSIONS` names by name — was classified as production code.

`is_doc_or_config_path()` matched `PurePosixPath(path).suffix` against the extension sets, but pathlib reports **no** suffix when a dotfile's only dot is the leading one (`PurePosixPath(".env").suffix == ""`). So the file literally named `.env` never matched the `".env"` entry while `foo.env` did: `file_population(".env")` returned `production`, and the knowledge graph's `_classify_file_type()` labelled it `"file"` for the same reason.

Fixes #2379.

## Changes

- `support_paths.classification_token(path)` — the token a path is classified by: the lowercased suffix, or the whole file name when there is no suffix. This keeps set entries that name a file rather than an extension (`".env"`) working.
- `is_doc_or_config_path()` now uses it, and `knowledge_graph._classify_file_type()` uses the same helper for its config check, so the duplicated idiom has one source of truth. The infra-name branch still keys off the real suffix, so `Dockerfile`/`Makefile` classification is unchanged.

## Test plan

- `tests/unit/test_support_paths.py`: `.env`, `proj/.env`, `packages/core/.env` are doc/config; `foo.env` keeps working; `.gitignore` — a dotfile the sets do not name — stays production; `file_population(".env") == "doc"`.
- `tests/unit/analysis/test_knowledge_graph.py::TestFileTypeClassification::test_config_by_dotfile_name`.
- Suites touching the two classifiers (`test_support_paths.py`, `test_knowledge_graph.py`, `generation/test_layers.py`, `test_no_doc_extension_copies.py`): 104 passed locally (Windows, Python 3.11).
- The issue's repro now prints `True True True` and `file_population(".env") == "doc"`.

Scope note: `.env.local`-style names (a second, real suffix) are unchanged — this PR only fixes the file named exactly `.env`, as the issue describes.


RaghavChamadiya:
Thanks @zhongxiao-chang, and I want to be straight with you quickly rather than leave this sitting.

This is the better of the two patches for #2379. The shared `classification_token()` helper used by both `is_doc_or_config_path` and `_classify_file_type` is the right shape, because those two functions having drifted apart is the actual reason this bug exists in two places, and your regression tests pin the over-reach case I would have asked for: `.gitignore` staying production, and `foo.env` still classifying by extension rather than by name.

It is not the one that is going to merge, though. @pamod-madubashana filed #2379 and opened #2381 four minutes later, which is the earliest and clearest claim on it, and in this repo the first claim binds even when a later patch is better. If I merged on quality here I would be telling people that claiming an issue counts for nothing next to being faster, and the people who lose that trade are the ones who write up a plan and wait. I have asked them on #2381 for the tests and the shared helper that yours already has.

What I would rather do with your work is point it at #2454, which you filed this morning and which nobody has answered. That is a genuinely better find than either PR here: a *third* classifier in `generation/categories.py` that still disagrees about `.env` for a different reason, resolving config off `is_config` or a `_CONFIG_LANGUAGES` membership test. Two surfaces disagreeing is a bug; three surfaces disagreeing three different ways is a design problem, and you are the only person who has noticed it. That one is yours, it is unclaimed, and I would review it quickly.

If it turns out #2381 stalls, I will come back to this PR rather than let the fix rot.

PR #2465, opened 2026-09-20
feat(decisions): ask the agent to record the choice it just committed

RaghavChamadiya:
The decision layer's first write-side hook. Every other surface delivers what
was already recorded; nothing asked for a record.

It shows. Of the records this repository holds from `decision add`, a person
wrote all of them and an agent wrote none — over four weeks in which agents
made most of the architectural choices in it.

After a successful, fresh `git commit` whose message carries two or more
decision signals, the PostToolUse hook offers the `decision add` line. **The
agent writes the record**, which is the point: it knows what it chose and what
it rejected, and a miner reading the message afterwards can only infer. The
agent-authored records in this store carry measurements and rejected
alternatives that no commit message contains.

## An agent cannot decline a hook, so the gates are the design

- **Off unless the repository turns it on.** `decisions.capture_prompt`,
  default false, no preset sets it, `repowise decision config
  capture-prompt --on`.
- **At most once per session**, claimed through `_claim_emission` before the
  state write so two hook processes cannot both print and the loser writes
  nothing.
- **Only `git commit`**, not a merge, a rebase or a `--dry-run`.
- **Only a fresh commit in this repository.** HEAD's committer timestamp must
  be within 15 minutes, and the git work tree must be the indexed one.
- **Only two or more decision signals.** Measured over 137 commits here: one
  keyword admits 58% of them, two admits 25%.
- **Never for a commit a record already cites.** An unreadable store counts as
  cited, so the failure mode is silence.

It proposes. Nothing is recorded without the agent running the command, and
what it records lands as a candidate for review, never as a rule.

## The measured tax

On the last eight days of this repository: 35 of 156 commits pass the gate,
against 82 sessions — **0.43 asks per session at most**, one per 2.3 sessions.

A model classifier over the same commits fires on 62% of them (1.18 per
session uncapped) at 88% precision. That measurement is why this gates on a
keyword count and a session instead, and leaves the judgement to the agent.
Write-up in `local-stash/decision-layer-research/hook-2026-09-20/`.

## Also

`decision add` gains `--evidence-commit`, and the prompt passes the sha.
Without it the already-recorded gate could never suppress a record this
feature creates, so the agent would be re-asked next session about a commit it
had just recorded.

`DECISION_SIGNAL_KEYWORDS` moves to `commit_signals`, which imports nothing:
the miner cannot be reached from the hook, and a second copy is how the two
would drift.

## Budget

No `repowise.core` at module scope, raw `sqlite3`, every failure swallowed.
`test_augment_hook_perf` passes unchanged. On the common path — not a git
commit — the cost is a `dict.get`, an `isinstance` and one substring scan. An
opted-out repository pays one config read and stops.

## Review

Two rounds, seventeen findings. Three only appear on real payloads:

1. **The PowerShell tool result carries no exit code**, so the success check
   fell back to scanning stdout for "error"/"fatal". `git commit -m "refactor
   to one error path; remove the old fatal handler"` — the exact shape this
   targets — read as a failed commit, and `nothing to commit` read as a
   success, prompting about a commit made days earlier.
2. **`git log -1 --no-merges` is not HEAD.** It filters merges out of the walk
   and returns an older ancestor, so finishing a conflicted merge with
   `git commit` gated on an unrelated commit.
3. **`.repowise` marks the indexed repository, not the git one.** This tree
   nests `backend/` and `frontend/` as separate repos, so a commit made in one
   of them was evaluated against the outer repo's HEAD.

Two findings are recorded rather than fixed, and are in the commit bodies:
`capture_prompt` is CLI-only (absent from `DecisionSettings` and the generated
types; the preset branch preserves it, so nothing silently resets it), and the
inner claim can let the *staleness* notice through twice in a race that needs
both hooks installed and a stale wiki.

28 tests, one per gate. 19 mutants, all killed, each by its own named test.


RaghavChamadiya:
### The trigger could never have fired — fixed in `468a16b5c`

A live audit of the installed hook state turned this up before merge.

`_handle_bash_post`, which this feature hangs off, is unreachable. `claude_config.py:229` narrowed the shared PostToolUse matcher to exclude the shell tools, on measurement — *"51% of hook invocations and 0.7% of emissions"* — and `_LEGACY_AUGMENT_MATCHERS` rewrites any install still carrying `Bash` back to the narrow form. Confirmed against this machine's ledger: `hook_runs` holds **zero** rows for `Bash` or `PowerShell`, ever, across 233 sessions.

The comment's escape hatch does not hold either — it keeps the branch "for Codex, the only surface a hook can reach it on", but a Codex install carries only the `PreToolUse` rewrite hook and no augment surface.

**The switch now installs its own prerequisite.** `capture-prompt --on` adds a separate `Bash|PowerShell` PostToolUse entry; `--off` removes it. Separate rather than widening the shared matcher, so the measured narrowing is not quietly reversed and a repository that never opts in pays nothing.

Two consequences worth stating: the cost is per *install* and the policy is per *repository*, so both commands now print what they did to the shared settings file and `--off` names the effect on other repositories. And the same audit shows the existing post-commit staleness notice has been dead for its whole life — `install_claude_code_hooks` still claimed PostToolUse "detects git staleness"; that claim is corrected here, the surface is revived for anyone who switches the prompt on, and reviving it generally is left as its own call.

swati510 (approved):

PR #2482, opened 2026-09-21
fix(cli): warn when init --resume reuses pages without populating decisions (#670)

### LINKED ISSUE #670: [Bug] init --resume skips decision extraction even when the persisted decision report is empty
## Describe the Bug

Decision records are only harvested inline during page generation. When `repowise init --resume` reuses a previously generated page (hash match, LLM call skipped), the inline harvest never runs, so decisions cannot be backfilled by a resume. A user who sees an empty Decisions page and runs `init --resume` (the natural "finish / repair" instinct) gets the same empty result and reasonably concludes decisions are broken, when the actual fix is a regeneration with a stronger model or higher coverage plus `--force`.

## Steps to Reproduce

1. Run an initial index where generation captures no decisions (for example a low-coverage run, or a model that emits nothing passing the harvest bar). `decision_records` stays 0.
2. Run `repowise init --resume ...` to completion.
3. `decision_records` is still 0. The Decisions surface renders its empty state.

## Expected Behavior

Either resume should be able to backfill decisions, or the tool should tell the user why resume will not populate them and what to do instead.

## Actual Behavior

Decision harvesting is inline in page generation, not a separate phase. In `packages/core/src/repowise/core/generation/page_generator/core.py` (around line 296), right after the LLM returns a page, `harvest_decisions(...)` pulls a trailing decision block out of that freshly generated content and strips it before the page is stored. It is gated on `harvest_decisions` and on the page type. On `--resume`, the incremental generator reuses prior pages by hash match and skips the LLM call (`_call_provider` returns the stored page early when `source_hash` matches). The harvest step still runs, but only over the reused page's already-stored content, which had its decision block stripped at original storage time, and the model is never re-invoked to emit a new one. Resume therefore cannot backfill decisions the original generation did not capture.

## Fix Direction

1. **Decouple harvest from generation** (best UX): make decision harvest a standalone, resumable pass over already-generated page content, so `--resume` or a dedicated `repowise decision harvest` can backfill without regenerating every page.
2. **Surface the limitation** (cheap, honest): when resume reuses pages and `harvest_decisions` is on but `decision_records` is empty, print a one-line notice, e.g. "decisions are harvested during generation; re-run with --force or a stronger model to populate."
3. Document that decisions come from the generation pass and will not appear via resume-reuse.

Option 2 is a small, self-contained starting point; option 1 is the durable fix.

## Additional Context

- `packages/core/src/repowise/core/generation/page_generator/core.py` (inline `harvest_decisions`, ~line 296)
- `packages/core/src/repowise/core/generation/models.py` (`harvest_decisions` config, default true)
- Related: the incremental update path also reuses decision records, so any backfill needs to run on resume/update, not just a fresh `init`.

rishu685:
Resolves #670.

Decision blocks are extracted inline during fresh page generation and stripped upon storage in `pages`. When `repowise init --resume` runs and reuses previously generated pages based on source hash matching, model calls are skipped and inline harvesting does not run over the already-stripped stored pages.

If an initial index produced 0 decisions (e.g. from low coverage or a model that emitted no valid decision blocks), running `repowise init --resume` leaves `decision_records` empty without explaining why or how to resolve it.

## Changes
- **Advisory Notice in `show_completion`**: Added `_render_resume_decision_notice` in `packages/cli/src/repowise/cli/commands/init_cmd/reporting.py` that alerts the user when a resumed full-mode run reused preserved pages and produced 0 decisions:


RaghavChamadiya:
Thanks @rishu685. Keeping the change to an advisory note, and leaving the resume skip alone, is a reasonable smallest version, and I am glad you went for a small version.

It does land in the trap I described on #670, though. The condition is:

```python
preserved = getattr(result, "preserved_page_ids", None)
if preserved and n_decisions == 0:
```

For a repo that genuinely has no decisions, that is true on every resumed run, forever, and the note tells the user to run `init --force`, which will produce zero decisions again. None of the four tests covers that case, and the docstring describes the old mechanism ("inline harvesting does not run on the stripped stored pages"), which is the one that no longer exists.

`preserved_page_ids` is also the wrong signal. It says generation reused pages. The bug is in the analysis phase: `resume_controller.can_skip(ResumePhase.ANALYSIS)` skipping decision extraction and rehydrating an empty report. What is needed is some way to tell "extraction ran and found nothing" from "extraction never ran", for example the persisted report recording that extraction completed (even with a count of zero), so that resume can re-run extraction only when it never ran. Then a test for each of the two cases.

Happy to talk through where that marker could live before you write it, if that helps.

rishu685:
Thanks for the clear feedback @RaghavChamadiya sir! That makes total sense — `preserved_page_ids` was at the wrong phase altitude (generation vs analysis), and warning on `n_decisions == 0` created an infinite `--force` loop for repos that legitimately contain 0 decisions.

Here is the updated plan to address this cleanly:

1. **Record Extraction Completion**: In `checkpoint_analysis` / `persist_analysis`, explicitly persist a marker indicating that decision extraction ran to completion (regardless of whether the extracted count is 0 or >0).
2. **Backfill on Resume**: On `init --resume`, if `rehydrate_decision_report` finds that decision extraction never completed, `orchestrator` will re-run `_run_decision_extraction` to backfill decisions instead of rehydrating an empty unextracted report.
3. **Drop Advisory Notice**: Remove `_render_resume_decision_notice` from CLI reporting, eliminating false positives on genuine 0-decision repos.
4. **Tests**: Add unit tests for both cases:
   - Genuinely 0-decision repo: resume recognizes completed extraction, does not re-extract, and produces no false warnings.
   - Unextracted/interrupted repo: resume detects missing extraction and backfills decisions.

Would love your thoughts on your preferred home for the completion marker (e.g. `state.json`, a ledger record, or an analysis report) so I can implement it in the exact right spot!

PR #2483, opened 2026-09-21
fix(claude): remove distill allow rules on uninstall and check in leftover reason (#2470)

### LINKED ISSUE #2470: [Bug] claude-code uninstall leaves an env pin and two permission rules, and reports settings.json clean
## Summary

`ClaudeCodeTarget.install` writes four things into `~/.claude/settings.json`. `ClaudeCodeTarget.uninstall` removes two of them, and the leftover probe that decides the exit code does not look at the other two.

Install, `packages/cli/src/repowise/cli/agent_targets/targets/claude_code.py:409-412`:

```python
register_with_claude_desktop(repo_path)
register_with_claude_code(repo_path)
install_claude_code_hooks()
enable_tool_search_in_claude_code()
```

Uninstall, `claude_code.py:459-468`, calls `uninstall_claude_code_rewrite_hook`, `uninstall_claude_code_augment_hooks`, `uninstall_claude_code_mcp_entry` and `uninstall_claude_desktop_mcp_entry`. There is no remover for `enable_tool_search_in_claude_code`, which sets `env.ENABLE_TOOL_SEARCH = "true"` (`editor_integrations/claude_config.py:199-202`).

Separately, `add_claude_code_distill_allow_rules` (`claude_config.py:424-455`) appends two entries to `permissions.allow`:

```python
DISTILL_ALLOW_RULES = (
    "Bash(repowise distill:*)",
    "PowerShell(repowise distill:*)",
)
```

Grepping `packages/cli` for `DISTILL_ALLOW_RULES` finds the definition and the writer. There is no remover, and `rewrite_uninstall` (`commands/hook_cmd.py:349-371`) does not touch them.

## Consequence

The exit code is the part that matters. `claude_code_leftover_reason` (`claude_config.py:725-752`) inspects `mcpServers` and `hooks` and nothing else, so after an uninstall it returns `None`, the run records `FileAction.REMOVED`, and the user is told settings.json is clean over a file that still carries a repowise env pin and two repowise permission rules.

That function's own docstring says why it exists:

> Collapsing those into "not found" told a user with an unparseable `settings.json` that we had looked and found nothing of ours, while our MCP registration and our hooks sat in it. That answer then fed the exit code, so the command printed "everything selected is gone" and exited zero over a file it had failed to touch.

The same sentence describes the current state for these two keys.

## One real subtlety

`ENABLE_TOOL_SEARCH` should probably not be deleted blindly. `enable_tool_search_in_claude_code` is deliberately non-destructive:

```python
if "ENABLE_TOOL_SEARCH" in env:
    return None
env["ENABLE_TOOL_SEARCH"] = "true"
```

so it never overwrites a value the user set, including a deliberate `false`. After the fact the file does not record who wrote it, so an uninstall cannot tell "we set this" from "the user set this to true and we left it alone". Deleting it unconditionally would silently change a setting the user owns.

The permission rules do not have that problem. They name `repowise distill` and nothing else writes them.

## Done looks like

The two permission rules are removed on uninstall, and `claude_code_leftover_reason` reports them when they survive, the way it already reports a surviving MCP entry.

For `ENABLE_TOOL_SEARCH`, a decision either way, written down. Leaving it is defensible given the ambiguity above, but then it belongs on the report as a named leftover rather than passing silently, which is the distinction `Item.blocked` already draws elsewhere in the uninstall code.

Two other uninstall paths leave something their install wrote: #2469 (`repowise uninstall` omits `.git/hooks/post-commit` from its inventory) and #2471 (`hook rewrite uninstall` leaves `distill.commands.enabled` set in the repo config). Same question in each case.

## Scope note

Verified by reading the install and uninstall methods, grepping `packages/cli` for every reference to `DISTILL_ALLOW_RULES` and `ENABLE_TOOL_SEARCH`, and reading `claude_code_leftover_reason` end to end. I did not run the uninstall. I did not check whether the plugin install path writes either key.

rishu685:
Fixes #2470.

When Claude Code integrations are installed, `add_claude_code_distill_allow_rules()` injects `DISTILL_ALLOW_RULES` (`Bash(repowise distill:*)`, `PowerShell(repowise distill:*)`) into `permissions.allow` in `~/.claude/settings.json`. However, during uninstallation, these rules were never removed, and `claude_code_leftover_reason()` did not inspect `permissions.allow`, causing it to falsely report `settings.json` clean when distill rules were still present.

### Changes
1. **Added `uninstall_claude_code_distill_allow_rules()`** in [`claude_config.py`](packages/cli/src/repowise/cli/editor_integrations/claude_config.py):
   - Safely removes only `DISTILL_ALLOW_RULES` from `permissions.allow`.
   - Prunes empty `allow` lists and empty `permissions` dicts while preserving user-defined rules and configurations.
2. **Updated `claude_code_leftover_reason()`** in [`claude_config.py`](packages/cli/src/repowise/cli/editor_integrations/claude_config.py):
   - Checks `permissions.allow` for lingering `DISTILL_ALLOW_RULES` and returns `"our distill permission rules were still present after the write"`.
3. **Wired into Uninstallation Flows**:
   - `ClaudeCodeTarget.uninstall()` in [`claude_code.py`](packages/cli/src/repowise/cli/agent_targets/targets/claude_code.py).
   - `ClaudeCodeAdapter.uninstall_rewrite_hook()` in [`claude_code.py`](packages/cli/src/repowise/cli/agent_adapters/claude_code.py).
4. **Unit Tests**:
   - Added `TestUninstallDistillAllowRules` in [`tests/unit/distill/test_allow_rule.py`](tests/unit/distill/test_allow_rule.py) covering rule removal, empty dict cleanup, preservation of user allow/deny rules, non-existent files, and malformed JSON files.


RaghavChamadiya (approved):
Thanks @rishu685. Filtering allow by exact membership in DISTILL_ALLOW_RULES means a user's own rules can never be caught, and test_preserves_user_allow_rules pins that. For the record, the bot's note about set_claude_code_capture_hook being removed is a false alarm: the diff adds to claude_config.py and deletes nothing, and a merge simulation keeps both functions.

PR #2485, opened 2026-09-21
fix(providers): make opencode subprocess concurrency configurable via shared resolver

### LINKED ISSUE #2460: [Bug] opencode provider still hardcodes Semaphore(1), serializing every subprocess call
`codex_cli` got a configurable subprocess concurrency in #2334, which closed #2266. The identical hardcoded serialization is still present in the `opencode` provider and has never had its own issue.

On current main:

```
packages/core/src/repowise/core/providers/llm/opencode.py:320
    self._subprocess_semaphore = asyncio.Semaphore(1)
```

against the now-fixed shape one file over:

```
packages/core/src/repowise/core/providers/llm/codex_cli.py:424
    self._subprocess_semaphore = asyncio.Semaphore(_resolve_concurrency())
```

Effect is the same one #2266 described for codex_cli: every subprocess call through this provider is serialized regardless of the configured concurrency, so `generate` on a large repo runs one page at a time and the concurrency setting silently does nothing.

## What the fix looks like

`_resolve_concurrency()` already exists at `codex_cli.py:97` and does the whole job: reads an env var, defaults sensibly, warns and falls back on an invalid value, clamps to a minimum of 1. The work here is to lift that helper somewhere both providers can import rather than copy it, then use it at `opencode.py:320` with its own env var name.

Copying it would leave two implementations to drift, which is the same trap `_declares_ruff_format` was in before #2444.

## Done looks like

A shared resolver used by both providers, an env var for opencode matching the codex_cli naming convention, and tests covering default, override, invalid value and the minimum-1 clamp. The invalid-value case is the one worth pinning, since the realistic bad input is someone setting it to `0` and expecting "unlimited".

Verified against current main today.

Aman-goel-04:
## Summary

- The `opencode` provider hardcoded `asyncio.Semaphore(1)`, so every subprocess call was serialized and the concurrency setting did nothing. It now reads `REPOWISE_OPENCODE_CONCURRENCY` (default 4).
- Moved the concurrency resolver out of `codex_cli.py` into a shared `providers/llm/_concurrency.py`. Both `codex_cli` and `opencode` use it, so there is one implementation instead of a copy. Invalid values fall back to the default, and values below 1 are clamped to 1.
- Left `claude_cli.py` on its own resolver to keep this scoped to the issue. It can be migrated in a follow-up if wanted.

## Related Issues

Fixes #2460

## Test Plan

- [x] Tests pass (`pytest`)

Ran `pytest tests/unit/test_providers` (328 passed) and `ruff check` on the changed paths. New tests cover the opencode default, an override, `0` clamped to 1, and an invalid value, plus direct tests for the shared resolver. The existing codex tests pass unchanged. The concurrency tests use a fake subprocess, not a real `opencode` run.

## Checklist

- [x] I have updated documentation if needed (added a Concurrency section to `website/opencode.md`)


RaghavChamadiya (approved):
Thanks @Aman-goel-04. Lifting the resolver into `_concurrency.py` and having codex_cli delegate to it, instead of copying it, is the part #2460 cared about most, and the zero-clamps-to-one test pins the realistic bad input. Two notes, neither blocking: the unset default for opencode moves from 1 to 4 by inheriting the shared default, which I think is right for parity with codex_cli, and `docs/agent/OPENCODE.md:134` still says subprocess execution is serialized via `Semaphore(1)`, so that line wants updating in a follow-up.

Aman-goel-04:
Thanks, @RaghavChamadiya. 

Since you're online rn, could you please also review #2160, or it might go conflicting again.

Aman-goel-04:
Oh nvm it has gone conflicting again

PR #2486, opened 2026-09-21
fix: include the post-commit hook in the uninstall inventory

### LINKED ISSUE #2469: [Bug] repowise uninstall never removes the post-commit hook that init installs
## Summary

`repowise uninstall` does not list or remove the git post-commit hook, although `repowise init` is one of the two things that installs it.

`packages/cli/src/repowise/cli/uninstall/inventory.py:20-34` declares the four groups the plan enumerates:

```python
class Group(StrEnum):
    AGENTS = "agents"
    REPO_FILES = "repo-files"
    INDEX = "index"
    GLOBAL = "global"
```

`.git/hooks/post-commit` is in none of them. Grepping the module for `post-commit` returns nothing.

`repowise.cli.hooks.uninstall` has exactly one caller in the tree:

```
packages/cli/src/repowise/cli/commands/hook_cmd.py:102
    from repowise.cli.hooks import uninstall
```

which is the explicit `repowise hook uninstall`. Meanwhile `init` installs it at `packages/cli/src/repowise/cli/commands/init_cmd/_interactive.py:239`.

## Consequence

A user runs `repowise init`, accepts the post-commit hook when offered, later runs `repowise uninstall --all`, and the marker block stays in `.git/hooks/post-commit`. Nothing in the output mentions it.

The block is inert once the index is gone, because it gates on `[ -f "$ROOT/.repowise/state.json" ]` (`packages/cli/src/repowise/cli/hooks.py:74`), so nothing breaks. The problem is the claim rather than the behaviour. The module docstring at `inventory.py:1-9` is explicit about what it is for:

> The full inventory of what repowise has written, computed without writing. This module is the trust claim. [...] So the plan enumerates every path in every group, present or not, and carries for each either the action a run would take or the reason it would not.

A path that install writes and uninstall neither removes nor names is the one outcome that design rules out. `Item` already carries a `blocked` field for exactly the case where a path is listed and deliberately not touched, so the shape for an honest answer is already there.

Three uninstall paths currently leave behind something their install wrote. The other two are #2470 (the Claude Code target leaves an env pin and two permission rules) and #2471 (`hook rewrite uninstall` leaves a repo config flag and, on some machines, an AGENTS.md block). They are separate files and separate commands, but a contributor picking up one may want to read the other two, since the question in each case is the same: what did install write that uninstall does not account for?

## Done looks like

`.git/hooks/post-commit` appears in the plan under `REPO_FILES`, with either the removal a run would perform or the reason it would not. Removing the repowise block while preserving any other hook content is already implemented in `hooks.uninstall` (`hooks.py:366-391`, including the "removed (other hook content preserved)" branch), so this is wiring rather than new logic.

Worth deciding in the same pass: whether removal belongs under `REPO_FILES` or under a group of its own. The hook lives in `.git/`, which is neither the repo's own tracked files nor `.repowise/`, and a user who wants the index gone but the auto-sync hook kept is not an obviously wrong user.

## Scope note

Verified against current main by reading the inventory groups, grepping for every caller of `hooks.uninstall`, and confirming `init` installs it. I did not run `repowise uninstall`, and I did not check whether the workspace uninstall path differs.

danielalanbates:
## Problem

`repowise init` can install a post-commit hook (auto-sync), but `repowise uninstall --all` never inventoried or removed it — the uninstall plan omitted a file written by init. Fixes #2469.

## What changed

- `hooks.py`: two small public helpers — `hook_path()` (reusing the existing worktree / `core.hooksPath` / husky-aware resolution so the inventory names exactly the path install/uninstall use) and `marker_present()` (safe marker check for the shared hook file).
- `uninstall/inventory.py`: the hook is now a `repo-files` row. A hook file carrying our marker block is removable; a hook file without it is listed but blocked with the reason `no repowise block in the hook, so it is left alone`. A repo with no hook file gets no row.
- `uninstall/runner.py`: removal goes through the existing marker-aware `hooks.uninstall()`, so other tools' hook content is preserved and the file is deleted only when it held nothing else.
- `uninstall_cmd.py`: the repo-files group description now mentions the post-commit hook.

## Tests

5 new tests in `tests/unit/cli/test_uninstall_cmd.py`: install via the same entry point `init` uses, confirm dry-run names the path and `uninstall --all` removes it; unrelated hook content survives; a hook file without our marker is named, reported as kept with a reason, and left untouched; no hook file means no row; dry-run path matches the removed path. Full uninstall + hooks suites: 72 passed, no regressions.

---
Disclosure: this PR was developed with AI assistance and reviewed by the author.


RaghavChamadiya:
Thanks @danielalanbates, and welcome. This is a careful first PR: reusing `hooks._hooks_dir` through `hook_path()` instead of reimplementing the worktree, `core.hooksPath` and husky resolution is right, and removing through `hooks.uninstall()` so a hook that also carries someone else's content only loses our block is the part that most needed getting right.

I have to hold it, though, and the reason is claim order, not the code. @Aman-goel-04 claimed #2469 on the 20th and, before writing anything, asked three scoping questions that I left unanswered. This PR landed twenty minutes after those questions. The repo's rule is that the first claim on an issue keeps it, and the person who stopped to ask is the last person I want to see overtaken because I was slow to reply.

Two of Aman's questions also affect this diff. I have answered them on #2469: when `core.hooksPath` points outside the repo, or when the hooks dir belongs to the main checkout of a worktree, the row should be `blocked` with a reason, because removing it there would remove the hook for other repos too. This PR removes it in both cases.

If Aman hands the issue over, this is the natural PR to land with that change. In the meantime, #2352 is open and unclaimed: `GRAPH.md` quotes benchmark figures that `BENCHMARKS.md` has since corrected. It is a documentation fix, and a good way to get something merged quickly.

PR #2488, opened 2026-09-21
fix(coverage): report no coverable lines as not applicable, not 0% covered

### LINKED ISSUE #2193: [Bug] A file with zero coverable lines (LF:0) is ingested as 0% coverage, so coverage_gradient imputes what it documents it never will

tacopiek:
## Summary

- Every coverage parser now reports a file with **no coverable lines** as `line_coverage_pct = None` instead of `0.0`, so `coverage_gradient`, `coverage_gap` and `untested_hotspot` stay silent on files that contain no executable code.
- `FileCoverage.line_coverage_pct` and `coverage_files.line_coverage_pct` widen to nullable (migration `0077`), matching what `branch_coverage_pct` has always done for "no branches".
- The repo rollup, the module aggregate and the worst-first file ordering step over the unmeasurable rows rather than treating them as 0%.

## Related Issues

Fixes #2193

## What changed, and why it is not only the parsers

@RaghavChamadiya's scope note on the issue was right about the chain being nullable already — `health/models.py`, `biomarkers/base.py`, `FileHealth` and `engine.py` all needed nothing. Three things came up that the note did not cover, and each is a deliberate decision rather than scope creep:

**1. `untested_hotspot` needed a guard after all.** Its `cov is None` branch is not a silent return — it is a documented fallback that flags a hotspot when neither a paired test file nor `reached_by_tests` says anything tests the file. Feeding it `None` therefore swapped one finding for another rather than removing it: the barrel in the issue would still have scored −2.00 critical, just with the reason *"no paired test file and no coverage data"*. The two causes of `None` call for opposite verdicts, and `total_coverable_lines` is 0 for both, so `FileContext` gained one field, `coverage_measured`, to tell them apart. Measured-and-empty returns early; unknown still falls through to the existing fallback, and there is a test pinning each side.

**2. The migration does not reach local SQLite stores, by design.** `init_db`'s reconciler is additive-only and nothing runs Alembic against a `.repowise/wiki.db`, so `line_coverage_pct FLOAT NOT NULL` survives every upgrade on an existing store — I confirmed this on one of mine. Relaxing the model alone would have turned the next `repowise coverage add` into an `IntegrityError` for everyone who has already ingested a report. `save_coverage_files` therefore reflects the live column and omits the not-applicable rows when it is still `NOT NULL`. Nothing observable diverges between the two store shapes: every consumer reads an absent row and a NULL percentage identically (`cov is None`), and the rollup weights each row by `total_coverable_lines`, which is 0 for exactly these rows. If you would rather have a runtime table rebuild in the `decision_migration` style so both shapes converge, say so and I will add it.

**3. `None`, not `100.0`** — taking the call from the issue thread. A vacuous `100.0` would quietly inflate any average instead of producing a visible false finding, which is the worse failure mode.

`file_count` and the repo-level `line_coverage_pct` now describe files with a *measurable* percentage. A report of nothing but type-only files reports `line_coverage_pct: None` rather than 0% — the same fix one level up, and the empty-table shape already returned `None` there, so every caller handles it.

## Test Plan

- [x] Tests pass (`pytest`) — `tests/unit/health` + `tests/unit/persistence`: **2911 passed, 1 skipped**. `tests/unit/server` + `tests/unit/pipeline`: **3666 passed**.
- [x] Lint passes (`ruff check packages/ tests/`) — clean. I deliberately did **not** run `ruff format`: it rewrites unrelated pre-existing lines in five of the files I touched (955 files repo-wide would be reformatted), and that churn does not belong in this diff.
- [x] `mypy packages/core/src packages/cli/src packages/server/src` — 690 errors before, 690 after, byte-identical output. No new ones.
- [x] The issue's repro lcov, against all four parsers.

New tests:

- `tests/unit/health/test_coverage_parsers.py` — parametrized over lcov / Cobertura / Clover plus a repowise-json case, each asserting both halves of the distinction (nothing coverable → `None`, genuinely uncovered → `0.0`).
- `tests/unit/health/test_coverage_biomarkers.py` — all three biomarkers silent on the issue's shape; `untested_hotspot` still firing when no report mentioned the file; `coverage_gradient` still firing on a real 0%.
- `tests/unit/health/test_coverage_discovery.py` — the hit-wins merge keeps `None` when neither side has coverable lines and resolves to a number as soon as one does.
- `tests/unit/persistence/test_coverage_not_applicable.py` — NULL round-trip, the rollup stepping over it without moving, the unknown repo-level answer, and the legacy `NOT NULL` store ingesting successfully.

**Pre-existing failures, not from this branch.** `tests/unit/cli` has ~8–10 failures (`test_agent_target_baseline`, `test_docs_pointer`, `test_ui_package`, `test_update_e2e`, `test_update_provider_prompt`, `test_workspace_docs_pointer`). They reproduce on a clean `upstream/main` checkout here, and the exact set shifts between runs on identical code, so they look order- or environment-dependent. None touch coverage.

## Checklist

- [x] I have updated documentation if needed — the docstrings that carry the contract (`FileCoverage`, `untested_hotspot`, the repowise-json schema note) state the new meaning.


RaghavChamadiya:
Thanks @tacopiek, and thanks for reporting #2193 in the first place, the write-up made the problem really clear.

Two things before this can go in. First, @modelpath-dev claimed the issue on the 10th and confirmed the approach on the 12th, so I want to give them first shot, same as I did on #2340. If they don't pick it up this week I'm happy to move forward with yours.

Second, a few things I spotted while testing it, so whichever PR lands can cover them:
- Local stores never run the migration, so on an existing `wiki.db` the N/A rows are skipped and the barrel from the issue still scores critical, just as "no coverage data". Writing 0.0 there and treating `total_coverable_lines == 0` as no coverage at load time would cover old stores and existing rows too.
- `file-coverage-tab.tsx` calls `pct.toFixed(1)` on `line_coverage_pct`, so a type-only file's Coverage tab would throw on `null`.
- `FileDetailCoverage` and `CoverageFileRow` in `packages/types` still type it as `number`.
- A report with only type-only files ends up with `file_count: 0`, and the CLI then says "No coverage ingested."

One small heads-up: a migration `0077` just landed on main, so this one will need to become `0078` with `down_revision = "0077"`. Happy to do that bit for you if it helps.

Really appreciate the careful parser tests here.

RaghavChamadiya:
@tacopiek one more thing I should have said in my comment yesterday: commit `7e811a42` carries a `Co-Authored-By: Claude Opus 5` trailer, and this repo does not merge commits with a coding agent as author or co-author. Whichever PR lands for #2193, if it is this one, that line needs to come off (`git commit --amend`, delete it, `git push --force-with-lease`). That is about authorship only and does not change anything about the review above.

PR #2489, opened 2026-09-21
fix(server): preserve workspace extraction diagnostics

### LINKED ISSUE #2026: [Bug] Workspace extraction diagnostics loses three of its fifteen reports at the API boundary
**Summary**

`ExtractionDiagnostics.to_dict()` emits fifteen keys. `WorkspaceExtractionDiagnostics`
declares twelve. The route constructs the response with `**`, and pydantic's
default `extra="ignore"` drops the difference without raising.

`packages/server/src/repowise/server/routers/workspace.py:643`

```python
return WorkspaceExtractionDiagnostics(**diagnostics)
```

The three keys that never arrive are `symbol_identity`, `schema_coverage` and
`code_api` (`packages/core/src/repowise/core/workspace/diagnostics.py:236-241`,
serialized at `:273-275`). They are computed on every workspace extraction and
reach no consumer: `GET /api/workspace/diagnostics` is the only route that
serves them, and nothing else in the repo reads them.

**Mechanism**

Three things have to line up for this to be silent, and all three do.

1. The schema is a strict subset. `schemas/workspace.py:254-268` lists twelve
   fields, ending at `http_consumer_coverage`. The three report fields were
   never added.
2. Nothing rejects the extras. There is no `model_config` in
   `schemas/workspace.py`, so pydantic v2's default `extra="ignore"` applies and
   construction succeeds.
3. There is no nested schema to add them to. `SymbolIdentity`
   (`diagnostics.py:115`), `SchemaCoverage` (`:143`) and `CodeApiCoverage`
   (`:184`) have no pydantic counterparts, so this is not a one-line field
   addition.

The core round trip is fine, which narrows where the fix belongs:
`ExtractionDiagnostics.from_dict` reads all three back
(`diagnostics.py:280-281`, `:329`). The loss is at the API boundary only.

Worth naming what is lost precisely, because these are not counters. Each of the
three carries **two denominators on purpose**, and the second one is the
diagnostic. From `SymbolIdentity`:

> *total* is the honest workspace-wide share. *bound_ratio_indexed* excludes the
> files with no parsed symbols at all [...] which is what says whether the
> binding rule itself is working: no rule can reach a file the parser never saw.

`SchemaCoverage` and `CodeApiCoverage` are built the same way and say so. All
three are documented "Reported, never asserted" - they exist to answer whether
extraction is working, which is the question this endpoint is for.

**Repro**

Run a workspace extraction, then:

```
GET /api/workspace/diagnostics
```

The response carries `total_providers` through `http_consumer_coverage` and
stops. No `symbol_identity`, no `schema_coverage`, no `code_api`.

The drop itself, without a server:

```python
from pydantic import BaseModel

class M(BaseModel):
    a: int = 0

M(**{"a": 1, "b": 2}).model_dump()   # {'a': 1} - 'b' is gone, no error
```

**Impact**

Every workspace, since the fields were added. The endpoint answers "is
extraction working" while withholding the three measurements built to answer it.
No error, no partial-response marker, and the response validates.

Unsized beyond that. No test asserts on the three keys in either direction, so
there was nothing to regress.

**Done looks like**

`symbol_identity`, `schema_coverage` and `code_api` reach the endpoint, which
needs pydantic models for the three nested reports first.

Two things worth deciding while fixing rather than after:

* Set `extra="forbid"` on `WorkspaceExtractionDiagnostics`, or construct it field
  by field. The `**` plus permissive-default combination is what made this
  silent, and it will re-silence the next field added to `to_dict()`.
* The same shape exists next door. `BreakingChange.provider_symbol_id`
  (`core/workspace/breaking_change.py:137`, populated at `:574`) is likewise
  absent from `WorkspaceBreakingChange` (`schemas/workspace.py:325-338`). Worth
  checking the rest of `schemas/workspace.py` against its core counterparts in
  the same pass.

Tests: assert the endpoint's key set against `ExtractionDiagnostics.to_dict()`'s
key set, so the next divergence fails instead of disappearing.

Shivang9983:
## Summary

- Preserve all workspace extraction diagnostics fields at the API boundary.
- Add strict schema validation with `extra="forbid"` to prevent silently dropped diagnostics.
- Add a regression test to keep the core diagnostics payload and API schema in sync.

## Related Issues

Fixes #2026

## Test Plan

- [x] Tests pass (`pytest`) — `64 passed`

## Checklist


RaghavChamadiya:
Thanks @Shivang9983, and sorry this sat three days without a reply. Adding the three nested models before setting `extra="forbid"`, in that order, is exactly the ordering I warned about on #2026, and deriving the models' fields from the core `to_dict()` shapes is right.

Two things before it can merge.

**The key-set test never runs.** In `tests/unit/server/test_workspace_router.py` it is indented one level too deep, so it is defined inside `test_reports_orphans_and_counts` instead of on the class:

```python
        def test_extraction_diagnostics_schema_matches_core_payload(self) -> None:
```

pytest does not collect nested functions, so CI is green without executing it. That test is the part of #2026 that stops the next field from vanishing silently, so it needs to sit at the same indentation as its siblings. It will then pass, since `to_dict()` emits exactly those keys.

**It needs a rebase.** Main has changed `schemas/workspace.py` and the same test file since you branched, so both conflict. The schemas one is trivial: main now has `from typing import Literal` above the pydantic import, and you need both that line and your `ConfigDict` import.

`provider_symbol_id` on `WorkspaceBreakingChange` has already landed on main in the meantime, so you can drop that from your list. Ping me once it is rebased and green and I will merge.

Shivang9983:
Thanks @RaghavChamadiya  for the detailed review! I've rebased the PR onto the latest main, fixed the indentation of the key-set test so pytest can collect it, and resolved the conflicts while keeping both Literal and ConfigDict. I also dropped provider_symbol_id since it is now present on main.

Ruff lint and formatting checks are passing, and I've force-pushed the rebased branch. The PR is ready for another look. Thanks!

PR #2490, opened 2026-09-21
fix(cli): read workspace list status from the configured database

### LINKED ISSUE #2368: [Bug] workspace list reads every repo as "not indexed" under a configured shared database

sloemo01:
Fixes #2368.

`workspace list` decided whether a repo was indexed from `.repowise/` on disk, and `_query_repo_counts` returned `(0, 0)` whenever `.repowise/wiki.db` was missing. Both describe the repo-local SQLite default and only it. Under a configured shared database those files are absent by design, so every member read `not indexed` with zero counts however well indexed it was, and the summary line said `0/N` for a fully indexed workspace.

The output was misleading rather than merely empty. The next step a reader draws from "not indexed" is to reindex, against a database that already holds the pages.

Both checks now consult `db_configured()`, the helper #2320 used for `repowise delete`, which is the shape the issue names.

## Measured

End to end against a seeded store, not from the tests alone:

```
no configured URL, no .repowise/        not indexed         0/1 repos indexed
REPOWISE_DB_URL, 5 files, 11 symbols    5 files, 11 symbols,
                                        up to date          1/1 repos indexed
```

The counts come from the store the engine was already built from, `get_db_url_for_repo`, so the guard only stops the query being skipped. With no configured URL the local-file behaviour is byte-identical, which the fourth test pins.

## What the tests cover

Four cases in `tests/unit/cli/test_workspace_list_shared_db.py`, following the `REPOWISE_DB_URL` fixture pattern from `test_update_stamp_configured_db.py`: a configured database is not reported as unindexed, its counts reach the row, the summary counts it, and no configured database keeps the old behaviour.

Each half of the fix is pinned by its own assertion. Reverting the indexed guard alone reds three tests; reverting the counts guard alone reds exactly the counts test; the full fix is 4/4. The counts assertion reads the row rather than the whole output, so a digit in a temp path cannot satisfy it.

## Scope

The issue's scope note asks whether other `workspace_cmd` subcommands share the assumption. They do not: after this change the file holds no other local-file proxy for indexed state. The remaining instances of the pattern are in other surfaces and stay open, `#1034` and `#2148`.

`indexed_count` feeds the summary line and moves with the per-row decision, as the issue asked. `check_repo_staleness` runs on that branch too, so a row that previously short-circuited now has its staleness evaluated rather than skipped.


sloemo01:
Full local run of `tests/unit/` on this branch, for the record:

```
3 failed, 21006 passed, 19 skipped, 4 xfailed   (32m37s)
```

The three failures are outside this change, and I checked that rather than assuming it:

- `tests/unit/cli/test_agent_target_baseline.py::test_integration_writes_match_baseline`
- `tests/unit/distill/test_savings_event.py::test_the_command_is_never_persisted`
- `tests/unit/ingestion/test_repo_totals_churn.py::test_a_rebased_branch_in_the_history_falls_back`

All three reproduce in isolation (so they are not concurrency flakes from the long run), and all three fail identically with `workspace_cmd.py` reverted to the base revision and my test file left in place. Nothing this diff touches is imported by them: it changes one CLI command module and adds one test module.

Two of them look environmental rather than code-level on this machine, which is why I am reporting rather than fixing them here: `test_a_rebased_branch_in_the_history_falls_back` dies on `git rebase -q master` (exit 128, the fixture expects a `master` branch), and `test_the_command_is_never_persisted` asserts something about a JSON payload shape.

I did not find an issue tracking any of the three. Happy to file them if that is useful, and I would rather not fold unrelated fixes into this PR.

sloemo01:
CI finished green on `39c538c4`, and it settles the local-failure note above.

```
Python 3.11 tests          success
Python 3.12 tests          success
Python 3.13 tests          success
Python lint (ruff)         success
Web UI lint + type check   success
VS Code extension test     success
tree-sitter floor          success
Structurizr DSL parses     success
Integration tests          skipped (as configured)
```

The three failures I reported from the local run do not reproduce in CI on any of the three versions. That matches the read I gave them at the time: environment-dependent rather than code-level, which is why I reported them instead of fixing them here. Nothing in this diff is implicated either way.

Remaining gate is maintainer review.

sloemo01:
Closing this in favour of #2403, which is the same fix and is already open.

I opened #2403 on Sep 18, claimed #2368 in the same session, and then forgot it. When I picked the issue back up this week I read the thread, saw no PR linked to it, and shipped #2490 as if it were the first. The claim comment at the top of #2368 and my own "shipped as #2490" comment below it are both mine, eleven minutes apart on different days. That is the mistake, and it is worth naming rather than dressing up as a scope call.

They conflict rather than compose, so this is not a case of keeping both. Two measurements decided which one survives.

**#2403's tests fail against this PR's code.** I ran its `tests/integration/test_cli.py` classes with `PYTHONPATH` pointed at each side:

```
TestWorkspaceListWithConfiguredDb::test_shared_db_repos_read_as_indexed
  #2403 code  -> passes
  #2490 code  -> fails: row_b reads "empty" where it should read "not indexed"
```

**This PR's tests fail against #2403's code**, for the opposite reason: two of the four assert `not indexed` is absent and `1/1 repos indexed` for a repository the shared store holds no row for. #2403 treats that case as not indexed on purpose, because it distinguishes `None` ("no store could answer") from `(0, 0)` ("store answered, nothing there"). This PR collapses the two, which is why a member the store has never seen comes back as indexed with a zero count.

That collapse is the same class of defect as the one being fixed: the row reads as a broken index instead of a missing one, and sends the reader looking in the wrong place. So #2403 keeps the fix.

**What I moved across before closing.** #2490's tests are at unit tier, which this repo runs on pull requests, and #2403's are in `tests/integration/`, which CI runs only on push. So the assertions here were the better-guarded ones even though the code was not. I ported them onto #2403 in `cd640f492` as `tests/unit/cli/test_workspace_list_store_states.py`, rewritten to #2403's semantics and covering the three states: a row the store holds, a path it does not, and no configured database. Sabotage-checked against two mutations, one of which is this PR's exact collapse, and each mutation reds the arm it names.

Nothing is lost by closing this other than the duplicate review load.

PR #2491, opened 2026-09-22
fix(cli): clear distill.commands.enabled and remove AGENTS.md on hook rewrite uninstall (#2471)

### LINKED ISSUE #2471: [Bug] hook rewrite uninstall leaves distill.commands.enabled set and skips AGENTS.md when Codex is gone
## Summary

`repowise hook rewrite uninstall` says it removes the rewrite hooks and the AGENTS.md awareness section. It leaves the repo-level flag that install set, and it skips the AGENTS.md removal entirely when Codex is no longer installed.

`packages/cli/src/repowise/cli/commands/hook_cmd.py:349-371`:

```python
def rewrite_uninstall(path: str | None, workspace: bool, no_workspace: bool) -> None:
    """Remove the rewrite hooks and the AGENTS.md awareness section."""
    ...
    removed = ClaudeCodeAdapter().uninstall_rewrite_hook()
    ...
    codex = CodexAdapter()
    if codex.detect():
        codex_removed = codex.uninstall_rewrite_hook()
        ...
        for repo_path in _target_repo_paths(target):
            if remove_agents_md_distill_section(repo_path):
                ...
```

Two things follow from that shape.

**The AGENTS.md removal is inside the Codex branch.** Install is not. `rewrite_install`'s own docstring says so at `hook_cmd.py:288`:

> the AGENTS.md awareness section installs regardless, because it [...]

So a user who installed while `~/.codex` existed, then removed Codex, cannot remove the section through this command. `codex.detect()` is false, the loop never runs, and nothing is printed about it.

**`distill.commands.enabled` is never cleared.** `rewrite_install` sets it, at `hook_cmd.py:273` and `:278`:

```python
save_distill_commands_enabled(abs_path, enabled=True)
```

`rewrite_uninstall` has no matching call. Grepping the file for `save_distill_commands_enabled` finds the two writes in install and none in uninstall.

## Consequence

For the AGENTS.md half: the managed block stays in a tracked file the user owns, with no command that removes it and no message saying it was skipped. Hand-editing is the only route left.

For the config half: `.repowise/config.yaml` keeps `distill.commands.enabled: true` after the user has uninstalled the thing it was set for. Any later path that installs the rewrite hook without re-asking, `init` among them, then resumes rewriting on a flag the user believes they turned off. The flag outliving the feature is what makes this more than cosmetic.

## Done looks like

The AGENTS.md removal runs on the same condition install ran on, which is unconditionally, and reports when it found nothing rather than when it was skipped. `remove_agents_md_distill_section` already returns a bool for that.

`distill.commands.enabled` is set back to `False` on uninstall, and the workspace branch does it for each repo the install branch wrote it to.

Worth checking in the same pass: whether the Codex rewrite hook removal itself should be gated on `codex.detect()`. A user who removed Codex still has whatever repowise wrote into `~/.codex/hooks.json`, and detection keys on the directory existing.

Two other uninstall paths leave something their install wrote: #2469 (`repowise uninstall` omits `.git/hooks/post-commit` from its inventory) and #2470 (the Claude Code target leaves an env pin and two permission rules in settings.json while reporting the file clean). Same question in each case: what did install write that uninstall does not account for?

## Scope note

Verified by reading `rewrite_install` and `rewrite_uninstall` in full and grepping the file for `save_distill_commands_enabled`. I did not run either command, and I did not check what `_target_repo_paths` returns for a workspace whose members were added after install.

rishu685:
Closes #2471.

When `repowise hook rewrite install` runs, it enables `distill.commands.enabled: true` in `.repowise/config.yaml` for each target repository (and all workspace members in workspace mode), and writes the `AGENTS.md` distill section if Codex is detected.

Previously, `repowise hook rewrite uninstall`:
1. Gated the `remove_agents_md_distill_section` cleanup inside `if codex.detect():`, skipping `AGENTS.md` cleanup entirely if Codex was removed/uninstalled.
2. Never cleared `distill.commands.enabled` back to `False` in `.repowise/config.yaml`.

---

### Changes
1. **Unconditional `AGENTS.md` distill section removal**:
   - In [`packages/cli/src/repowise/cli/commands/hook_cmd.py`](packages/cli/src/repowise/cli/commands/hook_cmd.py), moved `remove_agents_md_distill_section` outside the `if codex.detect():` check so it runs unconditionally across all target repositories.
2. **Clear `distill.commands.enabled`**:
   - Invokes `save_distill_commands_enabled(repo_path, enabled=False)` on each target repo (and every member repo in workspace mode).
3. **Tests**:
   - In [`tests/unit/distill/test_config_and_doctor.py`](tests/unit/distill/test_config_and_doctor.py):
     - Updated `test_uninstall_removes_hook_and_disables_repo_config` to assert `distill.commands.enabled` is `False`.
     - Added `test_uninstall_workspace_disables_all_repo_configs` to test workspace uninstall.
     - Added `test_uninstall_removes_agents_md_section_when_codex_not_detected` to test `AGENTS.md` removal when Codex is absent.


RaghavChamadiya (approved):
Thanks @rishu685. Moving the AGENTS.md cleanup out from under codex.detect() is exactly the fix, and it still goes through remove_agents_md_distill_section, so only the section repowise wrote is touched. The workspace test covering every member is a good addition.

